PR #1, opened 2022-06-29
refine docs

osoukup:
This is a draft for now and it is not ready for the review.

https://issues.redhat.com/browse/OSIDB-173


collectors/framework/README.md
```diff
@@ -1 +1,68 @@
...
+    """
+
+    from django.apps import AppConfig
+
+
+    class ExampleCollector(AppConfig):
+        """example collector"""
+
+        name = "collectors.example"
+
```
JakubFrejlach, line 18: Would wrap this into:
\`\`\`python
text
\`\`\`

collectors/framework/README.md
```diff
@@ -1 +1,68 @@
...
+    # as we set celery task bind=True by default
+    # we get the task object as the parameter here
+    # which is our collector
+    def example_collector(collector_obj):
+        """example collector"""
+        logger.info(f"Collector {collector_obj.name} is running")
+        # fake data to be up-to-date
+        collector_obj.store(updated_until_dt=timezone.now())
+        # optionally return some result to Celery
+        return f"The run of {collector_obj.name} was fun"
```
JakubFrejlach, line 47: Would wrap this into:
\`\`\`python
text
\`\`\`

collectors/framework/README.md
```diff
@@ -1 +1,68 @@
...
+
+Then, you need to register your Django application to `INSTALLED_APPS`
+in [settings](../../config/settings.py). Compose your OSIDB instance down and start again so the
+Celery hosts are up-to-date. Now you should see a new collector status being reported when running
+
+    # provide credentials and optionally set URL and port appropriately
+    export OSIDB_ACCESS_TOKEN=$(curl -H 'Content-Type: application/json' '-d' '{"username": USERNAME,"password": PASSWORD}' http://localhost:8000/auth/token | jq ."access" -r)
+    curl -H "Authorization: Bearer $OSIDB_ACCESS_TOKEN" -X GET "https://localhost:8000/collectors/api/v1/status" -w "\n" | jq
```
JakubFrejlach, line 55: Would wrap this into:
\`\`\`bash
text
\`\`\`

collectors/framework/README.md
```diff
@@ -1 +1,68 @@
...
+
+When working with the data models with ACLs applied, do not forget to set the collector permissions
+appropriately. Default ACL groups can be observed [here](../../config/settings_local.py).
+
+    # set permissions to public read-only access
+    from django.conf import settings
+    from osidb.core import set_user_acls
+    set_user_acls(
+        settings.PUBLIC_READ_GROUPS
+    )
```
JakubFrejlach, line 68: Would wrap this into:
\`\`\`python
text
\`\`\`

docs/developer/CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
...
+  expected behavior that is not occurring.
+
+### Did you write a patch that fixes a bug?
+
+* Open a new GitHub pull request with the patch.
+
+* Ensure the PR description clearly describes the problem and solution.
+  Include the relevant issue number if applicable.
+
+* Before submitting make sure that or linters and tests are passing.
```
JakubFrejlach, line 31: `that or linters and tests` -> `that linters and tests` ?

docs/user/AUTH.md
```diff
@@ -0,0 +1,42 @@
+# Kerberos authentication
```
Elkasitu, line 1: I think we should move this entire section to a Confluence page, it's not relevant to non-Red Hatters

Elkasitu (approved):
LGTM

JakubFrejlach (approved):
Couple of suggestions but otherwise LGTM!

docs/user/AUTH.md
osoukup: I am not entirely sure on this. True that it is Red Hat specific like eg. errata collector etc. At least for now. It is however defined here in the repo so the implementation is not internal. I would not bother much by moving it anywhere now but in the future we could either move it to some Red Hat internal place or generalize/parametrize it so it can be theoretically used by someone else too and offer it as one of the auth methods. For now I would extend the AUTH.md documentation with the local instance (the default) method.

docs/developer/CONTRIBUTING.md
osoukup: Good catch!

README.md
```diff
@@ -1,19 +1,100 @@
...
 
-### apps
-* [osim](apps/osim/README.md) - Open Security Issue Manager
+* [Introduction](#Introduction)
+* [Technologies](#Technologies)
+* [Structure](#Structure)
+* [Setup](#Setup)
+* [Usage](#Usage)
+* [Contributing](#Contributing)
+* [Status](#Status)
```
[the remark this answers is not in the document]
osoukup: Hm. You are correct that this should be the standard way. Probably GitHub allows the uppercase or does not care for the case at all as I tested that the links work and lead to where they should lead... Will lowercase it

docs/user/AUTH.md
Elkasitu: I don't mean the entire documentation, just the AUTH part, most of the things explained here give context into RH's internal kerberos implementation (which I will have to remove anyway) and only apply to RH, even if someone else were to use our kerberos + ldap module they most likely wouldn't encounter all the issues with `dns_canonicalize_hostname`, the different realms, etc.

osoukup: As of the chat discussion the Kerberos related chapter will be moved to `krb5_auth` and preserved there for now.

PR #2, opened 2022-07-01
Amend audit migration to correctly delete all data

Elkasitu:
The automatically generated migration by Django / PG History does not
correctly delete all the necessary items then recreate the deleted
tables with some required columns so that the actual migration can
proceed.


osoukup (approved):
What magic is this? LGTM but I barely understand what you do here. Some internal tables need a cleanup, that's all I understand here :smile:

JakubFrejlach:
LGTM. Did a quick search about the stuff you are deleting to understand it a little bit and it seems ok.

JakubFrejlach (approved):

PR #3, opened 2022-07-04
bzimport improvements

osoukup:
There are some minor BZ import improvements - namely disabling history fetching and better handling of restricted flaw related bugs when looking for tasks. More importantly there is however on quite important change. This was probably forgotten or I do not know. Basically the reliability of the BZ import is based on rerunning any batch sync which encountered some temporary failure. However the failures were so far not categorize in any way so all were just considered persistent - even any network outages, backend downtimes etc. This slight omission results in the reliability being on the same level as it was (zero) because we just skip all the errors. This PR introduces the BZ exception categorization into recoverable and non-recoverable and stops skipping the recoverable ones - failing hard there so the batch gets rerun.


collectors/bzimport/collectors.py
```diff
@@ -446,9 +446,12 @@ def collect(self):
...
             except Exception as e:
-                # TODO just for testing and should be removed
-                if "No CVE ID found in Bugzilla alias" in str(e):
-                    continue
+                # when we encounter an exception which can be
+                # recovered by the rerun we fail the sync of
+                # the whole batch so it can be fully rerun
+                if isinstance(e, RecoverableBZImportException):
+                    raise e
+
```
Elkasitu, line 464:
```python
except RecoverableBZImportException:
    raise
except Exception as e:
    ...
```

JakubFrejlach (approved):
Appart from the things raised by @Elkasitu, LGTM.

collectors/bzimport/exceptions.py
```diff
@@ -5,3 +5,17 @@
...
+
+
+class RecoverableBZImportException(BZImportException):
+    """
+    temporary exceptions to be recovered by a rerun
+    may be network issues or backend outages etc.
+    """
+
+
+class NoneRecoverableBZImportException(BZImportException):
```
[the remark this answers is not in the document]
osoukup: That is a funny typo :smile: thanks for spotting it!

collectors/bzimport/collectors.py
osoukup: Much better! Thank you

osoukup:
@Elkasitu all the requested changes should be hopefully incorporated. Feel free to merge this if it is OK now (and if it is possible on GitHub). I will be back at the end of July.

Elkasitu (approved):

PR #7, opened 2022-07-13
Unify logging + fix exporting the env variables

JakubFrejlach:
Unified logging across the OSIDB:
* Created custom formatter for the celery tasks to inject celery specific info. 
* separation of the `get_task_logger` and `logging.getLogger`
* unified logging inside the tests

minor: fixed exporting variables into containers

Closes https://issues.redhat.com/browse/OSIDB-33

### DIFF
--- a/collectors/errata/tests/test_core.py
+++ b/collectors/errata/tests/test_core.py
@@ -1,5 +1,3 @@
-import logging
-
 import pytest
 
 from osidb.models import Erratum, Tracker
@@ -15,7 +13,6 @@ from ..core import (
 BZ_CASSETTE = "TestErrataToolCollection.test_get_bz_trackers_for_erratum.yaml"
 JIRA_CASSETTE = "TestErrataToolCollection.test_get_jira_trackers_for_erratum.yaml"
 
-logger = logging.getLogger("collectors.errata")
 pytestmark = pytest.mark.unit
 
 
--- a/config/settings.py
+++ b/config/settings.py
@@ -164,9 +164,13 @@ LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
     "formatters": {
+        "verbose_celery": {
+            "()": "osidb.helpers.TaskFormatter",
+            "format": "%(asctime)s [%(levelname)s] %(task_name)s%(task_id)s: %(message)s",
+        },
         "verbose": {
             # exact format is not important, this is the minimum information
-            "format": "%(asctime)s [%(levelname)s] %(name)s: %(message)s"
+            "format": "%(asctime)s [%(levelname)s] %(name)s: %(message)s",
         },
     },
     "handlers": {
@@ -174,7 +178,7 @@ LOGGING = {
         "celery": {
             "level": "WARNING",
             "class": "logging.StreamHandler",
-            "formatter": "verbose",
+            "formatter": "verbose_celery",
         },
     },
     "loggers": {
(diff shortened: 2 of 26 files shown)


osidb/helpers.py
```diff
@@ -56,3 +58,23 @@ def get_unique_meta_attr_keys(model: Type[models.Model]) -> List[str]:
 def get_model_fields(model: Type[models.Model]) -> List[str]:
     """Get all field names of the model"""
     return [field.name for field in model._meta.get_fields()]
+
+
+class TaskFormatter(logging.Formatter):
+    """
+    Custom formatter based on celery 'celery.utils.log.TaskFormatter'
```
Elkasitu, line 65: can't we use `celery.utils.log.TaskFormatter` directly in the settings?

config/settings.py
```diff
@@ -188,54 +192,32 @@
...
             "level": "WARNING",
             "handlers": ["console"],
             "propagate": True,
         },
         "django_auth_ldap": {"level": "WARNING", "handlers": ["console"]},
+        # Collectors loggers
+        **{
+            collector_name: {
+                "level": "WARNING",
+                "handlers": ["console"],
```
Elkasitu, line 205: shouldn't it use the `celery` handler?

collectors/errata/tests/test_core.py
```diff
@@ -15,7 +15,7 @@
 BZ_CASSETTE = "TestErrataToolCollection.test_get_bz_trackers_for_erratum.yaml"
 JIRA_CASSETTE = "TestErrataToolCollection.test_get_jira_trackers_for_erratum.yaml"
 
-logger = logging.getLogger("collectors.errata")
+logger = logging.getLogger(__name__)
```
Elkasitu, line 18: do we even need logging in tests?

Elkasitu (commented):
Mostly questions

docker-compose.yml
```diff
@@ -125,6 +125,9 @@ services:
         BZIMPORT_BZ_API_KEY: ${BZIMPORT_BZ_API_KEY:?Variable BZIMPORT_BZ_API_KEY must be set.}
         JIRA_AUTH_TOKEN: ${JIRA_AUTH_TOKEN:?Variable JIRA_AUTH_TOKEN must be set.}
         JIRAFFE_AUTO_SYNC: ${JIRAFFE_AUTO_SYNC}
+        ET_URL: ${ET_URL}
+        PRODUCT_DEF_URL: ${PRODUCT_DEF_URL}
+        DASHBOARD_URL: ${DASHBOARD_URL}
```
Elkasitu, line 130: Should be added to `docker-compose.test.yml`

config/settings.py
```diff
@@ -188,54 +192,32 @@
...
+                "propagate": True,
+            }
+            for collector_name in [
```
osoukup, line 208: Could we make this alphabetically sorted?

osidb/helpers.py
JakubFrejlach: We can, but the base `celery.utils.log.TaskFormatter` injects `???` in log massages where `task_name` or `task_id` cannot be retrieved, eg. the code is executed outside the Celery task, it felt to me kinda disturbing to have bunch of question marks  in the logs and also it allows us to `task_name` which can be switched just for the name when it is missing (again when the code is executed outside of the celery task).

I am not strongly for or against this, so we can use the basic `celery.utils.log.TaskFormatter`, wdyt?

config/settings.py
JakubFrejlach: Yes it should, but it seems, that the `celery` logger handler overrides the handler for the collectors loggers anyway.

collectors/errata/tests/test_core.py
JakubFrejlach: Actually, when you brought this up, I guess we don't really make any use of them. Will remove them, is that ok?

docker-compose.yml
JakubFrejlach: Already done by @osoukup in #14

config/settings.py
JakubFrejlach: Yes, we definitely can.

osidb/helpers.py
Elkasitu: in that case I have no opinion either, but maybe it should inherit from the celery TaskFormatter instead of the generic one? not sure if the celery one does extra stuff

collectors/errata/tests/test_core.py
Elkasitu: sounds good

osidb/helpers.py
JakubFrejlach: The celery `TaskFormatter` is defined kinda weird, it inherits from some celery `ColorFormatter` but it overrides its coloring behavior and thus it basically inherits from the base `logging.Formatter` so I would say, that we don't need to inherit from the celery one. Also it is not a `celery.utils.log.TaskFormatter` but `celery.app.log.TaskFormatter`, just for the record.

Elkasitu (approved):

PR #9, opened 2022-07-20
Allow unauthenticated access to the OSIDB API

Elkasitu:
This commit changes the permission class of the main OSIDB endpoints to
IsAuthenticatedOrReadOnly which allows unauthenticated users to consume
the API in read-only fashion.

The commit also applies some changes to the auth and middleware hooks
that set the ACLs so that AnonymousUsers (=unauthenticated users) can
only read publically available information.

Closes OSIDB-313


osoukup (approved):
Makes sense. Especially now when the tool might potentially serve outside RH.

Elkasitu:
@RedHatProductSecurity/osidb-devs note that I tried to add some tests but it's non-trivial: we need to enable RLS in a local setting which is more complicated than I originally thought, but I have a WIP branch with some progress on that front

PR #12, opened 2022-07-25
Properly cleanup all FlawMeta objects on flaw sync

Elkasitu:
Before this commit, only certain types of FlawMeta objects were being
cleaned up / synchronized between OSIDB and upstream (BZ).

This led to issues such as typos in acknowledgments not being properly
corrected, as instead of acks with typos being replaced by acks without
typos all acks would be kept and displayed on CVE pages, including those
with typos.

This commit changes the behavior so that all types of FlawMeta are kept
in sync by removing any existing FlawMeta with a given flaw id, type and
meta_attr if it no longer exists upstream.

Closes OSIDB-315

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -246,6 +246,7 @@ class FlawSaver:
 
     def clean_affects(self):
         """clean obsoleted affects"""
+        # TODO: potentially optimize like clean_meta ?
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM. This is more tricky than I originally thought. Thank you for the improvement!

PR #19, opened 2022-08-02
Preparation for using Flower in stage / production

Elkasitu:
This PR includes some necessary changes to start using Flower in a meaningful way in stage / production, mainly as a way to see any errors emanating from collector runs, but also as a way to provide an API from which a myriad of things can be achieved (such as stats).

The main change is setting the collector exceptions to be of type `Retry`, which remove a lot of the noise from the Flower UI:

Screenshot from 2022-08-02 17-39-14
Screenshot from 2022-08-02 17-39-24
Screenshot from 2022-08-02 17-39-36

It also includes a small fix to the Jiraffe collector that I discovered while using Flower (proving that it works, huzzah!)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ tests/htmlcov
 .tox/
 /venv
 .vscode/
+*.db.gz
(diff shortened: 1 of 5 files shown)


JakubFrejlach (approved):
LGTM! This will make Flower a lot more useful.

osoukup (approved):
Very nice improvements! LGTM

PR #21, opened 2022-08-03
Validate combinations of the major incident flags

Elkasitu:
Certain combinations of the MAJOR_INCIDENT and MAJOR_INCIDENT_LITE
FlawMeta flags should never be valid, to ensure that this constraint is
enforced a validation method is added on the FlawMeta model and executed
on each save.

Closes OSIDB-329

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -9,7 +9,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.postgres import fields
 from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 from polymorphic.models import PolymorphicModel
@@ -1096,10 +1096,40 @@ class FlawMeta(TrackingMixin):
     def __str__(self):
         return str(self.uuid)
 
+    def _validate_major_incident_combos(self):
+        """
+        Checks that the combination of MAJOR_INCIDENT and MAJOR_INCIDENT_LITE is valid.
+        """
+        if self.type not in (
+            self.FlawMetaType.MAJOR_INCIDENT,
+            self.FlawMetaType.MAJOR_INCIDENT_LITE,
+        ):
+            return
+
+        INVALID_COMBOS = [("+", "+"), ("+", "?"), ("?", "+"), ("?", "-"), ("-", "?")]
+        maj_incident_flag = None
+        maj_incident_lite_flag = None
+
+        # must include self as it's potentially not yet included in flaw.meta.all()
+        for meta in list(self.flaw.meta.all()) + [self]:
+            if meta.type == FlawMeta.FlawMetaType.MAJOR_INCIDENT:
+                maj_incident_flag = meta.meta_attr.get("status")
+            if meta.type == FlawMeta.FlawMetaType.MAJOR_INCIDENT_LITE:
+                maj_incident_lite_flag = meta.meta_attr.get("status")
+            if maj_incident_flag and maj_incident_lite_flag:
+                break
+
+        flag_pair = (maj_incident_flag, maj_incident_lite_flag)
+        if flag_pair in INVALID_COMBOS:
+            raise ValidationError(
+                f"Flaw MAJOR_INCIDENT and MAJOR_INCIDENT_LITE combination cannot be {flag_pair}."
+            )
+
     def validate(self, *args, **kwargs):
         """validate model"""
         # add custom validation here
         super().clean_fields(*args, exclude=["meta_attr"], **kwargs)
+        self._validate_major_incident_combos()
 
     def save(self, *args, **kwargs):
         """save model override"""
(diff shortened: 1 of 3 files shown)


Elkasitu:
@RedHatProductSecurity/osidb-devs I've done things like this so that we can move forward, I am not at all happy with the result but the alternatives take way more time and thinking:

I think this goes back to changing the way we handle the check on whether a Flaw is a major incident or not, currently there's a lot of shenanigans to be done because we use the same representation as bugzilla (+, -, ?, ...), the correct way IMO would be to simply have the is_major_incident flag become an Enumeration that is set to a well-defined type on ingestion from bugzilla, something like:

```python
class MajorIncident(Enum):
    UNKNOWN                  # ( ,  )
    REJECTED                 # (-, -)
    REQUESTED                # (?, ?)
    ACCEPTED                 # (+, ) or (+, -)
    CISA                     # ( , +) or (-, +)
```

Then this validation doesn't even need to happen in the first place, only at ingestion from BZ, consumers who create flaws through OSIDB won't even be able to make a mistake

osidb/models.py
```diff
@@ -1096,10 +1096,36 @@ class Meta:
...
+        if flag_pair in INVALID_COMBOS:
+            raise ValidationError(
+                f"Flaw MAJOR_INCIDENT and MAJOR_INCIDENT_LITE combination cannot be {flag_pair}."
+            )
+
     def validate(self, *args, **kwargs):
         """validate model"""
         # add custom validation here
         super().clean_fields(*args, exclude=["meta_attr"], **kwargs)
+        self._validate_major_incident_combos()
```
JakubFrejlach, line 1128: Just a thought. If someone would modify the metas on flaw from the flaw side, eg. like flaw.meta.set(...), then this validation is not triggered at all since FlawMeta is basically not saved. I am not sure if we currently do this kind of `FlawMeta` manipulation but if we would use it in the future then this validation would not be 100% safe.

JakubFrejlach (commented):
Couple of minor things and one thing which we should really consider imho.

osidb/tests/test_flaw.py
```diff
@@ -397,3 +397,57 @@ def test_validate_choice(self, attr_name):
...
+    def test_major_incident_flag_invalid(self, flag_pair):
+        """
+        Test invalid combinations of hightouch / hightouch-lite flag values.
+        """
+        flaw = FlawFactory()
+        FlawMetaFactory(
+            type="MAJOR_INCIDENT",
+            meta_attr={"status": flag_pair[0]},
+            flaw=flaw,
+        ).save()
```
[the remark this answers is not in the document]
Elkasitu: that's what I thought but without the call to save the test fails

osidb/models.py
Elkasitu: I took it into consideration, but realistically you cannot create a `FlawMeta` without assigning it a flaw since the field is required, so you would have to create a `FlawMeta` with some `Flaw` already assigned to it, and then pass it to some other flaw through `Flaw.meta.set()`... I can implement it but not sure if it even makes sense

osidb/tests/test_flaw.py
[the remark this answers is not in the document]
Elkasitu: actually nevermind I figured out what the error is

Elkasitu:
CI fails due to `pip-audit` check, see #22

osidb/models.py
JakubFrejlach: Yeah, I am aware of that, I was just seeing it as a possible corner case when technically someone is able to manually force this into data and omit the validation. But probably no big deal, it is a really specific case.

Elkasitu: So I did some more research last night, and well the problem is that `set` and other `RelatedManager` methods do a direct SQL query to perform their actions (to be efficient) and that means that they skip save (but it's still saved into the db regardless).

I thought that maybe we can somehow override these `RelatedManager` methods but it seems like it's not possible [1] since the `RelatedManager` is created dynamically.

The only "solution" is using the `bulk=False` parameter when calling any of these methods [2], but it's not something we can enforce (technically speaking we can but that's more complicated / another story) so it'd have to be by convention, and at that point we might as well convene to simply not use `set()` and `add()` for `Flaw.meta`.

[1] https://stackoverflow.com/questions/4694767/overriding-djangos-relatedmanager-methods#comment5221073_4695692
[2] https://docs.djangoproject.com/en/4.0/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

Elkasitu: That said, I updated the tests to use `add` and `set` with `bulk=False` to confirm that the validation works in that case too

JakubFrejlach: Wow, I haven't thought that this would be that complicated. So I would say that the best idea is probably to just not use the `set()` and `add()` for `Flaw.meta` and this would prevent missing the validation.

JakubFrejlach (approved):

osidb/models.py
osoukup: Good to think about it in the future. Just not to use those methods.

osoukup (approved):
LGTM. I agree that this is a Bugzilla related issue which should not exist in the future when we get rid of it. For now it can serve as an indicator of what data should be fixed in Bugzilla.

PR #22, opened 2022-08-03
Update Django to the latest security revision

Elkasitu:
See https://www.djangoproject.com/weblog/2022/aug/03/security-releases/
for more details


JakubFrejlach (approved):
LGTM! You are really lucky that you run into it, few hours earlier and it would be me :grin:

PR #23, opened 2022-08-03
Fix retry logic

juspence:
@osidb-dev getattr will return the requests.response object, and it will be evaluated as a boolean. If the response had a 200ish code, it evaluates to True. If the code was 400ish or similar, it evaluates to False.

So any failed request will not enter the if block I changed in the diff, and we skip checking whether the code is retryable or not. Instead we just retry all requests, which is dangerous.


Elkasitu (approved):
LGTM, you can drop the django change as it was already merged in #22

juspence:
> LGTM, you can drop the django change as it was already merged in #22

Thanks for review! Going to merge as-is - I think if I force-push again it will dismiss your review and you'll have to approve again.

PR #24, opened 2022-08-04
Remove pip-audit checks

Elkasitu:
After a brief discussion we informally decided to drop pip-audit usage,
the main argument being that whenever a vulnerability was discovered in
one of our packages our entire development process is halted -- CI
cannot pass as long as the pip-audit check is failing, so we must
immediately update to the latest version in order to be able to continue
normal development and often times the vulnerabilities don't even affect
us since we may not use the feature that the security patch fixes.

Since we develop on GitHub now, we can simply defer this job to
Dependabot which can alert us once a week (by default) of potential
security vulnerabilities present in our codebase and let developers
gauge whether it's urgent or not to apply the security patches, without
disrupting workflow.

For reference, Dependabot is highly customizable through a
dependabot.yaml file inside the .github directory of the repository:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,18 +15,6 @@ check:
   script:
     - dnf install -y openldap-devel python3.9
     - tox --parallel --parallel-live
-  except:
-    refs:
-      - schedules
-
-
-# TODO: There's a better way, will fix this tomorrow if I have time
-pip-audit:
-  stage: check
-  image: quay.io/prodsecdev/fedora-latest:35  # includes tox
-  script:
-    - dnf install -y python3.9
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM. I would guess that **pip-audit** and **dependabot** should be interchangeable. One should be enough. I agree that the **dependabot** is more flexible and comfortable for our use

JakubFrejlach (approved):
LGTM! Having two scanning utilities seems like a too much burden, dependabot should do just fine.

PR #26, opened 2022-08-04
Validate differences between RH CVSS and NVD CVSS

Elkasitu:
The Red Hat CVSS score can contain certain differences when compared to
the NVD CVSS score as long as these differences are within a certain
threshold.

The constraints are:

* A score difference of less than 1.0 point between the two
* No severity difference between the two

Closes OSIDB-333

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - unified logging across the whole OSIDB
 - validate hightouch and hightouch-lite flag value combinations (OSIDB-329)
+- validate differences between Red Hat and NVD CVSS score and severity (OSIDB-333)
 
 ## [2.1.0] - 2022-08-02
 ### Changed
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -3,6 +3,7 @@ draft model for end to end testing
 """
 import logging
 import uuid
+from decimal import Decimal
 from typing import Union
 
 from django.conf import settings
@@ -19,7 +20,7 @@ from apps.exploits.mixins import AffectExploitExtensionMixin
 from apps.exploits.query_sets import AffectQuerySetExploitExtension
 from apps.osim.workflow import WorkflowModel
 
-from .constants import OSIDB_API_VERSION
+from .constants import CVSS3_SEVERITY_SCALE, OSIDB_API_VERSION
 from .core import generate_acls
 from .mixins import NullStrFieldsMixin, TrackingMixin
 from .validators import (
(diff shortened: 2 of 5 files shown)


osoukup (approved):
LGTM. I mentioned in the Jira issue that we might think about this more with respect to the workflows etc. so please ignore it. This is great for now and if we need a different approach when we eventually implement the workflows we can easily evolve it - the validation test would be the same anyway, maybe called under some condition or whatever but let us leave it for later.

osidb/models.py
```diff
@@ -592,10 +593,54 @@ def __str__(self):
         """return cve_id when str(flaw)"""
         return str(self.cve_id)
 
+    def _validate_rh_nvd_cvss_score_diff(self):
+        """
+        Checks that the difference between the RH and NVD CVSS score is not >= 1.0
+        """
+        if not self.cvss3_score or not self.nvd_cvss3:
```
JakubFrejlach, line 600: I am not sure, if you should check the `cvss3_score` directly. It is a float value and potentially it can be equal to `0.0`, in that case you are skipping the check even if the scores exists. You should probably either check it for `is not None` or check rather `cvss3`.

osidb/models.py
```diff
@@ -592,10 +593,54 @@ def __str__(self):
...
+            raise ValidationError(
+                f"RH and NVD CVSSv3 score differs by 1.0 or more - "
+                f"RH {rh_cvss3_score} | NVD {nvd_cvss3_score}"
+            )
+
+    def _validate_rh_nvd_cvss_severity_diff(self):
+        """
+        Checks that NVD and RH CVSS are not of a different severity.
+        """
+        if not self.cvss3_score or not self.nvd_cvss3:
```
JakubFrejlach, line 616: Potentially same issue as above.

JakubFrejlach (commented):
One remark otherwise LGTM.

docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - unified logging across the whole OSIDB
+- validate differences between Red Hat and NVD CVSS score and severity (OSIDB-333)
```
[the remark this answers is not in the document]
osoukup: Good point :smile: I would actually be for providing those Jira IDs when the change is based on a Jira issue. It is true that for any external contributor/watcher it has no meaning (also no harm I think) but we still have majority of our focus internally, we share our CHANGELOG there with every release, and this may clarify the changes etc. WDYT?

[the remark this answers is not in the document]
Elkasitu: I agree with @osoukup I think it's a good idea to have it (we can do it retroactively at some point), it could help internal stakeholders to follow the CHANGELOG better

osidb/models.py
Elkasitu: Good catch, we probably should have an extra validation that checks that `cvss3_score` matches the value from `cvss3`, because at the moment if you create a flaw from the API you can provide one and not the other, or maybe it should be a computed value altogether... that said don't know if it's worth the effort or if we should just wait until we refactor CVSS into a model

osoukup: It should be ideally tied together and not independent so it can never desync. But maybe let's wait until the refactor.

JakubFrejlach (approved):

PR #27, opened 2022-08-05
Validate public source embargoed flaw

Elkasitu:
An embargoed Flaw's source should not be public, thus this commit adds a
validation check for this explicit purpose.

Closes OSIDB-337

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -213,7 +213,7 @@
         "filename": "osidb/models.py",
         "hashed_secret": "c7e672880d394aa5dd924e04465c986652ba7291",
         "is_verified": false,
-        "line_number": 153,
+        "line_number": 158,
         "is_secret": false
       }
     ],
@@ -236,5 +236,5 @@
       }
     ]
   },
-  "generated_at": "2022-08-04T15:46:15Z"
+  "generated_at": "2022-08-05T13:29:24Z"
 }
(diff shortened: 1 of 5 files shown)


osoukup (approved):
LGTM. Only one thing I realized is that as this all goes to `osidb/models.py` which is IMO already too big we should I think eventually split it. There are already some validators in `osidb/validators.py` so maybe we could make it like `FlawValidator` in there? I am just not sure it could use `self` any more or it would have to be supper class or something. We will see.

Elkasitu:
> LGTM. Only one thing I realized is that as this all goes to `osidb/models.py` which is IMO already too big we should I think eventually split it. There are already some validators in `osidb/validators.py` so maybe we could make it like `FlawValidator` in there? I am just not sure it could use `self` any more or it would have to be supper class or something. We will see.

We can, if file length is a problem there is one Django pattern that we don't use in any of our projects but might be a good idea to start adopting:

Instead of having all model definitions in `models.py` you can have a `models/` directory and have individual files per model or closely-related models, e.g.:

```
osidb/
├─ apps/
├─ osidb/
│  ├─ models/
│  │  ├─ flaw.py
│  │  ├─ affect.py
│  │  ├─ tracker.py
├─ collectors/
```

Takes more time but it's very well-organized, but shifting to `validators.py` is also fine

osoukup:
Mode-wise organization looks nice. I am all for it. The current `models.py` is already hard to order and navigate.

PR #28, opened 2022-08-05
Add monitoring email

juspence:
@RedHatProductSecurity/osidb-devs Quick change I worken on since it's Friday. Note that OSIDB_MAILING_LIST and RED_HAT_EMAIL_SERVER will need to be defined in OpenShift environment variables before this will work. Those values shouldn't be needed for Gitlab or local envs, though.


osoukup (approved):
LGTM. Nice stuff! This is what you use in SDEngine?

Elkasitu:
@osoukup probably should've written it herre but while talking to Justin we agreed to not merge it straight away as we currently have many false positive or known errors being triggered by the collectors, in which case activating this task would generate a lot of noise, so we agreed to revisit it once we have better monitoring of tasks with Flower

osoukup:
@Elkasitu makes sense. I would await quite unnecessarily extensive error volume now.

config/settings.py
```diff
@@ -25,6 +26,17 @@
     ".redhat.com",
 ]
 
+OSIDB_MAILING_LIST = os.getenv("OSIDB_MAILING_LIST")
```
JakubFrejlach, line 29: Nitpick: You might consider using the `osidb.helpers.get_env` helper function so we do have unified approach.

config/settings.py
```diff
@@ -25,6 +26,17 @@
...
+# Mail these people on uncaught exceptions that result in 500 errors
+ADMINS = [("OSIDB developers", OSIDB_MAILING_LIST)]
+
+# Email settings - override for specific environments as needed
+SERVER_EMAIL = f"OSIDB <{OSIDB_MAILING_LIST}>"
+DEFAULT_FROM_EMAIL = SERVER_EMAIL
+EMAIL_PORT = 25
+EMAIL_HOST = os.getenv("RED_HAT_EMAIL_SERVER")
```
JakubFrejlach, line 37: Same as above.

osidb/tasks.py
```diff
@@ -0,0 +1,72 @@
+import logging
+
+from django.conf import settings
+from django.core.mail import EmailMessage
+from django.utils import timezone
+from django_celery_results.models import TaskResult
+
+from config.celery import app
+
+logger = logging.getLogger(__name__)
```
JakubFrejlach, line 10: Since this module contains celery tasks I would use rather `celery.utils.log.get_task_logger` instead of the basic `logging.getLogger` so we can utilize the task ID and task name in the log message.

JakubFrejlach (approved):
Couple of suggestions, otherwise LGTM.

osoukup:
@juspence a lot has changed since this was created and I think it is no more relevant. We should probably rather use Splunk+SignalFX for the notifications instead of handling it inside the app. WDYT?

osoukup:
> > @juspence a lot has changed since this was created and I think it is no more relevant. We should probably rather use Splunk+SignalFX for the notifications instead of handling it inside the app. WDYT?
> 
> First of all, I completely forgot this was open so thank you for the reminder. My opinion would be that SignalFX is good for infra-level alerts (pod restarts). I don't think it will work for this kind of task-level alerting, unless you've already done work to set up custom metrics or have Flower sending some celery.failed_tasks metric to SignalFX.
> 
> Splunk might be a solution if every task is wrapped in some try / except block, for example that logs the exception in a f"Task failed: {exc}" message. Otherwise, a task that raises an exception and dies won't log anything to Splunk, it will just create a failed TaskResult in the DB. That's why the separate monitoring email (which queries the TaskResult table) is needed in our other apps.
> 
> Of course it's your project so how you implement is up to you. Overall, I would say let me know if you like this / think it's useful based on above, and have fixed most of the collector errors that Adrian mentioned. In that case I'll fix up Jakub's comments (thank you for the review), fix the conflicts, and merge this. Or if you would prefer a different approach, feel free to just close this instead.

@juspence (@RedHatProductSecurity/osidb-devs FYI) so I am not sure about the approach here and the bits you have done here seem to be at least partially valid. The thing is that in OSIDB we are not too interested in Celery task level failures but the collector run level instead (though mostly it is probably the same). Splunk might be able to catch this as on every collector failure there is a log record being created

https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/framework/models.py#L494

however this needs a deeper analysis. We can leave this open and see later when we get to implementing some more clever monitoring.

osidb/tasks.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+
+@app.task(autoretry_for=(Exception,), retry_backoff=900, retry_jitter=False)
+def email_failed_tasks():
+    """Send email about failed Celery tasks within past 24 hours to OSIDB admins (mailing list)
+    If it failed to send, try again after 15 minutes, then 30 minutes, then give up"""
+    failed_tasks_threshold = get_last_success_for_task("osidb.tasks.email_failed_tasks")
+    failed_tasks = TaskResult.objects.filter(
+        status__in=(celery_states.FAILURE, celery_states.RETRY),
```
Elkasitu, line 45: Could you remove the `RETRY` state from the filters? We generate a lot of those because a few collectors can't run in parallel, furthermore even for the cases where it'd make sense it'll either eventually succeed or fail and will be caught then.

Elkasitu (approved):
LGTM just some minor comments / questions, thanks for the contribution.

Could you squash your commits into one? Thanks!

jobselko:
@juspence flake8 should be fine now, just rebase your branch.

PR #29, opened 2022-08-05
CVE-less flaws

osoukup:
Start syncing CVE-less flaws + some smaller improvements. As discussed with @Elkasitu we will not sync the existing CVE-less flaws by a migration as it would block the service for too long. I will be done semi-manually using the Python shell - observe the last commit with specified batch collection support.

OSIDB-25


osidb/models.py
```diff
@@ -459,10 +459,18 @@ class FlawManager(models.Manager):
...
+        this is meant for cases when UUID is not available - collector
+        """
         try:
-            flaw = Flaw.objects.get(cve_id=cve_id)
+            flaw = (
+                # CVE-less flaws are identified by Bugzilla ID
+                Flaw.objects.get(meta_attr__bz_id=bz_id)
+                if cve_id is None
+                else Flaw.objects.get(cve_id=cve_id)
+            )
```
Elkasitu, line 473: I would instead set cve_id and bz_id to be optional and have an early check to confirm that at least one of them is set:

E.g.:

```python
# keep the signature the same as before, so that old code doesn't have to be changed
def create_flaw(cve_id=None, bz_id=None, **extra_fields):
    # only one case when they're both equal: (None, None)
    assert cve_id != bz_id, "Either cve_id or bz_id must be provided"
    ...
```

osidb/tests/test_flaw.py
```diff
@@ -150,6 +150,37 @@ def test_create(self, datetime_with_tz, good_cve_id):
...
+    def test_create_flaw_method(self):
+        acls = [
+            uuid.uuid5(
+                uuid.NAMESPACE_URL,
+                "https://osidb.prod.redhat.com/ns/acls#data-prodsec",
+            )
+        ]
+        Flaw.objects.all().delete()
+        Flaw.objects.create_flaw(
+            bz_id="12345",
```
Elkasitu, line 162: If we pass it `bz_id` and it doesn't exist, `create_flaw` should pass the `bz_id` value to `meta_attr` so that we don't have to explicitly pass both `bz_id` and `meta_attr__bz_id`

collectors/bzimport/convertors.py
```diff
@@ -640,12 +640,20 @@ def get_meta_attr(self, cve_id):
 
     def get_nvd_cvss2(self, cve_id):
         """get NVD CVSS2"""
-        if cve_id in self._nvd_cvss and "cvss2" in self._nvd_cvss[cve_id]:
+        if (
+            cve_id is not None
```
Elkasitu, line 644: Can `_nvd_cvss` contain flaws with `None` as `cve_id`? I feel like adding the `cve_id is not None` is not necessary

Elkasitu (changes requested):

osidb/models.py
```diff
@@ -820,7 +821,6 @@ def get_queryset(self):
     def fts_search(q):
         """full text search using postgres FTS via django.contrib.postgres"""
         fields_to_search = (
-            "cve_id",
```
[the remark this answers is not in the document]
osoukup: Because `Affect` does not have `cve_id` attribute any more. It used to have it but it was removed in https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/migrations/0034_remove_affect_cve_id.py - CVE is obviously a flaw attribute not an affect one.

osidb/tests/test_flaw.py
```diff
@@ -132,8 +132,8 @@ def test_create(self, datetime_with_tz, good_cve_id):
         assert meta1 in all_meta
         assert meta2 in all_meta
 
-        vuln_2 = Flaw.objects.create_flaw(
-            "CVE-1970-12345",
+        vuln_2 = Flaw(
```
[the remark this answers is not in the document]
osoukup: I agree. I just wanted to get rid of `create_flaw` here and did not go any further.

osidb/models.py
osoukup: I am not sure how you mean instead as even with this assertion I need to query for the flaw object which is done differently based on Bugzilla ID and CVE ID.

osidb/tests/test_flaw.py
osoukup: That makes sense. I will extend the `create_flaw` method with it.

collectors/bzimport/convertors.py
osoukup: I think that you are right. I was overly cautious but the CVSS map should only contain CVE keys only - no None key.

osidb/models.py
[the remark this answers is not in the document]
Elkasitu: OK I see, I somehow missed that this was for affect, the commit message could describe what you wrote

osidb/models.py
Elkasitu: yes you will still have to do that, I just mean that the method's signature / api is more stable (i.e. existing code does not need to change) and convenient (no need to explicitly pass `None` values for either `cve_id` or `bz_id`)

e.g.

if cve_id but no bz_id:
`Flaw.objects.create_flaw(None, "CVE-2022-1234", my_arg1="foo", my_arg2="bar")`
vs
`Flaw.objects.create_flaw("CVE-2022-1234", my_arg1="foo", my_arg2="bar")`

if bz_id but no cve_id:
`Flaw.objects.create_flaw("1234567", None, my_arg1="foo", my_arg2="bar")`
vs
`Flaw.objects.create_flaw(bz_id="1234567", my_arg1="foo", my_arg2="bar")`

Elkasitu (approved):
LGTM, there's just the ongoing conversation about the `create_flaw` signature but it's nothing major

Elkasitu:
@osoukup one thing I just remembered: CHANGELOG update? 😄

osoukup:
@Elkasitu regarding https://github.com/RedHatProductSecurity/osidb/pull/29#discussion_r941230549 it makes sense to me. I will update the method interface with the default `None` values and remove `None` parameters from the method calls.

osidb/models.py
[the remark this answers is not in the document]
osoukup: I will extend the commit message. Thank you for pointing it out!

JakubFrejlach (approved):

PR #30, opened 2022-08-07
Use generally available images

JakubFrejlach:
Use postgres, redis and openldap public images from docker.io

For postgres the new Dockerfile is created to handle importing of the SSL certificate, see the commit with the explanation.

Closes OSIDB-170


JakubFrejlach:
Pipeline fails due to secrets update, will fix that.

osoukup (changes requested):
LGTM and works with one detail which is **openldap** image (already discussed in person).

Elkasitu (commented):
I have added some very high level comments on one the changes introduced by the PR but haven't gone into much detail, mostly because I think that we shouldn't go through with it.

Originally when creating this task the goal was to make simple changes to make it so that both Red Hatters and non-Red Hatters could easily spin up an instance of OSIDB with all the bells and whistles working, but I envisioned for example just a few new environment variables or maybe changing a registry image for some docker/quay image for local dev only.

The reason why I'm not completely onboard with these changes is because the complexity of maintaining these files becomes quadratic, I don't think we'll have the resources to keep them in sync or even remember to have them synchronized, we'll probably eventually forget and they'll essentially become dead code.

As long as the main Dockerfile can be built by external sources (which should already be the case), then we can let external users / contributors simply set up the rest as they wish (either by cooking up their own docker-compose files or by setting up their own service files for pg, etc.).

I very much appreciate the effort you put into this, and I'd be willing to merge it anyway if we agreed to as a team, but ultimately I think it'd be a mistake to merge it.

osoukup:
@Elkasitu I understand your point and agree that this will probably double the image maintenance effort needed. I would however see having some easy option how to run OSIDB using public resources only as a must if the opensourcing should have real meaning - though I know that we probably won't have any real external contributors for quite some time. I would rather be for merging this and eventually finding ways how to unify public and RH approach so the differences are minimal and the maintenance effort lower. I agree that we might easily forget to do a change on both places when we now all use the RH images though otherwise I do not think that the maintenance will be needed very frequently. I am not insisting on anything but I more for going with this so we make the project more open. Otherwise having it would be close to impossible to run it for out-comers.

Dockerfile-public-image
```diff
@@ -0,0 +1,53 @@
+FROM registry.fedoraproject.org/fedora:36
```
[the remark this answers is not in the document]
JakubFrejlach: Ok, so this was caused by lack of my knowledge, I thought that for the UBI you need subscription too, so in that case, we can drop this change and in the end only the redis and postgres images would need to be different.

Also there were some additional rpm packages installed so the pip install wouldn't complain, but it doesn't matter anyway, since we can easily drop this change.

JakubFrejlach:
I am much aware of the complexity which this change brings into the project and agree that this would not be ideal. If we make decision that we want to drop this, I am ok with that. I would like to just point out that after your clarification that the UBI image is publicly accessible, which I thought it isn't, we can drop good portion of the changes and go with just the postgres and redis changes which are not so extensive.

Elkasitu:
@JakubFrejlach what would it look like without the UBI changes? would it still require different docker-compose files?

What if we only kept the publicly available postgres, redis (and maybe openldap or any others too)? The images would be different than those we use for deployment, but IMO that doesn't matter since we already have the complexity of maintaining two sets of "compose" files (local and ops). We just need to make sure that the major versions match, and we'd probably need some docker account but that is fine (we already need one for registry.redhat.io and another for quay.io)

Very little reason to keep using the RH images locally as long as we keep using them for deployments.

osoukup:
We could also make those images public in **quay** so staying in quay (no docker account needed).

Elkasitu:
That could work, but then if we want the latest security patches or to change version we need to do manual work or to set up some sort of mirroring + webhook to have it synced with upstream

Dockerfile-public-image-postgres
```diff
@@ -0,0 +1,15 @@
+FROM docker.io/library/postgres:13
+
+COPY etc/pg /pg
+
+# Prepare init SQL scripts
+RUN cp pg/local-dev*.sql /docker-entrypoint-initdb.d
+# Prepare potgres config
+RUN cp pg/postgresql-public-image.conf /etc/postgresql/postgresql.conf
```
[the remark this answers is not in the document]
JakubFrejlach: No, we don't need them on both place, so your suggestion is correct, will change that.

Dockerfile-public-image-postgres
```diff
@@ -0,0 +1,15 @@
...
+
+# Set correct ownership and permissions to cert and key it needs to be
+# done via dockerfile becuase simply mounting the files as volumes
+# with the postgres ownership tampers with other containers
+# See: https://stackoverflow.com/questions/55072221/deploying-postgresql-docker-with-ssl-certificate-and-key-with-volumes
+RUN chown postgres:postgres /pg/local-server.* && \
```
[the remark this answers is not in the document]
JakubFrejlach: Thanks! Forgot about this possibility.

Dockerfile-public-image
```diff
@@ -0,0 +1,53 @@
...
+
+LABEL summary="OSIcDB" \
```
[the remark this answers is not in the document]
JakubFrejlach: This part of the code will be removed.

osoukup (approved):
LGTM

docker-compose.test.yml
```diff
@@ -4,7 +4,7 @@ services:
 
     testldap:
       container_name: testldap
-      image: quay.io/prodsecdev/openldap:2.5.7-debian-10-r57
+      image: docker.io/bitnami/openldap:2.5.13-debian-11-r19
```
Elkasitu, line 7: maybe we should target `2.5`? from looking at https://hub.docker.com/r/bitnami/openldap/tags it seems like that's the least specific version that should cover any future security updates

postgres.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM docker.io/library/postgres:13
+
+# Set correct ownership and permissions to cert and key it needs to be
+# done via dockerfile becuase simply mounting the files as volumes
+# with the postgres ownership tampers with other containers
+# See: https://stackoverflow.com/questions/55072221/deploying-postgresql-docker-with-ssl-certificate-and-key-with-volumes
```
Elkasitu, line 6: I don't think we need to do this, in fact the stack overflow post mentions that this is only necessary for windows, in linux permissions are kept between the host and container systems and in the `generate_local_pg_tls_cert` `make` entrypoint we already set the permissions correctly (0640 for postgres user/group), if this is not working then maybe we need to set the UID/GUID explicitly for the cert and key, which is apparently 999 [1]

[1] https://stackoverflow.com/questions/55072221/deploying-postgresql-docker-with-ssl-certificate-and-key-with-volumes

Elkasitu (approved):
LGTM overall, but I think we can do simpler for the PG image

postgres.Dockerfile
JakubFrejlach: This is what I initially though that would be enough. But 0640 wasn't enough:

```2022-08-25 09:05:38.354 UTC [43] DETAIL:  File must have permissions u=rw (0600) or less if owned by the database user, or permissions u=rw,g=r (0640) or less if owned by root```

It didn't work with the root, so I changed to the 0600 and postgres (999). Setting the UID/GUID explicitly of the cert and key to 999 ultimately failed all the containers with the:

```Error: lsetxattr /home/jfrejlac/work/osidb/etc/pg/local-server.crt: operation not permitted```

So in the end, the only way which worked for me was setting the chmod and chown in the Dockerfile and this also didn't tampered with the other containers.

File: docker-compose.test.yml (hunk 5 above)
Comment (JakubFrejlach) [reply docker-compose.test.yml:7]: True, we should be more general. `2.5` or `2.5-debian...` though? :thinking:

File: postgres.Dockerfile (hunk 6 above)
Comment (Elkasitu) [reply postgres.Dockerfile:6]: Yeah that's a SELinux issue, I managed to fix it by doing `podman unshare chown 999:999 etc/pg/local-server.*`, I will share with you my local diff, should probably add it to the docs or do it as part of the generate local tls make entrypoint, and chmod to 0600

File: docker-compose.test.yml (hunk 5 above)
Comment (Elkasitu) [reply docker-compose.test.yml:7]: `2.5` should be fine

File: postgres.Dockerfile (hunk 6 above)
Comment (JakubFrejlach) [reply postgres.Dockerfile:6]: I added code from your local diff and used that command and it works. Added it also to the Makefile rule which handles the pg tls certs so the ownership and the persmissions should now be correctly set and everything works as it should. If you agree, I can finally merge this.

File: postgres.Dockerfile [hunk 7]
```diff
@@ -0,0 +1,12 @@
+FROM docker.io/library/postgres:13
```
Elkasitu, line 1: we can get rid of this file now

Elkasitu (approved):
LGTM, good job!

PR #31, opened 2022-08-08
Verify that public source flaws don't have ack FlawMetas

Elkasitu:
It doesn't make sense to include acknowledgments for Flaws emanating
from public sources as said sources already include the acknowledgment
information.

Closes OSIDB-338


osoukup (approved):
LGTM

PR #33, opened 2022-08-11
AlertMixin and small flake8 invocation improvement

Elkasitu:
This PR introduces two changes:

* The addition of `AlertMixin` which allows the usage of alerts on any model that inherits from said mixin.
* A small fix to `flake8` so that failures are clearer

Closes OSIDB-324

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -1,3 +1,5 @@
+from enum import Enum
+
 from django.db import models
 from django.utils import timezone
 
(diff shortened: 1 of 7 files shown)


osidb/mixins.py
```diff
@@ -79,3 +81,49 @@ def clean(self):
...
+    The way that this mixin works is simple, any model that inherits from this mixin
+    will have a field in which alerts are stored in JSON, this field is re-populated
+    on each save when the validations are run.
+
+    The mixin provides a helper function for creating said alerts, this also serves
+    as an abstraction layer to enforce a schema on the JSON field and guarantee that
+    the alerts are somewhat constant in their content.
+    """
+
+    _alerts = models.JSONField(default=dict)
```
osoukup, line 99: What about adding some alert getter? The underscored attribute would be an unusual interface.

Elkasitu: Don't think we need to, there should ideally be only one case in which the attribute is accessed for read: exposing over API, and in that case it would be wrapped in a getter at the API level (something like `ReadOnlyField`)

PR #35, opened 2022-08-18
Bugzilla flaw create and update PoC

osoukup:
This pull request introduces PoC of Bugzilla flaw create and update functionality (plus some other minor changes). It is not yet integrated into OSIDB as it is just a PoC but it was successfully tested to create and update flaws. I am sending it on the review rather earlier so it is not so much stuff to look into and we can shape it soon enough. It is far from being complete as it does not handle some Bugzilla flaw attributes fully or at all. It can be tested in the OSIDB shell which something like

```
from osidb.models import Flaw
from apps.bbsync.save import BugzillaSaver

f = Flaw.objects.filter(cve_id__isnull=False).first()
bs = BugzillaSaver(f)
flaw = bs.create()
flaw.bz_id  # ID of newly created flaw in Bugzilla

f.title = "cool title"
bs = BugzillaSaver(f)
bs.save()
```

**Always switch to Bugzilla stage when testing write operations!**

Closes OSIDB-367

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ tests/htmlcov
 /venv
 .vscode/
 *.db.gz
+osidb_data_backup_dump*
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -1,0 +1,190 @@
+from collectors.bzimport.constants import ANALYSIS_TASK_PRODUCT
+from osidb.models import Flaw, FlawImpact
+
+
+class BugzillaQueryBuilder:
+    """
+    Bugzilla flaw bug query builder
+    to generate general flaw save query
+
+    https://bugzilla.redhat.com/docs/en/html/api/index.html
+    """
+
+    def __init__(self, flaw, old_flaw=None):
+        """
+        init stuff
+        parametr old_flaw is optional as there is no old flaw on creation
+        and if not set we consider the query to be a create query
+        """
+        self.flaw = flaw
+        self.old_flaw = old_flaw
+        self._query = None
+
+    @property
+    def query(self):
+        """
+        query getter shorcut
+        """
+        if self._query is None:
+            self.generate()
+
+        return self._query
+
+    @property
+    def creation(self):
+        return self.old_flaw is None
+
+    def generate(self):
+        """
+        generate query
+        """
+        self.generate_base()
+        self.generate_unconditional()
+        self.generate_description()
+        self.generate_resolution()
+        self.generate_alias()
+        self.generate_keywords()
+        self.generate_flags()
+        self.generate_groups()
+        self.generate_cc()
+        # TODO placeholder + has different groups
+        # TODO SRT notes
+        # TODO tracker links
+        # TODO prestage eligable date - deprecate
+        # TODO checklists
+        # TODO fixed_in
+        # TODO dupe_of
+        # TODO cf_devel_whiteboard
+        # TODO ARRAY_FIELDS_ON_CREATE = ("depends_on", "blocks", "cc", "groups", "keywords")
+        # TODO auto-requires doc text on create
+
+    def generate_base(self):
+        """
+        generate static base of the query
+        """
+        self._query = {
+            "product": ANALYSIS_TASK_PRODUCT,
+            "component": "vulnerability",
+            "op_sys": "Linux",
+            "platform": "All",
+            "version": "unspecified",
+        }
+
+    IMPACT_TO_SEVERITY_PRIORITY = {
+        FlawImpact.CRITICAL: "urgent",
+        FlawImpact.IMPORTANT: "high",
+        FlawImpact.MODERATE: "medium",
+        FlawImpact.LOW: "low",
+        FlawImpact.NOVALUE: "unspecified",
+    }
+
+    def generate_unconditional(self):
+        """
+        generate query attributes not requiring conditional processing
+        """
+        self._query["summary"] = self.flaw.title  # TODO handle prefixes
+        self._query["cf_release_notes"] = self.flaw.summary
+        self._query["status"] = self.flaw.state
+        self._query["severity"] = self.IMPACT_TO_SEVERITY_PRIORITY[self.flaw.impact]
+        self._query["priority"] = self.IMPACT_TO_SEVERITY_PRIORITY[self.flaw.impact]
+
+    def generate_description(self):
+        """
+        generate query for flaw description on create
+        """
+        if self.creation:
+            self._query["description"] = self.flaw.comments.first().text
+            self._query["comment_is_private"] = False
+            # TODO
+            # self._query["comment_is_private"] = True if ... else False
+
+    def generate_resolution(self):
+        """
+        generate resolution query
+        if status is CLOSED
+        """
+        if self.flaw.state == Flaw.FlawState.CLOSED:
+            self._query["resolution"] = self.flaw.resolution
+
+    def generate_alias(self):
+        """
+        generate add or remove CVE alias query
+        conditionally based on the changes and create|update
+        """
+        if self.creation:
+            if self.flaw.cve_id is not None:
+                # create query requires pure list
+                self._query["alias"] = [self.flaw.cve_id]
+
+        elif self.flaw.cve_id != self.old_flaw.cve_id:
+            self._query["alias"] = {}
+
+            if self.flaw.cve_id is not None:
+                self._query["alias"]["add"] = [self.flaw.cve_id]
+
+            if self.old_flaw.cve_id is not None:
+                self._query["alias"]["remove"] = [self.old_flaw.cve_id]
+
+    def generate_keywords(self):
+        """
+        generate keywords query based on creation|update
+        """
+        self._query["keywords"] = (
+            ["Security"] if self.old_flaw is None else {"add": ["Security"]}
+        )
+
+    def generate_flags(self):
+        """
+        generate query for Bugzilla flags
+        """
+        self._query["flags"] = []
+        # TODO needinfo and other flags
+        # hightouch | hightouch-lite | nist_cvss_validation | requires_doc_text
+
+    EMBARGOED_GROUPS = ["qe_staff", "security"]
+    DEADLINE_FORMAT = "%Y-%m-%d"
+
+    def generate_groups(self):
+        """
+        generate query for Bugzilla groups
+        which control the access to the flaw
+        plus eventually set deadline
+        """
+        # TODO groups handling is more complicated and involves
+        # product definitions processing but let us do it simple now
+
+        if self.flaw.embargoed:
+            if self.creation:
+                self._query["groups"] = self.EMBARGOED_GROUPS
+
+            if self.flaw.unembargo_dt:
+                self._query["deadline"] = self.flaw.unembargo_dt.strftime(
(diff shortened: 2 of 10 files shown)


apps/bbsync/save.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        return self.create() if self.flaw.bz_id is None else self.update()
+
+    def create(self):
+        """
+        create flaw in Bugilla
+        """
+        bqb = BugzillaQueryBuilder(self.flaw)
+        response = self.bz_conn.createbug(bqb.query)
+        self.flaw.meta_attr["bz_id"] = response.id
```
Elkasitu, line 41: I would hope that this doesn't change between the syncs, seems odd to do it

Elkasitu (changes requested):
See comments

collectors/bzimport/collectors.py
```diff
@@ -31,20 +31,8 @@
...
-    # 20 is the default but we set the limit to the maximum
-    # by explicitely giving it 0 value and the maximum is 1000
-    # this enhances the queries significantly
-    BZ_API_LIMIT = 1000
-
-    #######################
-    # BUGZILLA CONNECTION #
-    #######################
+class BugzillaConnector:
+    """Bugzilla connection handler"""
```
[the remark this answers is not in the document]
osoukup: Why the connector got split out of the querier? The querier has a lot of functionality that is not needed for create or update and just the connection is what is shared.

collectors/bzimport/collectors.py
```diff
@@ -38,7 +41,7 @@ class BugzillaConnector:
 
     def create_bz_conn(self) -> Bugzilla:
         """create Bugzilla connection"""
-        bz_conn = bugzilla.Bugzilla(url=BZ_URL, api_key=BZ_API_KEY, force_rest=True)
+        bz_conn = bugzilla.Bugzilla(url=BZ_URL, api_key=BZ_API_KEY)
```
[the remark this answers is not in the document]
osoukup: Fair point point. I have not noticed the deprecation note. I was getting `HTTP 400 Bad request` for create requests (not sure I tested update at this stage) which just worked flawlessly with XMLRPC and looking into `python-bugzilla` did not give me an answer what is the difference. I am not sure about the details. Probably better to investigate further and try to use REST.

.gitignore
```diff
@@ -30,3 +30,4 @@ tests/htmlcov
 /venv
 .vscode/
 *.db.gz
+osidb_data_backup_dump*
```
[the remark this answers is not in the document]
osoukup: The backup however also creates the files `osidb_data_backup_dump_git_commit_status.txt` and `osidb_data_backup_dump_known_migrations.txt`. Maybe I should have just replaced that line and not add a new one. I have no preference here but I think all the three files generated should be ignored by git.

apps/bbsync/save.py
osoukup: This is the creation so there is no Bugzilla ID before. I am not yet sure how will the save logic work and of course it can be done by the sync but for now it makes sense to me to update the flaw model with the new Bugzilla ID.

.gitignore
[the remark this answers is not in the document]
Elkasitu: ah right, I don't use those so I only have one `.db.gz` laying around, disregard then

collectors/bzimport/collectors.py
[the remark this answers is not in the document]
Elkasitu: ok, I somehow missed that it was being used elsewhere

osoukup:
@Elkasitu @JakubFrejlach would you mind another round of the review? This time there is less stuff as with `enforce_rest=True` there was no need to renew the cassettes and change the Bugzilla input handling.

apps/bbsync/save.py
```diff
@@ -0,0 +1,59 @@
...
+        returns an updated flaw instance (without saving)
+        """
+        return self.create() if self.flaw.bz_id is None else self.update()
+
+    def create(self):
+        """
+        create flaw in Bugilla
+        """
+        bqb = BugzillaQueryBuilder(self.flaw)
+        response = self.bz_conn.createbug(bqb.query)
```
JakubFrejlach, line 40: If `self.bz_conn.createbug(...)` fails for whatever reason, exception is raised inside or should we check the actual response? Does it include the status code, error message, etc.?

apps/bbsync/save.py
```diff
@@ -0,0 +1,59 @@
...
+        # and save back to Bugzilla so let us restrict this for now
+        if Flaw.objects.filter(meta_attr__bz_id=self.flaw.bz_id).count() > 1:
+            raise self.UnsaveableFlaw(
+                "Unable to save a flaw with multiple CVEs to Bugzilla "
+                "due to an ambigous N to 1 OSIDB to Buzilla flaw mapping"
+            )
+
+        old_flaw = Flaw.objects.get(uuid=self.flaw.uuid)
+        bqb = BugzillaQueryBuilder(self.flaw, old_flaw)
+        self.bz_conn.update_bugs([self.flaw.bz_id], bqb.query)
```
JakubFrejlach, line 58: Same question as with `self.bz_conn.createbug(...)` above.

apps/bbsync/save.py
```diff
@@ -0,0 +1,59 @@
...
+        which calls create or update handler to continue
+        returns an updated flaw instance (without saving)
+        """
+        return self.create() if self.flaw.bz_id is None else self.update()
+
+    def create(self):
+        """
+        create flaw in Bugilla
+        """
+        bqb = BugzillaQueryBuilder(self.flaw)
```
JakubFrejlach, line 39: I would be for some more descriptive name, `bz_query_builder` maybe?

apps/bbsync/save.py
```diff
@@ -0,0 +1,59 @@
...
+        # TODO flaws with multiple CVEs are non-trivial to update
+        # and save back to Bugzilla so let us restrict this for now
+        if Flaw.objects.filter(meta_attr__bz_id=self.flaw.bz_id).count() > 1:
+            raise self.UnsaveableFlaw(
+                "Unable to save a flaw with multiple CVEs to Bugzilla "
+                "due to an ambigous N to 1 OSIDB to Buzilla flaw mapping"
+            )
+
+        old_flaw = Flaw.objects.get(uuid=self.flaw.uuid)
+        bqb = BugzillaQueryBuilder(self.flaw, old_flaw)
```
JakubFrejlach, line 57: Same as above.

apps/bbsync/save.py
JakubFrejlach, line 41: Shouldn't we save the flaw afterwards so that the `bz_id` is not lost?

apps/bbsync/query.py
```diff
@@ -0,0 +1,184 @@
...
+        #
+        # - we need to store them on flaw fetch if we
+        #   want to do something more clever about them
+
+    def generate_cc(self):
+        """
+        generate query for CC list
+        """
+        # TODO now just empty CC list
+        self._query["cc"] = [] if self.old_flaw is None else {}
```
JakubFrejlach, line 184: This is some sort of a BZ magic that upon creation the `list` is expected and otherwise empty `dict`? :smile:

apps/bbsync/save.py
osoukup: I can provide a longer name. As it is just used here and on the next line I did not consider it necessary when the meaning can be read from the class name but no problem to put it there again :smile_cat:

apps/bbsync/save.py
osoukup: The idea is that the Bugzilla save is just one step of the general `flaw.save` so I wanted to leave the rest to be done there so this really relates to Bugzilla only. Otherwise it would be rather a side effect. One option is even to do the save by syncing the changed data back from Bugzilla which would be done completely separately.

apps/bbsync/query.py
osoukup: That is rather my mistake :laughing:

Elkasitu (approved):
Some small remarks and nitpicks, but overall LGTM

apps/bbsync/query.py
```diff
@@ -0,0 +1,184 @@
...
+            # TODO
+            # self._query["comment_is_private"] = True if ... else False
+
+    def generate_resolution(self):
+        """
+        generate resolution query
+        if status is CLOSED
+        """
+        if self.flaw.state == Flaw.FlawState.CLOSED:
+            self._query["resolution"] = self.flaw.resolution
```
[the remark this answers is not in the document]
osoukup: That should be 1:1 - just checking it and the values are uppercase in Bugzilla and we have all covered.

apps/bbsync/save.py
osoukup: There is not much info in the method documentation but it calls some validations which may definitely raise `BugzillaError` if the arguments are in the bad format (which should hopefully not happen after it is once correct) and there can definitely come errors from Bugzilla API either when the provided data are incorrect or there is some internal/processing error and also there can be errors like timeout I guess. Either case it would just quit the creation and float away to the caller. Right now I am not entirely sure how the caller will look and where is the correct place for addressing the errors but let it just propagate back to the API (in case of user call) might be a valid approach. To the return value it is always the created `Bug` instance so there is no need to check for status code or error message and so. I would leave it as it is for now and when we actually start calling it we will know better.

apps/bbsync/query.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        self._query["summary"] = self.flaw.title  # TODO handle prefixes
+        self._query["cf_release_notes"] = self.flaw.summary
+        self._query["status"] = self.flaw.state
+        self._query["severity"] = self.IMPACT_TO_SEVERITY_PRIORITY[self.flaw.impact]
+        self._query["priority"] = self.IMPACT_TO_SEVERITY_PRIORITY[self.flaw.impact]
+
+    def generate_description(self):
+        """
+        generate query for flaw description on create
```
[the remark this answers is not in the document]
osoukup: I think that the description (and any other comment) cannot be edited through API - the documentation I do not see it mentioned at all https://bugzilla.redhat.com/docs/en/html/api/core/v1/bug.html#update-bug. SFM2 also does not support any description/comment editing. I think the way is still just to go to Bugzilla and do it manually.

apps/bbsync/query.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        if self.flaw.state == Flaw.FlawState.CLOSED:
+            self._query["resolution"] = self.flaw.resolution
+
+    def generate_alias(self):
+        """
+        generate add or remove CVE alias query
+        conditionally based on the changes and create|update
+        """
+        if self.old_flaw is None:
```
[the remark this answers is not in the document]
osoukup: Good point! I will think about something more clever :smiley:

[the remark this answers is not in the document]
osoukup: I am introducing query `creation` so the logic is more clear. However after also abstracting it to `is_draft` here the draft terminology at this place seems to me unnecessarily abstract. The change here is not specifically tied to a draft status as there may be just CVE-X changed to CVE-Y without affecting the draft status. The change is really only about the CVE itself so I see the draft thing complicating the logic here.

PR #37, opened 2022-09-01
fix bug that Major Incident can be unset by unrelated BZ flag

osoukup:
The bogus convertor logic according to Bugzilla flag handling was just setting **is_major_incident** to either True or False on every Bugzilla flag processing which obviously wrong. We should only set it according to **hightouch** and **hightouch-lite** flags.

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -902,6 +902,8 @@ class FlawBugConvertor:
                     )
                 )
 
+        flaw.is_major_incident = False
+
         for flag in self.flags:
             flag_name = flag["name"]
             if flag_name not in self.FLAGS_META:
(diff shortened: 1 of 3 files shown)


Elkasitu:
@osoukup should target release-2.2.0 branch

osoukup:
> 

Shouldn't it rather first go to **master** and if it successfully get there be also brought to 2.2.0 ?

osoukup:
@Elkasitu I see, one less PR sounds nice :smile: I changed the base here and as **master** and release branch point to the same commit now there is no change

Elkasitu:
@osoukup just remembered, let's update the CHANGELOG

osoukup:
@Elkasitu thank you for the reminder! I am adding the CHANGELOG record

PR #38, opened 2022-09-02
Prevent libpq from attempting GSSAPI authentication

Elkasitu:
As it turns out, postgresql 12 introduced support for secure GSS
communication between libpq and pg over TCP/IP if libpq is compiled with
GSSAPI support through an option called gssencmode. This option, by
default, is set to "prefer", meaning that if libpq detects GSSAPI
credentials in the system it's running on, it will attempt to connect to
the database via encrypted GSSAPI connection.

Since we do not have GSSAPI connection properly set up between the OSIDB
service pod and the database pod, this will fail, always, leading to
wasted resources on every request to the database.

https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-CONNECT-GSSENCMODE

Closes OSIDB-424

### DIFF
--- a/config/settings_ci.py
+++ b/config/settings_ci.py
@@ -64,6 +64,8 @@ DATABASES = {
         "HOST": "postgres",
         "PORT": "5432",
         "ENGINE": "psqlextra.backend",
+        # prevent libpq from automatically trying to connect to the db via GSSAPI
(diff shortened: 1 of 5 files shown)


osoukup (approved):
This was a tricky one! LGTM

PR #41, opened 2022-09-07
Loosen the Red Hat vs NVD CVSSv3 validations

Elkasitu:
This commit converts both the RH vs NVD score difference and severity bound difference validations to no longer raise an error on save but instead create an alert of the respective object.

After some consideration and data analysis we came to the conclusion that roughly 7000 flaws currently exist and are currently being created with such differences (thus not a difference between legacy vs new data) which means that it's a valid use-case albeit an uncommon one.

Fixes OSIDB-426
Fixes OSIDB-427


osoukup (approved):
LGTM

PR #42, opened 2022-09-07
Improve memory footprint for flaw linking

skontar:
When EPSS objects are linked with flaws, intermediate map needs 500 MB of RAM. This change reduces memory footprint to ~1 MB.

### DIFF
--- a/apps/exploits/helpers.py
+++ b/apps/exploits/helpers.py
@@ -8,15 +8,21 @@ from osidb.models import Flaw
 def update_objects_with_flaws(exploit_objects):
     """
     Take a list of exploit like objects (Exploit or EPSS) which need to be linked with Flaw
-    objects. Make a one optimized query to the database instead of N queries (~180k for EPSS).
-    Create a link to Flaw object if it exists for the relevant CVE, otherwise set it to None.
+    objects. Create a link to Flaw object if it exists for the relevant CVE, otherwise set it to
+    None.
+
+    Optimizations:
+    * Make a one optimized query to the database instead of N queries (~180k for EPSS).
+    * Store just uuid (primary key) instead of whole object in the intermediate map, that saves
+      about 500 MB of RAM in case of EPSS and makes it faster
+    * Linking is done using 'flaw_id' instead of 'flaw' attribute, which accepts Flaw primary key
+      (uuid) instead of Flaw object
     """
     cves = set(exploit.cve for exploit in exploit_objects)
-    flaws = Flaw.objects.filter(cve_id__in=cves)
-    flaw_map = {flaw.cve_id: flaw for flaw in flaws}
-
+    flaws = Flaw.objects.filter(cve_id__in=cves).values("cve_id", "uuid")
+    flaw_map = {flaw["cve_id"]: flaw["uuid"] for flaw in flaws}
     for one_object in exploit_objects:
-        one_object.flaw = flaw_map.get(one_object.cve, None)
+        one_object.flaw_id = flaw_map.get(one_object.cve, None)
 
 
 def store_or_update_exploits(exploit_objects):


kgrant-rh (approved):
Looks good to me. Do I understand correctly that this is only used for transient modification of `exploit_objects`? So any changes that needed to be persisted to the database would be handled somewhere else?

skontar:
> Looks good to me. Do I understand correctly that this is only used for transient modification of `exploit_objects`? So any changes that needed to be persisted to the database would be handled somewhere else?

The linking is done for all exploit-like models before they are stored / updated in the database. For example see [EPSS collector](https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/epss/tasks.py#L60-L64) or [CISA collector](https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/exploits_cisa/tasks.py#L50-L54). As far as I know this does exactly the same thing as before, just both faster and using less memory.

apps/exploits/helpers.py
```diff
@@ -8,15 +8,21 @@
...
     """
     cves = set(exploit.cve for exploit in exploit_objects)
-    flaws = Flaw.objects.filter(cve_id__in=cves)
-    flaw_map = {flaw.cve_id: flaw for flaw in flaws}
-
+    flaws = Flaw.objects.filter(cve_id__in=cves).values("cve_id", "uuid")
+    flaw_map = {flaw["cve_id"]: flaw["uuid"] for flaw in flaws}
     for one_object in exploit_objects:
-        one_object.flaw = flaw_map.get(one_object.cve, None)
+        one_object.flaw_id = flaw_map.get(one_object.cve, None)
```
Elkasitu, line 25: According to the `values_list` documentation [1] you can request a flat list of all flaws' `uuid` and then use `get` to retrieve from said list the value for the object with `cve_id=one_object.cve`, it's kinda buried in the `values_list` documentation but I think it could help us get rid of the intermediate `flaw_map`:

```python
flaws = Flaw.objects.filter(cve_id__in=cves).values_list("uuid", flat=True)
for one_object in exploit_objects:
    one_object.flaw_id = flaws.get(cve_id=one_object.cve, None)
```
I know you told me you tried `values_list` first and wasn't helpful, but just double checking in case we can spare a few MB from building the map.

[1] https://docs.djangoproject.com/en/3.2/ref/models/querysets/#values-list

Elkasitu (approved):
LGTM

apps/exploits/helpers.py
Elkasitu: Though I must say I do not know how Django does this, it might be performing an extra query against the database which would be one step forwards and one backwards

skontar: **Original** (build map of objects)
```
    flaws = Flaw.objects.filter(cve_id__in=cves)
    flaw_map = {flaw.cve_id: flaw for flaw in flaws}
    for one_object in exploit_objects:
        one_object.flaw = flaw_map.get(one_object.cve, None)
```
_Memory: 533MB
Time: 10s_

**Current** (build map of tuples)
```
    flaws = Flaw.objects.filter(cve_id__in=cves).values("cve_id", "uuid")
    flaw_map = {flaw["cve_id"]: flaw["uuid"] for flaw in flaws}
    for one_object in exploit_objects:
        one_object.flaw_id = flaw_map.get(one_object.cve, None)
```
_Memory: 46MB
Time: 3s_

**Proposed** (+small changes, `get` does not have second parameter)
```
    flaws = Flaw.objects.filter(cve_id__in=cves).values_list("uuid", flat=True)
    for one_object in exploit_objects:
        try:
            one_object.flaw_id = flaws.get(cve_id=one_object.cve)
        except Flaw.DoesNotExist:
            one_object.flaw_id = None
```
_Not finished after half an hour, based on progress it would probably take 50 hours to complete_

I think it is so slow because it performs ~180k queries and is mostly equivalent to a naive solution I tried to avoid:
```
    for one_object in exploit_objects:
        one_object.flaw = Flaw.objects.filter(cve_id=one_object.cve).first()
```
...which actually finished
_Memory: 345MB
Time: 254s_

PR #44, opened 2022-09-15
Handle FlawSource that are both public and private differently

Elkasitu:
This commit changes the way existing and future validations based on
sources work, by providing an extra `is_private` method on FlawSource we
can now filter sources that are both public and private, and in those
cases the better course of action is probably to create an alert and not
raise any ValidationErrors.

Closes OSIDB-450


JakubFrejlach (approved):
LGTM.

osoukup (changes requested):
I do not think that this is correct. The sources in both categories are there for a reason. Let us look at a different validation (or maybe a workflow step but that does not make a difference in the source classes) https://issues.redhat.com/browse/OSIDB-339. It checks that every private source has an acknowledgement. However these amphibious sources does not need to have acknowledgements in the cases when considered public. If you just put them to the private class they will be always required to have the acknowledgement which is incorrect.

osidb/models.py
```diff
@@ -699,10 +717,17 @@ def _validate_embargoed_source(self):
         embargoed = self.acl_read == [
             uuid.UUID(acl) for acl in generate_acls([settings.EMBARGO_READ_GROUP])
         ]
-        if embargoed and FlawSource(self.source).is_public():
-            raise ValidationError(
-                f"Flaw is embargoed but contains public source: {self.source}"
-            )
+        if embargoed and (source := FlawSource(self.source)) and source.is_public():
+            if source.is_private():
+                self.alert(
```
osoukup, line 722: This is probably OK for now but in the future we should make the alerts actionable so people do not start to ignore them - when there is nothing one can do to suppress the alert they would just stop looking at them.

osoukup (approved):
LGTM. Thank you for the modifications!

osidb/models.py
Elkasitu: What's your idea of actionable? Maybe the alert could instead be some sort of "info" instead of a warning

osoukup: Here the only thing I can think about is the checkmark of some sort but maybe we could later solve it by converting into a workflow step. I would just leave it for now and think about this more generally in the future.

PR #47, opened 2022-09-20
Helper for manual flaw synchronization

Elkasitu:
This PR introduces a new custom Django management command which will hopefully be useful during live debugging sessions / production interventions / data fixes.

This command takes in an arbitrary amount of Bugzilla IDs as positional arguments and will sync them one by one, reporting successes and errors along the way.

It also makes some minor changes to the local compose file and adds an extra soft-dependency for the redis python client.

Closes OSIDB-389


osoukup (approved):
Looks handy! Nice improvements

PR #48, opened 2022-09-20
fix incorrect errata collector design

osoukup:
errata collector was implemented as two independent collectors one for the initial sync and the other for the periodic one which however results in two independent CollectorMetadata instances and as a result the data completeness context was not shared between the two and based on this the periodic run never started to run after the initial one as it kept thinking that the initial has not run yet despite it actually has

Fixes OSIDB-433

### DIFF
--- a/collectors/errata/tasks.py
+++ b/collectors/errata/tasks.py
@@ -20,12 +20,17 @@ logger = get_task_logger(__name__)
 
 
 @collector(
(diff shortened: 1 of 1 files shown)


Elkasitu (approved):
LGTM, do we maybe need to do some CollectorMetadata cleanup in stage/prod? as a migration maybe?

PR #50, opened 2022-09-21
Add php-pear to package exclusion list

skontar:
"php-pear" package needs to be added to an exclusion list until better mechanism is created.


osoukup (approved):
LGTM

PR #52, opened 2022-09-22
squash migrations

osoukup:
This is a prerequisite to https://github.com/RedHatProductSecurity/osidb/pull/49. I squashed all possible migrations (except the exploit ones as I do not touch this as Standa kinda owner). This is a first step to clean them up. Only after the release we can remove the old migrations. The squash was definitely not an automatic process. There was a lot of manual work included. The automatic output was more than 5k lines which I then manually sqaushed to 1.5k omitting unnecessary or later removed steps.


Elkasitu (commented):
LGTM but I would suggest doing the following to formally verify that the squashed migration is correct:

1. Delete all migrations in a separate branch
2. Call `makemigrations`
3. Compare by diff the single migration file generated by `makemigrations` with the squashed migrations file

osoukup:
> LGTM but I would suggest doing the following to formally verify that the squashed migration is correct:
> 
>     1. Delete all migrations in a separate branch
> 
>     2. Call `makemigrations`
> 
>     3. Compare by diff the single migration file generated by `makemigrations` with the squashed migrations file

I was running `makemigrations` on top of the squashed ones until it stopped adding or removing stuff so I considered it as a reference of equality but of course I can try the clean state and compare.

osidb/migrations/0001_initial_squashed_0059_flaw__alerts.py
```diff
@@ -0,0 +1,1494 @@
...
+                (
+                    "jira_user_id",
+                    models.CharField(blank=True, max_length=100),
+                ),
+            ],
+        ),
+        ######################
+        # ROW LEVEL SECURITY #
+        ######################
+        migrations.RunSQL(
```
JakubFrejlach, line 1361: Just a thought: since this SQL setting the RLS is irreversible the whole migration would be irreversible too which might not be convenient when playing around with migrations for whatever reason. Shouldn't we make this reversible? It can be done by supplying the argument `reverse_sql=migrations.RunSQL.noop` to `migrations.RunSQL`.

osidb/models.py
```diff
@@ -350,7 +350,9 @@ class FlawHistoryResolution(models.TextChoices):
     )
 
     # flaw severity, from srtnotes "impact"
-    impact = models.CharField(choices=FlawImpact.choices, max_length=20, blank=True)
+    impact = models.CharField(
+        choices=FlawImpact.choices, max_length=20, blank=True, default=""
```
JakubFrejlach, line 354: Nitpick: Would rather use `FlawImpact.NOVALUE`.

osidb/models.py
```diff
@@ -381,15 +387,19 @@ class FlawHistoryResolution(models.TextChoices):
 
     # , from srtnotes "mitigate"
     mitigated_by = models.CharField(
-        choices=FlawMitigate.choices, max_length=10, blank=True
+        choices=FlawMitigate.choices, max_length=10, blank=True, default=""
```
JakubFrejlach, line 390: This will fail the validation, since we do not have `""` as an option in the  `FlawMitigate.choices`.

osidb/models.py
```diff
@@ -953,14 +973,20 @@ class AffectType(models.TextChoices):
     ps_component = models.CharField(max_length=100)
 
     # from srtnotes affects/impact
-    impact = models.CharField(choices=AffectImpact.choices, max_length=500, blank=True)
+    impact = models.CharField(
+        choices=AffectImpact.choices, max_length=20, blank=True, default=""
```
JakubFrejlach, line 977: Rather `AffectImpact.Novalue`?

JakubFrejlach (changes requested):
Actually I am kinda confused right now, looking for example at the `Flaw.impact` treating in SFM2, we do not allow the empty value (None) anymore (the validations won't allow it so the only way is to force it through BZ), so why is the default necessary? :thinking: blank is allowed so the Flaws with missing impact values will be imported correctly but do we want the behavior that newly created flaws would get the empty `impact` by default? This thought also applies to some other attributes (eg. source, etc.). I might be missing some bigger picture, but this feels weird to me.

osidb/migrations/0001_initial_squashed_0059_flaw__alerts.py
osoukup: Good point! I will make add that reverse noop

osidb/models.py
osoukup: Good point

osidb/models.py
osoukup: That is true. Then I wonder why don't we have a `null=True` there

osidb/models.py
osoukup: Good point again

osidb/models.py
JakubFrejlach: Because of this https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.Field.null

That was one of the main points of reviewing the required fields. Basically you don't want to have `null=True` on string based fields (Char/Text) since then you have the two possible representations of the empty values (empty string vs. NULL/None)

osoukup: Then the default is needed as the value is obviously not mandatory so providing none should not result in a failure. Which means that the problem is not here but at the model. And as I look at it I am sure that missing empty value is not the only issue there. The real issue is that `mitigated_by` is multiple choice field and not just char. This might be probably wrong from the beginning.

osoukup:
@Elkasitu I removed all the migrations and generated them from scratch to compare with the squashed version. The result is that the freshly generated ones are low quality subset of the squashed ones. One thing is that they are sorted in a very dummy way so for example there is a lot of relations done in Add and not Create step as during the creation the related model does not exist yet. Second thing is that they are both differently formatted even though run through the same black command which is something I cannot explain but I did not dig in very much. The from-scratch-migration also miss unique together which is weird. Of course it miss the row level security. Otherwise they are functionally equivalent.

osidb/models.py
osoukup: Actually we are commenting the model and not the migration :smile: so yeah it is wrong but the default is not the main issue here.

osoukup: I actually recalled this https://issues.redhat.com/browse/OSIDB-126 so I will remove the field instead.

osoukup: @Elkasitu @JakubFrejlach this will make the next OSIDB release basically a breaking API change as `mitigated_by` will disappear from the API...

Elkasitu: @osoukup please make sure that the removal is N-1 compatible, even if the field is of no use we don't know who is using it and how, and then we should make a post to osidb-announce mentioning that it's being deprecated / removed

Elkasitu: or just keep the field and we will remove it during the next major version release

Elkasitu: there is probably a way to mark it as deprecated in the OpenAPI schema

JakubFrejlach: Oh, now I remember, this is the problematic multiplevalued field. Good that we are free to remove it.

This seems like the way how to mark [field deprecated](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema_serializer)

osoukup:
@Elkasitu @JakubFrejlach so please one more look. I created separated migration to remove `mitigated_by` field and put it before the squash so it can be squashed too. Then I reintroduced that field but only as a serializer field with deprecation mark and comment. I will still be provided but always empty (it is optional anyway). The weird is that the swagger behavior is to just remove it from the docs - Adrian has a good point what this means for the bindings ...

osoukup:
> Actually I am kinda confused right now, looking for example at the `Flaw.impact` treating in SFM2, we do not allow the empty value (None) anymore (the validations won't allow it so the only way is to force it through BZ), so why is the default necessary? thinking blank is allowed so the Flaws with missing impact values will be imported correctly but do we want the behavior that newly created flaws would get the empty `impact` by default? This thought also applies to some other attributes (eg. source, etc.). I might be missing some bigger picture, but this feels weird to me.

Maybe I am confused :smile: I thought that blank means that the value is not required on the API or Serializer level and null is for the database level. Which leads me to think that blank without default will result in the failure on the DB level when no value is provided which is lower than it should happen. I would await this to be caught by some validation or considered OK which would be not if there is no default as it would just fail. Do I understand it wrong? :smile:

JakubFrejlach:
> @Elkasitu @JakubFrejlach so please one more look. I created separated migration to remove `mitigated_by` field and put it before the squash so it can be squashed too. Then I reintroduced that field but only as a serializer field with deprecation mark and comment. I will still be provided but always empty (it is optional anyway). The weird is that the swagger behavior is to just remove it from the docs - Adrian has a good point what this means for the bindings ...

I will need to investigate, will try to regenerate bindings with the deprecated field in place and will report back, how it behaves.

JakubFrejlach:
> @Elkasitu @JakubFrejlach so please one more look. I created separated migration to remove `mitigated_by` field and put it before the squash so it can be squashed too. Then I reintroduced that field but only as a serializer field with deprecation mark and comment. I will still be provided but always empty (it is optional anyway). The weird is that the swagger behavior is to just remove it from the docs - Adrian has a good point what this means for the bindings ...

Well I've tried your branch myself and it is present crossed out, so I guess this behavior is correct
image

JakubFrejlach:
> @JakubFrejlach thank you for valuable comments! For the deprecated attribute in schema it really looks OK so maybe I did something wrong or really overlooked it (but I am convinced that I ctrl+f searched for it). For the default string value I actually really do not know what is correct here. I for some reason first thought that without `default` but with `blank` it should not work because something needs to be stored when it is not `null`. However it will actually store an empty string when no value is specified so it will work no problem. Therefore specifying `NOVALUE` which is actually an empty string as `default` is actually no change. I quite like that it is more explicit because the standard implicit `default` empty string is not very intuitive. So the change is just for the reader but makes no functional difference. I can remove it if you are against.

Also, I've just remembered that we do have a compatibility shim, which is doing this None -> empty string for us. Look at the `osidb/mixins.py::NullStrFieldsMixin`. This was done so the import from the BZ where we still do have the double empty representation (Null/None vs Empty string).

osoukup:
> > @JakubFrejlach thank you for valuable comments! For the deprecated attribute in schema it really looks OK so maybe I did something wrong or really overlooked it (but I am convinced that I ctrl+f searched for it). For the default string value I actually really do not know what is correct here. I for some reason first thought that without `default` but with `blank` it should not work because something needs to be stored when it is not `null`. However it will actually store an empty string when no value is specified so it will work no problem. Therefore specifying `NOVALUE` which is actually an empty string as `default` is actually no change. I quite like that it is more explicit because the standard implicit `default` empty string is not very intuitive. So the change is just for the reader but makes no functional difference. I can remove it if you are against.
> 
> Also, I've just remembered that we do have a compatibility shim, which is doing this None -> empty string for us. Look at the `osidb/mixins.py::NullStrFieldsMixin`. This was done so the import from the BZ where we still do have the double empty representation (Null/None vs Empty string).

It seemed to me that Django would do it automatically but I was not completely sure. I was playing in the console a bit. `null=True` fields obviously allow `NULL` which is not correct. `null=False` just fails with an error when it gets `None` but it has to be assigned explicitly. Even without `default` when I created an object without setting any values they were set to `""`. I also forget about the mixin which solves the issue with the explicit `None` so now my defaults are really extra. I will remove them.

osidb/migrations/0060_remove_mitigated_by.py
```diff
@@ -0,0 +1,22 @@
...
+    operations = [
+        migrations.RemoveField(
+            model_name='flaw',
+            name='mitigated_by',
+        ),
+        migrations.RemoveField(
+            model_name='flawhistory',
+            name='mitigated_by',
+        ),
+    ]
```
Elkasitu, line 22: This is not N-1 safe, see https://develop.sentry.dev/database-migrations/#deleting-columnstables for safe ways to remove fields

osidb/models.py
```diff
@@ -343,7 +343,12 @@ class FlawHistoryResolution(models.TextChoices):
     )
 
     # flaw severity, from srtnotes "impact"
-    impact = models.CharField(choices=FlawImpact.choices, max_length=20, blank=True)
+    impact = models.CharField(
+        choices=FlawImpact.choices,
+        max_length=20,
+        blank=True,
+        default=FlawImpact.NOVALUE,
```
Elkasitu, line 350: why are we adding default values?

Elkasitu (changes requested):
I think we should reduce the scope of this PR to just squash the migrations:

The `mitigated_by` is not a safe change, is not relevant and the risk > reward.
Adding a default can be a costly operation because Django will apply the default at the Python level and not at the SQL level, and I'm not even sure that adding it to a squashed migration will have any effect (the squashed migration will not be applied because it's already considered as applied).

I would instead have those changes as separate migrations, after the squashed one and depending on the squashed one

osoukup:
@Elkasitu @JakubFrejlach I removed `mitigated_by` removal and will create a another pull request for it with respect to N-1 compatibility and also to make this one more single purposed. I also removed unnecessary string default values. Hopefully now the changes are not controversial any more :smile_cat:

Elkasitu (approved):
LGTM, just one minor comment (my bad as I missed it in a previous review)

JakubFrejlach (approved):
LGTM now.

PR #53, opened 2022-09-23
diverse validation strategies

osoukup:
The validations are the way to ensure sufficient data quality. However for the historical data which are and never will be fully valid the validations should not be sync blocker. We need to eventually have the complete Bugzilla flaw data in sync in OSIDB.

This pull request introduces an approach (definitely not the only one) how to achieve this. It introduces an option (save parameter) to suppress all exceptions raised on the flaw (other model) save and just store them as alerts instead. The default behavior is to stop the save raising the first exception encountered.

Please review the code and also generally the approach taken.

OSIDB-442


Elkasitu (changes requested):
I still don't agree with the approach but let's go with it for the sake of progress, here are some remarks:

Given the current implementation, what should be the contract of the `_validation_*` methods? should they always raise a `ValidationError`? should they sometimes create alerts (thus bypassing the handling in `validate()`?

I have pushed a test that better exemplifies my concerns, it should pass but because the validation method always creates an alert (feel free to squash or discard said test).

Finally WRT ACL handling, we should maybe create a task to revamp it and abstract away as much as possible to prevent mistakes (in this case it's not very impactful but it could be in the future), I'm thinking something like a mixin + having groups be database records (like django groups, maybe we could reuse those too)

osidb/mixins.py
```diff
@@ -127,3 +128,45 @@ def alert(self, name, description, _type="warning", resolution_steps=""):
...
+        run standard Django validations first potentially raising ValidationError
+        these ensure minimal necessary data quality and thus cannot be suppressed
+
+        then custom validations are run either raising ValidationError on first issue
+        encountered (default) or storing all found issues as alerts at the end
+        according to the given raise_validation_error option
+        """
+        # standard validations
+        # exclude meta attributes
+        self.full_clean(exclude=["meta_attr"])
```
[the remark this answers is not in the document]
osoukup: Good question and I do not know the concrete reasons. It makes generally sense to me to permit whatever mess there and I guess that it was the reason of the exclusion. I just kept it unchanged for now.

osidb/mixins.py
```diff
@@ -127,3 +128,45 @@ def alert(self, name, description, _type="warning", resolution_steps=""):
...
     class Meta:
         abstract = True
+
+
+class ValidateMixin(AlertMixin):
+    """
+    generic validate mixin
+    """
+
+    def validate(self, raise_validation_error=True):
```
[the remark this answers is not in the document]
osoukup: Makes sense. I will look into it.

osidb/models.py
```diff
@@ -872,7 +872,7 @@ def fts_search(q):
 
 
 class Affect(
-    AlertMixin, TrackingMixin, AffectExploitExtensionMixin, NullStrFieldsMixin
+    ValidateMixin, TrackingMixin, AffectExploitExtensionMixin, NullStrFieldsMixin
```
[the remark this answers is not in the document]
osoukup: Sounds reasonable. I will have a look at it.

osoukup:
> I still don't agree with the approach but let's go with it for the sake of progress, here are some remarks:
> 
> Given the current implementation, what should be the contract of the `_validation_*` methods? should they always raise a `ValidationError`? should they sometimes create alerts (thus bypassing the handling in `validate()`?

My idea is that it would depend on the criticality level of the alert. The warnings would be always recorded as alerts in the validation test and the errors will be always raised as exceptions from the validation tests. The mixin will then contextually suppress the exceptions raised and only store them as alerts (of error level).

> 
> I have pushed a test that better exemplifies my concerns, it should pass but because the validation method always creates an alert (feel free to squash or discard said test).

LGTM

> 
> Finally WRT ACL handling, we should maybe create a task to revamp it and abstract away as much as possible to prevent mistakes (in this case it's not very impactful but it could be in the future), I'm thinking something like a mixin + having groups be database records (like django groups, maybe we could reuse those too)

Makes sense. Making a note here to create an issue for it.

>

osoukup:
@Elkasitu I refactored `AlertMixin` and `ValidateMixin` as you suggested on chat and it looks better and allow substituting of `save`-`validate` combo throughout the models. All the `save` methods together with corresponding `validate` ones could be removed from the model definitions.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -89,6 +89,8 @@ def get_history(self):
                 impact=FlawImpact.IMPORTANT,
                 title="historical title",
                 description="historical description",
+                pgh_created_at=timezone.now(),
+                pgh_label="user",
```
Elkasitu, line 93: ?

collectors/bzimport/convertors.py
```diff
@@ -346,10 +346,20 @@ def save(self):
...
                 if isinstance(part, TrackingMixin):
-                    part.save(auto_timestamps=False)
-                else:
-                    part.save()
+                    kwargs["auto_timestamps"] = False
+
+                # we want to store all the data fetched by the collector
+                # so we suppress the exception raising in favor of alerts
+                if isinstance(part, AlertMixin):
+                    kwargs["raise_validation_error"] = False
```
Elkasitu, line 359: I might be wrong but I think you don't need to be aware of the instance type, you can just always pass these as kwargs and they'll be ignored if it's not an `AlertMixin` / `TrackingMixin`

Elkasitu (approved):
LGTM, good job! 👏🏽

collectors/bzimport/tests/test_convertors.py
osoukup: The test does not pass without it now but I have not dig into it very much as the history functionality is not used. The difference is that there used to be just `clean` while now it is `full_clean` so some of the additional steps requires the values to be set - not sure why as it is `null=True` but not worth to investigate now as the history might be eventually completely different.

Elkasitu: ok, let's make a mental note of it and check again when we get rid of pg_history

collectors/bzimport/convertors.py
osoukup: If I leave it unconditionally (always) it fails at https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/mixins.py#L50 with `got an unexpected keyword argument` so I assume that `models.Model.save()` does not accept unexpected arguments.

PR #54, opened 2022-10-04
remove mitigated_by attribute (1/3)

osoukup:
This PR removes all the field related functionality and the model side of it. It however preserves the actual DB columns to be N-1 compatible and introduces temporary serializer-only attribute to keep API compatibility.

First part of OSIDB-126 according to https://develop.sentry.dev/database-migrations/#deleting-columnstables


JakubFrejlach (approved):
LGTM. I just wonder whether we should not mark somehow the deprecation also in the models (comment `# DEPRECATED - 3.0.0` maybe)

Elkasitu (approved):
LGTM, this might be a good time to give https://github.com/3YOURMIND/django-deprecate-fields a try.

Also you can do a delete in two steps:

1. Remove the field from django state in the migration + set the column to not null via an SQL statement (version N-1)
2. Remove the column from the database (version N)

Which is what the linked library does (I assume)

osoukup:
@Elkasitu (@JakubFrejlach) I tried to remove the manual one and create migrations again but the result is completely the same migration so I left the commits the way they were. The dummy serialize method is a different thing. There are to possible approaches I see. We can either collect the `deprecated_field` attributes until the major version and then remove them all or we can remove them in two consecutive minor releases and replace by the dummy serialize method. I am thinking what would be better. Keeping the deprecated fields would mean to have the model messy for longer. Dummy serialize methods means more code to create manually. I am not sure whether there are any other differences which matter. No strong preference from my side.

Elkasitu:
@osoukup I think `deprecate_field` is safer and more convenient:

* With the manual method, you need to remember to manually modify the migration and do it properly in multiple steps. With `deprecate_field` you just have to wrap the existing field and the migrations will be correctly generated.
* It is trivial to remove fields that use the `deprecate_field` wrapper (in fact, you can probably do it with a single call to the `sed` command) and can be part of a checklist for major releases, for the manual method we'd need to remember or mark dummy serializer methods to be removed in subsequent versions (whether minor or major doesn't really matter).
* While the model might look messy, the deprecated fields are easy to spot, and with the manual method we just move the "messiness" from the model to the serializer really.

osoukup:
Closing in favor of https://github.com/RedHatProductSecurity/osidb/pull/55

PR #55, opened 2022-10-10
remove mitigated by deprecate field only

osoukup:
I create this pull request as a cleaner version of https://github.com/RedHatProductSecurity/osidb/pull/54 after all the good discussions we had there. It still deprecates `mitigated_by` field but hopefully better. These are the commits introduced

- first just removes the related functionality
- second adds `deprecate-fields` lib dependency
- third uses it to deprecate the field + also migration is created and the API schema updated
- fourth deprecates the field on API level + again updates the schema

From the user point of view a few things happen. The field is marked read-only which is no difference as the whole OSIDB is read-only so far. The field is also marked deprecated which is probably just schema thing (not sure how the bindings cope with it). The field is still being returned in the query results it is just always null - which is no breaking change as the value was optional.

Process-wise with the next major release we should remove all remaining appearances of `mitigated_by` - deprecated field in model and serializer - and create migration and update API schema.


Elkasitu (approved):
LGTM, maybe we could add the process to the docs once we formalize it (i.e. doesn't need to be right now)

Elkasitu:
Should this mark OSIDB-126 as closed? I think there's no point in keeping it open as long as we remove all deprecations on major release

osoukup:
> Should this mark OSIDB-126 as closed? I think there's no point in keeping it open as long as we remove all deprecations on major release

Good point. Marking this as closing the issue.

PR #56, opened 2022-10-10
revert problematic tracker component regex refinement

osoukup:
an old merge request (228)
introduced a tracker component regex refinement with the following reasoning

    the component is exactly before the first colon (not any)
    I just encountered an exceptional tracker with

            component: something: text

    title which might cause us some troubles when parsing (greedy regex modifier +)

however it causes more troubles than good as it breaks parsing the title with RHEL modules or multiple components so let us just revert it back to SFM2 behavior and eventually deal with it if it is worthy

OSIDB-464

### DIFF
--- a/collectors/tests/test_utils.py
+++ b/collectors/tests/test_utils.py
@@ -1,0 +1,60 @@
+import pytest
+
+from collectors.utils import tracker_parse_update_stream_component
+
+pytestmark = pytest.mark.unit
+
+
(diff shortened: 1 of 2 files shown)


Elkasitu (approved):

Elkasitu (commented):
Actually, can we write a test that makes the current version fail?

osoukup:
@Elkasitu good point! I added some tests and signed a commit first time (+verified).

PR #57, opened 2022-10-11
document how to deprecate fields

osoukup:
https://github.com/RedHatProductSecurity/osidb/pull/55 follow up so we have the procedure documented


Elkasitu (approved):
LGTM

PR #58, opened 2022-10-13
store FlawMeta alerts on FlawMeta instead of on Flaw

osoukup:
we decided to store alerts on the same model when the validation is defined despite what is the parent or child and whether the alert relates to both etc.

if we later need all alerts to be gathered (parent and child) we will deal with it on the getter/serializer level


osidb/models.py
```diff
@@ -1268,14 +1268,11 @@ def _validate_public_source_no_ack(self):
 
         if (source := FlawSource(self.flaw.source)) and source.is_public():
             if source.is_private():
-                # FIXME: self.flaw is null=True, but in practice it doesn't happen
-                # and it doesn't even make sense
```
Elkasitu, line 1272: I need to remind myself to check this, does it make sense to have a FlawMeta without a Flaw attached? I would think not

Elkasitu (approved):

osidb/models.py
osoukup: It does not make sense as it is flaw metadata but I see that the foreign key is `null=True` for some reason. I am not sure why. I would consider a `FlawMeta` without `Flaw` an error. I am checking the production and there is no `FlawMeta` with `flaw__isnull=True` and I am also checking all occurrences in the codebase and whenever we create `FlawMeta` we provide a flaw. I think we could remove `null=True` so it is explicit that there must always must be a flaw. I will proceed with this, run tests, etc. and eventually add a commit here but it seems safe to me.

PR #60, opened 2022-10-18
Add reference to all CISA exploits

skontar:
Always provide exploit reference link. For CISA, who do not provide link, use generic link to CISA exploit catalog.

PSINSIGHTS-674


osoukup (approved):
LGTM

Elkasitu (approved):

PR #70, opened 2022-11-07
added costaconrado as a dummy user for local environment

costaconrado:
Added a new dummy user to LDAP local development environment.

Changed local-export.ldif to add the user 1017 (costaconrado) for development environment.


costaconrado:
MR wrongly created without githooks.

PR #75, opened 2022-11-14
Add Global CA certificate for AWS RDS database connection

Elkasitu:
This is a required change in order to migrate from a self-hosted PostgreSQL instance to AWS RDS while still keeping Encryption-in-Transit active.

The certificate is downloaded and installed per the instructions at https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html


osoukup (approved):
LGTM

PR #77, opened 2022-11-22
validate reported date

osoukup:
This PR closes OSIDB-332. The reported date validation has two parts. One is that it cannot be set in the future. This one is already present in OSIDB. It is not done by the validation framework but a simple strict Django validation but it seems to cause no issues (no collector data are restricted by it) so we can leave it as it is. One is that reported date (reported_dt attribute) should not be empty. The second is what this PR introduces.

Also the FlawFactory was extended with this attribute (was missing for some reason) and a number of old tests creating flaws with empty reported date has to be adjusted to comply with the new validation.

### DIFF
--- a/apps/osim/tests/test_models.py
+++ b/apps/osim/tests/test_models.py
@@ -64,7 +64,7 @@ class CheckDescFactory:
         ("has cwe", "cwe_id", ""),
         ("has unembargo_dt", "unembargo_dt", None),
         ("has source", "source", ""),
-        ("has reported_dt", "reported_dt", None),
+        # ("has reported_dt", "reported_dt", None),
         ("has cvss2", "cvss2", ""),
         ("has cvss2_score", "cvss2_score", None),
         ("has cvss3", "cvss3", ""),
--- a/osidb/tests/factories.py
+++ b/osidb/tests/factories.py
@@ -36,6 +36,7 @@ class FlawFactory(factory.django.DjangoModelFactory):
     cve_id = factory.sequence(lambda n: f"CVE-2020-000{n}")
     type = factory.Faker("random_element", elements=list(FlawType))
     created_dt = factory.Faker("date_time", tzinfo=UTC)
+    reported_dt = factory.Faker("date_time", tzinfo=UTC)
     state = factory.Faker("random_element", elements=list(Flaw.FlawState))
     resolution = factory.Faker("random_element", elements=list(FlawResolution))
     impact = factory.Faker("random_element", elements=list(FlawImpact))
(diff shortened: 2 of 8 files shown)


osidb/tests/factories.py
```diff
@@ -36,6 +36,7 @@ class Meta:
     cve_id = factory.sequence(lambda n: f"CVE-2020-000{n}")
     type = factory.Faker("random_element", elements=list(FlawType))
     created_dt = factory.Faker("date_time", tzinfo=UTC)
+    reported_dt = factory.Faker("date_time", tzinfo=UTC)
```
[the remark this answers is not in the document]
osoukup: I see your point. There are at least two different view on the flaw correctness. One is that such flaw can exist in the data and another one is that such flaw is valid according to the validations. I would prefer the latter and the factories would this way continuously get more and more strict or restricted as we add validations. However the resulting created flaws would pass all the validations then. This would be increasingly more work. The other approach would leave the test developer to manually fix all the fields to pass the validations. With this approach you can still create a validation-wise invalid flaw - eg. manually set `reported_dt` to `None` or create a raw `Flaw` without using the factory. WDYT?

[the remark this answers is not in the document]
JakubFrejlach: I definitely agree. I was not even disputing this I was just simply explaining my reasoning when I created the Factories some time ago. I agree with this approach as the Factory should definitely produce Flaw that can pass all the validation checks.

JakubFrejlach (approved):

costaconrado (approved):
LGTM

apps/osim/tests/test_models.py
```diff
@@ -64,7 +64,7 @@ class CheckDescFactory:
         ("has cwe", "cwe_id", ""),
         ("has unembargo_dt", "unembargo_dt", None),
         ("has source", "source", ""),
-        ("has reported_dt", "reported_dt", None),
+        # ("has reported_dt", "reported_dt", None),
```
Elkasitu, line 67: is there a point to leaving this as a comment instead of just deleting the line?

Elkasitu (approved):
LGTM, just one minor comment

apps/osim/tests/test_models.py
osoukup: I am actually not sure :smile: I made this test suite more than a year ago and I am already not very well familiar with all the intentions and I just did not want to spend time finding out the reasons why was it added and whether we can remove it entirely. I just made the test working :smile:

osoukup: Maybe the module will be eventually just removed.

PR #78, opened 2022-11-22
Ensure that the deprecated mitaged_by field always returns a string

Elkasitu:
This field had been returning null ever since 2.3.0 which is API breakage.

Closes OSIDB-614

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -604,7 +604,9 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
 
     # , from srtnotes "mitigate"
     mitigated_by = deprecate_field(
-        models.CharField(choices=FlawMitigate.choices, max_length=10, blank=True)
+        models.CharField(choices=FlawMitigate.choices, max_length=10, blank=True),
+        # required to keep backwards compatibility
+        return_instead="",
     )
 
     # , from srtnotes "cvss2"


osoukup:
LGTM. It is already out there :smile: but we can fix it until it is noticed

osoukup (approved):

JakubFrejlach:
I wonder whether should we somehow document the field deprecation process and explicitly state there that the string based field (CharField, TextField) should be deprecated with this `return_instead=""` param. 

Or a different idea, maybe create a wrapper around the deprecate_field which would check the field type and in the case of the CharField/StringField it would automatically add the `return_instead=""`? :thinking:

osoukup:
> I wonder whether should we somehow document the field deprecation process and explicitly state there that the string based field (CharField, TextField) should be deprecated with this `return_instead=""` param.
> 
> Or a different idea, maybe create a wrapper around the deprecate_field which would check the field type and in the case of the CharField/StringField it would automatically add the `return_instead=""`? thinking

It is documented at

https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#deprecate-fields

but the description is rather general/high-level and does not name specific cases for eg. string fields. We could add a note on `return_instead` there.

JakubFrejlach:
> > I wonder whether should we somehow document the field deprecation process and explicitly state there that the string based field (CharField, TextField) should be deprecated with this `return_instead=""` param.
> > Or a different idea, maybe create a wrapper around the deprecate_field which would check the field type and in the case of the CharField/StringField it would automatically add the `return_instead=""`? thinking
> 
> It is documented at
> 
> https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#deprecate-fields
> 
> but the description is rather general/high-level and does not name specific cases for eg. string fields. We could add a note on `return_instead` there.

Ah, my bad I was looking for the doc addition in the same PR as we deprecated the `mitigated_by` field and didn't find it there so I assumed that it is not documented. Still it might be good to document the usage. It is now clear since we do have an example in the code but once we remove the field for good the example will go away as well.

Elkasitu:
We're still figuring out how to deprecate things so this is a learning experience, we can think about how to make it better as we go along 💃🏽 

One idea to always verify that the schema corresponds is to write tests that verify that the api returns things consistent with the schema, something like reverse fuzzing

PR #79, opened 2022-11-23
Migration fixprovements

Elkasitu:
This PR contains two commits that touch on migrations:
* The first is a fix for managed databases (such as AWS Aurora) for which we might not want to provide a full `postgresql.conf` (which is the case for our local / self-managed databases).
* The second is a cleanup of no-longer-used migrations (squashed)


osoukup (approved):
Great cleanup! I was postponing this for a few weeks so thank you for finishing it :smile: also the collector migrations could be cleaned but those are not pressing.

Elkasitu:
@osoukup thanks for pointing it out, I added the rest of migrations to the last commit (weren't that many, just bzimport and collector framework)

PR #80, opened 2022-11-23
Validate public date

costaconrado:
This PR validate the if all public Flaws created have an unembargo date closing both OSIDB-344 and OSIDB-345.

To achieve it was create a new model validator and the FlawFactory was extended with `unembargo_dt` attribute and old tests that creates public flaws was edited to support this validation.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -12,6 +12,7 @@ from django.contrib.postgres import fields
 from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import models
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from django_deprecate_fields import deprecate_field
 from polymorphic.models import PolymorphicModel
@@ -700,11 +701,12 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
         """
         if not self.source:
             return
-        # TODO: make embargoed accessible from python code (property?)
-        embargoed = self.acl_read == [
-            uuid.UUID(acl) for acl in generate_acls([settings.EMBARGO_READ_GROUP])
-        ]
-        if embargoed and (source := FlawSource(self.source)) and source.is_public():
+
+        if (
+            self.is_embargoed
+            and (source := FlawSource(self.source))
+            and source.is_public()
+        ):
             if source.is_private():
                 self.alert(
                     "embargoed_source_public",
@@ -722,6 +724,16 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
         """
         if self.reported_dt is None:
             raise ValidationError("Flaw has an empty reported_dt")
+
+    def _validate_public_unembargo_date(self):
+        """
+        Check that an unembargo date (public date) exists and is in the past if the Flaw is public
+        """
+        if not self.is_embargoed:
+            if self.unembargo_dt is None:
+                raise ValidationError("Public flaw has an empty unembargo_dt")
+            if self.unembargo_dt > timezone.now():
+                raise ValidationError("Public flaw has a future unembargo_dt")
 
     # TODO this needs to be refactored
     # but it makes sense only when we are capable of write actions
@@ -814,6 +826,12 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
         """check that all trackers have resolution"""
         # TODO we have no tracker resolution for now
         return False
+
+    @property
+    def is_embargoed(self):
+        return self.acl_read == [
+            uuid.UUID(acl) for acl in generate_acls([settings.EMBARGO_READ_GROUP])
+        ]
 
 
 class AffectManager(models.Manager):
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -735,6 +735,19 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
             if self.unembargo_dt > timezone.now():
                 raise ValidationError("Public flaw has a future unembargo_dt")
 
+    def _validate_future_unembargo_date(self):
+        """
+        Check that an enbargoed flaw has an unembargo date in the future
+        """
+        if (
+            self.is_embargoed
+            and self.unembargo_dt is not None
+            and self.unembargo_dt < timezone.now()
+        ):
+            raise ValidationError(
+                "Flaw still embargoed but unembargo date is in the past."
+            )
+
     # TODO this needs to be refactored
     # but it makes sense only when we are capable of write actions
     # and we may thus actually do some changes to the embargo
--- a/osidb/tests/factories.py
+++ b/osidb/tests/factories.py
@@ -44,6 +44,11 @@ class FlawFactory(factory.django.DjangoModelFactory):
     description = factory.LazyAttribute(lambda c: f"Description for {c.cve_id}")
     statement = factory.LazyAttribute(lambda c: f"Statement for {c.cve_id}")
     embargoed = factory.Faker("random_element", elements=[False, True])
+    unembargo_dt = factory.Maybe(
+        "embargoed",
+        yes_declaration=factory.Faker("future_datetime", tzinfo=UTC),
+        no_declaration=factory.Faker("past_datetime", tzinfo=UTC),
+    )
     # cannot be set to ("random_element", elements=list(FlawSource)) because it could
     # inadvertently trigger validation errors in unrelated tests.
     source = ""
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -49,6 +49,7 @@ class TestFlaw:
             state=Flaw.FlawState.NEW,
             created_dt=datetime_with_tz,
             reported_dt=datetime_with_tz,
+            unembargo_dt=datetime_with_tz,
             type=FlawType.VULNERABILITY,
             title="title",
             description="description",
@@ -139,6 +140,7 @@ class TestFlaw:
             state=Flaw.FlawState.NEW,
             created_dt=datetime_with_tz,
             reported_dt=datetime_with_tz,
+            unembargo_dt=datetime_with_tz,
             type=FlawType.VULNERABILITY,
             title="title",
             description="description",
@@ -164,6 +166,7 @@ class TestFlaw:
         Flaw.objects.create_flaw(
             bz_id="12345",
             title="first",
+            unembargo_dt=tzdatetime(2000, 1, 1),
             description="description",
             acl_read=acls,
             acl_write=acls,
@@ -282,6 +285,7 @@ class TestFlaw:
             state=Flaw.FlawState.NEW,
             created_dt=datetime_with_tz,
             reported_dt=datetime_with_tz,
+            unembargo_dt=datetime_with_tz,
             type=FlawType.VULNERABILITY,
             title="title",
             description="description",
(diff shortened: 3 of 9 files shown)


osidb/models.py
```diff
@@ -592,7 +592,9 @@ class FlawMitigate(models.TextChoices):
     cwe_id = models.CharField(blank=True, max_length=50, validators=[validate_cwe_id])
 
     # date when embargo is to be lifted, from srtnotes "public"
-    unembargo_dt = models.DateTimeField(null=True, blank=True)
+    unembargo_dt = models.DateTimeField(
+        null=True, blank=True, validators=[no_future_date]
```
osoukup, line 596: This validator will not work correctly. Let me explain. The reason is that it will restrict all flaws to have no future `unembargo_dt`. This is OK for already unembargoed flaws as their unembargo time is obviously in the past. However for the flaws still under embargo it is not true. They still need to be unembargoed eventually so it will obviously happen in the future and as we want to plan it setting a future `unembargo_dt` is correct in these cases. As the condition is more complex you need to implement custom validation to check it.

osidb/tests/test_flaw.py
```diff
@@ -733,6 +740,58 @@ def test_validate_affect_ps_module_errors(self, bz_id, ps_module, should_raise):
             affect = AffectFactory(flaw=flaw, ps_module=ps_module)
             assert affect
 
+    @pytest.mark.parametrize(
+        "group,unembargo_date,should_raise",
+        [
+            (settings.PUBLIC_READ_GROUPS, None, True),
```
osoukup, line 746: Observe that `PUBLIC_READ_GROUPS` is an array while `EMBARGO_READ_GROUP` is string so you need to deal with them differently. Those ACLs are tricky :smile: and they work with many things without throwing any error but here one or the other case must be something else than intended.

osidb/tests/test_flaw.py
```diff
@@ -733,6 +740,58 @@ def test_validate_affect_ps_module_errors(self, bz_id, ps_module, should_raise):
...
+            (settings.PUBLIC_READ_GROUPS, tzdatetime(2000, 1, 1), False),
+            (settings.EMBARGO_READ_GROUP, None, False),
+            (settings.EMBARGO_READ_GROUP, tzdatetime(2000, 1, 1), False),
```
osoukup, line 749: For testing the time stuff you can use `freezegun` like eg. at https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/tests/test_flaw.py#L411 so you can safely assume what is past/present/future and it will not change with the real time.

osoukup (changes requested):
Nice progress! I have a few comments, some to consider but some require adjustments. Just ping me if something was not clear or you needed a re-review :wink:

osidb/tests/test_flaw.py
```diff
@@ -725,3 +726,55 @@ def test_validate_affect_ps_module_errors(self, bz_id, ps_module, should_raise):
...
+                            f"https://osidb.prod.redhat.com/ns/acls#{group}",
+                        )
+                    ],
+                    acl_write=[
+                        uuid.uuid5(
+                            uuid.NAMESPACE_URL,
+                            f"https://osidb.prod.redhat.com/ns/acls#{settings.EMBARGO_WRITE_GROUP}",
+                        )
+                    ],
+                )
```
Elkasitu, line 772: You can use `FlawFactory` instead and pass it `embargoed=True` as an attribute and will correctly set the ACLs IIRC, the benefit is that you'd only have to set `embargoed`, `reported_dt` and `unembargo_dt`.

osidb/tests/test_flaw.py
```diff
@@ -725,3 +726,55 @@ def test_validate_affect_ps_module_errors(self, bz_id, ps_module, should_raise):
...
+                        f"https://osidb.prod.redhat.com/ns/acls#{group}",
+                    )
+                ],
+                acl_write=[
+                    uuid.uuid5(
+                        uuid.NAMESPACE_URL,
+                        f"https://osidb.prod.redhat.com/ns/acls#{settings.EMBARGO_WRITE_GROUP}",
+                    )
+                ],
+            )
```
Elkasitu, line 792: same as above

Elkasitu (changes requested):
Aside from these comments, I think commit 3 should be merged into commit 2, and you should do a `git pull --rebase origin master` to avoid commit 4 (merge commit)

osidb/tests/factories.py
```diff
@@ -36,6 +36,7 @@ class Meta:
     cve_id = factory.sequence(lambda n: f"CVE-2020-000{n}")
     type = factory.Faker("random_element", elements=list(FlawType))
     created_dt = factory.Faker("date_time", tzinfo=UTC)
+    unembargo_dt = factory.Faker("date_time", tzinfo=UTC)
```
JakubFrejlach, line 39: This way the `FlawFactory` will always create a Flaw with `unembargo_dt` even when the Flaw it self is not embargoed, I would suggest to use `factory.Maybe` to decide based on the embargoed attribute also by default it should return some future date from now so the Flaw produced by factory can be correctly validated:

```python
unembargo_dt = factory.Maybe(
        "embargoed",
        yes_declaration=factory.Faker("future_datetime", tzinfo=UTC),
        no_declaration=None,
    )
```

JakubFrejlach (changes requested):
Nice progress. :+1:  Added couple of my suggestions.

osidb/models.py
```diff
@@ -723,6 +725,16 @@ def _validate_reported_date(self):
         if self.reported_dt is None:
             raise ValidationError("Flaw has an empty reported_dt")
 
+    def _validate_unembargo_date(self):
+        """
+        Check that an unembargo date (public date) exists if the Flaw is public
+        """
+        embargoed = self.acl_read == [
```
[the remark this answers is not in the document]
costaconrado: Then I will create an `is_embargoed` property (following the same name pattern of `is_major_incident`) and change the two validators that need this info.

osidb/models.py
costaconrado: I wil remove this commit and create a new one with a custom validator.

osidb/tests/test_flaw.py
costaconrado: Thank you for the feedback, changed in current PR.

osidb/tests/factories.py
costaconrado: Makes sense, I will edit it.

osidb/models.py
```diff
@@ -723,6 +725,22 @@ def _validate_reported_date(self):
...
             raise ValidationError("Flaw has an empty reported_dt")
 
+    def _validate_public_unembargo_date(self):
+        """
+        Check that an unembargo date (public date) exists if the Flaw is public
+        """
+        if not self.is_embargoed and self.unembargo_dt is None:
+            raise ValidationError("Public flaw has an empty unembargo_dt")
+
+    def _validate_future_unembargo_date(self):
```
osoukup, line 735: This check is still not correct. We want to validate that IF the flaw is embargoed AND has unembargo date THEN the date is NOT in the past. The reason is that the flaw should be made public at the point of unembargo date so when it is still embargoed the unembargo date cannot be passed.

osidb/tests/factories.py
```diff
@@ -44,6 +44,11 @@ class Meta:
     description = factory.LazyAttribute(lambda c: f"Description for {c.cve_id}")
     statement = factory.LazyAttribute(lambda c: f"Statement for {c.cve_id}")
     embargoed = factory.Faker("random_element", elements=[False, True])
+    unembargo_dt = factory.Maybe(
+        "embargoed",
+        yes_declaration=None,
+        no_declaration=factory.Faker("date_time", tzinfo=UTC),
```
osoukup, line 50: I wonder what date is generated here. I assume a random date. It should however be a date in the past only - if I understand it correctly that it is set only for public flaws. I am not sure it is ensured here. With a future date we would generate an invalid flaw.

osoukup (changes requested):
Getting closer :smile: but I still think that one of the validations should be fixed

osidb/tests/factories.py
costaconrado: makes sense, I will use `factory.Faker("past_datetime", tzinfo=UTC)`.

osidb/models.py
costaconrado: Sorry, I had misunderstood the story.

osoukup (approved):
LGTM. Thank you for addressing all my comments :+1:

osidb/tests/test_flaw.py
```diff
@@ -748,3 +755,39 @@ def test_validate_reported_date_non_empty(self):
...
+        else:
+            assert FlawFactory(unembargo_dt=unembargo_date, embargoed=embargoed)
+
+    @freeze_time(tzdatetime(2021, 11, 23))
+    def test_validate_future_unembargo_date(self):
+        """test unembargo_dt validation - no future date"""
+        future_dt = tzdatetime(2021, 11, 27)
+
+        with pytest.raises(ValidationError) as e:
+            FlawFactory(unembargo_dt=future_dt)
```
Elkasitu, line 785: you don't explicitly set `embargoed` here and by default it's random so this test will fail 50% of the time

osidb/models.py
```diff
@@ -723,6 +724,13 @@ def _validate_reported_date(self):
         if self.reported_dt is None:
             raise ValidationError("Flaw has an empty reported_dt")
 
+    def _validate_public_unembargo_date(self):
+        """
+        Check that an unembargo date (public date) exists if the Flaw is public
+        """
+        if not self.is_embargoed and self.unembargo_dt is None:
+            raise ValidationError("Public flaw has an empty unembargo_dt")
```
Elkasitu, line 733: @osoukup is this truly the case? what about 0-days? or do we assume that flaws within Red Hat will **always** start as embargoed and eventually become public?

osidb/tests/test_flaw.py
```diff
@@ -1,6 +1,7 @@
 import uuid
 
 import pytest
+from django.conf import settings
```
Elkasitu, line 4: unused import?

osidb/models.py
Elkasitu: @osoukup while we're at it (maybe this should be a different task) shouldn't we validate that a public flaw's unembargo_dt is always in the past?

Elkasitu (commented):
Tests are failing (one explained in the review, the other are probably missing adjustments to some Factory calls)

osidb/tests/test_flaw.py
costaconrado: sorry, that was mistake, fixing it.

osidb/models.py
osoukup: We always set public date whenever a flaw is not embargoed. For the 0-days we would set public date to the (estimated) time of the first public appearance (possibly predating the reported date).

The second comment is a good point and it makes sense to me to have such validation. That would be also some kind of leak safeguard as OSIDB would not permit an unembargo before the public date (one would have to modify the public date first so that would make her/him think twice). I know this is above the original task but could you @costaconrado add this validation too? It should be fairly similar.

costaconrado: Oh I see, then I think I should change the factory itself to something like 
```
    unembargo_dt = factory.Maybe(
        "embargoed",
        yes_declaration=factory.Faker("future_datetime", tzinfo=UTC),
        no_declaration=factory.Faker("past_datetime", tzinfo=UTC),
    )
```

I will create this new validation and change the factory to reflect it

osidb/tests/test_flaw.py
costaconrado: sorry, there were some other tests with the same problem, I have updated it.

Elkasitu (approved):
Good job!

PR #82, opened 2022-11-28
merge 2.3.2 to master

osoukup:
after release merge


Elkasitu (approved):
LGTM, I'd do a `git pull --rebase origin master` to remove that last merge commit 😜

osoukup:
@Elkasitu that would however bring all the master before 2.3.2 release which does not seem correct to me. That would make stuff already released appear after not yet released stuff. Is it really what we want?

Elkasitu:
Well the order of commits on master is not representative of the commits present in a given version, rebase is what we've done in the past but you can leave the merge commit if you prefer

osoukup:
@Elkasitu I agree that it would be no issue on master but if I rebase the branch I will change it in a way that if we potentially released a patch 2.3.3 it would also contain unrelated commits from master.

Elkasitu:
@osoukup if we need to release a 2.3.3 patch we'd base it off of the `2.3.2` tag as described in the patch release [documentation](https://git.prodsec.redhat.com/devops/osidb-ops/-/blob/master/README.md#deploy-patch-version-mid-sprint), meaning that nothing merged into master would get in

costaconrado (approved):
LGTM. Versions and Jira numbers make sense.

PR #85, opened 2022-12-01
Validate that flaw has cvss3

costaconrado:
This PR validate that flaw has a cvss3 string.

This PR creates a model validation for Flaw assuring it has a string cvss3. This PR also modifies the flaw factory and existing tests to reflect this validation. This PR closes OSIDB-335.

### DIFF
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -59,6 +59,7 @@ class TestFlaw:
             is_major_incident=True,
             acl_read=acls,
             acl_write=acls,
+            cvss3="3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N",
             # META
             meta_attr=meta_attr,
         )
@@ -149,6 +150,7 @@ class TestFlaw:
             resolution=FlawResolution.NOVALUE,
             acl_read=acls,
             acl_write=acls,
+            cvss3="3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N",
         )
         assert vuln_2.validate() is None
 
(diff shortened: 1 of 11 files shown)


osidb/tests/test_flaw.py
```diff
@@ -791,3 +798,13 @@ def test_validate_future_unembargo_date(self):
...
             FlawFactory(unembargo_dt=future_dt, embargoed=True)
             # no exception should be raised now
+
+    def test_validate_cvss3(self):
+        """
+        Test that the ValidationError is not raised when the flaw has a CVSS3 string
+        """
+        assert FlawFactory(cvss3="3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N")
+
+        with pytest.raises(ValidationError) as e:
```
osoukup, line 808: nitpick: the test case description does not mention this one :smile: I would maybe just split it into two different tests so when either flaw with or without CVSS string fails the test it is immediately clear which case failed

osoukup (approved):
One minor comment but otherwise LGTM

osidb/tests/test_flaw.py
costaconrado: makes sense, would a parameterized approach like this works or do you think it would be better two completely separated tests?
```
    @pytest.mark.parametrize(
        "cvss3,should_raise",
        [
            ("3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N", False),
            (None, True),
        ],
    )
    def test_validate_cvss3(self,cvss3,should_raise):
        """
        Test that the ValidationError is not raised when the flaw has a CVSS3 string
        """
        if should_raise:
            with pytest.raises(ValidationError) as e:
                FlawFactory(cvss3=cvss3)
            assert "CVSSv3 score is missing" in str(e)
        else:
            assert FlawFactory(cvss3=cvss3)
```

osoukup: I have no strong preference here so completely up to you :wink:

JakubFrejlach (approved):
LGTM!

PR #86, opened 2022-12-02
Add validations for flaw title

costaconrado:
This PR add validations for flaw title.

This PR checks if all embargoed flaws have the word "EMBARGOED" in its title and also validates if public flaws does not have it. This PR also change `FlawFactory` and multiple tests to adhere into this new validation.
This PR closes OSIDB-348 and OSIDB-349.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -779,6 +779,24 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
         # users from reviewing in the first place, in which case we don't
         # need to perform this validation
 
+    def _validate_public_flaw_title(self):
+        """
+        Check that the flaw's title does not contain the word "EMBARGOED" if the flaw is public.
+        """
+        if not self.is_embargoed and "EMBARGOED" in self.title:
+            raise ValidationError(
+                'Flaw title contains "EMBARGOED" despite being public.'
+            )
+
+    def _validate_embargoed_flaw_title(self):
+        """
+        Check that the flaw's title does contain the word "EMBARGOED" if the flaw is embargoed.
+        """
+        if self.is_embargoed and "EMBARGOED" not in self.title:
+            raise ValidationError(
+                'Flaw title does not contains "EMBARGOED" despite being embargoed.'
+            )
+
     # TODO this needs to be refactored
     # but it makes sense only when we are capable of write actions
     # and we may thus actually do some changes to the embargo
--- a/osidb/tests/factories.py
+++ b/osidb/tests/factories.py
@@ -40,8 +40,16 @@ class FlawFactory(factory.django.DjangoModelFactory):
     state = factory.Faker("random_element", elements=list(Flaw.FlawState))
     resolution = factory.Faker("random_element", elements=list(FlawResolution))
     impact = factory.Faker("random_element", elements=list(FlawImpact))
-    title = factory.LazyAttribute(lambda c: f"Title for {c.cve_id}")
     description = factory.LazyAttribute(lambda c: f"Description for {c.cve_id}")
+    title = factory.Maybe(
+        "embargoed",
+        yes_declaration=factory.LazyAttribute(
+            lambda c: f"EMBARGOED {c.cve_id} kernel: some description"
+        ),
+        no_declaration=factory.LazyAttribute(
+            lambda c: f"{c.cve_id} kernel: some description"
+        ),
+    )
     statement = factory.LazyAttribute(lambda c: f"Statement for {c.cve_id}")
     embargoed = factory.Faker("random_element", elements=[False, True])
     summary = factory.Maybe(
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -866,3 +866,38 @@ class TestFlawValidators:
                 flaw1.save()
         else:
             assert flaw1.save() is None
+
+    @pytest.mark.parametrize(
+        "title,should_raise",
+        [
+            ("CVE-2022-1234 kernel: some description", False),
+            ("EMBARGOED CVE-2022-1234 kernel: some description", True),
+        ],
+    )
+    def test_validate_public_flaw_title(self, title, should_raise):
+        """test that public flaws only accepts a valid title without the "EMBARGOED" word"""
+        if should_raise:
+            with pytest.raises(ValidationError) as e:
+                FlawFactory(embargoed=False, title=title)
+            assert 'Flaw title contains "EMBARGOED" despite being public.' in str(e)
+        else:
+            assert FlawFactory(embargoed=False, title=title)
+
+    @pytest.mark.parametrize(
+        "title,should_raise",
+        [
+            ("EMBARGOED CVE-2022-1234 kernel: some description", False),
+            ("CVE-2022-1234 kernel: some description", True),
+        ],
+    )
+    def test_validate_embargoed_flaw_title(self, title, should_raise):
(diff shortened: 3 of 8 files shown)


osidb/tests/factories.py
```diff
@@ -49,6 +49,15 @@ class Meta:
         yes_declaration=factory.Faker("future_datetime", tzinfo=UTC),
         no_declaration=factory.Faker("past_datetime", tzinfo=UTC),
     )
+    summary = factory.Maybe(
+        "embargoed",
+        yes_declaration=factory.sequence(
+            lambda n: f"EMBARGOED - CVE-2022-000{n} kernel-rt: kernel: race condition in example function [rhel-1.0.0.z]"
```
osoukup, line 55: nitpick: the correct flaw title (though we do not have all the checks implemented yet) looks a bit differently (https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/helpers.py#L22) - it does not contain dash between EMBARGOED and the CVE(s) and there is no square bracket at the end (unlike for the tracker title) - though the square bracket would not matter to the validations as the end part is a free text

osidb/models.py
```diff
@@ -750,6 +750,24 @@ def _validate_future_unembargo_date(self):
                 "Flaw still embargoed but unembargo date is in the past."
             )
 
+    def _validate_public_flaw_summary(self):
+        """
+        Check that the flaw's summary does not contain the word "EMBARGOED" if the flaw is public.
+        """
+        if not self.is_embargoed and "EMBARGOED" in self.summary:
```
osoukup, line 757: There is one tricky thing in OSIDB. It is not very easy to spot but you can see a comment at https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/models.py#L357 that what we used to call summary in the SFM2 terminology (and in Bugzilla) is actually called **title** in SFM2. The summary here needs to be replaced by title

osoukup: And not just here. Basically everywhere in the PR you need to use **title**

osidb/tests/test_flaw.py
```diff
@@ -791,3 +791,38 @@ def test_validate_future_unembargo_date(self):
         with freeze_time(future_dt):
             FlawFactory(unembargo_dt=future_dt, embargoed=True)
             # no exception should be raised now
+
+    @pytest.mark.parametrize(
+        "is_embargoed,summary,should_raise",
+        [
+            (False, "Valid summary", False),
+            (False, "EMBARGOED - Invalid summary", True),
```
osoukup, line 799: nitpick: when `is_embargoed` is always `False` there is no reason to parametrize it

osoukup (changes requested):
I have one important and two minor comments

osidb/tests/factories.py
```diff
@@ -49,6 +49,15 @@ class Meta:
         yes_declaration=factory.Faker("future_datetime", tzinfo=UTC),
         no_declaration=factory.Faker("past_datetime", tzinfo=UTC),
     )
+    summary = factory.Maybe(
+        "embargoed",
+        yes_declaration=factory.sequence(
```
JakubFrejlach, line 54: You don't have to use `factory.Sequence` as this field actually depends on `cve_id`. I would instead suggest to use `factory.LazyAttribute` which allows you to reference the generated `cve_id`. Tt is used for `title`, `description` and `statement` for example, you can take the usage idea from there.

osidb/tests/factories.py
```diff
@@ -49,6 +49,15 @@ class Meta:
...
+            lambda n: f"EMBARGOED - CVE-2022-000{n} kernel-rt: kernel: race condition in example function [rhel-1.0.0.z]"
+        ),
+        no_declaration=factory.sequence(
```
JakubFrejlach, line 57: Same case with the `factory.sequence` -> `factory.LazyAttribute`

osidb/tests/factories.py
JakubFrejlach: Actually after reading @osoukup comment the `title` which in fact is a `summary` from SFM2 already uses the `factory.LazyAttribute` so you can just tweak it accordingly.

osidb/tests/factories.py
JakubFrejlach: Actually after reading @osoukup comment the `title` which in fact is a `summary` from SFM2 already uses the `factory.LazyAttribute` so you can just tweak it accordingly.

JakubFrejlach (changes requested):
Apart from the `summary`/`title` already pointed out by @osoukup and some suggestion to factories, LGTM!

Elkasitu:
@osoukup I am wondering if this kind of validation should only apply when grabbing info from bugzilla, I think in OSIDB we want a "normal" title, everything else has its own dedicated field, so something like e.g.

`EMBARGOED - CVE-2022-1234 kernel-rt: kernel: race condition in example function [rhel-1.0.0.z]`

would be

```python
flaw = {
    "title": "kernel: race condition in example function",
    "cve_id": "CVE-2022-1234",
    "embargoed": True,
    "ps_product": "rhel-1.0.0.z",
    "ps_component": "kernel-rt",
}
```

We don't need to do it now but something to think about, and for syncing back to bugzilla I presume we can build an acceptable title string from the flaw's attributes

osidb/tests/factories.py
costaconrado: refactored to reflect the right format

osidb/tests/factories.py
costaconrado: makes sense, changed to reflect the suggestion.

osidb/models.py
costaconrado: Refactored the validation + tests to reflect the story correctly.

osidb/tests/factories.py
costaconrado: makes sense, changed to reflect the suggestion.

osoukup:
> 

I agree. It needs some changes on the collector and backwards sync side but then in OSIDB it would make sense to use the title just for the title and have the rest as the real attributes/metadata.

Elkasitu:
@costaconrado you should update the PR description + commit message to reflect the actual changes (title)

in the meantime I merged some commits that disable OSIM tests

osidb/tests/factories.py
```diff
@@ -40,8 +40,16 @@ class Meta:
...
     impact = factory.Faker("random_element", elements=list(FlawImpact))
-    title = factory.LazyAttribute(lambda c: f"Title for {c.cve_id}")
     description = factory.LazyAttribute(lambda c: f"Description for {c.cve_id}")
+    title = factory.Maybe(
+        "embargoed",
+        yes_declaration=factory.LazyAttribute(
+            lambda c: f"EMBARGOED {c.cve_id} kernel: {c.description}"
+        ),
+        no_declaration=factory.LazyAttribute(
+            lambda c: f"{c.cve_id} kernel: {c.description}"
```
JakubFrejlach, line 50: Just for clarification: `description` is actually a comment 0 in Bugzilla and it is usually a long text so making `description` part of the `title` is not something we do. It is OK here in the factories as a data filler but it might create a false impression that `description` is part of the `title`.

JakubFrejlach: For better understanding check this [flaw](https://osidb.prodsec.redhat.com/osidb/api/v1/flaws/CVE-2013-5093?include_fields=comments,description) in OSIDB and check for the comment 0 vs description, I've already filtered only those fields.

JakubFrejlach (approved):

osidb/tests/factories.py
costaconrado: Thank you for the info! I will remove from factory to not generate misunderstandings.

PR #90, opened 2022-12-06
Add validation for flaw's CWE

costaconrado:
This PR add validation for CWE format in the Flaw model.

Closes OSIDB-357.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Change logging of celery and django to filesystem (OSIDB-418)
+- Implement validation for CWE ID chain in a Flaw (OSIDB-357)
 
 ## [2.3.3] - 2022-12-13
 ### Changed
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2,6 +2,7 @@
 draft model for end to end testing
 """
 import logging
+import re
 import uuid
 from decimal import Decimal
 from typing import Union
@@ -808,6 +809,35 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin):
         old_flaw = Flaw.objects.get(pk=self.pk)
         if not old_flaw.embargoed and self.is_embargoed:
             raise ValidationError("Embargoing a public flaw is futile")
+
+    def _validate_cwe_format(self):
+        """
+        Check if CWE string is well formated
+        """
+        cwe_data = self.cwe_id
+        # First, remove possible [auto] suffix from the CWE entry
+        # [auto] suffix means value was assigned by a script during mass update
+        if len(cwe_data) > 6 and cwe_data.endswith("[auto]"):
+            cwe_data = cwe_data[:-6]
+
+        # Then split data on arrows ->, later we will parse the elements separately
+        arrow_count = len(re.findall("->", cwe_data))
+        parsed_elements = list(filter(None, cwe_data.split("->")))
+
+        # Ensure number of elements is one bigger then count of arrows, to catch
+        # stuff like: CWE-123->
+        if len(parsed_elements) > 0 and len(parsed_elements) != (arrow_count + 1):
+            raise ValidationError(
+                "CWE IDs is not well formated. Incorrect number of -> in CWE field."
+            )
+
+        # Ensure each element is well formed, i.e. one of:
+        #   * CWE-123
+        #   * (CWE-123)
+        #   * (CWE-123|CWE-456)
+        for element in parsed_elements:
+            if not re.match(r"^(CWE-[0-9]+|\(CWE-[0-9]+(\|CWE-[0-9]+)*\))$", element):
+                raise ValidationError("CWE IDs is not well formated.")
 
     # TODO this needs to be refactored
     # but it makes sense only when we are capable of write actions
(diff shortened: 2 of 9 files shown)


osidb/tests/test_flaw.py
```diff
@@ -914,3 +914,29 @@ def test_validate_embargoing_public_flaw(self):
...
+            ("CWE-123", False),
+            ("(CWE-123)", False),
+            ("CWE-123|CWE-1234", False),
+            ("CWE-123->CWE-12324", False),
+            ("CWE-123->(CWE-12324)", False),
+            ("(CWE-123->(CWE-12324))", False),
+            ("CWE-1->(CWE-2->CWE-3->CWE-4|(CWE-5|CWE-6))", False),
+            ("CWE-1+CWE-2", True),
+            ("cwe-1->CWE-2", True),
+            ("cwe-1|CWE-2|", True),
```
osoukup, line 930: I would make it all uppercase so we really test just the trailing `|`

osoukup (approved):
I have two minor comments but generally LGTM

Elkasitu (changes requested):
Please see the JIRA task, I think there's more viable alternatives / requires a bigger discussion

osidb/models.py
```diff
@@ -809,6 +810,82 @@ def _validate_embargoing_public_flaw(self):
...
+        """
+
+        def remove_valid_operation(cwe_token, ops):
+            """
+            Replace two CWE for a placeholder representing
+            the valid operation between them
+            (eg: CWE-1|CWE-2 become CWE-PlaceHolder)
+            """
+            if len(cwe_token) < 2 or len(ops) == 0:
+                raise ValidationError(f"CWE ID is not well formated.")
```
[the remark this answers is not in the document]
costaconrado: makes sense, I will update.

osidb/tests/test_flaw.py
costaconrado: Do you mean the test case itself only have one error scenario or the validation should be not case-sensitive?

osoukup: Oh I was definitely not clear here. I meant leave `("cwe-1->CWE-2", True),` as it is as it tests the uppercase letters but I would change `("cwe-1|CWE-2|", True),` to `("CWE-1|CWE-2|", True),` so it really tests the trailing `|`

costaconrado:
Refactored this PR entirely to reimplement SFM2 validation.

osidb/models.py
```diff
@@ -809,6 +810,35 @@ def _validate_embargoing_public_flaw(self):
...
+        if len(cwe_data) > 6 and cwe_data.endswith("[auto]"):
+            cwe_data = cwe_data[:-6]
+
+        # Then split data on arrows ->, later we will parse the elements separately
+        arrow_count = len(re.findall("->", cwe_data))
+        parsed_elements = list(filter(None, cwe_data.split("->")))
+
+        # Ensure number of elements is one bigger then count of arrows, to catch
+        # stuff like: CWE-123->
+        if len(parsed_elements) != (arrow_count + 1):
```
osoukup, line 829: CWE is not a mandatory flaw attribute. You can see at https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/validation/flaw.py#L2035 that for an empty CWE string we do not perform the validation in SFM2. It seems to me that with an empty CWE we would however fail here. Please check it and eventually fix

osoukup (commented):
I have on comment which should be probably addressed or let me know if I not correct

osoukup (approved):
LGTM

osidb/tests/factories.py
```diff
@@ -34,6 +34,7 @@ class Meta:
         model = Flaw
 
     cve_id = factory.sequence(lambda n: f"CVE-2020-000{n}")
+    cwe_id = "CWE-1"
```
JakubFrejlach, line 37: Nitpick: Since `cwe_id` is not a mandatory field I would be for generating also the cases where it is empty:

```python
factory.Faker("random_element", elements=["CWE-1", ""])
```

osidb/tests/test_flaw.py
```diff
@@ -914,3 +919,26 @@ def test_validate_embargoing_public_flaw(self):
             ]
             flaw.unembargo_dt = tzdatetime(2022, 1, 1)
             flaw.save()
+
+    @pytest.mark.parametrize(
+        "cwe,should_raise",
+        [
+            (None, False),
```
JakubFrejlach, line 926: We do not represent empty string based fields (CharField, TextField, etc.)  in OSIDB by `None` (NULL in db) since it creates a mess (you can read about it [here](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.Field.null)). Currently we do have a compatibility shim https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/mixins.py#L54 which converts all `None` values in string based fields to an empty string because Bugzilla and SFM2 stores  `None` values in db. This will go away one day and thus this test will then fail since it would violate the database integrity. I would drop the test with None.

JakubFrejlach (changes requested):
Overall LGTM, just couple of nitpicks and one possible future issue.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Change logging of celery and django to filesystem (OSIDB-418)
+- Implement validation for CWE ID chain in a Flaw
```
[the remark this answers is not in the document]
costaconrado: updated, thank you!

osidb/tests/factories.py
costaconrado: makes sense, updated.

osidb/tests/test_flaw.py
costaconrado: understood, I have updated, thank you!

Elkasitu (approved):

PR #92, opened 2022-12-07
Document Jira automation

JakubFrejlach:
This PR documents the Jira automation and its triggers so it is clear which keywords should be included in the future PRs so the automation would transfer the Jira issues status accordingly based on the PR status.

### DIFF
--- a/docs/developer/CONTRIBUTING.md
+++ b/docs/developer/CONTRIBUTING.md
@@ -31,7 +31,33 @@ functionality, or testability of the project will generally not be accepted.
 * Open a new GitHub pull request with the patch.
 
 * Ensure the PR description clearly describes the problem and solution.
-  Include the relevant issue number if applicable.
+
+* Include the relevant issue number if applicable. There is a Jira automation
+  which changes the state of the mentioned Jira issues accordingly. Expected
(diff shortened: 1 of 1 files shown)


docs/developer/CONTRIBUTING.md
```diff
@@ -31,7 +31,24 @@ functionality, or testability of the project will generally not be accepted.
...
+  format of the issue mention starts with word `Closes` or `Fixes` (**C or F must be uppercase**)
+  followed by a list of Jira IDs:
+
+  ```
+  Closes OSIDB-111 OSIDB-222 ...
+  ```
+  OR
+  ```
+  Fixes OSIDB-111 OSIDB-222 ...
+  ```
```
Elkasitu, line 46: Actually it can only parse one JIRA ID at a time, so you would have to do something like:

```
Closes OSIDB-123
Closes OSIDB-456
Fixes OSIDB-111
```

It might be possible to do it all in a single line but we'd have to try it to be sure, because regex capture groups work in a specific manner in JIRA automations.

JakubFrejlach: Ok, so lets rather be safe here, will make a note and try it in the future and if it works I will mention it additionally.

costaconrado:
LGTM.

PR #94, opened 2022-12-07
Ensure that flake8 exits with non-zero exit code on lint failure

Elkasitu:
The --exit-zero flag does the opposite of what we want it to do, it should return a non-zero exit code on lint failure so that the CI detects it and prevents a merge.


JakubFrejlach (approved):
LGTM!

PR #95, opened 2022-12-07
Change validate CVSS3 tests name

costaconrado:
This PR changes the name of CVSS3 tests for validators.

This PR explicitly distinguish the tests by field validation and model validation fixing the problem with both tests having the same name.


Elkasitu:
@costaconrado just fyi you need to mark PRs with the `technical` label if the changes don't introduce any functionality changes that the end-user might perceive

osoukup (approved):
LGTM

PR #96, opened 2022-12-08
link unacked PS update streams

osoukup:
During the tracker data analysis it was uncovered that the trackers filed against unacked PS update streams are not being linked to the affects. The reason was the missing link from unacked PS update stream to the corresponding PS module. These links were omitted to be introduced in the product definitions collector. Also the PS update stream Django model did not count on them. This PR corrects the situation so the trackers in question can be properly linked.

There is no data migration needed as the product definitions will just fix themselves on the next sync.

Closes OSIDB-629


Elkasitu:
@osoukup since this is due for a patch release, shouldn't it be based off of `2.3.2` branch and merged into a `2.3.3-release` branch?

osoukup:
Closing in favor of https://github.com/RedHatProductSecurity/osidb/pull/97

PR #97, opened 2022-12-08
link unacked PS update streams

osoukup:
During the tracker data analysis it was uncovered that the trackers filed against unacked PS update streams are not being linked to the affects. The reason was the missing link from unacked PS update stream to the corresponding PS module. These links were omitted to be introduced in the product definitions collector. Also the PS update stream Django model did not count on them. This PR corrects the situation so the trackers in question can be properly linked.

There is no data migration needed as the product definitions will just fix themselves on the next sync.

Closes OSIDB-629

This PR replaces https://github.com/RedHatProductSecurity/osidb/pull/96


osidb/migrations/0063_link_unacked_ps_update_stream.py
```diff
@@ -0,0 +1,29 @@
...
+
+    dependencies = [
+        ("osidb", "0062_alter_flawmeta_flaw"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="psmodule",
+            name="unacked_ps_update_stream",
+        ),
```
Elkasitu, line 17: this is not N-1 compatible, we should make it nullable and remove it in the next release

Elkasitu (changes requested):

osidb/migrations/0063_link_unacked_ps_update_stream.py
osoukup: That sounds weird to me. I am making the relation out of the former string field not just removing it. Do you mean that I should first make it nullable, then remove it, and only then introduce it again? That would be at least three releases :smile: I am also not sure about the incompatibility as I am not changing any existing functionality (just extending it so it creates one more type of relations) and it works - so it works with both the old and the new models.

osoukup: Looking at the code making the field nullable does not seem to me to change much. It would be still being set be the collector. I think making it N-1 compatible here would mean to specifically excluding the field eg. at 
https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/product_definitions/core.py#L125
then the release so the field would be unused and the next release could change it to the relation and the previous would not care. Or WDYT?

Elkasitu: My bad, I didn't see that the `related_name` was set for the FK, in that case I'm afraid that it's even more complicated [1].

N-1 compatibility is so that in case that something goes wrong with the new version (N) deployment we can safely go back to version N-1, if we go back the column will still be there in the database but the type will be completely different! (actually I'm not even sure that the column will still be there since the column will technically be in the `psupdatestream` table, don't think Django creates columns in the reverse relationship table).

You don't have to be fully compliant with the aforementioned link, I think the following steps are enough:

1. keep `unacked_ps_update_stream` and change to `null=True`
2. create the FK in `PsUpdateStream` but give it a different `related_name` (e.g. `unacked_ps_update_stream_{id,obj}`)
3. change all occurrences in the code of `unacked_ps_update_stream` for `unacked_ps_update_stream_{id,obj}.name`

Then in the next version we can remove `unacked_ps_update_stream`.

[1] https://develop.sentry.dev/database-migrations/#altering-column-types

osoukup: I still think that in this case it could be simpler because of the reason that the field is nearly unused - it is only being set in the collector but never read or used anyhow. If I exclude it from being set in the collector it will be completely unused. I can make it nullable on top of it.

Then releasing this as 2.3.3 would make this version not care about the field at all so I think it should work with the old and the new DB no problem. The next version 2.3.4 could be this PR. If the things go wrong during the migration we roll back to the old app with the old DB. If the things go wrong afterwards we have new DB and the old app which does not use the field. Finally we can have new DB and new app if everything goes well. Or is there any issue with this?

Elkasitu (approved):

osidb/migrations/0063_link_unacked_ps_update_stream.py
osoukup: Discussed on the Gchat...

osoukup:
@Elkasitu now it should be all according to our discussion. Thank you for helping to shape it and please have a one more quick look

PR #98, opened 2022-12-08
Validate empty affects

costaconrado:
This PR validates that flaws have an affect related.

To achieve it this PR adds a validation that allows Flaw creation without affect but raises error for any edition in flaw without affect, it also changes some tests enforcing the association of a new affect with flaw

Closes OSIDB-353.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -754,6 +754,17 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin, ACLMixi
             if not re.match(r"^(CWE-[0-9]+|\(CWE-[0-9]+(\|CWE-[0-9]+)*\))$", element):
                 raise ValidationError("CWE IDs is not well formated.")
 
+    def _validate_flaw_without_affect(self):
+        """
+        Check if flaw have at least one affect
+        """
+        # Skip validation at creation allowing to draft a Flaw
+        if self._state.adding:
+            return
+
+        if not Affect.objects.filter(flaw=self).exists():
+            raise ValidationError("Flaw does not contain any affects.")
+
     def _validate_no_placeholder(self):
         """
         restrict any write operations on placeholder flaws
--- a/osidb/tests/test_core.py
+++ b/osidb/tests/test_core.py
@@ -6,7 +6,7 @@ from osidb.api_views import get_valid_http_methods
 from osidb.core import set_user_acls
 from osidb.exceptions import OSIDBException
 from osidb.models import FlawMeta
-from osidb.tests.factories import FlawFactory, FlawMetaFactory
+from osidb.tests.factories import AffectFactory, FlawFactory, FlawMetaFactory
 from osidb.tests.models import TestAlertModel, TestAlertModelBasic
 
 pytestmark = pytest.mark.unit
@@ -26,6 +26,7 @@ class TestCore(object):
             type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
             meta_attr={"status": "+"},
         )
+        AffectFactory(flaw=flaw1)
         assert flaw1.save() is None
         assert "test" in flaw1.meta_attr
 
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -72,12 +72,11 @@ class TestFlaw:
             type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
             meta_attr={"status": "+"},
         )
-
-        assert vuln_1.save() is None
         assert vuln_1.is_major_incident
 
         affect1 = AffectFactory(flaw=vuln_1)
         all_trackers = affect1.trackers.all()
+        assert vuln_1.save() is None
         assert len(all_trackers) == 0
 
         affect2 = Affect.objects.create_affect(
@@ -175,7 +174,7 @@ class TestFlaw:
             )
         ]
         Flaw.objects.all().delete()
-        Flaw.objects.create_flaw(
+        flaw1 = Flaw.objects.create_flaw(
             bz_id="12345",
             cwe_id="CWE-1",
             title="first",
@@ -185,7 +184,10 @@ class TestFlaw:
             acl_write=acls,
             reported_dt=timezone.now(),
             cvss3="3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N",
-        ).save()
+        )
+        flaw1.save()
+        AffectFactory(flaw=flaw1)
+
         assert Flaw.objects.count() == 1
         assert Flaw.objects.first().meta_attr["bz_id"] == "12345"
         Flaw.objects.create_flaw(
@@ -878,6 +880,8 @@ class TestFlawValidators:
             is_major_incident=is_major_incident,
         )
         flaw1.save(raise_validation_error=False)
+        AffectFactory(flaw=flaw1)
+
         if req:
             FlawMetaFactory(
                 flaw=flaw1,
@@ -1005,3 +1009,14 @@ class TestFlawValidators:
 
         # exclude collectors from restriction
         flaw.save(raise_validation_error=False)
+
+    def test_validate_flaw_without_affect(self):
+        """test that flaws without affect raises an error on editing"""
+        flaw1 = FlawFactory()
+        AffectFactory(flaw=flaw1)
+        assert flaw1.save() is None
+
+        flaw2 = FlawFactory()
+        with pytest.raises(ValidationError) as e:
+            flaw2.save()
+        assert "Flaw does not contain any affects." in str(e)
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Added
 - Implement Bugzilla SRT notes builder in Bugzilla Backwards Sync (OSIDB-384)
+- Implement validation for flaw without affect (OSIDB-353)
 
 ### Changed
 - Change logging of celery and django to filesystem (OSIDB-418)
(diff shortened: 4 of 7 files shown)


osidb/serializer.py
```diff
@@ -536,8 +537,18 @@ def get_meta(self, obj):
...
         return serializer.data
 
+    from django.db import transaction
+
+    @transaction.atomic
     def create(self, validated_data):
-        return Flaw.objects.create(**validated_data)
+        affects = validated_data.pop("affects", [])
+        flaw = Flaw(**validated_data)
+        flaw.save(raise_validation_error=False)
```
osoukup, line 546: I am not very familiar with the serializers so maybe this question is silly :smile: but does this allow the user to create a flaw without it being validated? That would not be correct. Or if it is something different, could you please explain to me what is this suppressed error raising necessary for?

osidb/tests/test_flaw.py
```diff
@@ -914,3 +928,9 @@ def test_validate_embargoing_public_flaw(self):
             ]
             flaw.unembargo_dt = tzdatetime(2022, 1, 1)
             flaw.save()
+
+    def test_validate_flaw_without_affect(self):
+        """test that embargoed flaws without affect raises an error"""
```
osoukup, line 933: nitpick: I guess that the word `embargoed` is some leftover

osidb/tests/test_embargo.py
```diff
@@ -4,7 +4,7 @@
 
 from osidb.models import Flaw
 
-from .factories import FlawFactory
+from .factories import FlawFactory, AffectFactory
```
osoukup, line 7: needs to be in the alphabetic order (isort)

osoukup (commented):
I would like to understand the serializer change and the pipelines need to be satisfied. Otherwise the changes looks good and the factory enhancements are very nice!

osidb/serializer.py
Elkasitu: I'm not even sure that this works because nested writes (i.e. writing affects via the /flaws endpoint) is [not supported by default in DRF](https://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers) and here `affects` is not a list of `Affect` objects, in fact it isn't anything because `affects` is a read-only field meaning that it will never be found in `validated_data`

This needs some more thought, the solution isn't trivial so you may skip the endpoint tests that fail for now with the `pytest.mark.skip` decorator

osidb/tests/factories.py
```diff
@@ -97,6 +97,20 @@ class Meta:
...
+        "osidb.tests.factories.FlawCommentFactory", factory_related_name="flaw", size=1
+    )
+    meta = factory.RelatedFactoryList(
+        "osidb.tests.factories.FlawMetaFactory", factory_related_name="flaw", size=1
+    )
+    package_versions = factory.RelatedFactoryList(
+        "osidb.tests.factories.CVEv5PackageVersionsFactory",
+        factory_related_name="flaw",
+        size=1,
+    )
```
Elkasitu: I'm not convinced by this, as it's not representative of how the Django ORM works

Elkasitu (commented):
This needs to go back to the drawing board, we need to determine whether:

1. This validation should always be an alert (like rh vs nvd cvss score one)
2. We want nested writes or writes in different endpoints
3. We want to make `Affect.flaw` be `null=False`
4. We want to leave some validations up to the upper service layer (e.g. flaw w/o affects) depending on the state of a Flaw's processing (workflows?)

osidb/tests/factories.py
```diff
@@ -108,7 +122,8 @@ def _create(cls, model_class, *args, **kwargs):
         # but we want preserve it as writable factory attribute as it is easier to work with
         # than with ACLs so we need to remove it for the flaw creation and emulate annotation
         embargoed = kwargs.pop("embargoed")
-        flaw = super()._create(model_class, *args, **kwargs)
+        flaw = model_class(*args, **kwargs)
+        flaw.save(raise_validation_error=False)
```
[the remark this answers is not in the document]
costaconrado: I replaced `super()._create(model_class, *args, **kwargs)` because it call a save method with validation.
Being both Affect and Flaw dependent on each other I have to first save one without validation to generate an UUID and then be able to associate each other, if I don´t generate this UUID first, when the `RelatedFactoryList` is called to generate the affect (after the entire creation of the flaw) the database will raise error due the foreign key not existing.

Also it call a save method again with the validations, (I believe it is in the postgeneration method I wasn't able to find where in the docs) but if the `FlawFactory` is passed with an validation error like `FlawFactory(embargoed=True, title="test")`, even skiping validations in `_create` method we raise the validation error `Flaw title does not contains "EMBARGOED" despite being embargoed.`

osidb/serializer.py
costaconrado: I agree, we can have a separated task to create the nested writes, I will restore the files related to serialization and submit it again with the 2 creation tests skiped.

osidb/tests/test_flaw.py
costaconrado: That is right, sorry, I will fix it.

osidb/models.py
```diff
@@ -809,6 +812,14 @@ def _validate_embargoing_public_flaw(self):
         if not old_flaw.embargoed and self.is_embargoed:
             raise ValidationError("Embargoing a public flaw is futile")
 
+    def _validate_flaw_without_affect(self):
+        """
+        Check if flaw have at least one affects entry of a product.
+        """
+        affects = Affect.objects.filter(flaw=self)
+        if not affects.all():
```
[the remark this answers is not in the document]
costaconrado: refactored, thank you.

osidb/tests/test_core.py
```diff
@@ -19,14 +19,12 @@ def test_test(self):
...
-        FlawMetaFactory(
-            flaw=flaw1,
-            type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
-            meta_attr={"status": "+"},
+        flaw1 = FlawFactory(
+            is_major_incident=True,
+            meta__type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
+            meta__meta_attr={"status": "+"},
         )
-        assert flaw1.save() is None
```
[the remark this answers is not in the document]
costaconrado: I have not only only removed the `assert flaw1.save()` but also replaced the `flaw1.save(raise_validation_error=False)` for `FlawFactory( ... )` that saves raising validation errors. 
I will add `assert flaw1` for clarity of the test.

osidb/tests/test_embargo.py
costaconrado: it is locally enables, not sure why it fail to fail. Will change thanks.

osidb/tests/factories.py
costaconrado: Not sure if we want it in the project but it was added to being able to use short cuts like this to generate multiple nested entities
```
# short cut
flaw1 = FlawFactory(
    is_major_incident=True,
    meta__type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
    meta__meta_attr={"status": "+"},
)

# original version
flaw1 = FlawFactory.build(is_major_incident=True)
flaw1.save(raise_validation_error=False)
FlawMetaFactory(
    flaw=flaw1
    meta_attr={"status": "+"},
    type=FlawMeta.FlawMetaType.REQUIRES_DOC_TEXT,
)
```

but it also allow to get data from parent entity, for example, if we want to generate a valid major incident flaw in the tests already with the the meta_attr we could refer this info in the `FlawMetaFactory` with `factory.SelfAttribute("..flaw.is_major_incident")`

I saw the counter part of this features was already added `flaw = factory.SubFactory(FlawFactory)` and thought it would be useful for tests make it goes both ways.

osoukup:
> This needs to go back to the drawing board, we need to determine whether:
> 
>     1. This validation should always be an alert (like rh vs nvd cvss score one)
> 
>     2. We want nested writes or writes in different endpoints
> 
>     3. We want to make `Affect.flaw` be `null=False`
> 
>     4. We want to leave some validations up to the upper service layer (e.g. flaw w/o affects) depending on the state of a Flaw's processing (workflows?)

@Elkasitu in the current SFM2 world there cannot be legally created a flaw without any affects (it can of course be done in raw Bugzilla) so it would be a change. I am not saying it cannot make sense in some cases in the future but it should not be our decision. In SFM2 there are very few really mandatory attributes - summary (which is OSIDB title), impact, source, reported date, affects (at least one), and comment zero (which is OSIDB description).

Title and description are obviously the basic flaw describing fields. Source and reported date are always known - we know from who/what and when we find out about the flaw. We always want to estimate the impact so we know how urgently to deal with it. Finally we want the flaw has to affect something so we know who should deal with it and that it is even relevant to us.

We can later during the analysis find out that the preliminary affected product is actually not affected at the end for some reason (like the vulnerable method is not called etc.) so we set the affectedness to notaffected.

Later when we maybe start implementing the workflows it sounds relevant that eg. flaw draft might exist without any affects as it would be the first artifact a person sees. Assessing the affected products/components would be then part of the initial triage step. For now I would be rather for going the old way and do not permit affect-less flaws by the validation restrictions (which can be relieved by the flag - eg. when loading old data from Bugzilla).

Elkasitu:
@osoukup sure, but in that case this validation depends on some changes to the API: either allowing nested writes of affects or having some field (e.g. affect_ids) through which affects can be set for a flaw after being created individually through the /affects endpoint (this has the downside of allowing the creation of "orphaned" affects).

I don't like either approach to be honest, but in any case I don't think we can proceed with this validation until this is done... unless we disable the API tests which sucks.

FlawDraft and workflows make sense (e.g. a FlawDraft cannot be promoted to full Flaw unless there is at least one affect) but idk if we want to do this here, especially since we're uncertain about workflows and such (though it is becoming more evident that we might indeed need something like this).

osoukup:
@Elkasitu you are correct. There is that semantic approach I would be for but the technical side is quite tricky. In the context of workflow it makes sense to create flaw draft through flaw API and then assign affects through the affect API. I am thinking about how to approach it now. What if we checked in the validation the mode `._state.adding` and raise the error only when it is `False`. It would mean that we permit affect-less flaw creation but no other modifications until the affects are added. The the procedure would be to create flaw first and then create affects through the affect API. We would have the hard restriction that there can be no affect without a flaw also possible with this approach as they would always be created only when the flaw already exists.

costaconrado:
This PR was completely refactored to meet our new understanding of how we should handle flaws without affects. It does not changes the factories anymore and uses a simpler solution.

osoukup (approved):
Simple and clear! LGTM

PR #100, opened 2022-12-13
Avoid setting/updating old unacked_ps_update_stream field

Elkasitu:
We will now use the unacked_ps_update_stream_tmp field instead until we reach version N+1 in which we'll be able to rename the field to what it once was.

### DIFF
--- a/collectors/product_definitions/constants.py
+++ b/collectors/product_definitions/constants.py
@@ -34,5 +34,5 @@ PS_UPDATE_STREAM_RELATIONSHIP_TYPE = (
     "default_ps_update_streams",
     "eus_ps_update_streams",
     "aus_ps_update_streams",
-    "unacked_ps_update_stream",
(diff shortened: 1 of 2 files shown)


osoukup (commented):
I see that I forgot to filter out the old name from filtered data before the PS module is created. I however do not understand why the patch is complicated. Why just not to add something like 

`
            # TODO remove the following assignment in 2.3.4 or above
            filtered_module_data = {
                key: value
                for key, value in filtered_module_data.items()
                if key != "unacked_ps_update_stream"
            }
`

bellow the original line 148. There would be just two statements to remove, nothing to change. Removal is less complicated. In your solution the statement on the line 129 is refactored and will need to be refactored back. Additionally the constants will need an update.

osoukup:
@Elkasitu never mind, I am OK with this patch

PR #105, opened 2022-12-14
Validate trackers of embargoed flaw

costaconrado:
This PR add validation making sure a public tracker cannot be associated with an embargoed flaw.

In order to achieve that was creates a validation in  `Tracker` model. This PR changed the way `Tracker` model validates its fields from `ValidateMixin` to `AlertMixin`.
This PR also adds `ACLMixin` and `ACLManagerMixin` in order to make easier to have the annotation and the property with the embargoed status without the need to handle ACLs. The current models (`Tracker`, `FlawHistory`, `Flaw`) and its managers using this approach was also changed to make use of the mixins.

Closes OSIDB-350.


Elkasitu (commented):
Not sure I agree with the validation at multiple levels, my opinion is that only the Tracker-level validation makes sense because related objects operations do not call `save()` [1] by default and thus require an extra/manual call to `save()` or `validate()` which you could just do on the tracker itself.

If we think of the common case we will first create a Flaw, then its affects and finally its trackers, it would make sense then that it would look something like

```python
Tracker.objects.create_or_update(affect=affect, ...)
# or
Tracker(affect=affect, ...)
```

In which case the validation only makes sense at the tracker level.

In terms of usability, we might want to create a mixin for the `is_embargoed` stuff

[1] https://docs.djangoproject.com/en/4.1/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

osoukup:
@Elkasitu has a good point but I do not agree entirely. The standard procedure is as he describes but it is not rare that existing trackers are being linked to a flaw later then at the time of creation - some duplicates or extra streams tracked by the engineering. However I guess that the linking many-to-many relation does not trigger the validations so it does not help here.

Linking a random public tracker to an embargoed flaw is however not the issue here. There are two possible pitfalls. One is that we accidentally file a public tracker for an embargoed flaw so the flaw information is copied to a public (or at least Red Hat internal) place. This would be definitely a issue of the filing tool and I doubt these validations would help here as the visibility would be most probably not chosen by the filing user but by the syncing backend job. The other case is unembargoing a tracker linked to a public flaw. Here the tracker validation would prevent it if it was done in OSIDB. Of course it can be done directly in Bugzilla or Jira and we are screwed.

Which takes me to one thing which makes me positive about having this validation on multiple places. That is that it would be easier to spot. As this may lead to some embargo leakage it must be taken seriously. Though better way would be to implement some notification mechanism however this is for the future. I would be for accepting this for now so when the bad things happen we at least have a good visibility of it and refine it later.

Let us however discuss it first.

osoukup (changes requested):
@costaconrado so after a short discussion with @Elkasitu I accepted his point and I agree that flaw and affect validations have little value. The tracker one is the important one as it can actually prevent some damage unlike the other two. Those two might increase the visibility of the potential incident but we should rather later invent some better mechanism. Could you please drop the flaw and affect validations and preserve just the tracker one?

osidb/models.py
```diff
@@ -1130,6 +1139,19 @@ def _validate_ps_module_new_flaw(self):
...
                 )
 
+    def _validate_tracker_flaw_accesses(self):
+        """
+        Check whether an affect of an embargoed flaw has a public tracker.
+        """
+        if (
+            self.flaw
+            and self.flaw.is_embargoed
+            and Tracker.objects.filter(affects=self, embargoed=False)
```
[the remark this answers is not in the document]
costaconrado: I have removed this validation, but FYI why I decided to do it backwards, I have forgotten for a minute the relational field gets me a queryset and not a list and because the embargoed annotated field are only available through the ModelManager in direct queries I have made it the confusing way, but your solution is definitely better.

osidb/models.py
```diff
@@ -80,29 +85,12 @@ def search_helper(
     # Order remaining results from highest rank to lowest
 
 
-class FlawHistoryManager(models.Manager):
+class FlawHistoryManager(ACLMixinManager):
     """flaw history manager"""
 
     def get_queryset(self):
```
osoukup, line 91: I wonder whether this is even necessary or it would be just enough to inherit it

osoukup (approved):
LGTM. Very nice progress! The mixins saves a lot of code

osidb/tests/test_flaw.py
```diff
@@ -941,3 +956,31 @@ def test_validate_cwe_format(self, cwe, should_raise):
...
+        if should_raise:
+            with pytest.raises(ValidationError) as e:
+                tracker.affects.add(affect)
+                tracker.save()
+            assert "Tracker is public but is associated with an embargoed flaw." in str(
+                e
+            )
+        else:
+            tracker.affects.add(affect)
+            assert tracker.save() is None
```
Elkasitu, line 986: while technically correct, I would rather have it be the "common" way i.e.

```python
with pytest.raises(..):
    TrackerFactory(affect=affect, embargoed=embargoed_tracker)
```

Also let's try to use `pytest.raises(..., match=...)` instead of the `assert ... in str(e)` pattern from now on

osidb/models.py
```diff
@@ -1188,7 +1188,24 @@ def create_tracker(affect, external_system_id, _type, **extra_fields):
...
+                            uuid.UUID(acl)
+                            for acl in generate_acls([settings.EMBARGO_READ_GROUP])
+                        ],
+                        then=True,
+                    ),
+                    default=False,
+                    output_field=models.BooleanField(),
+                )
+            )
+        )
```
Elkasitu: As a general rule of thumb we try not to add stuff in one commit then remove it in another

osidb/models.py
Elkasitu: Agreed

Elkasitu (commented):
I think the PR is solid now, just needs some polishing

Elkasitu:
As mentioned in chat, this PR will need to be rebased on top of #107 as there's a migration conflict (and 2.3.4's migration has already been applied in prod so it takes precedence over the one introduced here).

osidb/tests/test_flaw.py
costaconrado: makes sense, refactored.

osidb/models.py
costaconrado: I believe I'm missing something here but I think the option of inherit only the `ACLMixin` does not work, let me take the `Flaw` model for example:
```
class ACLMixinManager(models.Manager):
    def get_queryset(self):
    ....
    
class ACLMixin(models.Model):
    objects = ACLMixinManager()
    ....
    
class FlawManager(models.Manager):
     def create_flaw(...):
     def fts_search(...):
     ...
     
class Flaw(... , ACLMixin):
    objects = FlawManager()
```

The `objects = FlawManager()` will replace the ACLMixin manager completely, poiting it to FlawManager's  `get_queryset` method (or to default if not set) and the models are using specific managers to override other methods besides `get_queryset`, making them necessary. 

Is there other solution available here?

osidb/mixins.py
```diff
@@ -198,3 +202,34 @@ def save(self, *args, **kwargs):
 
     class Meta:
         abstract = True
+
+
+class ACLMixin:
```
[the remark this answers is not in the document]
costaconrado: refactored the suggested solution.

osidb/models.py
costaconrado: makes sense, fixed.

osidb/models.py
Elkasitu: no no, I think what @osoukup meant is that we don't need to override the `get_queryset()` method since all we do is call `super`, so let it be inherited from `ACLManagerMixin`

osoukup: Exactly :smiley_cat:

costaconrado: oh now I see. Removed the method not in use.

osidb/models.py
```diff
@@ -36,7 +41,7 @@
 
 logger = logging.getLogger(__name__)
 
-
+# TODO conrado remove this comment before PR
```
Elkasitu, line 44: reminder 😄

docs/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Change logging of celery and django to filesystem (OSIDB-418)
 - Implement validation for CWE ID chain in a Flaw (OSIDB-357)
+- Implement validation for embargoed flaws not be able to have public trackers (OSIDB-350)
+
+## [2.3.4] - 2022-12-15
+### Changed
+- Change logging of celery and django to filesystem (OSIDB-418)
+- Implement validation for CWE ID chain in a Flaw (OSIDB-357)
```
Elkasitu, line 16: should be removed, probably came from a conflict

osidb/models.py
```diff
@@ -1193,6 +1177,17 @@ class Meta:
     def __str__(self):
         return str(self.uuid)
 
+    def _validate_tracker_flaw_accesses(self):
+        """
+        Check whether an public tracker is associated with an embargoed flaw.
+        """
+        if not self.is_embargoed and Flaw.objects.filter(
```
Elkasitu, line 1184: I would call `exists()` here after the `filter()` call, functionally it's the same but we need to be careful with performance as we're making expensive joins

osidb/tests/test_flaw.py
```diff
@@ -941,3 +956,26 @@ def test_validate_cwe_format(self, cwe, should_raise):
...
+        """test Tracker model validator making sure flaws can't have a public tracker"""
+        affect = AffectFactory(flaw__embargoed=embargoed_flaw)
+        if should_raise:
+            with pytest.raises(
+                ValidationError,
+                match="Tracker is public but is associated with an embargoed flaw",
+            ):
+                TrackerFactory(affects=[affect], embargoed=embargoed_tracker)
+        else:
+            assert TrackerFactory(affects=[affect], embargoed=embargoed_tracker)
```
Elkasitu, line 981: We should have an extra test with > 1 flaw linked to a tracker, in which both are embargo, both are public and 1 embargo 1 public since the relationship is M2M

osidb/models.py
```diff
@@ -88,10 +88,6 @@ def search_helper(
 class FlawHistoryManager(ACLMixinManager):
     """flaw history manager"""
 
-    def get_queryset(self):
-        """define base queryset for retrieving flaws"""
-        return super().get_queryset()
-
```
Elkasitu, line 106: this should be removed in the first commit, you can do this with an interactive rebase [1].

[1] https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_changing_multiple

Elkasitu (commented):
Mostly nitpicks, but we got rid of a lot of duplicated code :)

osidb/models.py
costaconrado: makes sense, moved to the first commit.

osidb/tests/test_flaw.py
costaconrado: added new test cases for M2M relationship.

osidb/models.py
costaconrado: makes sense, called exists after filtering.

docs/CHANGELOG.md
costaconrado: this is true, updated to reflect current state.

osidb/models.py
costaconrado: hahaha sorry, removed.

osoukup (approved):
LGTM now. Good job!

osidb/models.py
```diff
@@ -1318,7 +1255,7 @@ def get_queryset(self):
         return super().get_queryset()
 
 
-class FlawMeta(AlertMixin, TrackingMixin):
+class FlawMeta(AlertMixin, TrackingMixin, ACLMixin):
```
JakubFrejlach, line 1270: `FlawMeta` inherits from `ACLMixin` but `FlawMetaManager` is based on default `models.Manager` instead of `ACLMixinManager` and thus it lacks some embargoed annotation for quering based on the embargoed status.

docs/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Changed
+- Implement validation for embargoed flaws not be able to have public trackers (OSIDB-350)
+
+## [2.3.4] - 2022-12-15
+### Changed
```
JakubFrejlach, line 12: This still looks wrong. There is already version 2.3.4 in the changelog just below and I believe these changes you've tagged with this duplicate version are not released yet. :thinking:

JakubFrejlach (changes requested):
Found couple of things otherwise LGTM.

docs/CHANGELOG.md
costaconrado: that is true they are not released yet, sorry, I messed up this file during a rebase, now I believe it is fixed.

Elkasitu (approved):
Other than what @JakubFrejlach mentioned, LGTM.

Thanks!

osidb/models.py
costaconrado: makes sense, added `ACLMixinManager` to FlawMetaManager heritage.

JakubFrejlach (approved):
LGTM now!

PR #108, opened 2022-12-16
SRT notes builder part #2

osoukup:
1) generalize public date SRT notes builder to build reported date too

as these are completely same in the sense of SRT notes handling
the tests are generalized too to cover both

2) implement source SRT notes builder method and tests

relates to OSIDB-384


Elkasitu (approved):
LGTM!

PR #110, opened 2022-12-16
SRT notes builder part #3

osoukup:
- implement CVSS attributes SRT notes generator
- implement CWE attribute SRT notes generator
- implement statement attribute SRT notes generator
- implement affects array attribute SRT notes generator
- implement acknowledgments array attribute SRT notes generator
- implement SRT notes data validation
- plus some minor stuff

Closes OSIDB-384

### DIFF
--- a/apps/bbsync/constants.py
+++ b/apps/bbsync/constants.py
@@ -1,0 +1,14 @@
+"""
+BBSync constants
+"""
+import os
+
+from collectors.bzimport.constants import BZ_DT_FMT_HISTORY
+
+DATE_FMT = "%Y-%m-%d"
+# these two time formats are the same
+# thus spare us defining it again
+DATETIME_FMT = BZ_DT_FMT_HISTORY
+
+# JSON schema for SRT notes flaw metadata
+SRTNOTES_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), "./srtnotes-schema.json")
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -2,13 +2,13 @@ import json
 from ast import literal_eval
 from itertools import chain
 
-from collectors.bzimport.constants import ANALYSIS_TASK_PRODUCT, BZ_DT_FMT_HISTORY
+import jsonschema
+
+from collectors.bzimport.constants import ANALYSIS_TASK_PRODUCT
 from osidb.models import Flaw, FlawImpact, FlawMeta, PsModule, Tracker
 
-DATE_FMT = "%Y-%m-%d"
-# these two time formats are the same
-# thus spare us defining it again
-DATETIME_FMT = BZ_DT_FMT_HISTORY
+from .constants import DATE_FMT, DATETIME_FMT, SRTNOTES_SCHEMA_PATH
+from .exceptions import SRTNotesValidationError
 
 
 class SRTNotesBuilder:
@@ -34,6 +34,7 @@ class SRTNotesBuilder:
         """
         if self._json is None:
             self.generate()
+            self.validate()
 
         return json.dumps(self._json)
 
@@ -206,6 +207,19 @@ class SRTNotesBuilder:
         srtnotes = self.flaw.meta_attr.get("original_srtnotes")
         self._json = json.loads(srtnotes) if srtnotes else {}
         self._original_keys = list(self._json.keys())
+
+    def validate(self):
+        """
+        validation safeguard to ensure that we always create valid SRT notes JSON data
+        throws SRTNotesValidationError exception in the case of invalid SRT notes JSON
+        """
+        with open(SRTNOTES_SCHEMA_PATH) as schema_fp:
+            self.srtnotes_schema = json.load(schema_fp)
+
+        try:
+            jsonschema.validate(self._json, schema=self.srtnotes_schema)
+        except jsonschema.ValidationError:
+            raise SRTNotesValidationError("Invalid JSON produced for SRT notes")
 
 
 class BugzillaQueryBuilder:
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -64,6 +64,7 @@ class SRTNotesBuilder:
         """
         self.restore_original()
         self.generate_cvss()
+        self.generate_cwe()
         self.generate_date("unembargo_dt", "public")
         self.generate_date("reported_dt", "reported")
         self.generate_impact()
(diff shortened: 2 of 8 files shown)


JakubFrejlach:
Nitpick: second commit (293165587162935891ddfcac04eb85d19f9f441a) message is too long and is cut by the GitHub UI, according to the git best practices the maximal length of the git commit message should be 50 characters, we should probably try follow it.

apps/bbsync/query.py
```diff
@@ -143,6 +131,15 @@ def generate_source(self):
...
         self.add_conditionally("source", source)
 
+    def generate_string(self, flaw_attribute, srtnotes_attribute):
+        """
+        generate given string attribute
+        generic generator for string attributes with no special handling
+        """
+        self.add_conditionally(
+            srtnotes_attribute, getattr(self.flaw, flaw_attribute) or None
+        )
```
JakubFrejlach, line 193: So lets say in `cvss3` case where this field is empty, we call `getattr(self.flaw, "cvss3")` we got empty string `""` which evaluates to `False` in instead of that the `None` value is passed. Is there any reason for it? We have the empty representation, why do we want to pass over different empty representation? :thinking: Maybe I am missing something.

apps/bbsync/tests/test_query.py
```diff
@@ -85,6 +86,51 @@ def test_restore_original(self):
         for key in srtnotes_json.keys():
             assert cf_srtnotes_json[key] == srtnotes_json[key]
 
+    def test_generate_acknowledgments(self):
+        """
+        test generating of SRT acknowledgments attribute array
+        """
+        flaw = FlawFactory()
+        FlawCommentFactory(flaw=flaw)
+        FlawMetaFactory(flaw=flaw)  # not an ack
```
JakubFrejlach, line 95: Nitpick: I would probably set some actual type. I know that the `FlawMetaFactory` currently produces all the FlawMetas with `REFERENCE` type but we might change this in the future to generate from the allowable choices so I would explicitly state some other `FlawMetaType` here.

apps/bbsync/tests/test_query.py
```diff
@@ -85,6 +86,51 @@ def test_restore_original(self):
...
+        FlawMetaFactory(
+            flaw=flaw,
+            type="ACKNOWLEDGMENT",
```
JakubFrejlach, line 98: I would use `FlawMetaType` text choice instead of the string.

apps/bbsync/tests/test_query.py
```diff
@@ -85,6 +86,51 @@ def test_restore_original(self):
...
+            meta_attr={
+                "affiliation": "dear",
+                "from_upstream": False,
+                "name": "sir",
+            },
+        )
+        FlawMetaFactory(
+            flaw=flaw,
+            type="ACKNOWLEDGMENT",
```
JakubFrejlach, line 107: Would use `FlawMetaType` text choice instead of the string.

JakubFrejlach (changes requested):
Couple of nitpicks and one general question otherwise LGTM.

osoukup:
> 

Shortened.

apps/bbsync/query.py
```diff
@@ -68,6 +68,7 @@ def generate(self):
         self.generate_date("reported_dt", "reported")
         self.generate_impact()
         self.generate_jira_trackers()
+        # TODO implement generate_references when OSIDB-71 is implemented
```
Elkasitu, line 73: same as before

apps/bbsync/query.py
```diff
@@ -72,6 +72,8 @@ def generate(self):
         self.generate_source()
         self.generate_string("cvss2", "cvss2")
         self.generate_string("cvss3", "cvss3")
+        # TODO implement CVSS comment generator when OSIDB-377 is implemented
```
Elkasitu, line 77: ditto

apps/bbsync/query.py
```diff
@@ -80,6 +81,25 @@ def generate(self):
...
+        generate array of acknowledgments to SRT notes
+        """
+        self.add_conditionally(
+            "acknowledgments",
+            [
+                {
+                    "affiliation": meta.meta_attr["affiliation"],
+                    # hstore holds bolean values as strings containing True|False
+                    # so we need to explicitly convert it to the bolean value
+                    "from_upstream": meta.meta_attr["from_upstream"] == "True",
```
Elkasitu, line 96: might be better to use `ast.literal_eval`

apps/bbsync/constants.py
```diff
@@ -0,0 +1,14 @@
...
+
+from collectors.bzimport.constants import BZ_DT_FMT_HISTORY
+
+DATE_FMT = "%Y-%m-%d"
+# these two time formats are the same
+# thus spare us defining it again
+DATETIME_FMT = BZ_DT_FMT_HISTORY
+
+# JSON schema for SRT notes flaw metadata
+SRTNOTES_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), "./srtnotes-schema.json")
```
Elkasitu, line 14: does the schema change anymore? is it OK (from a security POV) to embed this into OSIDB?

apps/bbsync/query.py
```diff
@@ -201,6 +202,21 @@ def restore_original(self):
...
+        """
+        with open(SRTNOTES_SCHEMA_PATH) as schema_fp:
+            self.srtnotes_schema = json.load(schema_fp)
+
+        try:
+            jsonschema.validate(self._json, schema=self.srtnotes_schema)
+        except jsonschema.ValidationError as e:
+            raise SRTNotesValidationError(
+                f"Invalid JSON produced for SRT notes: {e}"
+            ) from e
```
Elkasitu, line 218: exception chaining is automatic in Python 3, i.e. you don't need to do the `from e` nor embed it into the exception string

Elkasitu (approved):
LGTM overall, just some nitpicks

apps/bbsync/query.py
JakubFrejlach: Ahhh, ok, I see now. Thanks for the explanation. LGTM then.

apps/bbsync/tests/test_query.py
```diff
@@ -82,6 +82,85 @@ def test_restore_original(self):
...
+    ):
+        """
+        test generating of SRT notes CVSS attributes
+        """
+        flaw = FlawFactory.build(
+            cvss2=osidb_cvss2,
+            cvss3=osidb_cvss3,
+            meta_attr={"original_srtnotes": srtnotes},
+        )
+        flaw.save(raise_validation_error=False)
```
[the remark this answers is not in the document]
JakubFrejlach: That's a good idea, I like that. Decorators are neat way to solve things like this.

JakubFrejlach (approved):

Elkasitu (approved):

PR #111, opened 2022-12-20
Misc fixes

Elkasitu:
This PR introduces the following changes:

* A fixup to the `osidb.acl` custom parameter being set on connection start
* Bump of the `Dockerfile`'s ubi8 version to 8.7 (latest)

### DIFF
--- a/config/settings_local.py
+++ b/config/settings_local.py
@@ -70,7 +70,7 @@ DATABASES = {
             "gssencmode": "disable",
             # this is a hack due to our inability to set a custom parameter either at
             # the database or role level in managed databases such as AWS RDS
-            "options": "-c osidb.acl=''",
+            "options": "-c osidb.acl=00000000-0000-0000-0000-000000000000",
         },
     }
 }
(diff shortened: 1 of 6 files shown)


config/settings_local.py
```diff
@@ -70,7 +70,7 @@
             "gssencmode": "disable",
             # this is a hack due to our inability to set a custom parameter either at
             # the database or role level in managed databases such as AWS RDS
-            "options": "-c osidb.acl=''",
+            "options": "-c osidb.acl=00000000-0000-0000-0000-000000000000",
```
osoukup, line 73: I trust you this is correct but could you please explain what this value is? :smile:

osoukup (approved):
LGTM

config/settings_local.py
Elkasitu: It's the UUID equivalent of `NULL` or empty `''` [1]

[1] https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID

osoukup: Oh I see! Some Postgres magic :smile_cat:

JakubFrejlach (approved):

PR #112, opened 2023-01-10
errata timestamps

osoukup:
The timestamps of the errata were not explicitly set on the collector sync so they were implicitly assigned to the current time without any relation to the actual erratum creation or update times. This PR fixes this plus refactors the collector and the tests accordingly.  One downside is that now for every erratum being synced we need to perform an extra API query however it was necessary as both the query for the initial sync and the one for the periodic sync contains only a single timestamp in response while need both of them.

I have to semi-manually refactor some cassettes as I had issues with the Kerberos auth in the container for some (not all) queries and at the same time some test cases would produce thousands of queries in the cassettes so I artificially cut them to some meaningful numbers. It should still test the same functionality just not five thousand times but eg. three times.

Closes OSIDB-453


costaconrado (approved):
LGTM. Thanks for putting expected warnings in silent mode.

collectors/errata/core.py
```diff
@@ -32,24 +32,36 @@ def get(path, return_json=True, session=None, **request_kwargs):
...
+def get_all_errata() -> list[dict]:
     """
     Fetches IDs for all Errata with CVEs when collector is run for first time
     """
-    all_errata_with_cves = get("/cve/list.json").items()
-    return [
-        (errata_id, errata["advisory"]) for errata_id, errata in all_errata_with_cves
-    ]
+    all_errata_with_cves = get("/cve/list.json").keys()
+    return [get_erratum(errata_id) for errata_id in all_errata_with_cves]
```
Elkasitu, line 56: Could we maybe raise a ticket against ET people to include `updated_at` in the list endpoint? For now we'll have to use this ofc

Elkasitu (approved):
Some minor comments, otherwise LGTM

collectors/bzimport/tests/conftest.py
```diff
@@ -70,3 +70,14 @@ def bz_tracker_collector():
 @pytest.fixture
 def bz_bug_requires_doc_text():
     return "2086753"
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
osoukup: @Elkasitu I am dropping that commit so you can go with your solution. No idea what is your way here :smile: but I did not want to just turn it off globally in the case there were some meaningful/important logs. Most of them are IMO just not very well written tests where something which the code wants to work with (mostly product definitions) is omitted to be defined. There are however also some test cases like the errata one I silenced which really test some error situation so the logs are expected and can be silenced.

collectors/errata/core.py
osoukup: I created a request for them. It is however not just this endpoint but also XMLRPC call we use. We will see how they will be willing to deal with it. If not we will just live with more queries.

osoukup: https://issues.redhat.com/browse/RHELWF-8322

PR #113, opened 2023-01-11
Jira tracker timestamps

osoukup:
The timestamps of the Jira trackers were not explicitly set on the collector sync so they were implicitly assigned to the current time without any relation to the actual tracker creation or update times. This PR fixes this plus extends the collector and the tests accordingly.

Closes OSIDB-14


costaconrado (approved):
LGTM.

Elkasitu (approved):

PR #114, opened 2023-01-11
restrict write operations on placeholder flaws

osoukup:
The placeholder flaws handling is non-standard (mainly due to visibility differences) and this concept is being deprecated. Currently there are 154 such flaws according to SFM2 central which is marginal.

Closes OSIDB-388

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1,6 +1,7 @@
 """
 draft model for end to end testing
 """
+import json
 import logging
 import re
 import uuid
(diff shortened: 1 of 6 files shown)


costaconrado (approved):
LGTM.

osidb/models.py
```diff
@@ -840,6 +842,28 @@ def _validate_cwe_format(self):
...
+
+    @property
+    def is_placeholder(self):
+        """
+        placeholder flaws contain a special Bugzilla keyword
+        """
+        return FLAW_PLACEHOLDER_KEYWORD in json.loads(
+            self.meta_attr.get("keywords", "[]")
+        )
+
```
Elkasitu, line 778: This effectively means that updates of placeholder flaws from Bugzilla will not succeed, at that point isn't it better to simply discard them in bzimport?

Basically I'm not sure what's the goal of these "placeholder flaws" in OSIDB, do we want to track them at all?

Elkasitu (approved):

osoukup:
@RedHatProductSecurity/osidb-devs just FYI if you care to review again I turned the safe restriction (a completely new mechanism) to just a validation which works the same way but is a simpler/cleaner approach.

PR #115, opened 2023-01-12
Testing setup quality of life improvements

Elkasitu:
This PR introduces two changes to improve the way that we do testing:

* Logs emitted by code that is run during tests are automatically captured by pytest and only shown if the test in question fails, if the test succeeds then the logs are silenced.
* OSIM tests are disabled in local testing entrypoints so as to match the CI behavior


Elkasitu:
@osoukup when the CI is done, you can take a look at the generated logs from pytest and let me know what you think, we can go back to ignoring the "warnings summary" but I believe that one is actually useful

osoukup (approved):
LGTM

PR #116, opened 2023-01-12
Validate high criticality flaw changes

costaconrado:
This PR validates unsupported changes in flaws with high criticicity.

Flaws that are `major incidents` or have impact level of `CRITICAL` or `IMPORTANT` cannot be changed to `non-major incident` / `LOW` / `MODERATE` if they have not `closed` trackers.

Closes OSIDB-347.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -765,6 +765,36 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin, ACLMixi
         if not Affect.objects.filter(flaw=self).exists():
             raise ValidationError("Flaw does not contain any affects.")
 
+    def _validate_unsupported_impact_change(self):
+        """
+        Check that an update of a flaw with open trackers does not change between
+        Critical/Important/Major Incident and Moderate/Low and vice-versa.
+        """
+        if self._state.adding:
+            return
+
+        old_flaw = Flaw.objects.get(pk=self.pk)
+        was_high_impact = (
+            old_flaw.impact in [FlawImpact.CRITICAL, FlawImpact.IMPORTANT]
+            or old_flaw.is_major_incident
+        )
+        is_high_impact = (
+            self.impact in [FlawImpact.CRITICAL, FlawImpact.IMPORTANT]
+            or self.is_major_incident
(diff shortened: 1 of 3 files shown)


osidb/models.py
```diff
@@ -796,6 +796,36 @@ def _validate_cwe_format(self):
...
+            self.impact in [FlawImpact.CRITICAL, FlawImpact.IMPORTANT]
+            or self.is_major_incident
+        )
+        if (
+            was_high_impact
+            and not still_high_impact
+            and Tracker.objects.filter(affects__flaw=self)
+            .exclude(status="Closed")
+            .exists()
+        ):
```
osoukup, line 822: I am checking the SFM2 validation and there is an important difference

`A change of impact from Critical/Important/Major Incident to Moderate/Low **and vice versa** is not supported by the tooling.`

where this validation does just one half (one direction). Do you plan to create another validation or does this one need to be adjusted? Basically instead of **and not** you might have **!=** but also the error text below would have to be adjusted accordingly so it applies also for eg. low-to-important change.

osidb/models.py
```diff
@@ -796,6 +796,36 @@ def _validate_cwe_format(self):
...
+            raise ValidationError(
```
osoukup, line 823: In the original SFM2 validation you can see that it starts by `@register.warning` which means that it is not a blocking error but rather a notification to the user that something potentially bad happened. The reason is that sometimes it is necessary to change the impact in an unsupported (by tooling) way. The user is then responsible for cleaning everything accordingly just manually. Instead of `raise` there should be just `self.alert(...`

osoukup (changes requested):
I have a two comments which should be addressed.

osidb/models.py
costaconrado: makes sense, submitting fixes.

osidb/models.py
costaconrado: Makes sense, will send fixes.

osoukup (approved):
I would maybe add a test case to verify that with closed trackers the validation is not raised but otherwise now LGTM. Thank you for incorporating my complains :smile:

PR #118, opened 2023-01-17
Add alerts for components affected by NOTABUG flaw

costaconrado:
This PR add validation alerting in case a component is being affected by a flaw closed as NOTABUG.

Closes OSIDB-363.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -807,6 +807,26 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin, ACLMixi
                 "OSIDB does not support write operations on placeholder flaws"
             )
 
+    def _validate_notabug_flaw_affected(self):
+        """
+        Alerts in case flaw is closed as NOTABUG but has affected components
+        """
+        if (
+            self.state != Flaw.FlawState.CLOSED
+            or self.resolution != FlawResolution.NOTABUG
+        ):
+            return
+
+        affected = self.affects.filter(
+            affectedness=Affect.AffectAffectedness.AFFECTED
+        ).first()
+        if affected:
+            self.alert(
+                "notabug_affect_ps_component",
+                f"Module {affected.ps_module} of component "
+                f"{affected.ps_component} is affected by a flaw solved as NOTABUG.",
+            )
+
     @property
     def is_placeholder(self):
         """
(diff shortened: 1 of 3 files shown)


osidb/models.py
```diff
@@ -807,6 +807,25 @@ def _validate_no_placeholder(self):
...
+        ):
+            return
+
+        affected = self.affects.filter(
+            affectedness=Affect.AffectAffectedness.AFFECTED
+        ).first()
+        if affected:
+            self.alert(
+                "notabug_affect_ps_component",
+                f"{affected.ps_component} is affected by a flaw solved as NOTABUG.",
```
osoukup, line 826: The component is not enough to uniquely identify the affect - from the user point of view here. It is usual that there are multiple affects with the same component but the different modules. There should be both the PS module and component in the warning message.

osidb/models.py
```diff
@@ -1050,6 +1069,20 @@ def _validate_ps_module_new_flaw(self):
...
+        Alerts in case a component is affected by a flaw closed as NOTABUG
+        """
+        if (
+            self.affectedness == Affect.AffectAffectedness.AFFECTED
+            and self.flaw.resolution == FlawResolution.NOTABUG
+            and self.flaw.state == Flaw.FlawState.CLOSED
+        ):
+            self.flaw.alert(
+                "notabug_affect_ps_component",
+                f"{self.ps_component} is affected by a flaw solved as NOTABUG.",
```
osoukup, line 1083: I would also add the PS module here for the same reason as above.

osoukup (approved):
I suggest a slight improvement of the alert messages but otherwise LGTM

osidb/models.py
costaconrado: makes sense, will add a fix.

JakubFrejlach (approved):
LGTM!

Elkasitu (approved):

osidb/models.py
```diff
@@ -807,6 +807,26 @@ def _validate_no_placeholder(self):
                 "OSIDB does not support write operations on placeholder flaws"
             )
 
+    def _validate_notabug_flaw_affected(self):
+        """
+        Alerts in case flaw is closed as NOTABUG but have affected components
```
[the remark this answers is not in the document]
costaconrado: sorry about it, fixed.

PR #119, opened 2023-01-19
Add validation for components from software collection

costaconrado:
This PR adds validation for components from software collection (RHSCL).

This PR also creates a constant in osidb/constants.py listing the components that does not follow the same naming rule of RHSCL component names.
Closes OSIDB-356.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -24,7 +24,12 @@ from apps.exploits.query_sets import AffectQuerySetExploitExtension
 from apps.osim.workflow import WorkflowModel
 from collectors.bzimport.constants import FLAW_PLACEHOLDER_KEYWORD
 
-from .constants import BZ_ID_SENTINEL, CVSS3_SEVERITY_SCALE, OSIDB_API_VERSION
+from .constants import (
+    BZ_ID_SENTINEL,
+    COMPONENTS_WITHOUT_COLLECTION,
+    CVSS3_SEVERITY_SCALE,
+    OSIDB_API_VERSION,
+)
 from .mixins import (
     ACLMixin,
     ACLMixinManager,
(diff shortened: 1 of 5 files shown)


osoukup:
BTW RHSCLs are quite pain :smile: I am dealing with them in two way sync and it has always very different special handling. This is a good example of product definitions which would need serious refactoring.

osidb/models.py
```diff
@@ -1050,6 +1055,43 @@ def _validate_ps_module_new_flaw(self):
...
+                "invalid_software_collection_component",
+                f"PSComponent {self.ps_component} for {self.ps_module} indicates collection "
+                "meta-package rather than a specific component in the collection",
+            )
+        elif not is_valid_component:
+            self.alert(
+                "invalid_software_collection_component",
+                f"PSComponent {self.ps_component} for {self.ps_module} "
+                "does not match any valid collection",
+            )
```
JakubFrejlach, line 1093: Is there a reason why originally two possible alerts which could occur at the same time (SFM2 behavior) this got transformed into only one possible alert? :thinking:

JakubFrejlach (changes requested):
I have one general question about the validation logic, otherwise LGTM.

osidb/models.py
costaconrado: oh, there is no real reason. It makes sense to be 2 separated alerts, I will submit a fix, thanks for pointing it out.

osidb/models.py
```diff
@@ -1050,6 +1055,43 @@ def _validate_ps_module_new_flaw(self):
...
+                "flaw_affects_rhscl_collection_only",
+                f"PSComponent {self.ps_component} for {self.ps_module} indicates collection "
+                "meta-package rather than a specific component in the collection",
+            )
+        elif not is_valid_component:
+            self.alert(
+                "flaw_affects_rhscl_invalid_collection",
+                f"PSComponent {self.ps_component} for {self.ps_module} "
+                "does not match any valid collection",
+            )
```
JakubFrejlach, line 1093: Well this still doesn't allow for both alerts to occur at the same time since there is the `if .. elif` statement. To completely mimic the SFM2 behavior you need to have it as a separate `if` statements as in:
https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/validation/flaw.py#L1996
https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/validation/flaw.py#L2010

JakubFrejlach (changes requested):
I believe the logic is still a bit off from the SFM2

osidb/models.py
costaconrado: Sorry I believed it would not be possible to both occurs at the same time because one validate exact `ps_component == valid_component` and the other compares `self.ps_component.startswith(valid_component + "-")` but having a list of valid components they may have both warnings, I will refactor.

costaconrado: I have refactored the validation and the tests to reflect the changes.

JakubFrejlach (approved):
LGTM now.

PR #120, opened 2023-01-19
Make first login into OSIDB fault-tolerant

Elkasitu:
On a user's first login we trigger a signal that creates a new Profile object which will link the kerberos login to Bugzilla and JIRA usernames, however this can fail for a number of reasons (expired token, service unavailable, network failure, etc.) and thus we should take care to handle such exceptions and simply let the user login without linking said credentials.

In the future we should try to sync at other points in time if possible.

### DIFF
--- a/osidb/signals.py
+++ b/osidb/signals.py
@@ -1,3 +1,4 @@
+import logging
 import os
 
 from bugzilla import Bugzilla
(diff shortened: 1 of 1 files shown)


costaconrado (approved):
LGTM.

JakubFrejlach (approved):
LGTM.

PR #124, opened 2023-01-24
Validate incomplete errata comment

costaconrado:
This PR adds validation on comments for released products without the product name.

In order to achieve it, this PR create validations in both `Flaw` and `Comment` saving process raising alerts on `Flaw` entity.
This PR also adds `AlertMixin` for `FlawComment` inheritance in order to validate during save time.
 
Closes OSIDB-359.


osidb/models.py
```diff
@@ -1496,6 +1520,22 @@ class Meta:
             "created_dt",
         )
 
+    def _validate_incomplete_errata_comment(self):
+        """
+        Alerts in case comment are missing affected product(s) for a released erratum
+        """
+        empty_erratum_comment = re.compile(r"This issue has been addressed.*\n{4}Via")
+
+        if "is_private" in self.meta_attr and self.meta_attr["is_private"] == "True":
```
Elkasitu, line 1529: same as before

osidb/models.py
```diff
@@ -832,6 +832,30 @@ def _validate_notabug_flaw_affected(self):
                 f"{affected.ps_component} is affected by a flaw solved as NOTABUG.",
             )
 
+    def _validate_incomplete_errata_comment(self):
+        """
+        Alerts in case comments are missing affected product(s) for a released erratum
```
Elkasitu, line 837: I think we need to include in these docstrings the reasoning behind the business logic, I don't quite understand what's going on here so I don't know whether the body is right or wrong

Elkasitu (changes requested):

osidb/models.py
```diff
@@ -832,6 +832,30 @@ def _validate_notabug_flaw_affected(self):
...
+        """
+        empty_erratum_comment = re.compile(r"This issue has been addressed.*\n{4}Via")
+
+        incomplete = []
+        for comment in self.comments.all():
+            if (
+                "is_private" in comment.meta_attr
+                and comment.meta_attr["is_private"] == "True"
+            ):
```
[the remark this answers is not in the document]
costaconrado: Submitted a fix.

osidb/models.py
costaconrado: Submitted a fix.

osidb/models.py
costaconrado: Added full description from SFM2 for clarification.

osidb/models.py
```diff
@@ -832,6 +833,34 @@ def _validate_notabug_flaw_affected(self):
...
+        product name. This issue has been fixed but such comments may still appear in older
+        flaws.
+        """
+        empty_erratum_comment = re.compile(r"This issue has been addressed.*\n{4}Via")
+
+        incomplete = []
+        for comment in self.comments.all():
+            if "is_private" in comment.meta_attr and ast.literal_eval(
+                comment.meta_attr["is_private"]
+            ):
```
Elkasitu, line 851: should use `get` as previously mentioned

osidb/tests/test_flaw.py
```diff
@@ -1194,3 +1196,84 @@ def test_flaw_affects_rhscl_invalid_collection(
         affect = AffectFactory(ps_module=ps_module, ps_component=ps_component)
         if alerts:
             assert set(alerts).issubset(affect._alerts)
+
+    @pytest.mark.parametrize(
+        "comment_attrs,should_raise",
+        [
+            ([("This issue has been addressed\n\n\n\nVia", "False")], True),
```
Elkasitu, line 1203: This is wrong, `"False"` should just be a boolean, that's why tests are failing and you shouldn't actually need `ast.literal_eval` since you're storing a scalar value

Elkasitu (changes requested):

osidb/models.py
Elkasitu: That helps, thanks

osoukup (commented):
I am not familiar with errata comments at all so I am not sure this validation is still needed. I asked Martin (the original author) on the ProdSec DEV chat so let us wait for his reply.

osidb/models.py
```diff
@@ -832,6 +833,34 @@ def _validate_notabug_flaw_affected(self):
...
+                continue
+
+            if empty_erratum_comment.match(comment.text):
+                incomplete.append("#" + str(comment.external_system_id))
```
JakubFrejlach, line 855: Also regarding my comment about the hstore key search, this might be also refactored to something like:
```python
empty_erratum_comment_pattern = r"This issue has been addressed.*\n{4}Via"

for comment in self.comments.filter(meta_attr__is_private=True, text__iregex=empty_erratum_comment_pattern):
    incomplete.append("#" + str(comment.external_system_id))
```

JakubFrejlach (changes requested):
Couple of suggestions

osidb/models.py
costaconrado: makes sense will fix both.

PR #127, opened 2023-01-26
Add validation for affects with wontreport status

costaconrado:
This PR adds validation for services related products with WONTREPORT resolution.

Closes OSIDB-362.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -29,6 +29,7 @@ from .constants import (
     COMPONENTS_WITHOUT_COLLECTION,
     CVSS3_SEVERITY_SCALE,
     OSIDB_API_VERSION,
+    SERVICES_PRODUCTS,
 )
 from .mixins import (
     ACLMixin,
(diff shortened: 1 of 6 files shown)


osoukup (changes requested):
I see that you extended the original validation a bit which is good but I would rather split it into several validation cases so we report these issues individually.

- one is that **wontreport** resolution can only be assigned to services
- second is that **wontreport** can only be assigned to **low** or **moderate** impact and we should consider first affect impact but when it has NOVALUE check the flaw impact instead
- third is that **wontreport** can only be combined with **affected** affectedness

For the third I would do something more general like https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/validation/flaw.py#L2138 (I hope there is a special Jira issue for it).

costaconrado:
I have submitted changes separating case 1 from 2. Case 3 is being handled by OSIDB-360 / PR125.

osidb/models.py
```diff
@@ -1157,6 +1158,39 @@ def delegated_resolution(self):
 
         return Affect.AffectFix.AFFECTED
 
+    def _validate_wontreport_products(self):
+        """
+        Validate that affected/wontreport only can be used for
+        to products associated with services
+        """
+        if (
+            self.affectedness == Affect.AffectAffectedness.AFFECTED
```
osoukup, line 1167: I would not check affectedness here at all. This should be really just about WONTREPORT resolution alone.

osidb/models.py
```diff
@@ -1157,6 +1158,39 @@ def delegated_resolution(self):
...
+                    "wontreport can only be associated with service products"
+                )
+
+    def _validate_wontreport_severity(self):
+        """
+        Validate that affected/wontreport only can be used for
+        low or moderate severity flaws
+        """
+        if (
+            self.affectedness == Affect.AffectAffectedness.AFFECTED
```
osoukup, line 1185: Same here

osoukup (changes requested):
I have still a small complain :angel:

osidb/tests/test_flaw.py
```diff
@@ -1194,3 +1195,79 @@ def test_flaw_affects_rhscl_invalid_collection(
...
+        ],
+    )
+    def test_validate_wontreport_products(
+        self, impact, affectedness, resolution, prefix, should_raise
+    ):
+        """
+        Tests that only products associated services, having a impact
+        of LOW or MODERATE, can be marked as AFFECTED by a WONTREPORT affect
+        """
+        SERVICES_PRODUCTS.append("service-product")
```
Elkasitu, line 1246: Use monkeypatch [1] instead, if you don't there's a risk that tests defined after this one that depend on `SERVICES_PRODUCTS` either directly or indirectly will use it with `"service-product"` as a valid value which might not be what we want.

[1] https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html

Elkasitu (approved):
Just some minor comments, otherwise LGTM

osidb/models.py
costaconrado: submitted a fixture.

osoukup (approved):
LGTM

PR #129, opened 2023-01-30
Add API endpoint for exploit report data

kgrant-rh:
Adds an API to export only the data necessary to generate the exploits report.


apps/exploits/api.py
```diff
@@ -344,6 +347,46 @@ def get(self, request):
...
+            .order_by("cve")
+        )
+
+        flaws = Flaw.objects.filter(exploit__in=exploits).distinct()
+        affects = Affect.objects.filter(flaw__in=flaws).distinct()
+        trackers = Tracker.objects.filter(affects_in=affects).distinct()
+
+        return Response(
+            {
+                "exploits": ExploitReportSerializer(exploits, many=True).data,
```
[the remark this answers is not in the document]
skontar: I thought that we could leverage `ListAPIView`.

apps/exploits/api.py
```diff
@@ -344,6 +347,46 @@ def get(self, request):
...
+                    "flaws": {"type": "array", "items": {"type": "object"}},
+                    "trackers": {"type": "array", "items": {"type": "object"}},
+                    "affects": {"type": "array", "items": {"type": "object"}},
+                },
+            }
+        }
+    )
+    def get(self, request):
+        exploits = (
+            Exploit.objects.filter(source="CISA")
```
skontar, line 371: We want all exploits irrespective of existing flaw, as reports need to tell if we have or do not have associated CVE. Also I am thinking if we want to filter to CISA here. Probably so for the time being.

apps/exploits/api.py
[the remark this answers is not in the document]
kgrant-rh: I see. My rationale behind providing them separately was that it would avoid duplicating information, but I suppose the only objects that would really be duplicated often are Affects. Depending on how much data we need from those, the cost of duplicated affects might even be less than the cost of including UUIDs for every object to rebuild the relationships later.

I agree that ListAPIView would be a better choice, and the only reason I didn't use it was because I had to think through how to paginate over various kinds of objects, but if the API just returns a list of exploits, that solves that problem too.

apps/exploits/api.py
```diff
@@ -344,6 +345,13 @@ def get(self, request):
         )
 
 
+class ExploitsReportData(ListAPIView):
+    """Export only the data required to generate the exploits report"""
+
+    serializer_class = ExploitReportDataSerializer
+    queryset = Exploit.objects.filter(source="CISA").order_by("cve")
```
Elkasitu, line 352: You should consider using `prefetch_related` / `select_related` [1] in combination with `only` [2] in order to optimize the query as much as possible.

The reasoning is that the endpoint's return value is deeply nested and that can result in a costly set of queries (Exploits * Flaws * Affects * Trackers).

Alternatively you can use references to IDs or hyperlinks in your serializer's relationships but that probably affects the tooling that depends on this endpoint.

[1] https://docs.djangoproject.com/en/4.1/ref/models/querysets/#prefetch-related
[2] https://docs.djangoproject.com/en/4.1/ref/models/querysets/#only

openapi.yml
```diff
@@ -2292,7 +2334,7 @@ components:
           nullable: true
         source:
           oneOf:
-          - $ref: '#/components/schemas/SourceEnum'
+          - $ref: '#/components/schemas/FlawSourceEnum'
```
Elkasitu, line 2337: some of these changes don't make sense (e.g. this PR doesn't even touch the `Flaw.source` field), are you using the version of `drf-spectacular` specified in the `requirements.txt` and did you regenerate the schema with the provided script / pre-commit hook?

kgrant-rh: I'm really not sure what happened there either. I didn't edit the schema by hand, it was generated by the pre-commit hook with the version of drf-spectacular installed by `make dev-env`. 

```
(venv) [kgrant@localhost osidb]$ pip list installed | grep drf-spectacular
drf-spectacular                   0.21.0

(venv) [kgrant@localhost osidb]$ python manage.py spectacular --settings=config.settings_local --version
3.2.15
```

If either of those is incorrect, I'll try to figure out why.

apps/exploits/serializers.py
```diff
@@ -21,6 +21,38 @@ class Meta:
...
+        fields = ["ps_module", "ps_component", "affectedness", "resolution", "trackers"]
+
+
+class FlawReportDataSerializer(serializers.ModelSerializer):
+    affects = AffectReportDataSerializer(many=True, required=False)
+
+    class Meta:
+        model = Flaw
+        # TODO: Is an OSIDB url or a BZ URL more useful here?
+        fields = ["cve_id", "api_url", "state", "resolution", "affects"]
```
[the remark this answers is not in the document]
kgrant-rh: @skontar Do you need a separate link for the flaw, or can that just be derived from the CVE ID?

apps/exploits/api.py
kgrant-rh: I added prefetching, which reduced the number of queries from 1581 to 4, and improved the performance quite a bit. I tried numerous methods of limiting the fields in the query with `only`, with nested `Prefetch()` objects and with snake specifiers like `flaw__affects__ps_module`, but they all raised exceptions due to either recursion limits or (I think) because `only` doesn't seem to work with many-to-many relationships. It seems like `defer` works, so in the worst case I think I can just include a very long defer list.

apps/exploits/serializers.py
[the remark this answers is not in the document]
skontar: I believe cve_id is enough identifier for a flaw.

apps/exploits/api.py
```diff
@@ -344,6 +347,46 @@ def get(self, request):
...
+    def get(self, request):
+        exploits = (
+            Exploit.objects.filter(source="CISA")
+            .filter(flaw__isnull=False)
+            .order_by("cve")
+        )
+
+        flaws = Flaw.objects.filter(exploit__in=exploits).distinct()
+        affects = Affect.objects.filter(flaw__in=flaws).distinct()
+        trackers = Tracker.objects.filter(affects_in=affects).distinct()
```
osoukup: There is only a single underscore so it is basically an unknown attribute **affects_in**. Plus I have no idea what does **in** operator applied on many-to-many relation so please double-check that this really works as you expect.

apps/exploits/api.py
```diff
@@ -16,17 +16,15 @@
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from osidb.models import Affect, Flaw, Tracker
```
osoukup: This is being added and removed in the same PR. Please use squash or fixup to keep the history clean so we can easily review what has really changed. I am pausing my review here until it is done as I see that I already spent time commenting stuff in the previous commit which is no more existing in this one.

osoukup (changes requested):
Stopping my review until squash/fixup is done.

openapi.yml
Elkasitu: It's weird but not something worth spending too much time on, I think functionally it's the same

apps/exploits/api.py
Elkasitu: That's good enough for me, the documentation does mention that `only()` and `defer()` are for very advanced use-cases so we can skip those given that `prefetch_related` already does a very good job, thanks!

Elkasitu (approved):
LGTM, needs a squash + rebase I think

P.S.: You should add an entry to the CHANGELOG as that's the bit the CI is complaining about

apps/exploits/serializers.py
```diff
@@ -21,6 +21,37 @@ class Meta:
...
+        model = Flaw
+        fields = ["cve_id", "state", "resolution", "affects"]
+
+
+class ExploitReportDataSerializer(serializers.ModelSerializer):
+    flaw = FlawReportDataSerializer()
+
+    class Meta:
+        model = Exploit
+        fields = ["date", "source", "reference", "flaw"]
```
skontar, line 52: **Tested for intended purpose:**
Please add missing fields "cve" (especially useful when flaw does not exist) and "maturity_preliminary" which will be needed for exploit sources different than CISA.

apps/exploits/api.py
```diff
@@ -344,6 +345,17 @@ def get(self, request):
         )
 
 
+class ExploitsReportData(ListAPIView):
+    """Export only the data required to generate the exploits report"""
+
+    serializer_class = ExploitReportDataSerializer
+    queryset = (
+        Exploit.objects.filter(source="CISA")
```
skontar, line 353: Change to `Exploit.objects.filter(Q(source="CISA") | Q(source="Metasploit"))` as we are quickly approaching ability to handle Metasploit.

skontar:
@kgrant-rh I am using this API for over a week as is and I think it works as expected. You can finish the MR and merge.

kgrant-rh:
@skontar @osoukup Please review.

skontar:
@skontar looks good to me. Does what I need.

osoukup (approved):
LGTM

apps/exploits/api.py
skontar: You are right and that is actually what we did in the end.

skontar:
@kgrant-rh please resolve the conflict. @osoukup please merge it when possible and proceed with release. Thanks you!

osoukup:
> @kgrant-rh please resolve the conflict. @osoukup please merge it when possible and proceed with release. Thanks you!

@skontar I see this is merged already so I guess you or Keith have the sufficient permissions. If not we should provide them.

PR #130, opened 2023-01-31
make UUID constant whenever possible

osoukup:
This PR changes the behavior of creating and deleting the flaws on Bugzilla sync. Previously we were deleting all the flaws first and then re-creating them. This was resulting in UUID changes as it was always newly generated. It is undesired when not really necessary. Basically whenever we are able to unambiguously map the existing flaws to the new ones we should consider them identical and preserve the UUID.

It the up-to-date workflow where we want to have zero or one CVE for each flaw this really means a constant UUID as the mapping is straightforward. It is problematic for the old multi-CVE flaws where there are cases when we cannot define an unambiguous mapping.

The problematic cases are when we assign multiple CVEs at once to a CVE-less flaw, when we remove one or more CVEs and add multiple CVEs at once, when we remove multiple CVEs and add one or more, and when we remove multiple CVEs at once resulting in a CVE-less flaw. For these cases we simply do not give any UUID guarantees. As it is all quite improbable operations and only applicable to very old flaws it is not a big deal.

Closes OSIDB-392

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -773,10 +773,11 @@ class FlawBugConvertor:
             for comment in self.flaw_comments
         ]
 
-    def get_flaw(self, cve_id):
+    def get_flaw(self, cve_id, full_match=False):
         """get Flaw Django model"""
         flaw = Flaw.objects.create_flaw(
             bz_id=self.bz_id,
+            full_match=full_match,
             cve_id=cve_id,
             type=FlawType.VULNERABILITY,
             meta_attr=self.get_meta_attr(cve_id),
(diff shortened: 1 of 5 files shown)


osidb/models.py
```diff
@@ -411,21 +411,25 @@ class FlawManager(ACLMixinManager):
...
+        full match means that we match on both Bugzilla and CVE ID
+        """
         try:
-            flaw = (
-                # CVE-less flaws are identified by Bugzilla ID
-                Flaw.objects.get(meta_attr__bz_id=bz_id)
-                if cve_id is None
-                else Flaw.objects.get(cve_id=cve_id)
-            )
+            cve_id = extra_fields.get("cve_id")
```
Elkasitu, line 422: `cve_id` can be `None`, would this be handled correctly in both scenarios?

Elkasitu: might require some extra tests

collectors/bzimport/convertors.py
```diff
@@ -976,26 +987,50 @@ def bug2flaws(self):
...
+        ###################
+
+        if len(self.cve_ids) == 1:
+            logger.debug(f"{self.__class__}: processing {self.cve_ids[0]}")
+
+            # if there was multiple flaws but now it is only one
+            # we remove all the previous as the mapping is ambiguous
+            # except the CVE itself where the mapping is straightforward
+            if existing_flaws.count() > 1:
+                existing_flaws.exclude(cve_id__in=self.cve_ids).delete()
```
Elkasitu, line 1001: since we know there's only one item in `self.cve_ids` it's probably much cheaper to convert this to:

```python
existing_flaws.get(cve_id=self.cve_ids[0]).delete()
```

Elkasitu: since you'll be using `self.cve_ids[0]` multiple times you might want to assign a variable to it too

docs/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix Jira tracker created and updated timestamps (OSIDB-14)
 - Fix errata created and updated timestamps (OSIDB-453)
 - Restrict write operations on placeholder flaws (OSIDB-388)
+- Make UUID constant whenever possible (OSIDB-392)
```
Elkasitu, line 24: I would maybe word this differently, something like "avoid recreating flaws on CVE id changes whenever possible" or something like that, to an end user the UUID is an implementation detail, what really changes is that flaws were being recreated unnecessarily.

When I read the PR title I thought there was something wrong with the UUIDs in the database (collision or something)

Elkasitu (changes requested):
Some minor changes required I think

costaconrado (approved):
I believe it will rely on bz_id only when cve_id is None, but I agree with @Elkasitu that a test case for this scenario would be nice. Besides that, LGTM!

collectors/bzimport/convertors.py
Elkasitu: my bad, the `exclude` remains but probably still cheaper to use direct comparison than the `IN` operator

osidb/models.py
osoukup: This is an expected and covered case. All the test cases where the alias contains no CVEs (either an empty array or some non-CVE alias) deal with it and hope I covered all possible cases here.

collectors/bzimport/convertors.py
osoukup: I will replace the IN operator

docs/CHANGELOG.md
osoukup: Good point. I will reword it with your suggestion

Elkasitu (approved):

PR #131, opened 2023-01-31
Add validation for invalid open trackers

costaconrado:
This PR implements validation for affects marked as `WONTFIX` or `NOTAFFECTED` with open trackers.

In order to achieve that was created validations in both `Affect` and `Tracker` model preventing user to associate open trackers in affects with invalid properties.

Closes OSIDB-364.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,11 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement validation for components affected by flaws closed as NOTABUG (OSIDB-363)
 - Implement validation for invalid components in software collection (OSIDB-356)
 - Implement Bugzilla metadata collector
-- Implement validation for for affects with exceptional combination of affectedness and resolution (OSIDB-361)
 - Implement validation for services related products with WONTREPORT resolution (OSIDB-362)
 - Implement validation for combinations of affectedness and resolution (OSIDB-360)
 - Implement a new API for getting a list of all supported products (PSINSIGHTS-593)
 - Implement CC list builder in Bugzilla backwards sync (OSIDB-386)
+- Implement validation for affects with exceptional combination of affectedness and resolution (OSIDB-361)
+- Implement validation for affects marked as WONTFIX or NOTAFFECTED with open trackers (OSIDB-364)
 
 ### Changed
 - Change logging of celery and django to filesystem (OSIDB-418)
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1141,7 +1141,7 @@ class Affect(
         ):
             self.alert(
                 "flaw_exceptional_affect_status",
-                f"Affect for {self.ps_module}/{self.ps_component} is in "
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} is in "
                 "a exceptional state having no affectedness.",
             )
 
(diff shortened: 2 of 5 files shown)


osoukup (approved):
LGTM

osidb/models.py
```diff
@@ -1291,6 +1319,11 @@ def fix_state(self):
                 return Affect.AffectFix.DEFER
         return Affect.AffectFix.AFFECTED
 
+    @property
+    def is_closed(self):
+        if self.status:
+            return self.status.upper() == "CLOSED"
```
Elkasitu, line 1325: Not sure whether this will cover all cases -- if SFM2 does it like this then OK, but there are definitely JIRA end states that are not called `CLOSED`

osidb/models.py
```diff
@@ -1147,6 +1147,34 @@ def _validate_exceptional_affectedness_resolution(self):
...
+        if (
+            self.resolution == Affect.AffectResolution.WONTFIX
+            and self.trackers.exclude(status="CLOSED").exists()
+        ):
+            self.alert(
+                "flaw_open_tracker_wontfix",
+                f"Affect for {self.ps_module}/{self.ps_component} is marked as "
+                "WONTFIX but has open tracker(s).",
+            )
+
```
Elkasitu, line 1177: Could we document why they're only alerts? I would expect someone changing an affect to `NOTAFFECTED` via the API to get an error if there are any open trackers, do we not want that? and if so, why?

osidb/models.py
costaconrado: It is how (SFM2 defined)[https://git.prodsec.redhat.com/devops/sfm2/-/blob/master/sfm2/dmodels/abstract/__init__.py#L344] it, but I can do further implementation, how I can find out JIRA statuses options?

osoukup: It is a wild west there but as this is the SFM2 implementation and nobody complains I would stick with it for now.

Elkasitu: Yep we should only care about feature parity right now which is why I asked, so we can keep this as-is, we can improve it in the future if there's an actual need.

JakubFrejlach, line 1325: `status` values are set to `null=False` and `blank=False` so if condition seems kinda redundant. Also in cases where `status` would allow blank fields this would ultimately result in returning `None` which doesn't seem right as property with naming convention `is_xxx` denotes the `bool` expectations.

JakubFrejlach: Also, what is the purpose of this property? I don't see it used anywhere.

JakubFrejlach (changes requested):
Couple of things, otherwise LGTM!

osidb/models.py
Elkasitu: Right, given that, the `exclude`s in the validations should do an ILIKE instead of an exact match

costaconrado: I have modified the filters as suggested, created new validations in trackers additional to the ones in affects and changed its behaviors for blocking validations (ValidationError) as discussed in GChat.

osidb/models.py
```diff
@@ -1296,6 +1322,31 @@ def _validate_tracker_flaw_accesses(self):
...
+        """
+        Check whether notaffected products have open trackers.
+        """
+        affect = self.affects.filter(
+            affectedness=Affect.AffectAffectedness.NOTAFFECTED
+        ).first()
+
+        if not self.is_closed and affect:
+            raise ValidationError(
+                f"Affect for {affect.ps_module}/{affect.ps_component} is marked as "
```
osoukup, line 1335: This is kinda nitpick but I am thinking about the situation that the tracker belongs to multiple affects and currently (may change in the future) it means that all of these affects have the same PS module and PS component but they belong to different flaws. Therefore PS module and PS component would not be enough to uniquely identify the bogus affect. We could either add its UUID or the related flaw UUID.

osoukup (approved):
LGTM

osidb/models.py
JakubFrejlach: This is still constructed that it may return `None` in some cases which is undesirable.

JakubFrejlach (changes requested):
Reopened one conversation.

osidb/models.py
costaconrado: Makes sense, removed condition.

osidb/models.py
costaconrado: Makes sense, added affect's uuid

docs/CHANGELOG.md
```diff
@@ -12,9 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement validation for components affected by flaws closed as NOTABUG (OSIDB-363)
 - Implement validation for invalid components in software collection (OSIDB-356)
 - Implement Bugzilla metadata collector
-- Implement validation for for affects with exceptional combination of affectedness and resolution (OSIDB-361)
 - Implement validation for services related products with WONTREPORT resolution (OSIDB-362)
 - Implement validation for combinations of affectedness and resolution (OSIDB-360)
+- Implement validation for affects with exceptional combination of affectedness and resolution (OSIDB-361)
```
Elkasitu, line 17: merge conflict mistake

Elkasitu (approved):

PR #134, opened 2023-02-01
Pin black dependency when running via tox (CI)

Elkasitu:
The dependency on black defined in tox.ini was not pinned, this meant that when running on CI the latest version of Black was always being pulled whereas the version was pinned to major version 22.3.0 in the pre-commit hook.

Black major version 23 was recently released meaning that there are some backwards-incompatible changes, this together with the CI and local version desyncs meant that all CI runs would fail on the black step.


costaconrado (approved):
LGTM!

osoukup (approved):
LGTM

PR #139, opened 2023-02-03
Add a new API for supported products

jobselko:
Supported products should have a non-empty `active_ps_update_streams` field. (PSINSIGHTS-593)

### DIFF
--- a/apps/exploits/api.py
+++ b/apps/exploits/api.py
@@ -16,7 +16,7 @@ from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from osidb.models import Flaw
+from osidb.models import Flaw, PsModule
 
 from .helpers import ExploitResult
 from .models import EPSS, Exploit
@@ -24,6 +24,7 @@ from .serializers import (
     AffectExploitReportSerializer,
     EPSSSerializer,
     ExploitsSimpleSerializer,
+    SupportedProductsSerializer,
 )
 
 
@@ -351,3 +352,14 @@ class EPSSRelevant(ListAPIView):
 
     serializer_class = EPSSSerializer
     queryset = EPSS.objects.exclude(flaw=None)
+
+
+class SupportedProducts(ListAPIView):
+    """
+    API endpoint for getting a list of all supported products.
+    """
+
+    serializer_class = SupportedProductsSerializer
+    queryset = PsModule.objects.filter(active_ps_update_streams__isnull=False).distinct(
+        "name"
+    )
--- a/apps/exploits/tests/test_api.py
+++ b/apps/exploits/tests/test_api.py
@@ -4,7 +4,12 @@ import pytest
 
 from apps.exploits.helpers import store_or_update_exploits
 from osidb.models import Affect
-from osidb.tests.factories import AffectFactory, FlawFactory
+from osidb.tests.factories import (
+    AffectFactory,
+    FlawFactory,
+    PsModuleFactory,
+    PsUpdateStreamFactory,
+)
 
 pytestmark = pytest.mark.unit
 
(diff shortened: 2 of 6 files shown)


osoukup (commented):
I am not approving yet and I am suggesting a slight change but if you can justify the status qua I am not strongly insisting on it.

apps/exploits/tests/test_api.py
```diff
@@ -60,6 +65,10 @@ def load_data():
     objects = process_data_epss(EPSS_TEST_FILE)
     store_objects(objects)
 
+    ps_module = PsModuleFactory(name="ps-module")
+    PsUpdateStreamFactory(active_to_ps_module=ps_module)
+    PsUpdateStreamFactory(active_to_ps_module=None)
```
skontar, line 70: I may be wrong, but I do not think this combination testes the logic/query. It seems to me that you create one module with an update stream, and update stream without any associated module.

I think it should be more something like:
```python
    # This module is supported
    ps_module = PsModuleFactory(name="ps-module-1")
    PsUpdateStreamFactory(active_to_ps_module=ps_module)

    # This module is not supported, as there is no active update stream
    PsModuleFactory(name="ps-module-2")
```

apps/exploits/api.py
```diff
@@ -351,3 +352,12 @@ class EPSSRelevant(ListAPIView):
 
     serializer_class = EPSSSerializer
     queryset = EPSS.objects.exclude(flaw=None)
+
+
+class SupportedProducts(ListAPIView):
```
skontar, line 357: I am unsure about the name. I know I am always describing the expected result as "Supported products", which they in a way are, but it may be worth talking to @osoukup @Elkasitu if we really should describe it as "Supported ps_modules", "Supported modules" or something else, to be consistent with the PS view of product definitions.

osoukup:
@skontar @RedHatProductSecurity/osidb-devs this is not too related to the work here but I would like to open one topic. OSIDB generally has quite a strict policy for API compatibility between the versions and the related release procedures. Basically any breaking changes should be very rare and very carefully communicated etc. This is because we have several external data consumers who have their own development cycles and we cannot just break their applications randomly (in the past we already did) but we do not have complete knowledge of what they use and how.

On the other hand we were not and probably should not be generally considering this much in the case of exploits as it is just an application module above OSIDB and it is used by the same people who develop it. I mean that it would be unnecessary overhead to think whether this or that exploits API piece is future proof, announce every change to all the channels, do the code freezes to let everyone test it etc.

However we do not have any formalization of this different approach. We should maybe have a different requirements on OSIDB core functionality and the apps. Something like that the apps are experimental or at least do not guarantee such stability - or only from some specific denoted point in time/life-cycle. Maybe stating this in the README would be enough. Or maybe it should be denoted in the API schema. WDYT?

skontar:
@osoukup I fully agree to mark exploits APIs as experimental in whatever way you guys think makes sense. We definitely need some more flexibility, at least for now. We are still experimenting and prototyping.

apps/exploits/api.py
skontar: This is supposed to "replace" https://github.com/RedHatProductSecurity/osidb/blob/master/apps/exploits/constants.py#L45-L109 , only expected to be used for exploit report generation. So I guess it does not matter. I also usually consider **rhel-8** and **rhel-9** to be two different products.

skontar:
**Tested for intended purpose:**
The API returns duplicate objects. This is not a problem in practice, but it may be useful to either add fields which make these objects distinct or modify the queryset to:
```python
queryset = PsModule.objects.filter(active_ps_update_streams__isnull=False).distinct("name")
```

Elkasitu:
I think there's a myriad of ways to handle experimental APIs, none are native as the OpenAPI spec does not support some sort of `experimental` attribute [1] like the `deprecated` one, we could play around with tags [2], docstrings, OpenAPI extensions [3] and Swagger UI plugins [4], but I think ultimately the most pragmatic solution is a combination of documentation + using API versioning e.g. `/exploits/api/v1alpha/cve_map`

[1] https://github.com/OAI/OpenAPI-Specification/issues/432
[2] https://swagger.io/docs/specification/grouping-operations-with-tags/
[3] https://swagger.io/docs/specification/openapi-extensions/
[4] https://swagger.io/docs/open-source-tools/swagger-ui/customization/plugin-api/

apps/exploits/tests/test_api.py
```diff
@@ -60,6 +65,10 @@ def load_data():
...
+
```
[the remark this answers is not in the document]
jobselko: Updated.

apps/exploits/tests/test_api.py
jobselko: Updated.

apps/exploits/api.py
jobselko: Ok, so we can leave the name as it is.

@skontar Should I also update some parts in constants.py in this PR?

skontar: No other updates, CISA Exploit Monitor is dependent on the old API and functionality provided by constants.

osoukup (approved):
LGTM. I am not so knowledgeable of what is the desired functionality here so judging just code-wise

skontar (approved):

PR #140, opened 2023-02-03
Use GitHub actions for CI

Elkasitu:
See individual commits / tasks for better information on the changes

Note: the "test" workflow cannot be triggered until this is merged to master, as it depends on a workflow that must exist in master.

For OSIDB-283 see GChat

Closes OSIDB-292
Closes OSIDB-283

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,84 @@
+name: Lint
+on: [push]
+jobs:
+  detect-secrets:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: pip install tox==3.24.4
+      - name: Check for secret lekeage
+        run: tox -e secrets
+  black:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: pip install tox==3.24.4
+      - name: Check that code is formatted with black
+        run: tox -e black
+  isort:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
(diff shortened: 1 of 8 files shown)


osoukup (changes requested):
This is great! I have one note. I went through the pipelines and noticed that `migrations` and `schema` both show an error in the check output even though they seem to be successful so please check this. Maybe it is just that the `venv` checker complains and the check actually succeeds.

Elkasitu:
@osoukup yeah, basically it makes a venv check but it doesn't stop the script (i.e. it's a non-terminating error) so schema/migrations check works regardless of the error, I was debating removing it or not but might as well to avoid confusion

osoukup (approved):
Now it is perfect!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,84 @@
...
+          python-version: 3.9
+      - name: Install dependencies
+        run: pip install tox==3.24.4
+      - name: Scan codebase for security issues
+        run: tox -e bandit
+  migrations:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install required ubuntu packages
+        run: sudo apt install libkrb5-dev build-essential python3-dev libldap2-dev libsasl2-dev libssl-dev
```
JakubFrejlach, line 63: Just out of curiosity, why are these ubuntu packages not needed in the previous steps? These are django specific?

JakubFrejlach (approved):
LGTM! I am surprised how simple GH actions are.

.github/workflows/lint.yml
Elkasitu: Not really specific to Django, but the code needs to be executed in order for these two checks to run, that includes some kerberos and ldap libs that we use, otherwise the installation of the requirements will fail (if you check tox.ini these are the only two that install the entire requirements.txt et al).

The previous ones are all static checks, maybe in the future we can separate them and better take advantage of caching and such, but I was rushing so I could technically have it "done" by the end of the sprint 💃🏽 

I'm also not exactly happy with reusing the Fedora image in `test.yml` but it works for now, so the rest will be future improvements

JakubFrejlach: Ah, I see.

You mean using the fedora image inside ubuntu? :smile:

Elkasitu: yep

PR #141, opened 2023-02-06
Fix Schema YAML link and missing closing element </div>

jobselko:
Fixes two small issues in `osid/osidb/templates/index.html` - Schema YAML link pointed to JSON instead of YAML, and closing element `</div>` was missing.


osoukup (approved):
LGTM

PR #142, opened 2023-02-07
Add validation on flaw title format

costaconrado:
This PR adds validation on flaw title format.

This PR enforces all flaws to have a title formatted in the following pattern: 
`[EMBARGOED] [CVE IDs [...]] <component name>: <description of vulnerability nature>`.
In order to achieve that some tests were edited to use the correct format.

Closes OSIDB-331.

### DIFF
--- a/collectors/bzimport/tests/test_collectors.py
+++ b/collectors/bzimport/tests/test_collectors.py
@@ -33,7 +33,7 @@ pytestmark = pytest.mark.unit
 class TestBugzillaQuerier:
     def test_remove_testing(self):
         flaw1 = FlawFactory(
-            title="regular flaw", embargoed=False, meta_attr={"bz_id": "321"}
+            title="regular: flaw", embargoed=False, meta_attr={"bz_id": "321"}
         )
         flaw2 = FlawFactory(
             title="testing: flaw", embargoed=False, meta_attr={"bz_id": "123"}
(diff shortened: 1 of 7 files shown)


osoukup (approved):
LGTM.

I have a small complain that I like the SFM2 error messages a bit more. You are correct that anything missing (or not recognized) means an invalid title. However it seems to me that better is to be more specific. In the first case really mention that the component is missing (or was not recognized) and in the second one that the description is missing/empty.

Anyway in the future when we get rid of Bugzilla there will be no more need to have EMBARGOED in the title same as the list of CVEs and the component should become an ordinary attribute so this should be all temporary. Then it would be enough to just check non-emptiness.

Elkasitu (changes requested):
I don't think this should be a validation at all, I have already expressed why but the tl;dr is that whenever we enable writing directly to OSIDB (which is presumably coming soon since we've done the bbsync work) we'll have to remove this validation because with it clients can't create proper titles (e.g. "Buffer overflow in libcurl/curl")

We can check that the title conforms to the bugzilla "spec" by performing the check in bzimport

osoukup:
> I don't think this should be a validation at all, I have already expressed why but the tl;dr is that whenever we enable writing directly to OSIDB (which is presumably coming soon since we've done the bbsync work) we'll have to remove this validation because with it clients can't create proper titles (e.g. "Buffer overflow in libcurl/curl")
> 
> We can check that the title conforms to the bugzilla "spec" by performing the check in bzimport

I agree that for the clients it is an unnecessary burden. And I agree that it is not future proof. You are maybe right and we should reconsider it. It would mean two things. The bzimport collector has to decompose the Bugzilla summary and strip some unneeded parts away (not sure what to do in the case they are not valid) and the BBSync would compose the Bugzilla summary from the title, CVE(s), embargo status, and the component. This would require adding the component as the model attribute. Maybe @costaconrado hold on this and let us discuss it.

osoukup:
@costaconrado this can be closed in favor of https://github.com/RedHatProductSecurity/osidb/pull/163

PR #143, opened 2023-02-09
Simplify GitHub Testing workflow

Elkasitu:
This commit changes the "Testing" GitHub workflow to simply run on push, just like the "Linting" one.

Previously it had been set up to only run after the "Linting" workflow completed and only if it was successful, the idea behind this was to not waste resources on testing (heavy) if the linting (not-so-heavy) failed.

While this is a very noble goal, GH workflows are not really conceived to work this way and thus when pushing the "Testing" workflow wouldn't be attached to any branch, where one would expect it to be attached to the branch that triggered the dependent workflow, this makes it hard to track the result of the "Testing" workflow.

Since we have unlimited resources for GH actions, the benefits of the previous approach do not outweigh the consequences, so we can simply run both workflows in parallel, enjoy a reduction in PR overhead and be able to easily track the result of all workflows.

### DIFF
--- a/.github/workflows/meta.yml
+++ b/.github/workflows/meta.yml
@@ -5,8 +5,7 @@ on:
     branches:
       - master
 jobs:
-  lint:
-    name: Linting
+  changelog:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM

JakubFrejlach (approved):
LGTM! I start to like these GH actions over GL CI

PR #145, opened 2023-02-10
Bugzilla save preparation

osoukup:
This PR contains various prerequisites to enable saving data through Bugzilla and generally enable the write operations through the API. There are also various random changes which I did along the way. I hope the commit messages explains the changes enough. I am creating this as a separate PR to bring it to the review sooner and in smaller pieces.

Except the random stuff there are three main types of changes.

- One is preparation for the Bugzilla API key substitution from the user to the proxied Bugzilla query (in BBSync) so the change is done by the user and not by OSIDB service.
- Second is restricting some API operations which does not make sense now. I am doing it in the backend not to break any API as it will be in the future allowed again. Please comment if you think it should be done other way.
- Third is introducing API view validations. Some things are not available on the model level - namely user data and query parameters - and some are not mandatory going from the collector unlike from the user (ACLs).

It may be not clear enough how the BBSync part and the API view part will be connected in the middle - coming soon.

Closes OSIDB-692


osidb/api_views.py
```diff
@@ -96,6 +97,123 @@ def perform_update(self, serializer) -> None:
...
+        """
+        self.validate()
+        return super().create(request, *args, **kwargs)
+
+    def update(self, request, *args, **kwargs) -> Response:
+        """
+        update model instance if valid
+        """
+        self.validate()
+        return super().update(request, *args, **kwargs)
```
Elkasitu, line 155: Please use DRF's validation mechanism [1] we do not need more custom code, in fact you should also convert the bz_api_key parameter to a a serializer field (that does not get written back to the model/db)

[1] https://www.django-rest-framework.org/api-guide/validators/#validators

osidb/api_views.py
```diff
@@ -96,6 +97,123 @@ def perform_update(self, serializer) -> None:
...
 
 
+class ValidateModelViewSetMixin(ModelViewSetMixin):
+    # mixin class defining mechanism of API validations
+    # which are unlike model validations tied to query itself
+    #
+    # TODO this is not a proper class doc string as it would generate
+    # a ton of unrelated API endpoint descriptions in OpenAPI schema
+
+    _mandatory_attrs = None
```
Elkasitu, line 107: can be handled at the serializer level with `required=True`

osidb/api_views.py
```diff
@@ -96,6 +97,123 @@ def perform_update(self, serializer) -> None:
...
+    _mandatory_params = None
```
Elkasitu, line 108: query params should never be required, just add a common serializer field that is `write_only=True` and use that for storing the `bz_api_key` like done here: https://stackoverflow.com/a/52532194

Elkasitu (changes requested):
The logic is sound, but it requires some rework to take out the custom code and use the abstractions provided by the libraries that we use unless there's a strong reason not to do so

osidb/api_views.py
```diff
@@ -315,6 +315,14 @@ def perform_create(self, serializer):
...
         serializer.save(acl_read=acls, acl_write=acls)
 
+    def destroy(self, *args, **kwargs):
+        # TODO in Bugzilla it is not possible to delete an already existing flaw
+        # so deal with it when it is no more the authoritative source of the flaw data
+        raise NotImplementedError(
+            "Bugzilla nature does not allow to delete an existing flaw "
+            "and OSIDB is not yet the authoritative source of the flaw data"
+        )
+
```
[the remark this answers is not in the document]
osoukup: You can technically achieve the delete behavior already. You can set the summary to start with `testing:` which makes both SFM2 and OSIDB ignore that flaw bug (and OSIDB additionally to delete it). However this is kinda unofficial to really exclude some testing crap and there is no process defining when one could legitimately delete a flaw.

osidb/tests/test_endpoints.py
```diff
@@ -1126,11 +1126,15 @@ def test_flaw_delete(self, auth_client, test_api_uri):
         response = auth_client.get(f"{test_api_uri}/flaws/{flaw.uuid}")
         assert response.status_code == 200
 
-        response = auth_client.delete(f"{test_api_uri}/flaws/{flaw.uuid}")
-        assert response.status_code == 204
+        with pytest.raises(NotImplementedError) as e:
+            response = auth_client.delete(f"{test_api_uri}/flaws/{flaw.uuid}")
+        assert "Bugzilla nature does not allow to delete an existing flaw" in str(e)
```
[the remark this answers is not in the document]
osoukup: Good point. I will have a look.

osidb/api_views.py
```diff
@@ -33,6 +33,74 @@
...
+        """
+        perform create by save
+        """
+        self.perform_save(serializer)
+
+    def perform_update(self, serializer) -> None:
+        """
+        perform update by save
+        """
+        self.perform_save(serializer)
```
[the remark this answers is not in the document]
osoukup: They do one thing. They unify the two cases of `create` and `update` into a single `save` so I can later define the functionality - which is equal - on just single place.

osidb/api_views.py
```diff
@@ -403,15 +403,21 @@ def perform_create(self, serializer):
...
+        """
+        raise NotImplementedError(
+            "Tracker write operations are not yet supported in OSIDB"
+        )
+
+    def destroy(self, *args, **kwargs):
+        """
+        just redirect
+        """
+        self.perform_save()
```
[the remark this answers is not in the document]
osoukup: Good point. I think that the two-way sync will possibly require a major version anyway. And I agree that code-wise it is a breaking change. From the user point of view I think we could maybe consider it a non-breaking (at least no to worry about breaking their stuff) as the servers (prod and stage) are set read-only. Of course from the bindings point of view and with respect to the local instances we should probably raise the major version number next time.

osidb/api_views.py
osoukup: Will have a look. Thank you for referencing!

osidb/api_views.py
osoukup: I will read through it. The query parameter will however somehow need to be required to successfully process the write request due to the Bugzilla audit record.

Elkasitu: @osoukup what I mean is that in DRF you can have fields that are not mapped in the model and only in the API, in this way you can have a serializer with a `bz_api_key` field that will be accepted in the JSON payload sent to the endpoint and then you can do whatever you want with it, without having it in the query param which I would argue is not even secure

osidb/api_views.py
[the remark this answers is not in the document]
Elkasitu: Agreed

osoukup:
Closing in favor of other better designed PRs. Though this was an important PoC.

PR #147, opened 2023-02-16
ACL validations

osoukup:
This MR implements ACL validation mixin class. It also makes the ACLs set-able through the API and serializers (so I can make the integration tests passing the ACL validations). There are also a lot of modified tests to pass the new validations and some new tests validating the validations themselves.

I also incorporated a few suggestions from https://github.com/RedHatProductSecurity/osidb/pull/145 so that one will probably be rebased on this one.

Closes OSIDB-691
Closes OSIDB-692

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -1,5 +1,6 @@
 import uuid
 from enum import Enum
+from functools import cached_property
 
 from django.conf import settings
 from django.contrib.postgres import fields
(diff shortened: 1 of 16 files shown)


osidb/serializer.py
```diff
@@ -287,18 +289,65 @@ class Meta:
...
-class MetaSerializer(TrackingMixinSerializer):
+@extend_schema_field(
+    {
+        "type": "array",
+        "items": {"type": "string", "minLength": 1},
+        "minItems": 1,
+        "uniqueItems": True,
+    }
+)
+class ACLSerializer(serializers.ListField):
```
Elkasitu, line 300: What's the rationale behind letting API users set ACLs? I don't think they should be able to or at least not directly, right now we should only care about setting it for them based on whether a flaw is embargoed or not

Elkasitu: In the future we might want to let them add/remove groups that can view/edit a flaw/etc. but not completely override `acl_read` and `acl_write` directly, I consider those as backend-only fields.

osidb/mixins.py
```diff
@@ -241,3 +242,179 @@ def is_embargoed(self):
 
     class Meta:
         abstract = True
+
+
+class ACLAlertMixin(ACLMixin, AlertMixin):
```
Elkasitu, line 247: Is it really necessary to couple these two? because if not, I would just implement the validations in `ACLMixin` and let implementors choose whether they want alerts or hard errors in any given model by choosing to inherit from it or not.

Elkasitu: The whole point of mixins is that you can mix and match as you choose, and in the cases in which you added this `ACLAlertMixin` the `AlertMixin` was already present, so I don't think there's a big benefit to coupling them together

Elkasitu (changes requested):
Some questions about the principle of some things but overall the logic seems sound.

osidb/serializer.py
osoukup: Sure that now we could make it with public/embargoed but this is not future proof. Later there can be arbitrary groups being probably dynamically changed by the users in some kind of product definitions (product/component related groups etc.).

What do you see in between capability of add/remove groups that can view/edit a flaw/etc. and completely overriding the ACLs? I do not see a difference.

I see the responsibility of the backend to ensure that the user only assigns the groups (s)he is allowed to to the entities (s)he is allowed to. This is my motivation to make the API very general but introduce the ACL validations.

osidb/mixins.py
osoukup: You have a valid point. When I chose to make the merging ACLAlertMixin it was based on the fact that it requires functionality of both parent mixins. I was not sure whether this functionality is always needed in the models of the parent mixins. I see that ACLMixin is still left there in the models while the AlertMixin was fully replaced. I thinking whether we might might need a model with alerts which will not have ACLs. Currently it is not a case so I may merge them. On the other hand building the functionality one on another looks more readable to me here.

osidb/serializer.py
Elkasitu: I just think that something like encapsulation works better in this particular case, if we ever need to change the `acl_read`, `acl_write` fields we can do so without breaking API, by providing an interface or abstraction at the API level to add/remove only a subset of groups we guarantee that we can change the backend without affecting the API.

osidb/mixins.py
Elkasitu: It doesn't really require functionality of the alert mixin, raising an error is perfectly fine, if you want it to treat it as an alert sometimes then you can add `AlertMixin` to the mix, it's the whole goal of composability.

If we do not want to have ACL without alerts, then might as well make `ACLMixin` inherit from `AlertMixin` and use that instead.

`FlawComment` seems to inherit from `ACLMixin` alone, btw, not sure if that's intended.

osidb/serializer.py
osoukup: So how would you rather see this implemented? Something like public/embargoed then extended to eg. some team/product terminology? The user needs to somehow specify the access/visibility and if it is not directly then indirectly through some mapping - the mentioned abstraction.

Currently I can imagine abstracting to something like embargoed boolean field. However that would be definitely broken in the future. We do not know much about what will come so I am not sure how to abstract it. Would it be like embargoed+OCP+RHEL? How would we define read vs write? To many unknowns for me. Therefore I came with not abstracting.

I am not sure how should this approach lead to breaking API. That would require that we reimplement the backend ACL mechanism so it is no more based on `acl_read` and `acl_write` which is probably big enough change worth breaking anything. We do not plan any such change.

osidb/mixins.py
osoukup: One important point of the `AlertMixin` is that you can explicitly suppress the error raising by `raise_validation_error=False` and it is used a lot in the tests and on the collector side so because of this I think any model with the custom validations should be the `AlertMixin`. If the native Django validations are enough the it can be just a `ValidateMixin` but I see a clear hierarchy there.

To the composability having ACLs is a different feature than having validation or alerts so IMO it is correct to have it defined as an independent mixin. However having validated ACLs is again a composition - of having ACLs and alerts. And as it is quite a typical situation that we require validated ACLs I made it a mixin.

Should I make it as an independent mixin? I think that not as it clearly depends on the functionality of the two. Should I make it a part of one of the existing mixins? I am not sure. It is an extended composition of them. Why would this be against composability? For your model you can choose whatever composition you want except the options violating the dependencies.

Or maybe I do not understand well now :smile_cat: The `FlawComment` is not much being handled. There is no way how to set it so I do not care about it much in the context of two-way sync.

osoukup: @Elkasitu I could imagine if we agree that every ACLs should be valid that we would merge `ACLMixin` and `ACLAlertMixin`. The resulting let us say `ACLMixin` would definitely require `AlertMixin` as a base. Then we would have a clear linear hierarchy. It would be less composable but simpler.

osidb/serializer.py
Elkasitu: > So how would you rather see this implemented? 

At the API level, we would have a single `groups` API-only field which would serve as the abstraction, a response with this field could look something like:

```json
{
    "groups": {
        "redhat": {"read": true, "write": false}
    }
}
```

When reading, the field would **not** include any of the various security-related groups, such as embargo read/write, osidb-admin groups, etc.

When writing, the field should refuse/ignore whenever a client passes any of the security-related groups that I mentioned previously, but accept any other valid LDAP group (whether the group is valid or not is up to the client), instead those security-related groups should **only** be set via a different way (like an embargo boolean).

I don't want clients to be able to set the embargo/public/admin groups or even read them directly because **they don't need to**, even if we have validations that prevent those things from happening bugs also can happen and the risk of letting them set `acl_read` and `acl_write` directly just isn't worth it.

> I am not sure how should this approach lead to breaking API.

It doesn't, but by having encapsulation we can prevent API breakage in the future, if we ever completely change the ACL system we can keep the public API as a list/object of ldap group names and we can change the backend however we want. If we just let users directly modify `acl_read` and `acl_write` then if we do decide to change anything in the future it might be more painful.

That said that's just my opinion, there's nothing technically wrong with the implementation just not what I'd prefer from a design pov.

osidb/mixins.py
Elkasitu: For me it doesn't make sense:

With your patch you have to either choose between ACL with no validations (does that even make sense, why would you allow duplicate groups, semantic errors, etc.) or ACL with validation and alerts (what if you don't want alerts?).

You can still achieve the same with something like

```python
class ACLAlertMixin(ACLMixin, AlertMixin):
    pass
```

if you define validations in `ACLMixin` and you really want the shortcut class with both components.

osoukup: I agree that ACLs without any validations makes a little sense. On the other hand I am not sure we should have validations without alerts. The reason is that such validations would always result in an exception so such model eg. could not be loaded by a collector as the some broken data are expected there but the exceptions could not be suppressed. On the other hand it could be implemented your way and all the models fetched by collectors would just use the composite `ACLAlertMixin`. I could do it this way.

osidb/serializer.py
osoukup: @Elkasitu the hard thing is that the future usage of the groups is not very clear. So far we have public or embargoed visibility and that is it. If we had embargoed plus this and that team is it still embargoed?

What would you say on the following temporary solution? I will refactor the UI to accept only the embargoed boolean for now - no ACL groups. The serializer would translate it to the corresponding ACLs. For now there would basically be two possible sets (all public read and write or all embargoed read and write groups). We would deal with the higher granularity later when we know more.

osidb/mixins.py
osoukup: I reworked the PR so now only the `ACLMixin` is extended and now new mixin is introduced.

osidb/serializer.py
osoukup: I reworked the PR so that now the API accepts `embargoed` attribute (and is mandatory) and sets the ACLs appropriately. No other options/granularity are allowed now - will be dealt with when we know more.

osidb/mixins.py
```diff
@@ -229,15 +230,186 @@ def get_queryset(self):
...
+        for acl in self.acl_read:
+            if acl not in self.acls_read:
+                raise ValidationError(
+                    f"Read ACL contains non-read ACL group: {self.acl2group(acl)}"
+                )
+
+        for acl in self.acl_write:
+            if acl not in self.acls_write:
+                raise ValidationError(
+                    f"Write ACL contains non-write ACL group: {self.acl2group(acl)}"
```
Elkasitu, line 367: nitpick: I would probably do two different validations here

osidb/tests/test_mixins.py
```diff
@@ -14,21 +16,177 @@
...
+            FlawFactory(acl_read=[])
+        assert "'acl_read': ['This field cannot be blank.'" in str(e)
+
+    def test_empty_acl_write(self):
+        """
+        test that an empty write ACL is not allowed
+        """
+        with pytest.raises(ValidationError) as e:
+            FlawFactory(acl_write=[])
+        assert "'acl_write': ['This field cannot be blank.'" in str(e)
```
Elkasitu, line 57: nitpick: we should try to use `pytest.raises(..., match="foo")` instead of `"foo" in str(e)`

Elkasitu (approved):
LGTM, good work!

osidb/mixins.py
osoukup: Good point. I will spit it.

osidb/tests/test_mixins.py
osoukup: Good suggestion. Will do.

PR #148, opened 2023-02-16
Validate unknown component

costaconrado:
This PR adds validation for flaws affecting components not tracked in Bugzilla.

This PR also makes CC list the following methods static for reuse purpose, changing some validations that could make uses of them:
- `apps::bbsync::cc.py::CCBuilder::is_community`
- `apps::bbsync::cc.py::CCBuilder::is_notaffected`
- `apps::bbsync::cc.py::CCBuilder::is_rhscl`
- `apps::bbsync::cc.py::CCBuilder::is_unknown`

Closes OSIDB-355.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -231,7 +231,7 @@
         "filename": "osidb/models.py",
         "hashed_secret": "c7e672880d394aa5dd924e04465c986652ba7291",
         "is_verified": false,
-        "line_number": 146,
+        "line_number": 148,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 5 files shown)


osoukup (changes requested):
I am suggesting a few nitpicks and one important correction. I also wonder whether it would be actually better to bring some of those affect properties really to the `Affect` model instead of having them in the CCs. I should have maybe already done it while implementing the `CCBuilder` - at least for those we might expact to  be more reusable.

osoukup (approved):
Much better! LGTM

PR #149, opened 2023-02-16
Remove prestage_eligible_date field from schemas

costaconrado:
This PR removes the field `prestage_eligible_date`.

Since this data is not used anymore in any IR process or automated process it has been removed from schemas and internal API.

Closes OSIDB-695.


osoukup (approved):
LGTM

PR #150, opened 2023-02-16
Add validation for no ACKed Flaw with private src

costaconrado:
This PR implements validation for flaws with private source without acknowledgment.
 
Closes OSIDB-339.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -856,6 +856,29 @@ class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin, ACLMixi
                     "are marked as special handling but flaw does not contain statement.",
                 )
 
+    def _validate_private_source_no_ack(self):
+        """
+        Checks that flaws with private sources have ACK.
+        """
+        if (source := FlawSource(self.source)) and source.is_private():
+            if self.meta.filter(type=FlawMeta.FlawMetaType.ACKNOWLEDGMENT).exists():
+                return
+
+            if source.is_public():
+                alert_text = (
+                    f"Flaw source of type {source} can be public or private, "
+                    "ensure that it is public since the Flaw has no acknowledgments."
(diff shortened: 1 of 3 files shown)


osidb/models.py
```diff
@@ -856,6 +856,27 @@ def _validate_special_handling_modules(self):
...
+                return
+
+            if source.is_public():
+                alert_text = f"Flaw source of type {source} can be public or private, "
+                alert_text += (
+                    "ensure that it is public since the Flaw has no acknowledgments."
+                )
+            else:
+                alert_text = f"Flaw has no acknoledgments but source of type {source} is private, "
+                alert_text += "include them in acknowledgments."
```
osoukup, line 874: nitpick: This is probably rather cosmetic by I wonder why are there two assignments instead just one eventually with a multi-line string. Maybe I am just more use to the other option :smile_cat: I am not sure about any advantages of one above the other.

osidb/tests/test_flaw.py
```diff
@@ -1512,3 +1512,12 @@ def test_special_handling_modules(self):
 
         assert "special_handling_flaw_missing_summary" in flaw2._alerts
         assert "special_handling_flaw_missing_statement" in flaw2._alerts
+
+    def test_validate_private_source_no_ack(self, private_source, public_source):
```
osoukup, line 1516: I would also add a test case for that twofold private-public source.

osoukup (approved):
LGTM. Just one minor suggestion

osidb/models.py
costaconrado: I don't have a strong preference on that, I update for a single assignment as suggested.

osidb/tests/test_flaw.py
costaconrado: Makes sense, added another test case.

PR #151, opened 2023-02-17
revise the allowed API view HTTP methods on models

osoukup:
as the OSIDB is still not an authoritative source and therefore the flaws cannot really be deleted in Bugzilla and for the trackers we do not have any write integration with the backends at all

this is a part of OSIDB-240 and re-introduces part of https://github.com/RedHatProductSecurity/osidb/pull/145 in a better way

Closing OSIDB-748

### DIFF
--- a/osidb/tests/test_endpoints.py
+++ b/osidb/tests/test_endpoints.py
@@ -1141,10 +1141,9 @@ class TestEndpoints(object):
         assert response.status_code == 200
 
         response = auth_client.delete(f"{test_api_uri}/flaws/{flaw.uuid}")
-        assert response.status_code == 204
-
-        response = auth_client.get(f"{test_api_uri}/flaws/{flaw.uuid}")
-        assert response.status_code == 404
+        # this HTTP method is not allowed until we leave Bugzilla and
+        # define the conditions under which a flaw can be deleted
+        assert response.status_code == 405
 
     def test_list_flaws_tracker_ids(self, auth_client, test_api_uri):
         """
@@ -1273,15 +1272,9 @@ class TestEndpoints(object):
         response = auth_client.post(
             f"{test_api_uri}/trackers", tracker_data, format="json"
         )
-        assert response.status_code == 201
-        body = response.json()
-        created_uuid = body["uuid"]
-
-        response = auth_client.get(f"{test_api_uri}/trackers/{created_uuid}")
-        assert response.status_code == 200
-        body = response.json()
-        assert body["status"] == "foo"
-        assert body["resolution"] == "bar"
+        # this HTTP method is not allowed until we integrate
+        # with the authoritative sources of the tracker data
+        assert response.status_code == 405
 
     def test_tracker_update(self, auth_client, test_api_uri):
         """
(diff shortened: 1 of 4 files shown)


osidb/tests/test_endpoints.py
```diff
@@ -1127,10 +1127,13 @@ def test_flaw_delete(self, auth_client, test_api_uri):
...
-        assert response.status_code == 204
+        # this HTTP method is not allowed until we leave Bugzilla and
+        # define the conditions under which a flaw can be deleted
+        assert response.status_code == 405
 
-        response = auth_client.get(f"{test_api_uri}/flaws/{flaw.uuid}")
-        assert response.status_code == 404
+        # TODO resurrect eventually
+        # response = auth_client.get(f"{test_api_uri}/flaws/{flaw.uuid}")
+        # assert response.status_code == 404
```
costaconrado, line 1136: it may be only a personal preference and thus may be ignored if you will, but  I would like to suggest maybe remove this code inside comments and open a task linking this code from any other commit, like:

https://github.com/RedHatProductSecurity/osidb/blob/de526048739ebeb72c56f17aceee1f8d7b91744a/osidb/tests/test_endpoints.py#L1132

osidb/tests/test_endpoints.py
```diff
@@ -1259,15 +1262,18 @@ def test_tracker_create(self, auth_client, test_api_uri):
...
+        assert response.status_code == 405
+        # TODO resurrect eventually
+        # body = response.json()
+        # created_uuid = body["uuid"]
+        #
+        # response = auth_client.get(f"{test_api_uri}/trackers/{created_uuid}")
+        # assert response.status_code == 200
+        # body = response.json()
+        # assert body["status"] == "foo"
+        # assert body["resolution"] == "bar"
```
costaconrado, line 1276: same as above.

osidb/tests/test_endpoints.py
```diff
@@ -1286,9 +1292,12 @@ def test_tracker_update(self, auth_client, test_api_uri):
...
         )
-        assert response.status_code == 200
-        body = response.json()
-        assert original_body["resolution"] != body["resolution"]
+        # this HTTP method is not allowed until we integrate
+        # with the authoritative sources of the tracker data
+        assert response.status_code == 405
+        # TODO resurrect eventually
+        # body = response.json()
+        # assert original_body["resolution"] != body["resolution"]
```
costaconrado, line 1300: same as above.

osidb/tests/test_endpoints.py
```diff
@@ -1300,7 +1309,10 @@ def test_tracker_delete(self, auth_client, test_api_uri):
...
-        assert response.status_code == 204
+        # this HTTP method is not allowed until we integrate
+        # with the authoritative sources of the tracker data
+        assert response.status_code == 405
 
-        response = auth_client.get(tracker_url)
-        assert response.status_code == 404
+        # TODO resurrect eventually
+        # response = auth_client.get(tracker_url)
+        # assert response.status_code == 404
```
costaconrado, line 1318: same as above.

costaconrado (approved):
Its only a personal preference but I would like to suggest managing commented code on Jira, otherwise LGTM.

Elkasitu (approved):

PR #152, opened 2023-02-17
uncomment integration tests of relative timestamp based queries

osoukup:
These were commented out temporarily but can be already put back to the test suite.


costaconrado (approved):
LGTM.

PR #153, opened 2023-02-17
remove unused Jiraffe exceptions

osoukup:
This is some historical leftover which I noticed when checking the test coverage as it has zero coverage as it is apparently not being used at all.


costaconrado (approved):
LGTM.

PR #154, opened 2023-02-23
Temporary NVD collector

JakubFrejlach:
This is a temporary solution for OSIDB-632. Collector simply runs once a day during midnight and synchronizes all desynced NVD CVSS.


JakubFrejlach:
> I wonder if wouldn't make more sense to "squash" those two commits into a single one, but LGTM.

My silly habit to include changelog into a separate commit, should get rid of it. Thanks! :smile:

osoukup (approved):
LGTM. Nice to have at least this!

Elkasitu (approved):
LGTM

PR #155, opened 2023-02-23
BBSync integration

osoukup:
This PR completes the BBSync integration into the flaw and affect save logic and introduces few more write-related changes. It is based on https://github.com/RedHatProductSecurity/osidb/pull/147. The heart of the logic is `BugzillaSyncMixin` which connects all together and holds the main conditional decisions.

Closes OSIDB-368, OSIDB-240
Closes OSIDB-104, OSIDB-105, OSIDB-106

### DIFF
--- a/apps/bbsync/mixins.py
+++ b/apps/bbsync/mixins.py
@@ -1,0 +1,46 @@
+from django.db import models
+
+from .constants import SYNC_TO_BZ
+
+
+class BugzillaSyncMixin(models.Model):
+    """
+    mixin for syncing the model to the Bugzilla
+
+    the sync between Bugzilla and the internal DB is done in a way
+    that the model is first stored to the Bugzilla and then fetched
+    back from there to make sure that the timestaps are kept in sync
+    (otherwise there might be a few seconds differences)
+
+    this mixin does not directly call Model.save
+    because it is done when fetched from Bugzilla
+    """
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, bz_api_key=None, **kwargs):
+        """
+        save the model by storing to Bugzilla and fetching back
+
+        when no Bugzilla API key is provided it is considered to be a call
+        done by a collector or test and thus we perform just regular save
+
+        Bugzilla sync is also conditional based on environment variable
+        """
+        # preliminary save to link and annotate and
+        # to make sure the validations are performed
+        super().save(*args, **kwargs)
+        self = self.__class__.objects.get(pk=self.pk)
+
+        # check BBSync conditions are met
+        if SYNC_TO_BZ and bz_api_key is not None:
+            self.bzsync(*args, bz_api_key=bz_api_key, **kwargs)
+
+    def bzsync(self, *args, bz_api_key, **kwargs):
+        """
+        Bugzilla sync of a specific class instance
+        """
+        raise NotImplementedError(
+            "Inheritants of BugzillaSyncMixin must implement the bzsync method"
+        )
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,6 +75,7 @@ services:
         ET_URL: ${ET_URL}
         PRODUCT_DEF_URL: ${PRODUCT_DEF_URL}
         DASHBOARD_URL: ${DASHBOARD_URL}
+        BBSYNC_SYNC_TO_BZ: ${BBSYNC_SYNC_TO_BZ}
       command: ./scripts/setup-osidb-service.sh
       volumes:
         - ${PWD}:/opt/app-root/src:z
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -7,6 +7,7 @@ from urllib.parse import urljoin
 
 import pkg_resources
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
@@ -352,6 +353,21 @@ class AffectView(ModelViewSet):
     filterset_class = AffectFilter
     http_method_names = get_valid_http_methods(ModelViewSet)
     permission_classes = [IsAuthenticatedOrReadOnly]
+
+    def perform_destroy(self, instance):
+        """
+        override the default behavior to proxy the delete to Bugzilla
+        """
+        flaw = instance.flaw
+        instance.delete()
+
+        # serialize Bugzilla API key and check it was provided
+        serializer = self.get_serializer(data=self.request.data)
+        bz_api_key = serializer.initial_data.get("bz_api_key")
+        if bz_api_key is None:
+            raise ValidationError({"bz_api_key": "Field is required"})
+
+        flaw.save(bz_api_key=bz_api_key)
 
 
 # until we implement tracker write operations
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -20,6 +20,7 @@ from polymorphic.models import PolymorphicModel
 from psqlextra.fields import HStoreField
 
 from apps.bbsync.constants import RHSCL_BTS_KEY
+from apps.bbsync.mixins import BugzillaSyncMixin
 from apps.bbsync.models import BugzillaComponent
 from apps.exploits.mixins import AffectExploitExtensionMixin
 from apps.exploits.query_sets import AffectQuerySetExploitExtension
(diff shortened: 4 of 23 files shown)


osoukup:
@RedHatProductSecurity/osidb-devs this PR is rebased with the ACL validation refactoring and ready for a review.

apps/bbsync/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Affect needs to be synced through flaw
+        elif isinstance(self, Affect):
+            self.flaw.save(*args, bz_api_key=bz_api_key, **kwargs)
+
+        else:
+            # no other model can be synced
+            raise UnsavableModelError(
+                f"Instance of {self._meta.model} cannot be synced to Bugzilla"
+            )
```
Elkasitu, line 67: A meta class, mixin, abstract class shouldn't be aware of its implementors, here is what I propose instead:

```python
class BugzillaSyncMixin(...):
    def save(...):
        self.bzsync(...)

    def bzsync(...):
        raise NotImplementedError("Inheritants of BugzillaSyncMixin must implement the bzsync method")


class Flaw(...):
    def bzsync(...):
        # do the sync here
        ...

class Affect(...):
    def bzsync(...):
        self.flaw.save(...)
```

Elkasitu: Note however that I don't think syncing on flaw save is a good idea, if bz is down or there are network issues we can't save a flaw, even under normal circumstances it would be too slow.

I would rather queue a bbsync job

docker-compose.yml
```diff
@@ -75,6 +75,7 @@ services:
         ET_URL: ${ET_URL}
         PRODUCT_DEF_URL: ${PRODUCT_DEF_URL}
         DASHBOARD_URL: ${DASHBOARD_URL}
+        BBSYNC_SYNC_TO_BZ: ${BBSYNC_SYNC_TO_BZ}
```
Elkasitu, line 78: requires an update to the operational repo too

Elkasitu: both PSI and MP+

osidb/api_views.py
```diff
@@ -353,6 +353,14 @@ class AffectView(ModelViewSet):
     http_method_names = get_valid_http_methods(ModelViewSet)
     permission_classes = [IsAuthenticatedOrReadOnly]
 
+    def perform_destroy(self, instance):
+        """
+        override the default behavior to proxy the delete to Bugzilla
+        """
+        flaw = instance.flaw
+        instance.delete()
+        flaw.save(bz_api_key=self.request.data.get("bz_api_key"))
```
Elkasitu, line 362: you should use a write-only serializer field instead of grabbing from the request data alone

osidb/serializer.py
```diff
@@ -366,6 +367,35 @@ def update(self, instance, validated_data):
...
 
+    def validate(self, data):
+        """
+        validate that the current user is member of all LDAP groups (s)he provides
+        with the access so there is no access expansion beyond the user permissions
+        """
+        data = super().validate(data)
+
+        # get the resulting ACLs from the embargo status
+        embargoed = self.context["request"].data.get("embargoed")
```
Elkasitu, line 378: if you manually instantiate this serializer (because of a custom view or something like that) then `request` won't be included in the context by default, DRF does it when instantiating it but this can easily turn into a mess, I'm not sure if we can have a serializer-only field that accepts both read and write operations, but worth a look maybe

Elkasitu (changes requested):
Not every review comment is a blocker, but a few things could be improved

apps/bbsync/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+
+        when no Bugzilla API key is provided it is considered to be a call
+        done by a collector or test and thus we perform just regular save
+
+        Bugzilla sync is also conditional based on environment variable
+        """
+        # preliminary save
+        # to link and annotate
+        super().save(*args, **kwargs)
+        self = self.__class__.objects.get(pk=self.pk)
```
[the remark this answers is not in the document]
osoukup: There are two main reasons. One is to get flaw annotation `embargoed` in place so it is possible to do things like `if flaw.embargoed:`. The second is to link the affect to the flaw so it is possible to do things like `self.flaw...`.

You are correct that the instance has `objects` attribute but accessing it results in `AttributeError: Manager isn't accessible via Flaw instances` which leads me to use the `__class__` here.

apps/bbsync/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    mixin for syncing the model to the Bugzilla
+
+    the sync between Bugzilla and the internal DB is done in a way
+    that the model is first stored to the Bugzilla and then fetched
+    back from there to make sure that the timestaps are kept in sync
+    (otherwise there might be a few seconds differences)
+
+    this mixin does not actually call Model.save at all
+    because it is done when fetched from Bugzilla
```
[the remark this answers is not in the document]
osoukup: I can rephrase it in the way that it does not call it directly.

apps/bbsync/mixins.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class BugzillaSyncMixin(models.Model):
+    """
+    mixin for syncing the model to the Bugzilla
+
+    the sync between Bugzilla and the internal DB is done in a way
+    that the model is first stored to the Bugzilla and then fetched
+    back from there to make sure that the timestaps are kept in sync
+    (otherwise there might be a few seconds differences)
```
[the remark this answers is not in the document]
osoukup: I hope that the yesterday demonstration has shown clearly that it is not a minute but about two seconds. Of course it may be theoretically longer with some network issues and would be impossible with Bugzilla outage but that is the thing of the design incorporating Bugzilla. SFM2 would has completely same potential issues.

apps/bbsync/mixins.py
osoukup: I already tried to explain this in the last comment. This is basically the same what SFM2 does so no degradation. Until we leave Bugzilla we can hardly do it better. The risk of having desynced data between Bugzilla and OSIDB is IMO more serious that two seconds of waiting.

Otherwise your suggestion of unaware abstract class makes sense. I will rework it your way.

docker-compose.yml
osoukup: Good point. However do you think we still need PSI part here? I mean we only plan to enable Bugzilla backwards sync in the stage now and it is already in MP+ while the production enabling will happen maybe in late Q2 when we should already have everything in MP+.

osoukup: https://gitlab.corp.redhat.com/product-security/dev/osidb-ops/-/merge_requests/22

osidb/api_views.py
osoukup: Good point. That would be much cleaner approach. Will do.

docker-compose.yml
Elkasitu: I assume if the env variable isn't present in PSI that it will be treated as false so LGTM, good catch

osidb/serializer.py
Elkasitu: I would just document it for now, it's nothing too serious and I have no better suggestions for now, I'd suggest putting a comment/warning somewhere and having a JIRA ticket for tracking this as an improvement

apps/bbsync/mixins.py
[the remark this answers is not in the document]
Elkasitu: Re: just saw your comment below on the overhead of two network calls

Your argument is somewhat valid, it'll be as fast/slow as SFM2 however I still think the WebUI guys will not appreciate it much 😄

osidb/models.py
```diff
@@ -444,7 +445,16 @@ def fts_search(q):
         # If search has no results, this will now return an empty queryset
 
 
-class Flaw(WorkflowModel, TrackingMixin, NullStrFieldsMixin, AlertMixin, ACLMixin):
+class Flaw(
+    ACLMixin,
+    AlertMixin,
+    # Bugzilla sync must take place after
+    # the validations not to bypass them
```
Elkasitu, line 452: This makes me a bit uneasy, maybe `BugzillaSync` shouldn't be a mixin but a subclass, WDYT? You can have alerting/validation without bzsync but you cannot have bzsync without alerting/validation

Elkasitu (approved):
LGTM, good work! 💪🏽 I only have one concern with the MRO but otherwise this can be merged

osidb/models.py
osoukup: @Elkasitu actually I am not sure that my comment is actually up-to-date here. You can see in `BugzillaSyncMixin` that the first thing it does actually is the call of `super().save(*args, **kwargs)` which means that no matter how the mixins are ordered the validations will run. I will update the comments and to be really safe add a test case specifically checking that the validations are run when syncing to Bugzilla.

osoukup: Done.

PR #157, opened 2023-03-07
implement non-empty impact validation

osoukup:
During the two-way sync testing I found out that it is possible to create a flaw with an empty impact. This is not correct and requires an additional validation plus some updates of the existing tests.

Closes OSIDB-758

### DIFF
--- a/apps/bbsync/tests/test_query.py
+++ b/apps/bbsync/tests/test_query.py
@@ -389,9 +389,10 @@ class TestGenerateSRTNotes:
         """
         test generating of SRT notes impact attribute
         """
-        flaw = FlawFactory(
+        flaw = FlawFactory.build(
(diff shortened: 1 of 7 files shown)


costaconrado (approved):
LGTM.

osoukup:
@costaconrado thank you for the quick review

Elkasitu:
This is OK for now, but shouldn't we fix the model instead to have it be not null?

osoukup:
> This is OK for now, but shouldn't we fix the model instead to have it be not null?

@Elkasitu I tried to explain it in the comment. The problem here are the historical data. We have to store them but they often have an empty impact. Enforcing this on the model level would prevent us from storing them.

Elkasitu:
@osoukup actually I just checked and the impact field is already not null, I think I got confused because here empty means the empty string or blank, not `None`, so LGTM sorry about that

PR #158, opened 2023-03-09
Fix missing query params in schema

JakubFrejlach:
Trackers and Affects API endpoints, more precisely their GET methods do accept `include_fields`, `exclude_fields` and `include_meta_attr` query params but these params were missing from the schema. Main purpose of this PR is that osidb_bindings and griffon are then not able to use these params as they treat it as unknown params.

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -3,6 +3,7 @@ implement osidb rest api views
 """
 
 import logging
+from typing import Type
 from urllib.parse import urljoin
 
 import pkg_resources
@@ -17,7 +18,12 @@ from rest_framework.permissions import AllowAny, IsAuthenticatedOrReadOnly
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from rest_framework.viewsets import ModelViewSet, ReadOnlyModelViewSet, ViewSet
+from rest_framework.viewsets import (
+    ModelViewSet,
+    ReadOnlyModelViewSet,
+    ViewSet,
+    ViewSetMixin,
+)
 
 from .constants import OSIDB_API_VERSION, PYPI_URL, URL_REGEX
 from .filters import AffectFilter, FlawFilter, TrackerFilter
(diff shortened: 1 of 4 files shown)


osoukup (approved):
LGTM

osidb/api_views.py
```diff
@@ -155,45 +155,53 @@ def get_valid_http_methods(cls: ViewSet, excluded: list[str] = None) -> list[str
...
-                    "Dot notation can be used to specify meta_attr keys on related models. "
-                    "Example: `include_meta_attr=bz_id,affects.ps_module,affects.trackers.bz_id`"
-                    "Use wildcards eg. `include_meta_attr=*,affects.*,affects.trackers.*` "
-                    "for retrieving all the keys from meta_attr. "
-                    "Omit this parameter to not include meta_attr fields at all. "
-                ),
-            ),
+            include_fields_param,
+            exclude_fields_param,
+            include_meta_attr,
```
Elkasitu, line 204: should be done in the retrieve section as well

costaconrado (approved):
LGTM

Elkasitu (approved):
This works, but I'd rather check whether it's possible to include this schema information in the `IncludeExcludeFieldsMixin` serializer, as it stands it would be easy to add a new serializer that inherits from the aforementioned mixin but forget to extend the schema in the view.

Maybe a view mixin or custom decorator could ease this burden too.

osidb/api_views.py
JakubFrejlach: Unfortunately you cannot do this from a serializer point of view. `extend_schema` and  `extend_schema_view` works only on api views and `extend_schema_serializer` has limited functionality for deprecating fields, examples, etc.

We can create mixin, eg:

```python
@extend_schema_view(
    list=extend_schema(
        parameters=[include_fields_param, exclude_fields_param, include_meta_attr],
    ),
    retrieve=extend_schema(
        parameters=[include_fields_param, exclude_fields_param, include_meta_attr],
    ),
)
class ParamsModelViewSet(ModelViewSet):
    pass
```

but in some cases we also need some additional params as well and purpose of this would be defeated.

JakubFrejlach: Well apparently these `extend_schema_view` decorators can be stacked as they don't overwrite themselves, so something like this is possible:

```python
@extend_schema_view(
    list=extend_schema(
        parameters=[include_fields_param, exclude_fields_param],
    ),
    retrieve=extend_schema(
        parameters=[include_fields_param, exclude_fields_param],
    ),
)
class ParamsModelViewSet(ModelViewSet):
    pass


@extend_schema_view(
    list=extend_schema(
        parameters=[include_meta_attr],
    ),
    retrieve=extend_schema(
        parameters=[include_meta_attr],
    ),
)
class AffectView(ParamsModelViewSet):
  ...
```

JakubFrejlach:
Changed the PR so it introduces new decorator shortcuts for include/exclude fieds and include meta_attr query params.

osoukup (approved):
LGTM

PR #159, opened 2023-03-09
remove deprecated mitigated_by field (OSIDB-753)

osoukup:
Done by the procedure defined at

https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#deprecate-fields

(the second part after **do release**)

Closes OSIDB-753


Elkasitu (commented):
I think we need to do this in two stages, no?

First remove the field from the API so that it's truly no longer used in version N-1 and version N, then we can remove completely the field from the database in version N+1

osoukup:
> I think we need to do this in two stages, no?
> 
> First remove the field from the API so that it's truly no longer used in version N-1 and version N, then we can remove completely the field from the database in version N+1

This is the procedure described by the **django-deprecate-fields** library. My understanding is that by using **return_instead** we already disconnected the app and the DB (in N-1) and so we can remove them at the same time (in N).

Elkasitu:
> > I think we need to do this in two stages, no?
> > First remove the field from the API so that it's truly no longer used in version N-1 and version N, then we can remove completely the field from the database in version N+1
> 
> This is the procedure described by the **django-deprecate-fields** library. My understanding is that by using **return_instead** we already disconnected the app and the DB (in N-1) and so we can remove them at the same time (in N).

Ah you are correct! In that case LGTM!

Elkasitu (approved):

JakubFrejlach (approved):
LGTM

PR #161, opened 2023-03-13
refactor bugzilla mixin not to save new flaw

osoukup:
I would like to perform this refactoring of the `BugzillaSyncMixin` before the freeze. It seems that for now it is not so important but will be for the future. When implementing OSIDB-382 I realized that the current design of the model `save` is not well scalable. The reason is this line

https://github.com/RedHatProductSecurity/osidb/blob/ef88c13c82385bf47eee489f4c25107eb387c360/apps/bbsync/mixins.py#L33

Its purpose was three-fold. One was to get `embargoed` annotation, then to perform the validations, and finally to link the affect to the corresponding flaw. The issue is that it later makes the old and the new flaw in BBSync being the same

https://github.com/RedHatProductSecurity/osidb/blob/ef88c13c82385bf47eee489f4c25107eb387c360/apps/bbsync/save.py#L53

which I unfortunately have not realized as an issue. The reason is that I was taking the data from the stored `meta_attr` wherever possible as I was first not sure whether the new and old flaw will be possible to be kept different.

This PR makes it possible substituting `embargoed` with `is_embargoed` which is not an annotation but just a property, then explicitly performing the validations in the BugzillaSyncMixin instead of the whole `save`, and finally performing the `save` in `Affect.bzsync` to link it. This effectively replaces what the `save` did in BugzillaSyncMixin so the new flaw save is now not necessary and it is kept different from the old flaw.

### DIFF
--- a/apps/bbsync/tests/test_query.py
+++ b/apps/bbsync/tests/test_query.py
@@ -870,7 +870,7 @@ class TestGenerateGroups:
         )
 
         new_flaw = Flaw.objects.first()
-        new_flaw.embargoed = False
+        new_flaw.acl_read = []  # make it whatever but embargoed
 
         bbq = BugzillaQueryBuilder(new_flaw, old_flaw=flaw)
         query = bbq.query
--- a/apps/bbsync/cc.py
+++ b/apps/bbsync/cc.py
@@ -355,7 +355,7 @@ class CCBuilder:
             add_cc = all_cc
             remove_cc = []
 
-        elif self.old_flaw.embargoed and not self.flaw.embargoed:
+        elif self.old_flaw.embargoed and not self.flaw.is_embargoed:
             # on unembargo we add and remove all but do not add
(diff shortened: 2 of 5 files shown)


apps/bbsync/tests/test_query.py
```diff
@@ -870,7 +870,7 @@ def test_unembargo(self):
         )
 
         new_flaw = Flaw.objects.first()
-        new_flaw.embargoed = False
+        new_flaw.acl_read = []  # make it whatever but embargoed
```
JakubFrejlach, line 873: What about `acl_write`? :thinking:

JakubFrejlach (approved):
LGTM!

apps/bbsync/tests/test_query.py
osoukup: The `is_embargoed` property really cares only about `acl_read` so for the sake of the test `acl_write` is not important.

PR #165, opened 2023-03-20
minimize mid-air collisions

osoukup:
This PR has two main parts.

The first one is to completely prevent the mid-air collisions on the API. You can imagine user having a flaw update page opened in a browser for half a day and then hitting the save button. There might have happened a lot in the meantime. By requiring and checking the timestamps we can prevent it.

The second one is to minimize the mid-air collision window in BBSync. We will fetch the flaw right before the update and check that we have the Bugzilla last change timestamp up-to-date. This should minimize the collision window to maybe a second. We cannot completely avoid it as there will simply always be some network delay.

The PR also introduces a few related changes in tests and factories to deal with the timestamps.

@skontar there is also one a bit unrelated change you should know about. By introducing the changes in the test factories the `created_dt` do not necessarily correspond to the order in which you create the flaws in the test.  This should bring us more to reality as we cannot assume that the flaw with `CVE-2000-0001` was created before the one with `CVE-2000-0002`. 

However the exploit tests were counting on it and therefore I assume the API user might too. I added an explicit ordering by `cve_id` but be aware of that it is not perfect and will lead to `CVE-2000-123456` being before `CVE-2000-9999`. I am not sure this is what you need and I do not want to go beyond fixing the tests here :smile: so if you need something better please improve the exploit app.

Closes OSIDB-765


osidb/tests/factories.py
```diff
@@ -126,14 +127,29 @@ class Meta:
...
     )
-    meta_attr = factory.Dict({"test": "1"})
+    # valid flaw is expected to have certain meta attributes present
+    meta_attr = factory.LazyAttribute(
+        lambda c: {
+            "bz_id": getattr(c, "bz_id", "12345"),
+            "last_change_time": c.updated_dt
+            if isinstance(c.updated_dt, str)
+            else c.updated_dt.strftime(DATETIME_FMT),
+            "test": "1",
```
costaconrado, line 137: NIT: This is a minor thing. I believe the field `test` was used as placeholder to see if the factory would create meta_attrs, therefore it can be removed as we have other meta_attr being created by default if you will.

costaconrado (approved):
Left some minor optional considerations but LGTM.

skontar:
The change is OK from exploits API standpoint.

osidb/tests/factories.py
osoukup: Good point. I will remove it

apps/bbsync/save.py
```diff
@@ -52,5 +58,39 @@ def update(self):
...
+        one last preventative check that Bugzilla last_change_time
+        really corresponds to the stored one so there was no collision
+        """
+        if self.actual_last_chante != self.stored_last_change:
+            raise DataInconsistencyException(
+                "Save operation based on an outdated model instance"
+            )
+
+    @property
+    def actual_last_chante(self):
```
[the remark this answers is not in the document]
osoukup: Oh those typos :smile: I will fix it

apps/bbsync/save.py
```diff
@@ -52,5 +58,39 @@ def update(self):
...
+        self.check_collisions()  # check for collisions right before the update
         self.bz_conn.update_bugs([self.flaw.bz_id], bugzilla_query_builder.query)
         return self.flaw
+
+    def check_collisions(self):
+        """
+        one last preventative check that Bugzilla last_change_time
+        really corresponds to the stored one so there was no collision
+        """
+        if self.actual_last_chante != self.stored_last_change:
```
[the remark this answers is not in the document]
osoukup: and the typo is copy-pasted ...

osidb/serializer.py
```diff
@@ -216,6 +220,42 @@ class Meta:
...
+        so the TrackingMixin can later perform the comparison
+
+        it is not actually used when saving the instance to DB
+        as that timestamp is being obtained automatically
+        """
+        instance.updated_dt = make_aware(
+            datetime.strptime(
+                self.context["request"].data.get("updated_dt"), DATETIME_FMT
+            )
+        )
```
JakubFrejlach, line 235: I am afraid this won't be so straightforward. `updated_dt` is readOnly field but you are counting on it to be provided here. This is a schema for this current branch and `updated_dt` is not there and thus it is not very clear that this value should be send in the request body. One workaround might be to treat it the similar way as I treated it the `embargoed` in ACL mixin by creating a custom `EmbargoedField`. :thinking: 

image

osidb/serializer.py
```diff
@@ -216,6 +220,42 @@ class Meta:
...
+        """
+        validate that the updated_dt timestamp was provided
+        because it is used to detect the mid-air collisions
+        """
+        # the updated_dt field is required but read-only
+        # so it is not present as part of the serialized data
+        updated_dt = self.context["request"].data.get("updated_dt")
+        # check that the instance exists so this is an update
+        if self.instance and updated_dt is None:
+            raise ValidationError({"updated_dt": "Field is required"})
```
JakubFrejlach, line 255: If we use the approach which I've suggested in the previous comment this might not be necessary as the validation of field will do it anyway.

osidb/tests/test_endpoints.py
```diff
@@ -1145,6 +1146,41 @@ def test_flaw_update(self, auth_client, test_api_uri):
...
+                "type": flaw.type,
+                "title": f"{flaw.title} appended test title",
+                "description": flaw.description,
+                "state": flaw.state,
+                "resolution": flaw.resolution,
+                "impact": flaw.impact,
+                "source": flaw.source,
+                "embargoed": flaw.embargoed,
+                "updated_dt": flaw.updated_dt - timedelta(days=1),  # outdated timestamp
+                "bz_api_key": "SECRET",
```
JakubFrejlach, line 1189: use header instead of body request

osidb/mixins.py
```diff
@@ -59,6 +59,32 @@ def save(self, *args, auto_timestamps=True, **kwargs):
...
+class TrackingMixinManager(models.Manager):
+    """
+    TrackingMixin companion changing the QuerySet accordingly
+    """
+
+    def get_or_create(self, defaults=None, **kwargs):
+        """
+        filter out auto_timestamps from the defaults
+        """
+        defaults.pop("auto_timestamps", None)
```
JakubFrejlach, line 71: If `defaults` are `None` this will result in exception.

osidb/mixins.py
```diff
@@ -59,6 +59,32 @@ def save(self, *args, auto_timestamps=True, **kwargs):
         super().save(*args, **kwargs)
 
 
+class TrackingMixinManager(models.Manager):
```
JakubFrejlach, line 62: I am kinda confused by this manager, can you give me a quick explanation please?

osidb/tests/factories.py
```diff
@@ -127,7 +127,15 @@ class Meta:
             )
         ]
     )
-    meta_attr = factory.Dict({"test": "1"})
+    # valid flaw is expected to have certain meta attributes present
+    meta_attr = factory.LazyAttribute(
```
JakubFrejlach, line 131: note: If both `bz_id` and `meta_attr` parameters are passed to the factory, `bz_id` is `overwritten` by the passed `meta_attr`, is this expected?

JakubFrejlach (changes requested):
Couple of things and suggestions:

osidb/tests/test_endpoints.py
osoukup: Again the old code here...the conflicts below would force me anyway :smile:

osidb/tests/factories.py
osoukup: This is a side effect I am not sure we can easily prevent. It would be however a developer mistake. I introduced `bz_id` factory attribute as a kind of shortcut (similar to `embargoed`) which ease the work of writing the factory. The lazy attribute then makes the instance produced by the factory as we generally expect it so again easing the work for the test developer. If one needs some more complicated `meta_attr`, (s)he simply has to write it all by hand and then it makes no sense to provide `bz_id` attribute also.

osoukup:
@JakubFrejlach so I made couple of changes with respect to the API - mainly in `serializer.py`.  I made `updated_dt` read-only no more so it normally appear in the schema for POST and PUT. I made it not required in the schema. It makes no sense to require it generally as on creation there was no last update.

The complicated thing is that it is in fact not required for the POST (create) but required for the PUT (update) as for the existing model we have to check the mid-air collisions. The OpenAPI schema allows HTTP method granularity so we theoretically could make it required for PUT only but the issue is that DRF does not count with such a granularity. So we cannot auto-generate such schema. Therefore I am enforcing the timestamp by the custom validation even though it is not required by the schema. I do not see any better option here. We have to require it to prevent the collisions.

Is this solution acceptable from the bindings point of view? I mean will it send the value even when not required but available? Otherwise I am not sure how to solve this :smile:

osidb/mixins.py
JakubFrejlach: Ahhh, now I get it, thank you!

osidb/tests/factories.py
JakubFrejlach: Understandable. Seems ok.

JakubFrejlach:
> @JakubFrejlach so I made couple of changes with respect to the API - mainly in `serializer.py`. I made `updated_dt` read-only no more so it normally appear in the schema for POST and PUT. I made it not required in the schema. It makes no sense to require it generally as on creation there was no last update.
> 
> The complicated thing is that it is in fact not required for the POST (create) but required for the PUT (update) as for the existing model we have to check the mid-air collisions. The OpenAPI schema allows HTTP method granularity so we theoretically could make it required for PUT only but the issue is that DRF does not count with such a granularity. So we cannot auto-generate such schema. Therefore I am enforcing the timestamp by the custom validation even though it is not required by the schema. I do not see any better option here. We have to require it to prevent the collisions.
> 
> Is this solution acceptable from the bindings point of view? I mean will it send the value even when not required but available? Otherwise I am not sure how to solve this smile

Yeah, I see the problem and it is not easy to handle this kind of thing which is kinda meant to be consistent and without much of a custom logic to comply to the custom logic. I have actually an idea how to get rid of the `updated_dt` from the request body in schema for POST.  I propose to annotate the Flaw view create (POST) method with like this in the `extend_schema_view`:

```python
create=extend_schema(
        request=extend_schema_serializer(exclude_fields=["updated_dt"])(FlawSerializer),
    ),
 ```
 
 This basically applies the `extend_schema_serializer` decorator and excludes the `updated_dt` field from the FlawSerializer just for this one case. Does this make sense?

JakubFrejlach (approved):
One suggestion regarding the better schema handling, otherwise LGTM!

Elkasitu:
> The first one is to completely prevent the mid-air collisions on the API. You can imagine user having a flaw update page opened in a browser for half a day and then hitting the save button. There might have happened a lot in the meantime. By requiring and checking the timestamps we can prevent it.

I don't think that this should be OSIDB's responsibility, it's common for frontends to simply display a banner saying that the data is outdated and should be refreshed. Case in point: GitHub will display a small banner when it detects that what you see is outdated and prompts you to refresh via a button.

Having to submit a timestamp with every update is annoying for both clients and servers.

It is up to clients to make sure that the info they are submitting is as up-to-date as possible.

osoukup:
@Elkasitu I cannot agree. Moving safeguards to the clients means that they will either have to be everywhere consistently (which is not possible for eg. `curl`, random scripts) or they will just stop existing. We will eventually become an authoritative source of vital security data so I think it is pretty much our responsibility to enforce the consistency. The clients may do mistakes. Just imagine eg. someone hitting `griffon` command to add a comment accidentally twice. Or imagine one person modifying the same attribute through `griffon` and the other through OSIM at the same time. These and similar examples would lead in an inconsistent data or unexpected behavior (from the user point of view). If the client (`griffon`, OSIM, ...) can just simply handle this for the user, there is no decrease in the user comfort for them, just increase in the data consistency. Other clients like `curl` will simply have to obey the same rules not to mess with the data or they will not work.

Elkasitu:
> Just imagine eg. someone hitting griffon command to add a comment accidentally twice. 

And how does this patch prevent that? I can submit a comment with `griffon` or `curl` then go back on my shell history to the last command and press on enter again, griffon will send a different timestamp and the request will go through.

> Or imagine one person modifying the same attribute through griffon and the other through OSIM at the same time.

I see your point but this patch does not prevent it, because anyone can just lie about the timestamp, we can control OSIM and Griffon but we cannot control literally any other client (curl, requests, etc.), I'm not even sure that this would work on OSIM / Griffon because then you have to keep track of the last time the user queried / opened / read the flaw and send that as timestamp.

osoukup:
> > Just imagine eg. someone hitting griffon command to add a comment accidentally twice.
> 
> And how does this patch prevent that? I can submit a comment with `griffon` or `curl` then go back on my shell history to the last command and press on enter again, griffon will send a different timestamp and the request will go through.
> 

Maybe `griffon` would update the timestamp itself, so not a great example. But `curl` would not - the timestamp in the DB is updated with the first call so the second one with the old timestamp will simply fail.

> > Or imagine one person modifying the same attribute through griffon and the other through OSIM at the same time.
> 
> I see your point but this patch does not prevent it, because anyone can just lie about the timestamp, we can control OSIM and Griffon but we cannot control literally any other client (curl, requests, etc.), I'm not even sure that this would work on OSIM / Griffon because then you have to keep track of the last time the user queried / opened / read the flaw and send that as timestamp.

We are not defending against an evil actor, that is the auth thing. If two parallel requests happen, one will simply go first and the other will fail because of an already out-dated timestamp, so this is prevented.

PR #172, opened 2023-03-23
make API requests transactional (OSIDB-232)

osoukup:
This was really as easy as turning the option on in the DB settings. The tests should be hopefully demonstrative enough to see how this functionality works. The motivation is simple - we do not want to leave the DB in some half-way state in the case that a part of a more complex operation failed only after something related was already changed.

This is of course still not solving all the transaction issues. Namely the changes to the backends will of course not be rolled back. We can eg. ended up in a situation when the flaw was successfully updated in DB and the followup Jira task update failed. This should however still be a step forward dealing at least with what we can be easily dealt with.

I am setting the transactions to be atomic in all the environments (stage, prod, local, ...) as I think it is better to keep them behaving similarly. I am opened to any objections against it.

Closes OSIDB-232


osoukup:
@JakubFrejlach I see that for some reason I originally tagged only Conrado as a reviewer :smile: but still as this is an important change I would be glad to have your eyes on it too

JakubFrejlach (approved):
LGTM!

PR #174, opened 2023-03-29
Add validation for allowed flaw sources

jobselko:
This PR adds validation for allowed flaw sources. It also updates FlawFactory to use only allowed flaw sources.

Related to "extend and refine flaw model" (OSIDB-73).


osoukup (approved):
One small suggestion but overall LGTM

PR #175, opened 2023-03-29
Update vcrpy in devel-requirements

jobselko:


JakubFrejlach (approved):

osoukup (commented):
I am not against this change but I am still not convinced it solves the issue

JakubFrejlach:
> I am not against this change but I am still not convinced it solves the issue

If not it will probably resurface in the future and then we will investigate more I think. For now it seems this is sufficient and it unblocks new developers running tests locally.

PR #177, opened 2023-03-30
Increase the maximum length of "cwe_id" field in Flaw

jobselko:
This PR increases the maximum length of "cwe_id" field in Flaw from 50 to 255.

Related to "extend and refine flaw model" (OSIDB-73).


osoukup (approved):
LGTM. One side not. When you pick the reviewers, choose `osidb-devs` (plural, not the bot account) so the team is notified and anyone can do the review.

PR #179, opened 2023-03-30
Expose REST API for Taskman

costaconrado:
This PR exposes Taskman as REST API.

In order to do that, this PR also introduce serializers mapping Jira entities to generate OpenAPI entires.

Closes OSIDB-811.

### DIFF
--- a/apps/taskman/api.py
+++ b/apps/taskman/api.py
@@ -1,0 +1,225 @@
+"""
+Taskman API endpoints
+"""
+import logging
+
+from drf_spectacular.utils import OpenApiParameter, OpenApiResponse, extend_schema
+from rest_framework.generics import GenericAPIView
+from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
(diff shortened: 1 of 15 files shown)


apps/taskman/api.py
```diff
@@ -0,0 +1,225 @@
...
+        return JiraTaskmanQuerier(
+            token=request.headers.get("JiraAuthentication")
+        ).create_group(
+            name=serializer.validated_data["name"],
+            description=serializer.validated_data["description"],
+        )
+
+
+@jira_token_description
+class task_group(GenericAPIView):
```
osoukup, line 146: Is there a reason why `task_group` and `task_group_new` need to be two separate classes? I am also not sure about the split of `task_comment` and `task_comment_new`.

osoukup (approved):
LGTM. One minor comment but generally very nice work!

apps/taskman/api.py
osoukup: No problem then. I think that passing the key in the body would be worse. This is good :+1: I was just not clear on the split but when there is a reason then leave it as it is

PR #180, opened 2023-03-31
Update readme and makefile to check quay login

jobselko:
Extends the DEVELOP.md file and the checks.mk script to check if a user is logged in to quay.

### DIFF
--- a/docs/developer/DEVELOP.md
+++ b/docs/developer/DEVELOP.md
@@ -74,6 +74,13 @@ Username: yourusername
 Password: ************
 ```
 
+Log in to **quay.io** using [Quay](quay.io) credentials
+```
+> podman login quay.io
+Username: yourusername
+Password: ************
+```
+
 Optional: Log in to **docker.io** for higher image pull limit
 ```
 > podman login docker.io
--- a/mk/checks.mk
+++ b/mk/checks.mk
@@ -4,12 +4,13 @@
 
 
 #***********************************
-### Check podman registry login
+### Check podman registries login
 #***********************************
 .PHONY: check-reg
 check-reg:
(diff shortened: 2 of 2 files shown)


docs/developer/DEVELOP.md
```diff
@@ -65,7 +65,7 @@ Note that your `.env` file contains secrets you should not share. Make sure that
 
 ### Image registries
 
-Before starting osidb the first time, or when updating to newer images, you need to log in to registries.
+Before starting osidb the first time, or when updating to newer images, you need to log in to registries and quay.
```
osoukup, line 68: This change is unnecessary as `quay` is a registry

mk/checks.mk
```diff
@@ -4,12 +4,14 @@
 
 
 #***********************************
-### Check podman registry login
+### Check podman registry login and podman quay login
```
osoukup, line 7: Both `registry.redhat.io` and `quay.io` are container registries so this change does not make sense to me. You can update the text by just putting the `registry` word to plural making it `registries`

mk/checks.mk
```diff
@@ -4,12 +4,14 @@
...
 #***********************************
 .PHONY: check-reg
 check-reg:
 	@echo ">Checking podman registry login"
 	@$(podman) login --get-login registry.redhat.io > /dev/null
+	@echo ">Checking podman quay login"
```
osoukup, line 13: Same here.

osoukup (changes requested):
Definitely a valuable change but some of the changes does not make sense to. I suggested corrections

osoukup (approved):
LGTM now. Thank you for addressing my feedback :smile_cat:

PR #181, opened 2023-04-03
serializers and api fixes

JakubFrejlach:
This PR brings couple of fixes to serializers and API:

- it changes HTTP status code returned from DELETE method from 204 to 200 becuase 204 has no content and thus no body and thus we fail to return injected info about OSIDB instance
- it adds custom conversion from str to bool for embargoed field (see the commit description for more info)
- fixed method HTTP method check in serializer


osoukup (approved):
Good set of fixes! Thank you and LGTM

PR #183, opened 2023-04-05
Add task rejection

costaconrado:
This PR include rejecting task in Taskman.

This PR also adds validation alerting user in case he rejects a flaw being affected.
Closes OSIDB-74.

### DIFF
--- a/apps/taskman/api.py
+++ b/apps/taskman/api.py
@@ -158,7 +158,7 @@ class task_group(GenericAPIView):
             OpenApiParameter(name="task_key", required=True, type=str),
         ],
         description="Add a task into a group",
-        responses={204: OpenApiResponse(description="Modified.")},
+        responses={200: OpenApiResponse(description="Modified.")},
     )
     def put(self, request, group_key):
         serializer = TaskKeySerializer(data=request.data)
@@ -194,7 +194,7 @@ class task_flaw(GenericAPIView):
 
     @extend_schema(
         description="Update a task in Jira from a Flaw",
-        responses={204: OpenApiResponse(description="Modified.")},
+        responses={200: OpenApiResponse(description="Modified.")},
     )
     def put(self, request, flaw_uuid):
         """Update a task in Jira from a Flaw"""
@@ -228,7 +228,7 @@ class task_status(GenericAPIView):
             ),
         ],
         description="Change a task workflow status",
-        responses={204: OpenApiResponse(description="Modified.")},
+        responses={200: OpenApiResponse(description="Modified.")},
     )
     def put(self, request, task_key):
         serializer = StatusSerializer(data=request.data)
@@ -261,7 +261,7 @@ class task_assignee(GenericAPIView):
             OpenApiParameter(name="task_key", required=True, type=str),
         ],
         description="Assign a task to a user",
-        responses={204: OpenApiResponse(description="Modified.")},
+        responses={200: OpenApiResponse(description="Modified.")},
     )
     def put(self, request, user):
         """Assign a user for a task"""
--- a/apps/taskman/api.py
+++ b/apps/taskman/api.py
@@ -23,7 +23,7 @@ from .serializer import (
     TaskGroupSerializer,
     TaskKeySerializer,
 )
-from .service import JiraTaskmanQuerier, TaskStatus
+from .service import JiraTaskmanQuerier, TaskResolution, TaskStatus
 
 logger = logging.getLogger(__name__)
 
@@ -209,7 +209,22 @@ class task_status(GenericAPIView):
     @extend_schema(
         parameters=[
             OpenApiParameter(
-                name="status", required=True, type=str, enum=TaskStatus.values
+                name="status",
+                required=True,
+                type=str,
+                enum=TaskStatus.values,
+            ),
+            OpenApiParameter(
+                name="resolution",
+                type=str,
+                enum=TaskResolution.values,
+                description="Resolution of a CLOSED task.",
+            ),
+            OpenApiParameter(
+                name="reason",
+                type=str,
+                enum=TaskStatus.values,
+                description="Reason of status change. Mandatory for rejecting a task.",
             ),
         ],
         description="Change a task workflow status",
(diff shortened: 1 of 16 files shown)


apps/taskman/service.py
```diff
@@ -107,10 +127,35 @@ def create_or_update_task(self, flaw: Flaw, fail_if_exists=False) -> Response:
...
+
+            issue = self.jira_conn.issue(issue_key)
+            uuid_labels = [
+                label for label in issue.fields.labels if "flawuuid:" in label
+            ]
+            flaw_uuid = uuid_labels[0].split(":")[1]
+            flaw = Flaw.objects.get(uuid=flaw_uuid)
+            flaw.meta_attr["jira_task_status"] = status
+            flaw.meta_attr["jira_task_resolution"] = resolution
+            flaw.save()
```
osoukup, line 158: I am thinking about this metadata. It is definitely handy but this way it is not persistent. The issue is that on any flaw change it is synced from Bugzilla and the `meta_attr` is created from scratch. To make this working the Jira task metadata fetch would also have to happen on flaw sync in `collectors.bzimport.convertors.py`. You could either just remove this `meta_attr` creation for now or make it also a part of flaw sync. We should think about this a bit more - whether the Jira task information is needed in OSIDB and what eventually what parts are necessary. We could maybe start with no information in OSIDB (metadata does not matter) and just fetch everything on demand and we will see whether it is going to be reasonable in the future.

osoukup (approved):
LGTM. Very good work! I have one comment which rather something to think about and not really a complain.

apps/taskman/service.py
costaconrado: I have have removed the validation from Flaw and added in the status change operation, please take a look again to evaluate if this would be a better solution of if we should discuss an alternative solution. Thank you! @osoukup

apps/taskman/service.py
```diff
@@ -107,10 +126,43 @@ def create_or_update_task(self, flaw: Flaw, fail_if_exists=False) -> Response:
...
+    ) -> Response:
         """Transition a task to a new state"""
         try:
-            self.jira_conn.transition_issue(issue=issue_key, transition=status)
+            if status == TaskStatus.CLOSED and resolution == TaskResolution.WONT_DO:
+                issue = self.jira_conn.issue(issue_key)
+                uuid_labels = [
+                    label for label in issue.fields.labels if "flawuuid:" in label
+                ]
+                flaw_uuid = uuid_labels[0].split(":")[1]
```
osoukup, line 142: I thinking about a case when someone manually removes the label in Jira. That should be rare and would be a human error but in such case we would end up with an **IndexError** which would not tell us much about what is actually the issue. Maybe we should check that the label was actually found and raise a concrete error if not?

osoukup: I hope that a situation when someone adds a label with non-existing UUID is too imaginary :smile:

apps/taskman/tests/test_integration.py
```diff
@@ -102,6 +102,24 @@ def test_task(self, monkeypatch, user_token, auth_client, test_api_uri):
         assert "assignee" in issue3["fields"]
         assert "customfield_12311140" in issue3["fields"]
 
+        response7 = auth_client.put(
+            f"{test_api_uri}/task/{issue['key']}/status",
+            data={"status": TaskStatus.CLOSED, "resolution": TaskResolution.DONE},
+            format="json",
+            **headers,
+        )
+        assert response7.status_code == 204
```
osoukup, line 111: I do not remember the exact details but I think that the Python bindings have some issues with an empty response... I wish @JakubFrejlach could explain it :smile:

osoukup (approved):
LGTM

openapi.yml
```diff
@@ -4212,6 +4212,32 @@ paths:
           type: string
         description: User generated token for Jira authentication.
         required: true
+      - in: query
```
Elkasitu, line 4215: why are these query params? it's a put operation, why not have these be in the body of the payload?

osoukup: It should be done the same way as the Bugzilla API key is. Neither in the query nor in the body but in the headers. There was some issue regarding the bindings when the Bugzilla API key was in the body and I think the headers are generally the standard method of transferring the keys.

Elkasitu: Header is acceptable, but I'd like to know more as to why it's not feasible to have it in the body (perhaps @JakubFrejlach can shed some light).

apps/taskman/serializer.py
```diff
@@ -21,8 +21,19 @@ class TaskKeySerializer(serializers.Serializer):
...
             raise serializers.ValidationError("Invalid status option.")
+        if data["status"] == TaskStatus.CLOSED and not data["resolution"]:
+            raise serializers.ValidationError("Closing a task requires a resolution.")
+        elif (
+            data["status"] == TaskStatus.CLOSED
+            and data["resolution"] == TaskResolution.WONT_DO
+            and "reason" not in data
+        ):
+            raise serializers.ValidationError("Rejecting a task requires a reason.")
+
```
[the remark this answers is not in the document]
costaconrado: I will remove the first validation for Jira. The second one I believe it still necessary to enforce people rejecting tasks to leave a comment explaining why, what do you think?

apps/taskman/service.py
costaconrado: Makes sense, I will hope for the best and expect for the worse and raise an error :smile:

costaconrado: I have added a 500 with error description for this case.

apps/taskman/tests/test_integration.py
costaconrado: I have removed all 204 from taskman, returning the edited issue and changing for 200 status.

apps/taskman/api.py
```diff
@@ -218,6 +233,15 @@ class task_status(GenericAPIView):
     def put(self, request, task_key):
         serializer = StatusSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
+
+        if "reason" in serializer:
+            JiraTaskmanQuerier(
+                token=request.headers.get("JiraAuthentication")
```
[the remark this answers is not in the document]
costaconrado: Duplication removed.

apps/taskman/service.py
Elkasitu: What about 409? It's called in a `PUT` anyway so I feel like it should be OK, 500 implies server mistake where in this case we're assuming that someone or something other than us removed the label, or the passed-in key is incorrect

costaconrado: Makes sense, changed for 409.

Elkasitu (approved):
LGTM!

PR #185, opened 2023-04-06
Update overlooked test with wrong HTTP response status code

jobselko:
This PR continues the changes done in #181, and updates overlooked test with wrong HTTP response status code (204 -> 200).


jobselko:
@RedHatProductSecurity/osidb-devs Should I update CHANGELOG in this case?

osoukup:
Oh I remember @JakubFrejlach was changing the delete behavior as the bindings could not easily handle 204 **No Content** response. Thank you for the fix! This does not require changelog item (neither user visible nor important change). I already added the **technical** label so the CI should stop complaining.

osoukup (approved):
LGTM

PR #186, opened 2023-04-11
implement proper NVD CVSS score collector

osoukup:
The last version of the NVD CVSS score collector was already really periodically collecting the data but still not very well using the collector framework (eg. not much meaningful collector status handling) and was dependent on the data from the old Krakken Dashboard. This one should be hopefully a proper one.

This PR also splits the NVD CVSS score fetching functionality from the Bugzilla flaw sync.

Obviously after the CVSS attribute-into-model refactoring this collector will need some adjustments.

Closes OSIDB-632


Elkasitu (approved):
LGTM, just wanted to bring the following lib: https://github.com/vehemont/nvdlib to our attention, it might reduce some of the overhead we have here.

And also we can request API keys from NIST, as by default there is some heavy throttling for unauthenticated requests, this could improve the performance by a lot, see https://nvd.nist.gov/developers/start-here#divRateLimits (from 5 requests per 30s to 50 requests per 30s)

osoukup:
@Elkasitu thank you for pointing me out to the lib! My impression was that the API calls are pretty simple and using a library (and introducing a new dependency) is not necessary but I did not noticed the information on the throttling at all. I experience some forbidden request time to time and was just considering it being some API instability. Obviously I was hitting the limits. Reading through the lib and NVD docs it does not seem to me to be necessary to deal with the API key but deal with the throttling would be good. Therefore I will try to rewrite the queries using the lib. It ensures there are 6s delays between the non-authenticated requests which is enough. There will most probably be always only a single request every ten minutes. The exception is the initial sync when for huge batches it may be exceeded. The lib should help there.

PR #190, opened 2023-04-12
patch flaw collector with jira server

costaconrado:
This PR fixes FlawCollector to get default values for Jira server / token from parent class.


osoukup (approved):
LGTM

osoukup:
@costaconrado this change should actually be a patch against **release-3.1.0** branch and after we do the patch release it would get merged to master from there. (hope I am correct with the procedure @Elkasitu :smile:)

jobselko:
@costaconrado Nitpick: Wouldn't it be better to put the `__init__` on top of the class for better readability?

collectors/bzimport/collectors.py
```diff
@@ -539,6 +539,10 @@ def collect(self, batch=None):
         msg += " Nothing new to fetch." if not flaw_ids else ""
         return msg
 
+    def __init__(self) -> None:
+        Collector.__init__(self)
+        JiraQuerier.__init__(self)
+
```
Elkasitu, line 545: I think the better solution is to change `JiraQuerier` to work like `BugzillaQuerier` (i.e. without an `__init__`, as a singleton)

Elkasitu (commented):
The commit message should explain what the problem was in the first place, how and why does this fix it, etc. Maybe a regression test would be useful too.

osoukup:
I like this version with the default values more :+1: and great you added the tests!

Elkasitu (approved):

PR #192, opened 2023-04-16
Add task ownership assignment

costaconrado:
This PR introduces task ownership.

Since we have relationship of 1:1 of task and flaw, assigning a user to a task may be enough to fulfill task ownership without storing user data in OSIDB. Please doing the review take an extra time to see if this approach would be adequate in OSIDB-69 scenario.

Closes OSIDB-69.

### DIFF
--- a/apps/taskman/api.py
+++ b/apps/taskman/api.py
@@ -223,3 +223,44 @@ class task_status(GenericAPIView):
         ).update_task_status(
             issue_key=task_key, status=serializer.validated_data["status"]
         )
+
+
+@jira_token_description
+class task_assignee(GenericAPIView):
+    @extend_schema(
+        responses=JiraIssueQueryResultSerializer,
+    )
+    def get(self, request, user):
+        """Get a list of tasks from a user"""
+        return JiraTaskmanQuerier(
+            token=request.headers.get("JiraAuthentication")
+        ).search_tasks_by_assignee(user)
+
+    @extend_schema(
+        parameters=[
+            OpenApiParameter(name="task_key", required=True, type=str),
+        ],
+        description="Assign a task to a user",
+        responses={204: OpenApiResponse(description="Modified.")},
+    )
+    def put(self, request, user):
+        """Assign a user for a task"""
+        serializer = TaskKeySerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        issue_key = serializer.validated_data["task_key"]
+
+        return JiraTaskmanQuerier(
+            token=request.headers.get("JiraAuthentication")
+        ).assign_task(task_key=issue_key, assignee=user)
+
+
+@jira_token_description
+class task_unassigneed(GenericAPIView):
+    @extend_schema(
+        responses=JiraIssueQueryResultSerializer,
+    )
+    def get(self, request):
+        """Get a list of tasks without an user assigned"""
+        return JiraTaskmanQuerier(
+            token=request.headers.get("JiraAuthentication")
+        ).search_tasks_by_assignee(None)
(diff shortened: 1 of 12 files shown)


osoukup (approved):
LGTM and I think that the approach is correct. We definitely want the flaw owner to be the task assignee and therefore to be stored and managed by Jira and not OSIDB.

When searching for a resource one will choose the API depending on what is the goal. If searching by assignee or status it would be done through the **taskman** and then the dots of the flaw data would be connected through flaw API. If searching on impact or affectedness (etc.) one will go through the flaws first and then optionally connect it with owners. The queries like give me all my tasks affecting OCP will have to be done in two steps by their nature but it OK as we do not want to make a task DB.

I am thinking about **taskman** API endpoints. There is one for the statuses and now one for assignees. However I thing that a very common use case will be getting all my opened tasks or all team unowned tasks. Both of these will basically created the queues - personal and team queue. What API endpoint would one hit to get these?

apps/taskman/api.py
```diff
@@ -223,3 +223,29 @@ def put(self, request, task_key):
         ).update_task_status(
             issue_key=task_key, status=serializer.validated_data["status"]
         )
+
+
+@jira_token_description
+class task_assignee(GenericAPIView):
```
Elkasitu, line 229: why are class names snake case?

apps/taskman/api.py
```diff
@@ -223,3 +223,29 @@ def put(self, request, task_key):
...
+        ],
+        description="Add a task into a group",
+        responses={204: OpenApiResponse(description="Modified.")},
+    )
+    def put(self, request, user):
+        """Assign a user for a task"""
+
+        return JiraTaskmanQuerier(
+            token=request.headers.get("JiraAuthentication")
+        ).assign_task(task_key=request.data["task_key"], assignee=user)
```
Elkasitu, line 251: This is a problem when you don't use serializers to validate input:

`task_key` is required but there is nothing enforcing that requirement, there is documentation (OpenAPI spec) that mentions that it is required, but what happens when someone calls this endpoint without a `task_key`?

They **should** get a 400 error explaining that `task_key` is a required attribute and that they did not provide it, what actually happens here is that you'll get a `KeyError` because there's no `task_key` in `request.data` and the server will return a cryptic error message with a 500 status which is not correct.

Please fix this and write a regression test, this potentially applies to the other views in taskman so I'd suggest checking for that too.

Elkasitu (changes requested):

apps/taskman/api.py
costaconrado: both `collectors/framework/api.py` and `apps/osim/api.py` are snake case, I though it would be the guideline for our API classes, but I'm up to change if you prefer.

apps/taskman/api.py
costaconrado: makes sense, I will submit a fix for that.

apps/taskman/api.py
Elkasitu: eh whatever, I'd prefer to stick to Python's conventions (which we do in OSIDB's api views) but unless we change all of them it doesn't change much here... odd that the linters don't pick this up

apps/taskman/tests/cassettes/test_integration/TestIntegration.test_comment.yaml
```diff
@@ -1,935 +1,4 @@
...
-      Connection:
-      - keep-alive
-      Content-Type:
-      - application/json
-      User-Agent:
-      - python-requests/2.26.0
-      X-Atlassian-Token:
-      - no-check
-    method: GET
-    uri: https://squid.corp.redhat.com:3128/rest/api/2/field
```
Elkasitu, line 20: A lot of cassettes still contain this URL, probably to delete in a separate PR

Elkasitu (approved):
LGTM just one minor comment

PR #195, opened 2023-04-21
Optimize /flaws endpoint and filters that traverse relationships

Elkasitu:
This commit increases the speed of the /flaws endpoint and some of its filters through two actions, namely:

- FlawView.queryset now includes a call to `prefetch_related` which will automatically load all affects and trackers for the relevant flaws being retrieved in as few SQL queries as possible. Without `prefetch_related`, Django will perform N queries to the affects table where N is the amount of flaws being requested (100 by default), with the call to `prefetch_related` the amount of queries is reduced to 2 per page.

- There are two new indexes on the affects table based on the `flaw` field and the `ps_module` and `ps_component` fields, this should, in theory, make filtering by `ps_module` / `ps_component` faster from the `/flaws` endpoint.

From some preliminary testing, this should increase speed by at least a factor of two and possibly make these queries less prone to crashing.

Closes OSIDB-851


osoukup (approved):
LGTM. This sounds as a potentially significant speed boost. I wonder whether it may solve the pod issues. I would expect that those are caused by the memory issues. Couldn't this actually cause more memory issue when we prefetch the related models also in the cases when they are actually not needed? Or do you think the memory is not the actual problem here?

Elkasitu:
@osoukup I did a test with the problematic query while looking at the OCP metrics and I see only a spike in CPU usage for the database and web service pods and very little to no impact on memory, so while this patch would increase memory usage, I don't think it'll make the problem worse.

That said, we won't know for sure until we try it I guess

PR #196, opened 2023-04-24
Jiraffe into collector

osoukup:
This PR introduces a number of changes mainly regarding the Jira tracker sync:

- removes some obsoleted tests and constants
- rewrites the Jiraffe core to contain only Jira connection and query functionality
- implements multi-issue and tracker specific queries
- fixes the tracker ACL handling
- introduces a shared mixin for a common ACL stuff
- splits tracker convertor class (the main reason is the ACLs are differently handled in Bugzilla and Jira)
- integrate Jiraffe with the collector framework to have better status reporting and recoverablity
- this also removes the last dependance on the Krakken Dashboard

Closes OSIDB-576


costaconrado (approved):
I left a comment about the behavior of 2-way sync, I think we still want to have a way to disable it in local development environment (not sure), but overall LGTM.

collectors/jiraffe/tasks.py
```diff
@@ -1,60 +1,23 @@
...
-        if affect_uuids:
-            for affect_uuid in affect_uuids:
-                process_affect.delay(
-                    affect_uuid=affect_uuid,
-                    groups_read=settings.PUBLIC_READ_GROUPS,
-                    groups_write=[settings.PUBLIC_WRITE_GROUP],
-                )
+def jira_tracker_collector(collector_obj):
+    logger.info(f"Collector {collector_obj.name} is running")
+    return collector_obj.collect()
```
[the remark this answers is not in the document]
osoukup: Correct. Do you think we should have it conditional? If so it would be probably good to have some general mechanism of setting the collectors off with having them on by default.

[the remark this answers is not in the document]
osoukup: Currently if not configured it would just fail in the backend once a minute and if configured with the wrong environment it would probably not link the stuff and just have separate production flaws and stage trackers. Otherwise it should do no harm as the collectors do not write to the other systems.

If it was for some reason not good to have all collectors on locally I would be for creating some more general mechanism of turning them of as part of the collector framework. Eg. we could have registering the collector conditional at https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/framework/models.py#L542 with something like
```
if get_env("COLLECTOR_{name.upper()}", default="True", is_bool=True):
```
(just an idea, not tested)

collectors/jiraffe/core.py
```diff
@@ -39,125 +37,120 @@ def _get_jira_connection(server, token) -> JIRA:
...
-                AND labels="{affect.flaw and affect.flaw.cve_id}" \
-                AND labels="pscomponent:{affect.ps_component}" \
-                AND labels="SecurityTracking" \
-                AND type="Bug"'
-    return conn.search_issues(jql_query, maxResults=False)
+    ###########
+    # HELPERS #
+    ###########
 
+    def datetime2jira_str(self, timestamp: timezone.datetime, inc: bool = False) -> str:
```
Elkasitu, line 50: I would expect the JIRA lib to convert datetimes to the right format, is it not the case?

collectors/jiraffe/convertors.py
```diff
@@ -0,0 +1,180 @@
...
+from osidb.core import generate_acls, set_user_acls
+from osidb.models import Tracker
+
+from ..utils import (
+    tracker_parse_update_stream_component,
+    tracker_summary2module_component,
+)
+
+
+class TrackerConvertor:
```
Elkasitu, line 18: kinda weird that this common base class is in Jiraffe

collectors/jiraffe/convertors.py
```diff
@@ -0,0 +1,180 @@
...
+        tracker.updated_dt = (
+            self.tracker_data["updated_dt"] or self.tracker_data["created_dt"]
+        )
+        return tracker
+
+    def convert(self, affect=None) -> Tracker:
+        return self._gen_tracker_object(affect)
+
+
+class TrackerIssueConvertor(TrackerConvertor):
```
Elkasitu, line 88: I'd just call them `BugzillaTrackerConvertor` and `JiraTrackerConvertor`, because that's what you want to convey but make it less understandable by using `Issue` and `Bug`.

collectors/jiraffe/collectors.py
```diff
@@ -0,0 +1,91 @@
...
+
+        on every run the next batch of Jira trackers is fetched and store
+        while proceeding forward by the timestamp of the last update
+
+        tracker_id param makes the collector to sync a tracker of the given ID only
+        """
+
+        logger.info("Fetching Jira trackers")
+        start_dt = timezone.now()
+        desync = []
```
Elkasitu, line 50: should probably be called `updated_trackers` or similar instead

Elkasitu (commented):
LGTM mostly, just some minor comments

collectors/jiraffe/convertors.py
osoukup: I was thinking whether **bzimport** or **Jiraffe** but the other way there was a cyclic import so this way it was simpler. We could find some common place between the collectors.

collectors/jiraffe/core.py
osoukup: The thing is that `search_issues` expects a string (https://jira.readthedocs.io/api.html#jira.client.JIRA.search_issues). Another reason of this method is the rounding which is more related to the defining the sync window.

collectors/jiraffe/convertors.py
osoukup: The reason is to prevent a cycle. The **bzimport** collector uses the **Jiraffe** and not the opposite. We could maybe create some base module for the trackers but as they only share a part of conversion logic I am not sure it is worth.

collectors/jiraffe/convertors.py
osoukup: I will rename.

collectors/jiraffe/collectors.py
osoukup: I will rename.

PR #197, opened 2023-04-25
account for the Bugzilla flaw data migration in 2021 (OSIDB-858)

osoukup:
as it makes the bzimport more prone to transient sync errors which can significantly prolong the initial flaw data sync

(the detailed analyses can be found at https://issues.redhat.com/browse/OSIDB-858)

Closes OSIDB-858

### DIFF
--- a/collectors/bzimport/collectors.py
+++ b/collectors/bzimport/collectors.py
@@ -242,57 +242,80 @@ class BugzillaQuerier(BugzillaConnector):
         return products[0] if products else None
 
 
+# TODO tzinfo should be set implicitly but my current
+# attempts do not work - leaving explicit for now
+TIMEZONE = timezone.get_current_timezone()
+
+
 class FlawCollector(Collector, BugzillaQuerier, JiraQuerier):
     """Bugzilla flaw collector"""
 
     # date before the first flaw was created
-    BEGINNING = timezone.datetime(2000, 1, 1, tzinfo=timezone.get_current_timezone())
+    BEGINNING = timezone.datetime(2000, 1, 1, tzinfo=TIMEZONE)
 
     # this should be considered the minimum amount of flaws to be synced
     # per collector run and not a hard limit per batch, as each batch will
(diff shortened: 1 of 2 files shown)


osoukup:
@Elkasitu @skontar this should hopefully help with the initial sync. It will not make the best case better in any way but it should significantly lower the probability of the worst case where the sync cycle trough the huge batch over and over due to some transient network or Bugzilla error as it splits the problematic batches into a number of smaller ones.

collectors/bzimport/collectors.py
```diff
@@ -242,57 +242,80 @@ def get_product_with_components(self, name):
...
+# attempts do not work - leaving explicit for now
+TIMEZONE = timezone.get_current_timezone()
+
+
 class FlawCollector(Collector, BugzillaQuerier, JiraQuerier):
     """Bugzilla flaw collector"""
 
     # date before the first flaw was created
-    BEGINNING = timezone.datetime(2000, 1, 1, tzinfo=timezone.get_current_timezone())
+    BEGINNING = timezone.datetime(2000, 1, 1, tzinfo=TIMEZONE)
```
skontar, line 254: Bugzilla ignores timezone at this time anyway. However, the documented format is "%Y-%m-%dT%H:%M:%SZ", so I think if you want to be 100% compliant with BZ API documentation, you should probably convert datetime to string before sending it to BZ library.

skontar: I think you actually do not want any other timezone than UTC.

osoukup: The timezone stuff is not much to be review here :smile: nothing functionally changes in it, I was just moving furniture around. The timezones are more for the Python reasons like not compare timezone-aware and timezone-naive timestamps. I am not sure about the format here, I thought the magic is done by the **bugzilla**. Anyway again nothing which would be changed in this PR.

Elkasitu (approved):

PR #199, opened 2023-04-25
Add FlawReference and FlawReferenceManager models

jobselko:
Related to "implement adding of external links to flaws" (OSIDB-71).


osidb/models.py
```diff
@@ -1788,6 +1788,58 @@ class Meta:
...
+    # internal primary key
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    type = models.CharField(
+        choices=FlawReferenceType.choices,
+        default=FlawReferenceType.EXTERNAL,
+        max_length=50,
+    )
+
+    url = models.CharField(max_length=50)
```
jobselko, line 1828: This should probably be bigger. Maybe `200`?

osoukup: Good point! Links can be definitely long so 50 is not enough.  I am not sure what might be a reasonable length :smile: your suggestion sounds OK

docs/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Added
 - Introduce flaw ownership through task management system (OSIDB-69)
+- Introduce FlawReference and FlawReferenceManager models (OSIDB-71)
```
osoukup, line 10: As this PR does not bring anything visible to the users and not something extra large or of some user interest, I would omit mentioning the change in the changelog. The better timing is when there is finally some API.

osoukup (changes requested):
I am suggesting a few small changes. (ignore the `meta_attr` comment)

osidb/models.py
```diff
@@ -1788,6 +1789,55 @@ class Meta:
...
+
+
+class FlawReference(TrackingMixin, ACLMixin):
+    """Model representing flaw references"""
+
+    class FlawReferenceType(models.TextChoices):
+        """allowable references"""
+
+        VULN_RESPONSE = "VULN_RESPONSE"
+        EXTERNAL = "EXTERNAL"
```
Elkasitu, line 1815: I assume the name `vuln response` is historical: to me it means nothing, what differentiates these two types? should we rename them?

osidb/models.py
```diff
@@ -1788,6 +1789,55 @@ class Meta:
...
+    # internal primary key
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    type = models.CharField(
+        choices=FlawReferenceType.choices,
+        default=FlawReferenceType.EXTERNAL,
+        max_length=50,
+    )
+
+    url = models.CharField(max_length=200)
```
Elkasitu, line 1826: make it a `TextField`, see https://stackoverflow.com/a/417184

osidb/models.py
```diff
@@ -1788,6 +1789,55 @@ class Meta:
...
+
+    # one flaw can have many references
+    flaw = models.ForeignKey(Flaw, on_delete=models.CASCADE, related_name="references")
+
+    objects = FlawReferenceManager()
+
+    class Meta:
+        """define meta"""
+
+        unique_together = ["flaw", "type", "url"]
```
Elkasitu, line 1838: does it make sense to have the same URL twice, once per type? if not, I'd just have it be unique on flaw and URL

Elkasitu (changes requested):
A more existential question: Does this warrant a new model? Can we not have it be a e.g. an array of text fields? The only thing that makes this warrant a model is the description field, is it really necessary?

Also, we might want to have this be a type of `FlawMeta` instead.

osidb/models.py
Elkasitu: We should at least document the implications of each type in the code itself

osidb/models.py
Elkasitu: we might also want to provide field validation to make sure that it looks like a URL, but that depends on the existing data

osidb/models.py
osoukup: Probably it does not. Why should we have the same URL as two different types... Makes sense to me to simplify.

osidb/models.py
jobselko: A validation is a good point. I can use `URLField`, which is already validated by `URLValidator`.

osidb/models.py
```diff
@@ -1823,7 +1823,7 @@ class FlawReferenceType(models.TextChoices):
         max_length=50,
     )
 
-    url = models.CharField(max_length=200)
+    url = models.URLField()
```
Elkasitu, line 1826: `URLField` is a `CharField` underneath with a `max_length=200`, maybe we should bump it to `2048` to be compliant with most web servers? I feel like we're gonna have to bump this eventually otherwise.

osidb/models.py
Elkasitu: Sure, but what does `vuln_response` mean? What's the difference between an `external` and `vuln_response` link?

jobselko: I checked that links on sfm2/cve page, and it seems we have to differentiate between them. If I understand it correctly, `vuln_response` represents "Security Bulletin" and `external` represents "External References" in e.g. the following [cve page](https://access.redhat.com/security/cve/cve-2022-0847). 

So I added a description of these links, and I am also going to add validations for `vuln_response`. It should start with https://accesss.redhat.com/, and IMHO every flaw/cve should have only one `vuln_response` link (already discussed with @osoukup). Do you agree with it, @Elkasitu?

Elkasitu: Maybe checking that the internal / vuln_response URL is a `redhat.com` subdomain is good enough and a bit more future-proof?

jobselko: @Elkasitu I would change `vuln_response` to `security_bulletin_url` and `external_references` leave as it is. Do you agree?

Elkasitu: ok let's keep it as `security_bulletin` and `external` then

jobselko: Updated.

osidb/models.py
```diff
@@ -1788,6 +1804,100 @@ class Meta:
...
+
+
+class FlawReference(TrackingMixin, ACLMixin, AlertMixin):
+    """Model representing flaw references"""
+
+    class FlawReferenceType(models.TextChoices):
+        """
+        Allowable references:
+
+        SECURITY_BULLETIN:
```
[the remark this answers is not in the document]
Elkasitu: What about `official_statement`? I guess that's the purpose for differentiating between "external" and "something else", the "something else" is Red Hat's official statement regarding the vulnerability.

Anyway both of you seem to know the difference between both better than me so feel free to pick, I just wanted to push the idea of giving it a more meaningful name.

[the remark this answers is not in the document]
jobselko: In the end I simply used `article`.

osidb/models.py
```diff
@@ -1731,6 +1746,7 @@ def create_flawcomment(flaw, external_system_id, comment, **extra_fields):
             )
             flawcomment.meta_attr = comment
             return flawcomment
+
```
Elkasitu, line 1749: you might be using a different version of black locally / in your IDE

osidb/models.py
```diff
@@ -1788,6 +1804,98 @@ class Meta:
...
+        EXTERNAL_REFERENCES:
+            URL links to other trustworthy sources of information about this
+            vulnerability. A link should not point to a missing resource.
+            Since these links are displayed on the CVE page of the flaw, we only
+            want to include respectable sources (such as upstream advisories,
+            analysis of security researches, etc.).
+        """
+
+        ARTICLE = "ARTICLE"
+        EXTERNAL_REFERENCES = "EXTERNAL_REFERENCES"
```
Elkasitu, line 1846: calling the type `EXTERNAL_REFERENCES` seems redundant when the class is already called `FlawReference`

osidb/tests/test_flaw.py
```diff
@@ -1427,6 +1443,62 @@ def test_validate_allowed_source(self):
...
+    def test_validate_article_link(self):
+        """
+        Tests that an article link not starting with https://accesss.redhat.com/
+        raises an alert.
+        """
+        flawreference = FlawReferenceFactory(
+            type=FlawReference.FlawReferenceType.ARTICLE,
+            url="https://httpd.apache.org/link123",
+        )
+        assert "wrong_article_link" in flawreference._alerts
```
Elkasitu, line 1455: you don't test the opposite case, in which the article link *does* start with `https://access.redhat.com/`

also, there's a typo in the docstring `accesss` vs `access`

osidb/models.py
```diff
@@ -1788,6 +1804,98 @@ class Meta:
...
+        """define meta"""
+
+        unique_together = ["flaw", "url"]
+
+    def _validate_article_link(self):
+        """
+        Checks that an article link begins with https://accesss.redhat.com/.
+        """
+        if self.type == self.FlawReferenceType.ARTICLE and not self.url.startswith(
+            "https://accesss.redhat.com/"
```
Elkasitu, line 1876: is it a typo or is it actually `accesss` with 3 s?

osidb/tests/test_flaw.py
Elkasitu: P.S.: I see that you do use `access.redhat.com` in the tests below, but never explicitly check that the alerts are cleared/empty

Elkasitu (approved):
Mostly nitpicks, good job

osidb/models.py
jobselko: Just a typo, thanks!

osidb/models.py
jobselko: Do you prefer to change a constant name or class name? `EXTERNAL` instead of `EXTERNAL_REFERENCES` does not sound good to me (maybe `EXTERNAL_LINKS` to keep it different?).

osoukup (approved):
LGTM

osidb/models.py
Elkasitu: `external`, `generic`, `other` are all fine, but links/references in the name makes no sense as each object represents a single link

jobselko: Oh, you are right; this reasoning makes sense. Thanks for the recommendation!

PR #200, opened 2023-04-26
parallelize bzimport (OSIDB-864)

osoukup:
Simple parallelization of **bzimport** flaw sync using **joblib** and threads. It seems to show nearly linear thread-number-wise improvement on the local machine so should help us prevent waiting several day until an instance gets in sync. I tested how it behaves with the exception raised in one of the threads and simply fails all which is what we want as the collector framework will ensure the rerun.

Closes OSIDB-864


Elkasitu (changes requested):
As discussed in chat, nothing wrong with the code itself but I'm not sure this will work outside of a local dev environment because parallelization does not work well with cloud environments (multi-processing is the golden standard and its easier to scale horizontally than vertically).

osoukup:
@Elkasitu so what do you suggest? Using processes instead of the threads? I guess we do not care much. The real parallelism is probably not needed. We just need to open multiple Bugzilla queries and wait for the responses in parallel. Sure we could come up with some cleaner solution using Celery. I am not sure it is worth the investment now. As this is a pure backend matter we can rewrite it any time without causing some major release etc.

PR #203, opened 2023-04-27
Call NVD collector once per Flaw collector run

Elkasitu:
Previously it was being called once per thread per Flaw collector run, which could lead to rate-limiting, memory, etc. issues.

Since all threads will start and finish at roughly the same time, it should be OK to cache the NVD results once per collector run (which was originally the case anyway).


osoukup (approved):
Looks reasonable. LGTM

PR #204, opened 2023-04-28
double check whether a release contains API breaking changes

osoukup:
preventative measures taken after OSIDB-854 incident


Elkasitu (approved):

Elkasitu:
This can be included (in the post-mortem as well): https://github.com/RedHatProductSecurity/osidb/pull/207

osoukup:
@Elkasitu I merged the postmortem before reading your note... I think it does not matter. The important is that the safeguard is actually in place

PR #208, opened 2023-05-08
Openapi diff testcase

costaconrado:
Do not merge. The only intent of this PR is test PR207.


costaconrado:
API breaking changes validation worked as expected.

PR #209, opened 2023-05-09
new flaw comment add API

jsvob:
This PR introduces new flaw comment add API.

Closes OSIDB-81.


jsvob:
Open questions, could you please answer these during the review?

* Should the new `temporary_pending` field be marked for DB indexing for performance?
* Should the new comment functionality be described in TUTORIAL.md?

Thanks!

Elkasitu (changes requested):
Some general remarks not related to the code:

- We should use a separate endpoint for writing comments instead of using PUT/PATCH on the flaw directly, something like e.g. `/flaws/<uuid/cve_id>/comment`.
- I think we should do the sync differently, the `FlawComment` class should probably write to Bugzilla directly on save and then we have two choices:
    * We keep the created object in OSIDB, and then on bzimport sync overwrite all comments
    * We discard the created object in OSIDB and force a flaw sync (there is already precedence for this in `Flaw.bzsync()`

Note that neither of these options guarantee object permanence, meaning we cannot keep a comment's UUID as a reference because it might change between calls, but I don't see any way that we can guarantee this without completely reworking the sync mechanism

osidb/serializer.py
```diff
@@ -502,6 +516,66 @@ class Meta:
...
+    """
+    serializer mixin class implementing special handling of pending new comments
+    """
+
+    def to_representation(self, instance):
+        """
+        Perform normal to_representation, but first delete stale pending
+        comments.
+        """
+        return super().to_representation(instance)
```
jobselko, line 529: Based on our discussion, something is probably missing here.

jsvob: Good catch. This code will go away based on Adrian's feedback, but I fixed it nonetheless, so that the current state is consistent at least.

osidb/mixins.py
```diff
@@ -234,6 +235,35 @@ class Meta:
         abstract = True
 
 
+class PendingFlawCommentMixinManager(models.Manager):
```
osoukup, line 238: This functionality is unnecessary. Every API call is resolved as atomic (https://github.com/RedHatProductSecurity/osidb/blob/master/config/settings_prod.py#L74 - other non-prod environments have the same setting) so there cannot be a situation when the request crashes half way (eg. when calling Bugzilla) and there is a dangling `FlawComment` left. Also no other parallel thread should see anything half way done in the DB because of the atomicity (should be handled by Django).

collectors/bzimport/convertors.py
```diff
@@ -246,6 +246,19 @@ def clean_affects(self):
                 # affect does not exist any more
                 old_affect.delete()
 
+    def clean_comments(self):
```
osoukup, line 249: I agree that the cleanup here is necessary to replace the previously added temporary comment but we do not need any special functionality to discover it. Simply the temporary comment has an empty `external_system_id` so here you can just remove all such comments and that should be it.

apps/bbsync/query.py
```diff
@@ -289,3 +293,20 @@ def generate_srt_notes(self):
         """
         srt_notes_builder = SRTNotesBuilder(self.flaw, self.old_flaw)
         self._query["cf_srtnotes"] = srt_notes_builder.content
+
+    def generate_comment(self):
+        with transaction.atomic():
```
osoukup, line 298: All the API calls are atomic by default and the Django is responsible for it
https://github.com/RedHatProductSecurity/osidb/blob/master/config/settings_prod.py#L74

osoukup (changes requested):
Very nice progress! I added some suggestions of what to improve and how. Ping me if anything was not clear or meaningful :smile:

osoukup:
> Open questions, could you please answer these during the review?
> 
>     * Should the new `temporary_pending` field be marked for DB indexing for performance?

Should be removed as I explained in the review comments.

>     * Should the new comment functionality be described in TUTORIAL.md?

I guess the tutorial is overall a bit outdated :smile: so any update is definitely welcome but it is not essential.

> 
> Thanks!

osidb/mixins.py
jsvob: TIL!

osidb/serializer.py
```diff
@@ -449,6 +451,18 @@ class Meta:
         ] + TrackingMixinSerializer.Meta.fields
 
 
+class NewCommentSerializer(serializers.Serializer):
+    """
+    new_comment serializer, write-only.
+    """
+
+    # max length: https://bugzilla.redhat.com/show_bug.cgi?id=1112318
+    text = serializers.CharField(write_only=True, max_length=65534, required=True)
```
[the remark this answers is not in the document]
jsvob: The model [already uses `TextField`](https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/models.py#L1789). It [seems](https://www.django-rest-framework.org/api-guide/fields/#charfield) the serializer counterpart is `serializers.CharField`.

Ack on removing length check. Tested on today's new version. Bugzilla indeed responds with a 400 and ATOMIC_REQUESTS keep the database consistent without a lingering FlawComment instance.

jsvob:
Thank you both for the feedback. Please re-review.

I believe the updated PR addresses all feedback except [the first point](https://github.com/RedHatProductSecurity/osidb/pull/209#pullrequestreview-1420441728) - _something like e.g. `/flaws/<uuid/cve_id>/comment`_.

- The first point is still standing, right? Based on the discussion in https://github.com/RedHatProductSecurity/osidb/pull/215 I presume that instead of `FlawCommentView`, I should use `@action` inside `FlawView`, right?

  - Which methods should be supported for comments? With a separate viewset, implementing list and retrieve in addition to create seemed straightforward and intuitive from a presumed user's perspective. But with an ad-hoc non-viewset endpoint, should there be purely create/POST support and nothing else for comments?

- Are there any issues with the schema? I'm not familiar with how the schema should look like or how to properly use the schema-generating decorators on the serializer and viewset, so providing feedback by pointing me in the right direction for further research and learning would be valuable, please.

osidb/serializer.py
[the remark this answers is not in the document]
osoukup: Great you tested it and thanks for the explanation!

osidb/models.py
```diff
@@ -1753,6 +1767,25 @@ def create_flawcomment(flaw, external_system_id, comment, **extra_fields):
             flawcomment.meta_attr = comment
             return flawcomment
         except ObjectDoesNotExist:
+            # The "text" field is required for full validation, so save it if the caller forgets
+            # to explicitly copy it over from meta_attr.
```
osoukup, line 1771: How is it meant that the caller could forget to copy from  `meta_attr`? I would expect that the `text` field to be vital here and the `meta_attr` is rather a some extra/debugging/optional information.

osidb/models.py
```diff
@@ -1753,6 +1767,25 @@ def create_flawcomment(flaw, external_system_id, comment, **extra_fields):
...
+            extra_fields["order"] = extra_fields.get("order") or comment.get("count")
+
+            # In-place upgrade of a temporary pending comment to preserve uuid.
+            pending = FlawComment.objects.pending().filter(
+                flaw=flaw, order=extra_fields["order"]
+            )
+            if pending.exists():
+                flawcomment = pending.first()
+                flawcomment.external_system_id = external_system_id
+                flawcomment.meta_attr = comment
```
osoukup, line 1784: Is there any additional information in the comment JSON above `text`, `order`, `external_system_id`, timestamps (private) which is worth saving?

osidb/models.py
```diff
@@ -1743,6 +1745,18 @@ def _validate_public_source_no_ack(self):
...
+        """
+        Get new pending comments that are about to be submitted to bugzilla.
+
+        Note that consumers of this might break if ATOMIC_REQUESTS is not used
+        because long-forgotten uncleaned pending comments might suddenly be
+        posted to bugzilla.
+        """
+        pending_comments = super().get_queryset().filter(external_system_id="")
+
+        return pending_comments
```
osoukup, line 1758: nitpick: Why not just `return super().get_queryset().filter(external_system_id="")` ?

osidb/models.py
```diff
@@ -1808,6 +1845,76 @@ class Meta:
...
+    def _validate_single_pending_comment(self):
+        """
+        Check that this is at most the first pending comment for the flaw.
+        This comment must not be a second pending comment for the flaw. If it
+        is, it means that API calls are not ATOMIC_REQUESTS anymore, or that
+        the temporary pending comments are not cleaned up properly, or that
+        something makes concurrent requests.
+        """
+        tmp_cmnts = FlawComment.objects.pending().filter(flaw=self.flaw)
+        if tmp_cmnts.count() > 1:
```
osoukup, line 1870: I am not sure this is necessary. If the atomicity does not work, we are screwed anyway when the scheduler stops the thread before this line and lets another thread to get to the same point (after the check but before the actual save). I would just assume that the atomicity works and Django is fully responsible for it.

Otherwise we could question many other things like do the Django really stores the model to the DB after `save` and rather check it by a pure SQL select to be truly sure and we could think about other examples.

osidb/models.py
```diff
@@ -1808,6 +1845,76 @@ class Meta:
...
+        else:
+            proper_following_order = (
+                1 + self.flaw.comments.aggregate(Max("order"))["order__max"]
+            )
+        if self.external_system_id == "" and proper_following_order != self.order:
+            raise ValidationError(
+                f"FlawComment ({self.uuid}) for {self.flaw} is not of the correct "
+                f"order. FlawComment.order ({self.order}) but flaw has "
+                f"({self.flaw.comments.count()}) comments (starting with 0). "
+                "Either OSIDB's DB is outdated or the client has an outdated "
```
osoukup, line 1898: Can the DB be more outdated than the client when the client gets data from the DB and not vise versa?

osidb/models.py
```diff
@@ -1808,6 +1845,76 @@ class Meta:
...
+                "idea of the Flaw."
+            )
+
+    def _validate_nonblank_extid_for_nonpending(self):
```
osoukup, line 1902: This should not be validated here. The user can do nothing about it and would probably not even understand what is wrong. If we have a blank external ID of an already collected (from Bugzilla) comment it is not even a data error but a bug. I am also not sure why do we operate here with 1 minute. The flaw (and all the related data) is force synced after every update so we do not have to wait for the periodic sync to have it up-to-date.

osidb/serializer.py
```diff
@@ -806,3 +806,47 @@ class Meta:
...
+    text = serializers.CharField(required=True)
+    type = serializers.CharField(max_length=100, required=False)
+    order = serializers.IntegerField(allow_null=True, required=True)
+
+    def create(self, validated_data):
+        """
+
+        Requires ATOMIC_REQUESTS so that temporary pending comment instances are
+        not left over after unsuccessful calls, as they could have unintended
+        consequences such as submitting long-forgotten comments to bugzilla later.
```
osoukup, line 829: I am not sure we have to mention this requirement all around. It is a general requirement for the whole OSIDB API to work properly that the requests are atomic.

osoukup (changes requested):
I raised some comments but generally it is on the good path forward.

apps/bbsync/query.py
```diff
@@ -307,3 +312,30 @@ def generate_srt_notes(self):
...
+    def generate_comment(self):
+        pending_comments = FlawComment.objects.pending().filter(flaw=self.flaw)
+
+        # These checks are not triggerable with ATOMIC_REQUESTS, but
+        # are present as a failsafe because posting old pending comments
+        # might have bad consequences.
+        if pending_comments.count() > 1:
+            raise BadRequest(
+                "Cannot add more than one comment with one update! "
+                "New comment requests running in parallel on the same flaw."
```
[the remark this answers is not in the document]
jsvob: Thank you for raising a warning about this. After some preliminary investigation with submitting concurrent requests, I see that these checks are insufficient to prevent posting multiple comments with the same `order`. The experiment resulted in two conflicting comments actually ending up being saved in bugzilla and all other requests resulting in uncleaned temporary `FlawComment` objects, which borked the OSIDB instance (and prevented future flaw syncs) until I deleted them manually from the DB.

 It seems that `ATOMIC_REQUESTS` in itself is insufficient. The reason is most probably the default `isolation_level` is happy to run concurrent operations even if each operation needs to have the most up-to-date read view of the database. I yet have to find out why ATOMIC_REQUESTS sufficiently fails a single non-concurrent request that gets a 40x from bugzilla (and cleans up the temporary FlawComment instance), but not when done in parallel.

I'll investigate whether there's a way to signal to Django that the given view method must not run concurrently (a lock? a different isolation_level?) and make sure a bugzilla 40x makes the osidb request fail and the database properly rolled back. Any pointers, please?

This should make it possible to remove the checks you commented on.

apps/bbsync/query.py
```diff
@@ -307,3 +312,30 @@ def generate_srt_notes(self):
...
+            )
+        time_minute_ago = timezone.now() - timedelta(minutes=1)
+        if (
+            pending_comments.exists()
+            and pending_comments.first().created_dt < time_minute_ago
+        ):
+            raise BadRequest(
+                "Cannot add a new pending comment that has been submitted more "
+                "than a minute ago. New comment request running too long."
+            )
```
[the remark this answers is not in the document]
jsvob: I will remove the check. Related to discussion https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388

apps/bbsync/query.py
```diff
@@ -307,3 +312,30 @@ def generate_srt_notes(self):
...
+
+        if pending_comments.exists():
+            self._query["comment"] = {
+                "body": pending_comments.first().text,
+                "is_private": False,
```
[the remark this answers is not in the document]
jsvob: You also provided this feedback: https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1191123381

Please specify how exactly `is_private` should be handled in this PR.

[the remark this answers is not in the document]
jsvob: Seems that the link gets mangled by github.

```
https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1191123381
```

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: This is bizarre, github mangles the link to a meaningless one. Here's the comment's source:

```
I will remove the check. Related to discussion https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388
```

osidb/models.py
jsvob: I agree `text` is vital, hence the lines 1770-1775. This became necessary with the addition of `BugzillaSyncMixin` to `FlawComment`'s ancestors, since `BugzillaSyncMixin` forces `validate()` on `save()`.

AFAIK `meta_attr` is actually the primary and original set of information from bugzilla. Parts of it are duplicated into normal model fields. Here's the breakdown:

* the dict that is stored into `meta_attr` https://github.com/RedHatProductSecurity/osidb/blob/f41cf220cbdf0cb197dc404b9fef48996e5d0393/osidb/models.py#L1759 comes from the 3rd argument `comment`
* this `comment` dictionary comes from a list in `[FlawConvertor instance].flaw_comments` https://github.com/RedHatProductSecurity/osidb/blob/f41cf220cbdf0cb197dc404b9fef48996e5d0393/collectors/bzimport/convertors.py#L705 which comes from `self._flaw_comments` which is set in `__init__()`
* that is provided in https://github.com/RedHatProductSecurity/osidb/blob/f41cf220cbdf0cb197dc404b9fef48996e5d0393/collectors/bzimport/collectors.py#LL460C17-L460C17
* and that is loaded from bugzilla in https://github.com/RedHatProductSecurity/osidb/blob/f41cf220cbdf0cb197dc404b9fef48996e5d0393/collectors/bzimport/collectors.py#L97

If you try to delete line 1772 and rerun tests, you will see an example; I think it's `test_flaw.py:121`.

osidb/models.py
jsvob: I don't know. Is this a part of the ask in the ticket?

osidb/models.py
jsvob: Very slightly easier debugging but will do as you suggest.

osidb/models.py
jsvob: I agree that we have to act as if Django works well enough so that we don't fall into a rabbit hole of checks. Hopefully a Django config change will effect an actually working mutually exclusive/locked transactions. Just tested the check and it doesn't ensure atomicity in concurrent situations anyway. Related to discussion https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388 (`https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388`)

osidb/models.py
jsvob: I imagine three scenarios are possible:

- Client has information from OSIDB that matches OSIDB database, but OSIDB itself has an outdated information about the flaw (bugzilla was updated from elsewhere recently).

- Client has information from OSIDB, but after it got that information, OSIDB fetched an update that the client doesn't have yet.

- The two above, combined.

Does that seem right?

osidb/models.py
jsvob: The checks are here basically to prevent invalid objects to be saved during concurrent operations, but it turns out it doesn't work.

I'll delete it.

Related to discussion https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388 (`https://github.com/RedHatProductSecurity/osidb/pull/209/#discussion_r1209214388`)

osoukup:
> Thank you all for the feedback. Please re-review.
> 
> I believe the updated PR addresses all feedback except [the first point](https://github.com/RedHatProductSecurity/osidb/pull/209#pullrequestreview-1420441728) - _something like e.g. `/flaws/<uuid/cve_id>/comment`_.
 
I think that this URL composition under flaw is the correct way.

>     * The first point is still standing, right? Based on the discussion in [Add a new FlawReference API #215](https://github.com/RedHatProductSecurity/osidb/pull/215) I presume that instead of `FlawCommentView`, I should use `@action` inside `FlawView`, right?

I have no experience with `@action` so I would trust @Elkasitu but it seems to me based on the feedback from @jobselko that the implementation is not so straightforward. The library approach with the nested router looked simpler to me but if the library is not well maintained I understand that it would be a problematic dependency.
       
>       * Which methods should be supported for comments? With a separate viewset, implementing list and retrieve in addition to create seemed straightforward and intuitive from a presumed user's perspective. But with an ad-hoc non-viewset endpoint, should there be purely create/POST support and nothing else for comments?

Regardless of the approach taken there should be **get** on both `/flaws/{uuid}/comments` and `/flaws/{uuid}/comments/{uuid}` and **post** on the first one. No **put** or **delete** as it is obviously not possible in Bugzilla.
 
>     * Are there any issues with the schema? I'm not familiar with how the schema should look like or how to properly use the schema-generating decorators on the serializer and viewset, so providing feedback by pointing me in the right direction for further research and learning would be valuable, please.

Schema LGTM.

apps/bbsync/query.py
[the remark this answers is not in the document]
osoukup: Interesting analysis! I would say that we should balance the safeness and the complexity. It is not such a big deal if there are two concurrent comments really heading to Bugzilla without properly catching the collision. What would be a serious issue however is if OSIDB got in some blocked state when no one can add comments.

We can pretty much (probably not totally perfectly) mitigate this on our API as the potential collision time between hitting the API and checking the comment order in the DB should be very small. It is harder between OSIDB and Bugzilla where the minimum collision window will be maybe 1s all the time. I would just accept it and not go crazy about it :smile:

The real probability of two concurrent comments is then very small so I think an acceptable risk and eventual duplicite comment is not such a big deal.

apps/bbsync/query.py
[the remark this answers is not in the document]
osoukup: I see! So it is a misunderstanding as I am talking about `existing comment changes`. When creating a new comment there should be an option of making it public or private. It is also in the Jira task description.

osidb/models.py
osoukup: I am not sure I get it but it is not such a big deal. You can leave it as it is.

osidb/models.py
osoukup: Maybe just a personal preference :smile:

osidb/models.py
osoukup: As I mentioned above, let us not go crazy with ensuring the mutual exclusion. Just minimize it but that should be done with already existing checks. To much complexity is IMO not worth as the risk is already low anyway.

osidb/models.py
osoukup: I probably did not understand the alert. I thought it is just between the client and OSIDB. If not so then never mind and you can leave it as it is.

Elkasitu (commented):
Some high-level comments / remarks

There should be no `/flaw_comments` endpoint as there's no use in having a view with all the comments for every single view.

The view / action should not include filters with backreferences (e.g. you can do `flaws?comments__external_system_id=` but not `flaws/CVE-2022-1234/comments?flaw__uuid=`).

I don't think we should require "order" as an attribute when creating comments, it's a BZ mechanic that has no place in OSIDB (the model should probably be ordered by `["order", "created_dt", "uuid"]`) the only reason we have it is for backwards-compatibility, but once we're through with Bugzilla there is no point keeping track of it.

I will review the actual code once these remarks are addressed, thanks!

osoukup:
> I don't think we should require "order" as an attribute when creating comments, it's a BZ mechanic that has no place in OSIDB (the model should probably be ordered by `["order", "created_dt", "uuid"]`) the only reason we have it is for backwards-compatibility, but once we're through with Bugzilla there is no point keeping track of it.

@Elkasitu we discussed the `order` matter with @jsvob and I was for incorporating it as a prevention of duplicite/outdated comments. Basically when the client has either outdated data and there was someone else commenting in the meantime or (s)he accidentally sent the commenting request twice the `order` mismatch will prevent the action. The other approach would be to send the flaw `updated_dt` together with the comment data but that sounds worse to me as this timestamp is not part of the same model. I think that we should have some mechanism to prevent these collisions.

osoukup:
@Elkasitu you have valid points but the reasoning of modern and historical approach does not sound valid to me. The important is what we want or have to achieve. Creating duplicite comment is not a big deal. I can however easily imagine a situation where the mid-air collision is pretty big deal.

Imagine that a manager raises a Major Incident based on customer escalation and awaits that all the machinery will soon start to react. In the meantime an analyst with outdated browser data fixes a typo in the statement but also accidentally removes the Major Incident flag because it is not in the outdated data. I am sure we will have a serious escalation pretty soon and followed by a strong request to implement mid-air collision prevention.

Maybe Jira can afford it. I do not think we can. Of course this does not completely prevent a malicious actor but that malicious actor would have to be behind VPN and logged in and authorized to OSIDB. When such an actor occurs we have much bigger problem then the mid-air collisions. These safeguards are to prevent mistakes or unexpected behavior like that your already submitted change is silently discarded by the next update.

Clients just have to deal with it. They can keep the data updated without user even noticing and the UX can be still outstanding. Or if the client is very simple/stupid it can just fail on the collision. This is up to the user which client to pick and up to the client developer how to make the use comfortable. The important is that it is independent of the data collisions safeguards so we are in control of all the clients.

Elkasitu:
You assume malicious intent but in most cases people will just retry with a different setting to bypass the check, in this case `order`, and that's it, and if it goes through it's not their fault but ours but implementing a check that is so easy to skip, it's bad API design.

Creating comments is a different API endpoint so there is no way that your scenario can happen here.

To avoid what you mention we can simply allow PATCH requests (aka partial updates) or let frontends manage state on their own (which they do already most of the time), in any case the burden lies downstream because we cannot 100% know in advance whether the submitted data is based off of stale data or not.

Thinking about it, this check doesn't even work for Griffon because as a CLI application it typically does not manage state, please explain how can Griffon know when was the last time a user viewed the data before submitting changes? Writing to a file on disk? Having it be a daemon? Seems overkill to comply with such a requirement. What if a user looks at the data from BZ or OSIM but submits it through Griffon?

I have yet to see a REST API that requests such obnoxious parameters for something that only works sometimes, and OSIDB is not a special application, it's a REST API like any other. Good API design makes the interface convenient for its consumers, not for its producers.

osidb/models.py
jsvob: Perhaps by `caller` you mean an API user, whereas I meant it as a caller (i.e. python code) of this method. So I'll change the comment to say "caller of this method" to be clear.

osidb/models.py
jsvob: I'm enhancing the docstring of this validation based on this discussion.

jsvob:
https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4 addresses all review feedback except for today's discussion about POST idempotence. It's just a follow-up commit for ease of review. There will be another one for `order` once we come to an agreement.

My understanding was that we want to force the client to have a sufficiently up-to-date idea of the flaw to prevent human confusion with out-of-date comments posted through an out-of-sync client.

I think there's a precedent for this practice of caution already. `git` works the very same way - multiple people working on the same branch can't magically bifurcate the head of a branch by pushing commits that share the same ancestor. One of them succeeds and the other one has to rebase or force-push. I imagine an OSIDB client, even a CLI-based one, should force the human to re-evaluate what the human wants to say if the discussion has already moved past the human's memory of the discussion. I think in the situations OSIDB will be used, a comment that is written without the knowledge of the preceding _n_ comments might be unwanted from an organizational or process perspective. Perhaps if the human really wants to post a comment no matter what, they can either create a shell-based workaround.

Or perhaps better, we can make the `order` field optional in the POST method? That way those users who want the cautious check can provide `order` and those users who want just an ordinary non-idempotent POST that just simply posts the comment with the accompanying risks can omit `order`. Would that work @Elkasitu @osoukup ?

As for the new API endpoint paths, it exactly follows the guidance 

> Regardless of the approach taken there should be **get** on both `/flaws/{uuid}/comments` and `/flaws/{uuid}/comments/{uuid}` and **post** on the first one. No **put** or **delete** as it is obviously not possible in Bugzilla.

However, it does so without `@action`, going against what we previously discussed off-github. The reason is that I ended up with quite a lot of code that essentially duplicated what ViewSet already does. The current approach IMO fits OSIDB code style (or rather code structure?) more than the use of `@action` would. We can have a videocall where I show a working prototype with `@action` for comparison if you want.

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: Resolved off-github.

osidb/models.py
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

osidb/models.py
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

osidb/models.py
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

osidb/models.py
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

docs/user/TUTORIAL.md
```diff
@@ -555,3 +555,181 @@ assert response.ok
...
+
+## Fetching flaw comments
+
+### Fetching all flaw comments
+
+With cURL
+
+```bash
+$ curl -H "Authorization: Bearer <access_token>" \
+       ${SERVICE_URL}/osidb/api/v1/flaw_comments
```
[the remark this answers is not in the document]
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: Resolved in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/f9ebfa5badd141532da083c3d6506b796e4868b4

osoukup:
I do not insist on the comment order as a required/existing parameter as mixed or duplicate comments are not such a big deal.  Generally the collision prevention is IMO must have but not for the comments.

osidb/api_views.py
```diff
@@ -360,6 +361,44 @@ def whoami(request: Request) -> Response:
     return Response(UserSerializer(request.user).data)
 
 
+@include_meta_attr_extend_schema_view
+@include_exclude_fields_extend_schema_view
+@extend_schema_view(
+    create=extend_schema(
+        request=FlawCommentSerializer,
```
osoukup, line 368: I think that this is not necessary as `serializer_class` defines the default serializer and is equal.

osidb/migrations/0076_auto_20230526_1959.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.19 on 2023-05-26 19:59
```
osoukup, line 1: Could you give the file a non-auto generated name? (something describing the change) It is handy when later navigating through the migrations when trying to find something.

osidb/models.py
osoukup: We can live with some extra duplicate data. You can leave it as it is. Generally the tasks are quite vague and rarely describe technical details so it is up to the implementor.

osidb/models.py
```diff
@@ -1807,6 +1841,54 @@ class Meta:
...
+
+        Usually, a flaw has a comment number 0 as its description. However,
+        if the description is missing, the first non-description comment must
+        start with number 1.
+
+        This validation is skipped for comments that have external_system_id,
+        which bugzilla always provides, so it's used just for client-submitted
+        new comments.
+        """
+        if self.flaw.comments.count() == 0:
```
osoukup, line 1878: I wonder whether the count can be actually zero. If you `self.flaw` exists then that flaw has at least one comment which is `self`. Maybe you need to exclude comments without the `external_system_id`?

osoukup (approved):
Just a few minor comments but generally LGTM. Great job!

osidb/migrations/0076_auto_20230526_1959.py
jsvob: Before the final squash and rebase, I'll make sure to regenerate the migration(s) with a meaningful name.

osidb/models.py
jsvob: I think I saw flaws that had zero comments. It might be just in tests, but in principle it's probably possible. Without this condition, 53 tests fail. Maybe caused by uncommitted instances? This is what I tried to run with:

```
        (...)
        This validation is skipped for comments that have external_system_id,
        which bugzilla always provides, so it's used just for client-submitted
        new comments.
        """
        proper_following_order = (
            1 + self.flaw.comments.aggregate(Max("order"))["order__max"]
        )
        if self.external_system_id == "" and proper_following_order != self.order:
            raise ValidationError(
        (...)
```

osoukup: It is definitely possible that the new functionality breaks existing tests and often it is just because they become wrong in the new context. Failing tests do not have to mean that your change is wrong.

It can happen that there is a flaw with no comments. My confusion however is how could it happen when there is an existing comment (`self`) and its `flaw` attribute is an existing flaw. Then it seems logical to me that this exact flaw has at least one comment (`self`) and thus I do not understand when the number could be zero.

osidb/tests/test_endpoints.py
```diff
@@ -1258,6 +1258,52 @@ def test_flaw_update_collision(self, auth_client, test_api_uri):
...
+
+        def new_flaw():
+            flaw = FlawFactory(embargoed=False)
+            AffectFactory(flaw=flaw)
+            response = auth_client.get(f"{test_api_uri}/flaws/{flaw.uuid}")
+            assert response.status_code == 200
+            assert not FlawComment._base_manager.filter(flaw=flaw).exists()
+            return flaw
+
+        def get_response(flaw, new_comment):
```
jobselko, line 1274: A better name would be `new_comment` or `create_comment`. `get_response` is too general.

osidb/models.py
```diff
@@ -1743,6 +1744,16 @@ def _validate_public_source_no_ack(self):
 class FlawCommentManager(ACLMixinManager, TrackingMixinManager):
     """flawcomment manager"""
 
+    def pending(self):
```
jobselko, line 1747: I am unsure whether we need to handle `pending` comments and whether we should assume that `ATOMIC_REQUESTS` may be False. No action here for you, I am just unsure about it.

jobselko (dismissed):
I apologize if any of my comments have already been discussed and decided. It is hard for me to get oriented in a big PR. Next time, I would recommend to work on smaller pieces to make the review easier.

osidb/models.py
jobselko: I agree with Ondrej.

apps/bbsync/tests/test_query.py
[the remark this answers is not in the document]
osoukup: I guess miss the context. If it required modifications, the tests would fail.

osidb/models.py
```diff
@@ -1753,6 +1764,25 @@ def create_flawcomment(flaw, external_system_id, comment, **extra_fields):
             flawcomment.meta_attr = comment
             return flawcomment
         except ObjectDoesNotExist:
+            # The "text" field is required for full validation, so save it if the caller of
+            # this method forgets to explicitly copy it over from meta_attr.
+            extra_fields["text"] = extra_fields.get("text") or comment.get("text")
+
+            # The "order" field is required for full validation, but bugzilla calls it "count".
+            extra_fields["order"] = extra_fields.get("order") or comment.get("count")
+
```
Elkasitu, line 1773: @osoukup 's comment at https://github.com/RedHatProductSecurity/osidb/pull/209#discussion_r1209402917 is still valid, it is expected of callers of this method to provide all the necessary fields explicitly in extra_fields, and in fact we do so in bzimport https://github.com/RedHatProductSecurity/osidb/blob/f41cf220cbdf0cb197dc404b9fef48996e5d0393/collectors/bzimport/convertors.py#L702-L719

apps/bbsync/tests/test_query.py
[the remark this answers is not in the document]
jobselko: I am sorry, I thought `apps/bbsync/tests/test_integration.py`. It contains e.g. `test_affect_create()`, `test_affect_update()`, `test_affect_delete()` so I would expect similar tests also for `comment`.

osidb/models.py
```diff
@@ -1753,6 +1764,25 @@ def create_flawcomment(flaw, external_system_id, comment, **extra_fields):
...
+            # In-place upgrade of a temporary pending comment to preserve uuid.
```
Elkasitu, line 1774: I don't think we'll be able to do this if we don't require order, but that's ok, we can just delete all pending comments and recreate them with the data provided by bzimport.

osidb/tests/factories.py
```diff
@@ -350,6 +351,35 @@ class Meta:
...
+        """
+        if "order" not in kwa:
+            if kwa["flaw"].comments.count() == 0:
+                # Comment 0 would be the flaw description.
+                # Creating a proper comment here, so starting with 1.
+                order = 1
+            else:
+                # Description is optional, so just using .count() doesn't work consistently.
+                order = 1 + kwa["flaw"].comments.aggregate(Max("order"))["order__max"]
+            kwa["order"] = order
```
Elkasitu, line 367: Can't this just be https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Sequence ?

apps/bbsync/tests/test_query.py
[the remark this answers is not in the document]
osoukup: I am leaving this to Jakub. Or you can request it if you think it should be there :smile_cat:

osidb/tests/factories.py
```diff
@@ -350,6 +351,35 @@ class Meta:
...
+        if "text" not in kwa:
+            kwa["text"] = "Test comment {}".format(kwa["order"])
```
Elkasitu, line 369: the text doesn't have to be dynamic, please look at the other Factories as an example and just declare a text field

Elkasitu (changes requested):

osidb/urls.py
```diff
@@ -19,7 +19,9 @@
 
 router = routers.DefaultRouter(trailing_slash=False)
 router.register(r"flaws", FlawView)
-router.register(r"flaw_comments", FlawCommentView)
+router.register(
+    r"flaws/(?P<flaw_id>[^/.]+)/comments", FlawCommentView, basename="flawcomments"
```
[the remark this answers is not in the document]
jobselko: This router works. However, I cannot say it will not cause any errors/warnings.

Btw, when I manually ran the command to generate openapi schema last time, there already were about 50 warnings/errors.

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: Why? This code hasn't been touched and this PR specifically doesn't support private comments.

apps/bbsync/tests/test_query.py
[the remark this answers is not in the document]
jsvob: No cassettes need modification AFAIK.

There's a test in test_endpoints, so I think there's no need to also test in test_integration.

osidb/migrations/0076_auto_20230526_1959.py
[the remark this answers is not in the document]
jsvob: Yes. Before the final squash and rebase, I'll make sure to regenerate the migration(s) with a meaningful name. I'm creating individual commits as the conversation in the review goes on. I presume nobody wants to re-review a squashed and rebased commit and guess what did or didn't change since the last time they did the review (I wouldn't for sure).

osidb/serializer.py
```diff
@@ -806,3 +806,43 @@ class Meta:
             "groups",
             "profile",
         ]
+
+
+class FlawCommentSerializer(
```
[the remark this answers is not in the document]
jsvob: There will be 3 comment-related serializers (3rd added in `76a8581033d95a8a77a598c876457e2b162a158f` purely for correct schema generation). `CommentSerializer` has a docstring that says what it is for. `CommentSerializer` and `FlawCommentSerializer` are each used for very different purpose. IMO deduplicating them will result in a serializer that will be too complicated or behaving in unintended ways. Apart from that, this question is for @osoukup and @Elkasitu , not me, to answer.

osidb/models.py
jsvob: Good catch, Ondrej already told me to remove mentions of `ATOMIC_REQUESTS`. I'll remove the paragraph.

The method is used in `BugzillaQueryBuilder` so I think we still need it.

osidb/api_views.py
```diff
@@ -369,11 +369,34 @@ def whoami(request: Request) -> Response:
...
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def get_queryset(self):
+        flaw = self.get_flaw()
+        return FlawComment.objects.filter(flaw=flaw)
+
+    def create(self, request, *args, **kwargs):
+        flaw_uuid = self.get_flaw().uuid
+        Flaw.objects.select_for_update().get(uuid=flaw_uuid)
```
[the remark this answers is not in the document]
jsvob: That was to ensure two concurrent transactions don't each add a new comment. Bugzilla tends to not handle concurrency well. This effectively serialized the POST calls. It's not necessary anymore. I'll remove it.

osidb/models.py
```diff
@@ -1844,6 +1841,11 @@ class Meta:
             "external_system_id",
             "created_dt",
         )
+        constraints = [
+            models.UniqueConstraint(
```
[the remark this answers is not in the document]
jsvob: Removed in https://github.com/RedHatProductSecurity/osidb/pull/209/commits/76a8581033d95a8a77a598c876457e2b162a158f

osidb/models.py
jsvob: My goal was to make the POST endpoint return the actual uuid of the new comment. But yeah, if concurrent requests are made, multiple pending comments are going to be created. Perhaps a simple way to resolve that would be to just delete all pending comments in FlawSaver cleanup. Pending comments are marked by blank `external_system_id` and it wasn't blank-able until now, so I think bugzilla will never provide a blank value in there.

osidb/tests/factories.py
jsvob: It can! Since we are removing the "order" validation, it doesn't matter what's in there. I can remove the 3 new methods and make the factory changes really simple then.

osidb/api_views.py
```diff
@@ -369,11 +369,34 @@ def whoami(request: Request) -> Response:
...
+        request.data["flaw"] = str(flaw_uuid)
```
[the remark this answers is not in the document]
jsvob: I tried multiple ways to pass the flaw to the serializer and still make most of the stack behave as expected and do its automagic without a lot of code and I'm not aware of an elegant way to do that. If there is one, I'm interested to see it. I guess if it breaks (e.g. requests.data starts being immutable in some future version), we'll deal with it.

osidb/serializer.py
```diff
@@ -817,20 +817,16 @@ class FlawCommentSerializer(
...
-
-        Requires ATOMIC_REQUESTS so that temporary pending comment instances are
-        not left over after unsuccessful calls, as they could have unintended
-        consequences such as submitting long-forgotten comments to bugzilla later.
+        Create FlawComment instance by deserializing input. Force empty
+        external_system_id to force submitting and redownloading the
+        new comment through bugzilla.
         """
 
         validated_data["external_system_id"] = ""
```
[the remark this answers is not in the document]
jsvob: Nice! That works.

osidb/models.py
jsvob: I'll fix the invalid call in `test_flaw.py` and remove these copy operations then.

osidb/models.py
jsvob: Resolved in commit 76a8581033d95a8a77a598c876457e2b162a158f

osidb/api_views.py
jsvob: Resolved in commit 76a8581033d95a8a77a598c876457e2b162a158f

osidb/tests/factories.py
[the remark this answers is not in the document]
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/tests/factories.py
```diff
@@ -350,6 +351,35 @@ class Meta:
         "creation_time": "2006-03-30T11:56:45Z",
     }
 
+    @staticmethod
+    def prepare_order_text(kwa):
```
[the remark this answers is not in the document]
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/tests/test_endpoints.py
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/api_views.py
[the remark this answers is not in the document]
jsvob: In the end, I found that during history, `request.data` wasn't always modifiable, so I did it in a bit more verbose but hopefully elegant enough way - https://github.com/RedHatProductSecurity/osidb/pull/209/commits/a3415f9b5cde43c7e4fb3ac8476e8791fd710365#diff-4a51610181bd496db117d73122150acb06a494798165c70ce3c1b15252430ceaR396-R403

Further improvement feedback welcome.

osidb/serializer.py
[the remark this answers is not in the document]
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/models.py
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/tests/factories.py
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/tests/factories.py
jsvob: Resolved in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365

osidb/models.py
jsvob: Cleanup added in commit a3415f9b5cde43c7e4fb3ac8476e8791fd710365 - https://github.com/RedHatProductSecurity/osidb/pull/209/commits/a3415f9b5cde43c7e4fb3ac8476e8791fd710365#diff-ebe1c224bdd0ffc9f247b35caecd1b03746276b968e7cbf57200af9c5958cd76R253

osidb/urls.py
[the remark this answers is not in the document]
jsvob: Yes, there are additional errors when compared against the `master` commit this branch is based on:

```
Warning #8: FlawCommentView: could not derive type of path parameter "flaw_id" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:flaw_id>) or annotating the parameter type with @extend_schema. Defaulting to "string".
Warning #9: FlawCommentView: Failed to obtain model through view's queryset due to raised exception. Prevent this either by setting "queryset = Model.objects.none()" on the view, checking for "getattr(self, "swagger_fake_view", False)" in get_queryset() or by simply using @extend_schema. (Exception: 'flaw_id')
Warning #10: FlawCommentView: could not derive type of path parameter "comment_id" because it is untyped and obtaining queryset from the viewset failed. Consider adding a type to the path (e.g. <int:comment_id>) or annotating the parameter type with @extend_schema. Defaulting to "string".
```

I'll look into it on Monday to try and fix these errors. @Elkasitu @osoukup in case you already did something like that in the past, any tips or pointers, please?

osidb/models.py
jsvob: Also found this gem https://github.com/bugzilla/bugzilla/blob/2f203a360eab075742d42fd228a4233a3d45d221/Bugzilla/Bug.pm#L3760

jsvob:
In commit 3e83eb9f8ce1ad03a96205e7cf069655a7c1fd86 I reintroduce flaw locking and UniqueConstraint. I performed numerous experiments with various combinations of locking, constraints, cleanup and debug output, and this seems to be the most reasonable solution to me.

- Cleanup of duplicate instances is of no use. The reason is that duplicate objects raise `MultipleObjectsReturned` in https://github.com/RedHatProductSecurity/osidb/blob/188b2c9fa0c302e71895d8876c6a78f5cc929003/osidb/models.py#L1782 and handling that would carry multiple issues (let me know if you want details).
- Without the locking introduced in commit 3e83eb9f8ce1ad03a96205e7cf069655a7c1fd86, duplicate instances can appear as a result of concurrent transactions (POST requests) where each transaction in itself is perfectly fine, but their composition after they are committed results in an invalid state. This is enabled by the non-`SERIALIZABLE` default value of [`isolation_level`](https://docs.djangoproject.com/en/2.2/ref/databases/#isolation-level) in osidb's db django config. Setting it to SERIALIZABLE might carry multiple issues (let me know if you want details).
- The most proper solution, AFAIK, would be to do `django.db.connection.cursor().execute(f"LOCK TABLE {model._meta.db_table} IN ACCESS EXCLUSIVE MODE")` but this doesn't seem right to me (not portable to other db backends, also sidesteps django so future compatibility might be questionable).
- Not handling concurrency will cause OSIDB to be unusable in automated or bulk operations that create comments.

osidb/api_views.py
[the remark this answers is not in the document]
jsvob: Further slightly improved by a comment in commit 3e83eb9

jsvob:
At this point, I think there are only 4 outstanding TODOs marked by the respective open conversations:

- @jsvob to fix new warnings in schema generation (will do on monday)
- @jsvob to deduplicate migrations and ensure they have descriptive names (will do after you finish review)
- @jsvob to squash and rebase before merge
- @osoukup and @Elkasitu to answer @jobselko's question about seemingly duplicated serializers

osidb/models.py
jsvob: Actually, in commit 3e83eb9 I removed the cleanup and added locking with a comment that explains it. Delete & recreate without that locking wouldn't help because multiple concurrent transactions would simply delete & recreate in a way that would be correct from each transaction's perspective, but invalid in composition after transactions are committed. The reason is that transactions in django don't lock the table, again detailed in the comment in commit 3e83eb9 and in comment https://github.com/RedHatProductSecurity/osidb/pull/209#issuecomment-1574889681

osidb/models.py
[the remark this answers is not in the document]
jsvob: And reintroduced in commit 3e83eb9 but this time with an explanatory comment. Additional context in https://github.com/RedHatProductSecurity/osidb/pull/209#issuecomment-1574889681

osidb/api_views.py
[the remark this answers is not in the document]
jsvob: And added back in commit 3e83eb9. Context in https://github.com/RedHatProductSecurity/osidb/pull/209#issuecomment-1574889681

osidb/models.py
osoukup: Detective work :smile:

apps/bbsync/query.py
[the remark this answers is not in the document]
osoukup: This is probably some outdated artifact I created when it was not yet decided whether we will support the private Bugzilla comments or not. We do not need to so really outdated piece.

osidb/serializer.py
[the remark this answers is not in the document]
osoukup: The naming may be confusing but otherwise if you need a different/extended behavior for the comment endpoint (like various mixins) the approach seems valid to me.

osidb/urls.py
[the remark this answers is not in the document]
osoukup: I know we have some outstanding warnings in the schema generation and I am not sure all of them can be easily resolve (not too familiar with that). The important is whether the resulting schema looks as we expect. You can check the generated documentation on the `swagger` web page of your local instance and check that there are all expected fields and no more and that they are mandatory or read-only as they should.

osidb/serializer.py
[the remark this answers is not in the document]
Elkasitu: You can do a number of things here, but I think at most you only need two serializers:

- One for the `/flaws` view, which is essentially always read-only
- One for the `/flaws/<id>/comments` view which allows both read and write operations

The first one is probably OK as-is.
The second one just needs to have the `flaw`, `order` and `external_system_id` fields declared as read-only attributes, and I would even go as far as having `order` and `flaw` omitted because their value is either obvious or uninteresting to clients.

You can probably also simplify the new serializer to inherit from the existing CommentSerializer so that it fits your needs with minimal code duplication, but I leave that as a wishlist item as I don't think it's such a big deal.

Elkasitu:
> In commit [3e83eb9](https://github.com/RedHatProductSecurity/osidb/commit/3e83eb9f8ce1ad03a96205e7cf069655a7c1fd86) I reintroduce flaw locking and UniqueConstraint. I performed numerous experiments with various combinations of locking, constraints, cleanup and debug output, and this seems to be the most reasonable solution to me.
> 
>     * Cleanup of duplicate instances is of no use. The reason is that duplicate objects raise `MultipleObjectsReturned` in https://github.com/RedHatProductSecurity/osidb/blob/188b2c9fa0c302e71895d8876c6a78f5cc929003/osidb/models.py#L1782
>        and handling that would carry multiple issues (let me know if you want details).
> 
>     * Without the locking introduced in commit [3e83eb9](https://github.com/RedHatProductSecurity/osidb/commit/3e83eb9f8ce1ad03a96205e7cf069655a7c1fd86), duplicate instances can appear as a result of concurrent transactions (POST requests) where each transaction in itself is perfectly fine, but their composition after they are committed results in an invalid state. This is enabled by the non-`SERIALIZABLE` default value of [`isolation_level`](https://docs.djangoproject.com/en/2.2/ref/databases/#isolation-level) in osidb's db django config. Setting it to SERIALIZABLE might carry multiple issues (let me know if you want details).
> 
>     * The most proper solution, AFAIK, would be to do `django.db.connection.cursor().execute(f"LOCK TABLE {model._meta.db_table} IN ACCESS EXCLUSIVE MODE")` but this doesn't seem right to me (not portable to other db backends, also sidesteps django so future compatibility might be questionable).
> 
>     * Not handling concurrency will cause OSIDB to be unusable in automated or bulk operations that create comments.

I don't understand any of this, in my mind creating a comment is pretty simple:

1. User A and User B submit POST request to create a comment at the same time
2. Both comments exist in OSIDB only, their ordering is perhaps non-deterministic if `created_dt` is exactly the same but still no big deal.
3. Comments are synchronized to BZ, if either comment fails at this point I believe we're still within the request-response lifecycle so an error will be presented to the user and the comment rolled back.
4. On next sync from bugzilla all pending comments will be dropped and the ones that succeeded will be recreated with proper ordering.

So I still don't get why we need to lock anything or even have a unique together constraint.

osidb/urls.py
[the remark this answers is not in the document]
Elkasitu: It's not necessary to clean up the warnings if it's only warnings, my questions was more whether this could introduce some sort of confusion between the two views when resolving the route, which does not seem to be the case.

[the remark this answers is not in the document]
jsvob: In the end I fixed the new warnings because I was too deep in learning how schema generation works. The new part of the schema was somewhat nonsensical, so it needed a fix anyway.

Fixed in commit 7ff735b65a892a63e3be3f14668e70356743c595.

As for the route resolving, it should work fine. http://127.0.0.1:8000/osidb/api/v1/blablabla shows the generated routes and they seem fine (thanks @jobselko for the tip).

jsvob:
> I don't understand any of this, in my mind creating a comment is pretty simple:

With a simple app it would be. The main problem here is that osidb calls bugzilla. IMO, if you want to make the comment sync code in this PR any simpler, you have to give up on one or both of these qualities:

- Client is able to issue multiple POST requests shortly after each other or simultaneously.
- The Response to the POST request contains the correct data for the new comment.

> User A and User B submit POST request to create a comment at the same time

Wait, so OSIDB can have multiple users? I thought it's a localhost-only thing. That makes even a stronger argument for having the ability to let clients issue concurrent POST requests to OSIDB.

> Both comments exist in OSIDB only, their ordering is perhaps non-deterministic if created_dt is exactly the same but still no big deal.

If they can exist outside of a transaction, you need a cleanup during a subsequent sync and need to make sure that comment listing before that cleanup happens doesn't contain non-bz comments. If you want that cleanup to be simple, you probably can't have useful data in the Response to each POST request (or please teach me how to solve that concurrency issue).

> Comments are synchronized to BZ, if either comment fails at this point I believe we're still within the request-response lifecycle so an error will be presented to the user and the comment rolled back.

No, we are not within that lifecycle and no, no rollback is guaranteed unless you configure `isolation_level` to `SERIALIZABLE`.

Please read the explanation I already provided

>> duplicate instances can appear as a result of concurrent transactions (POST requests) **where each transaction in itself is perfectly fine,** but their composition after they are committed results in an invalid state

or provide data to the contrary.

> On next sync from bugzilla all pending comments will be dropped and the ones that succeeded will be recreated with proper ordering.

We can do that without much complication (I hope; except for about an extra week of work) if we want to introduce these two qualities to OSIDB:

- No useful Response to POST requests
- Comment listing might contain temporary or duplicate comments (and if you don't want that, it adds complexity back into the code, but in a way that goes against feedback that was already provided in this PR).

> So I still don't get why we need to lock anything or even have a unique together constraint.

Please schedule a call or try to change the code to your liking and then test concurrent POST requests. Maybe I don't see something.

Thanks!

jsvob:
I'll squash and rebase (and deduplicate migrations) tomorrow, and ask for a merge tomorrow if no further feedback before then.

apps/bbsync/query.py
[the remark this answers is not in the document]
jsvob: Removed in commit 08de75ecf2228d4c0a09bed6e40aa9948474725b

jsvob:
Squashed and rebased.

@osoukup @Elkasitu Please merge or edit and merge as you deem appropriate.

docs/user/TUTORIAL.md
```diff
@@ -555,3 +555,161 @@ assert response.ok
...
+flaw_id = "CVE-2005-0001"
+comment_id = "acef1a6c-890b-4abb-85c1-d81e76cd7858"
+headers = {"Authorization": "Bearer <access_token>"}
+response = requests.get(f"${SERVICE_URL}/osidb/api/v1/flaws/{flaw_id}/comments/{comment_id}", headers=headers)
+assert response.ok
+```
+
+### Creating a flaw comment
+
+Please note that multiple comments are created in parallel are not guaranteed to keep their OSIDB-internal `uuid`.
```
jobselko, line 680: Redundant `are`.

docs/user/TUTORIAL.md
```diff
@@ -555,3 +555,161 @@ assert response.ok
...
+```
+
+With python
+
+```python
+import requests
+
+flaw_id = "2fe16efb-11cb-4cd2-b31b-d769ba821073"
+headers = {"Authorization": "Bearer <access_token>"}
+response = requests.get(f"${SERVICE_URL}/osidb/api/v1/flaws/{flaw_id}/comments", headers=headers)
```
jobselko, line 580: Redundant `$` (also on lines 627, 650, 674, 709).

Btw. this readme file is outdated and contains more known errors/inaccuracies (they will be fixed sometimes).

docs/user/TUTORIAL.md
```diff
@@ -555,3 +555,161 @@ assert response.ok
...
+    "text": "A jacket was found.",
+    "embargoed": False,
+}
+response = requests.post(f"${SERVICE_URL}/osidb/api/v1/flaws/{flaw_id}/comments", headers=headers, json=data)
+assert response.ok
+```
+
+### Updating or deleting flaw comments
+
+Updating or deleting flaw comments is not supported.
```
jobselko, line 715: I would add `in Bugzilla` at the end.

osidb/api_views.py
```diff
@@ -376,6 +397,84 @@ def whoami(request: Request) -> Response:
...
+    retrieve=extend_schema(
+        description="Retrieve a single existing comments for a given flaw. Beware that freshly "
+        "created comments are not guaranteed to keep their original UUIDs, especially "
+        "if multiple comments are created simultaneously.",
+        parameters=[
+            flaw_id,
+            comment_id,
+        ],
+    ),
+)
```
jsvob, line 439: @Elkasitu  Please review the three new `description`s for schema generation.

jsvob:
@Elkasitu IDK if "Partially verified" commits are fine. If not, please re-sign your commits (and I guess I'll then have to re-sign the third commit in the sequence).

jsvob:
Test `test_flawreference_create` fails. I guess I'll need to wait a few days for `master` to get a fix and then rebase, right?

jobselko:
> Test `test_flawreference_create` fails. I guess I'll need to wait a few days for `master` to get a fix and then rebase, right?

My apologies. The fix for master is in #235.

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement statement validation for Major Incident flaw (OSIDB-657)
 - Introduce new module for creating trackers in Jira (OSIDB-93)
 - Introduce aditional metadata in tasks generated from Taskman (OSIDB-861)
+- Implement adding new flaw comments (OSIDB-81)
```
jobselko, line 20: This should be in the `Unreleased` section.

jsvob: D'oh! Thanks for catching it.

Elkasitu (approved):
Maybe @osoukup wants to give it a quick review? Especially the two new commits since they affect BBSync

osoukup:
> Maybe @osoukup wants to give it a quick review? Especially the two new commits since they affect BBSync

@Elkasitu @jsvob the last two commits LGTM

PR #212, opened 2023-05-15
make CVSSv3 score mandatory no more

osoukup:
It should be possible to create a flaw without CVSS3 provided.
This was identified as a discrepancy in the required OSIDB functionality during the IR acceptance testing

https://docs.google.com/spreadsheets/d/1JXjtwpCbpQesXWOQCkuh8rzkR3MbweM2zuEcBlX-bpk/edit?usp=sharing

Closing OSIDB-901

### DIFF
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -846,22 +846,17 @@ class TestFlawValidators:
             # no exception should be raised now
 
     @pytest.mark.parametrize(
-        "cvss3,should_raise",
+        "cvss3,should_alert",
         [
             ("3.7/CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N", False),
             (None, True),
         ],
     )
-    def test_validate_cvss3_model(self, cvss3, should_raise):
+    def test_validate_cvss3_model(self, cvss3, should_alert):
         """
-        Test that the ValidationError is not raised when the flaw has a CVSS3 string
+        Test that an alert is not raised when the flaw has a CVSS3 string
         """
-        if should_raise:
-            with pytest.raises(ValidationError) as e:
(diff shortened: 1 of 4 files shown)


Elkasitu (commented):
The code LGTM but does this mean that a Flaw can be published without cvss2/cvss3? Or is it more of a draft flaw problem?

osoukup:
> The code LGTM but does this mean that a Flaw can be published without cvss2/cvss3? Or is it more of a draft flaw problem?

Definitely a draft flaw problem. If we had the workflow (OSIM module) we could solve this properly. Actually as I am more thinking about it we might rather turn this into a soft alert instead of completely removing it. I will ask Marian

osoukup:
@Elkasitu so after discussing with Marian I am turning this into a non-blocking alert

osidb/tests/test_flaw.py
```diff
@@ -846,22 +846,17 @@ def test_validate_future_unembargo_date(self):
...
-        Test that the ValidationError is not raised when the flaw has a CVSS3 string
+        Test that an alert is not raised when the flaw has a CVSS3 string
         """
-        if should_raise:
-            with pytest.raises(ValidationError) as e:
-                FlawFactory(cvss3=cvss3)
-            assert "CVSSv3 score is missing" in str(e)
-        else:
-            assert FlawFactory(cvss3=cvss3)
+        assert should_alert == bool("cvss3_missing" in FlawFactory(cvss3=cvss3)._alerts)
```
Elkasitu, line 859: the `in` operator always returns a boolean value

Elkasitu (approved):

osidb/tests/test_flaw.py
osoukup: True... I just copy-pasted another test. I will simplify

PR #213, opened 2023-05-16
Gate OSIDB response revision information behind env var

Elkasitu:
This new env var allows us to disable the call to the git rev-parse command that is appended to every single OSIDB response, the default behavior is to disable.

This is due to a new revision of git including a security check that makes any git command fail depending on the repository permissions, meaning that it can break every single server response with the following error message: "fatal: detected dubious ownership".

A proper fix will follow.

### DIFF
--- a/osidb/renderers.py
+++ b/osidb/renderers.py
@@ -9,6 +9,7 @@ from django.utils import timezone
 from rest_framework.renderers import JSONRenderer
 
 import osidb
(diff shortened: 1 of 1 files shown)


costaconrado (approved):
LGTM.

osoukup (approved):
LGTM. Thank you for handling it!

PR #215, opened 2023-05-18
Add a new FlawReference API

jobselko:


openapi.yml
```diff
@@ -1649,6 +1649,194 @@ paths:
                   version:
                     type: string
           description: ''
+  /osidb/api/v1/flaw_references:
```
jobselko, line 1652: My local instance has kinda broken Bugzilla-related things, so I am working with partially disabled githooks. Therefore, I cannot guarantee that the `openapi.yml` file is correct, so I rely on CI tests here.

osidb/api_views.py
```diff
@@ -331,6 +333,18 @@ def create(self, request, *args, **kwargs):
         return response
 
 
+@extend_schema_view(
```
jobselko, line 336: I am not sure if I should add here also `@include_meta_attr_extend_schema_view` and `@include_exclude_fields_extend_schema_view`.

openapi.yml
osoukup: Generating the API documentation is based on DRF stuff and should not be blocked by broken Bugzilla integration or failing Bugzilla tests. Just run `make update-schema` and `podman exec -it testrunner tox -e schema`

openapi.yml
```diff
@@ -5303,11 +5502,86 @@ components:
...
+      - updated_dt
+      - url
+      - uuid
+    FlawReferencePost:
+      type: object
+      description: FlawReference serializer
+      properties:
+        description:
+          type: string
+        flaw:
```
osoukup, line 5557: This should be rather in path like here https://github.com/RedHatProductSecurity/osidb/blob/master/openapi.yml#L4368

openapi.yml
```diff
@@ -5875,6 +6169,11 @@ components:
       - JIRA
       - BUGZILLA
       type: string
+    Type154Enum:
```
osoukup, line 6172: I was discussion with Marian that they would like to also have a type `REPRODUCER` and `PATCH`. Has he reached out to you already? That should be pretty small change and it would cover maybe 10 additional acceptance test cases.

osidb/api_views.py
osoukup: The schema extension is only for the sake of documentation and it should correspond to the serializer. There is no `meta_attr` in the model so no need for that one and so the serializer correctly does not inherit `IncludeMetaAttrMixin`. Also the serializer does not inherit the `IncludeExcludeFieldsMixin` and I think that is OK for such a small model. Only if we wanted to have it everywhere for the sake of consistence. I am not sure what is the good practice here. WDYT @Elkasitu ?

osoukup: I see that even `ErratumSerializer` with almost no fields inherit `IncludeExcludeFieldsMixin` so maybe we want to have it everywhere so we are consistent.

osoukup (changes requested):
I have a few comments. Generally you are on the right path :+1:

openapi.yml
jobselko: He told me about it this week, but I would prefer to create a new Jira with a detailed description (e.g. if we need validations etc.).

osidb/api_views.py
jobselko: OK, let's wait for @Elkasitu's opinion.

openapi.yml
osoukup: I do not have any preferences on how to track it. It would be good to deliver it as part of this initiative (can be a different PR, does not matter) so we can claim the reference functionality done as a whole.

osoukup (approved):
LGTM. Only rebase the commits so the intermediate changes which were later reconsidered are not in the Git history

osidb/serializer.py
```diff
@@ -621,6 +622,31 @@ def to_representation(self, value):
...
+            + ACLMixinSerializer.Meta.fields
+            + TrackingMixinSerializer.Meta.fields
+        )
+
+
+@extend_schema_serializer(exclude_fields=["updated_dt"])
+class FlawReferencePostSerializer(FlawReferenceSerializer):
+    # extra serializer for POST request as there is no last update
+    # timestamp but we need to make the field mandatory otherwise
+    pass
```
Elkasitu, line 648: this wouldn't be necessary if `updated_dt` was a simple `read_only=True` field, not that there's anything you can do here but see my comment at https://github.com/RedHatProductSecurity/osidb/pull/165#issuecomment-1559125237

osidb/urls.py
```diff
@@ -21,12 +23,16 @@
 router.register(r"affects", AffectView)
 router.register(r"trackers", TrackerView)
 
+nested_router = NestedDefaultRouter(router, r"flaws", lookup="flaw")
```
Elkasitu, line 26: This is interesting, but given that it's already possible to create the same behavior with vanilla DRF [1] I'd prefer not to add this dependency

[1] https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

osidb/api_views.py
```diff
@@ -331,6 +333,19 @@ def create(self, request, *args, **kwargs):
...
+    create=extend_schema(
+        request=FlawReferencePostSerializer,
+    ),
+)
+class FlawReferenceView(ModelViewSet):
+    queryset = FlawReference.objects.all()
+    serializer_class = FlawReferenceSerializer
+    http_method_names = get_valid_http_methods(ModelViewSet)
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
Elkasitu, line 346: If you use actions you won't need the view at all

openapi.yml
jobselko: I asked Marian to create a Jira with all details, so we can discuss it later.

osoukup: The reproducer is definitely an internal thing so it will never go to the CVE pages or so. I am not sure about the patch. The purpose should be described by IR team. There are already those types of links being added probably often but just as the external type.

Elkasitu: My question is basically "will anyone actually use this?" we probably don't even need a reference type, I am sure SDEngine could tell which one is a security bulletin or not just by looking at the URL

osoukup: We do not know whether and how is this going to use. IR team would like to have it and it is a little work to add it so I do not think this is a big deal. If it was a new functionality then the reasoning should be more precise.

Maybe SDEngine could recognize them but 1) the restriction that it begins with `https://access.redhat.com/` does not ensure it (CVE pages and much more is also there) 2) the location may change which should not make the historical articles (or other references) not being articles any more 3) the restriction is not even enforcing and based on analysis Jitka did the actual data are wild west so auto-recognition would be unreliable.

osidb/urls.py
jobselko: I have not been able to get all `@action` working yet, so at least I removed the external library and updated the router directly.

When manually running the command to generate openapi schema, l get only this warning: `Warning #8: FlawReferenceView: could not derive type of path parameter "id" because model "osidb.models.FlawReference" contained no such field. Consider annotating parameter with @extend_schema. Defaulting to "string".`

openapi.yml
jobselko, line 1: @Elkasitu @osoukup Do you happen to know what could be the problem in openapi that is causing the tests to fail?

Elkasitu (approved):
LGTM

PR #220, opened 2023-05-24
reflect Bugzilla attribute removals (OSIDB-910)

osoukup:
On number of places an attribute removal in Bugzilla was not reflected in OSIDB. The reason was always the same. There was an if condition checking the attribute presence so it is only assigned when present but there was no else branch so when not present the old value was simply left as it is. The correct approach is to empty the value when the attribute in Bugzilla is removed as it is still the authoritative source of the data.

Closes OSIDB-910


Elkasitu (approved):
LGTM!

PR #221, opened 2023-05-25
unembargo date tests

osoukup:
I have written a few tests trying to reproduce OSIDB-912 but it turned out that the issue was actually with the Bugzilla stage missing some accounts... I am submitting at least the tests.

### DIFF
--- a/osidb/tests/test_endpoints.py
+++ b/osidb/tests/test_endpoints.py
@@ -1225,6 +1225,101 @@ class TestEndpoints(object):
         assert body["embargoed"] == embargoed
         assert body["cve_id"] == new_cve_id
 
+    @pytest.mark.parametrize(
+        "embargoed,old_date,new_date,alerts",
+        [
+            (
+                False,
+                datetime(2011, 1, 1, tzinfo=timezone.utc),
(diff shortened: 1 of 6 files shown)


osidb/tests/test_endpoints.py
```diff
@@ -1225,6 +1225,97 @@ def test_flaw_update_cve(
...
+                False,
+                datetime(2011, 1, 1, tzinfo=timezone.utc),
+                datetime(2012, 1, 1, tzinfo=timezone.utc),
+                False,
+            ),
+            (
+                False,
+                datetime(2021, 1, 1, tzinfo=timezone.utc),
+                datetime(2022, 1, 1, tzinfo=timezone.utc),
+                True,
```
Elkasitu, line 1241: why is this different from the above? maybe we can explain in the test method docstring each case?

Elkasitu (approved):
LGTM but confused about the test cases (don't remember the validations)

osidb/tests/test_endpoints.py
osoukup: I will add some explanation. Basically the issue is that you are assigning a future public date to an already public flaw which is obviously not correct.

PR #225, opened 2023-05-30
Add extra flaw metadata for tasks

costaconrado:
This PR introduce new metadata in Tasks needed for more meaningful Jira dashboards.
 
Closes OSIDB-940.

### DIFF
--- a/apps/taskman/service.py
+++ b/apps/taskman/service.py
@@ -10,7 +10,7 @@ from jira.exceptions import JIRAError
 from rest_framework.response import Response
 
 from collectors.jiraffe.core import JiraQuerier
-from osidb.models import Affect, Flaw
+from osidb.models import Affect, Flaw, PsProduct
 
 from .constants import JIRA_TASKMAN_PROJECT_KEY, JIRA_TASKMAN_URL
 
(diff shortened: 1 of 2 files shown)


apps/taskman/service.py
```diff
@@ -93,12 +93,26 @@ def create_or_update_task(self, flaw: Flaw, fail_if_exists=False) -> Response:
...
+        labels = [f"flawuuid:{str(flaw.uuid)}", f"impact:{flaw.impact}"]
+        for product in products:
+            if product.team:
+                labels.append(f"team:{product.team}")
+        if flaw.is_major_incident:
+            labels.append("major_incident")
+        if flaw.cve_id in flaw.title:
+            summary = flaw.title
+        else:
+            summary = f"{flaw.cve_id} {flaw.title}"
```
osoukup, line 107: CVE ID may be blank at the first stages of the flaw life cycle. Will this work with such flaw?

osoukup (approved):
LGTM

Elkasitu (approved):

apps/taskman/service.py
costaconrado: Double checked and the answer is no, it will add an extra space in summary, nice catch, will fix.
(`"" in "some string"` is always true)

PR #226, opened 2023-05-31
Add additional validations for Major Incident flaw

jobselko:
This PR adds mitigation and statement validations for Major Incident flaw.

Closes OSIDB-656
Closes OSIDB-657

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce flaw ownership through task management system (OSIDB-69)
 - Implement task rejection in Taskman (OSIDB-74)
 - Implement article validation for Major Incident flaw (OSIDB-655)
+- Implement mitigation validation for Major Incident flaw (OSIDB-656)
+- Implement statement validation for Major Incident flaw (OSIDB-657)
 
 ### Changed
 - Integrate Jira tracker collector with collector framework (OSIDB-576)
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce flaw ownership through task management system (OSIDB-69)
 - Implement task rejection in Taskman (OSIDB-74)
 - Implement article validation for Major Incident flaw (OSIDB-655)
+- Implement mitigation validation for Major Incident flaw (OSIDB-656)
+- Implement statement validation for Major Incident flaw (OSIDB-657)
+
```
osoukup, line 14: nitpick: extra blank line

osidb/models.py
```diff
@@ -725,6 +725,24 @@ def _validate_major_incident_article(self):
                 "A flaw marked as Major Incident does not have an article."
             )
 
+    def _validate_major_incident_mitigation(self):
+        """
+        Tests that a Flaw that is Major Incident has a mitigation.
+        """
+        if self.is_major_incident and self.mitigation == "":
```
osoukup, line 732: I think that these alerts (both) should be rather non-blocking. I see that it is not clearly specified in the task and it has no SFM2 counterpart to compare. There is a similar alert checking that a major incident has summary (doc text) which is also non-blocking so I would consider this on the same level of restriction.

osoukup (approved):
LGTM but I would make the alerts non-blocking.

osidb/models.py
jobselko: I updated all article, mitigation and statement validations to use alerts.

Currently, `validate_major_incident_summary` uses exceptions instead of alerts. I will update this last one once the MI model is reworked (doing it now would not make sense).

PR #229, opened 2023-06-02
BugzillaSaver generalization and two tracker validations

osoukup:
This PR closes my day and it brings two pieces on the way to the Bugzilla trackers. The validations are actually applicable to the Jira trackers too.


costaconrado (approved):
LGTM.

PR #232, opened 2023-06-08
Implement CORS rules in OSIM for specific domain access

sfxworks:
This pull request introduces changes to your Django settings to enable Cross-Origin Resource Sharing (CORS) for a specific domain. This change allows the UI being built at [OSIM](https://github.com/RedHatProductSecurity/osim) to make API requests against the OSIDB project.

The changes involve:

Adding 'corsheaders' to the INSTALLED_APPS section in our settings. This is to ensure Django recognizes this as a valid app and loads it correctly.

Inserting 'corsheaders.middleware.CorsMiddleware' into our MIDDLEWARE list. This middleware handles the addition of the appropriate CORS headers to responses from your server. To ensure all responses are processed, this middleware has been placed as high as possible in the middleware stack, specifically before Django's CommonMiddleware.

Specifying our allowed origins for CORS in a new CORS_ALLOWED_ORIGINS list. This list includes the domain for the OSIM UI.

This change is crucial to our ongoing development efforts, as it paves the way for the OSIM UI to interact effectively with the OSIDB project. This interaction is key to providing users with a seamless and integrated experience across both platforms.

Please review the changes and provide your feedback.


config/settings.py
```diff
@@ -256,3 +258,7 @@
 DEFAULT_REQUEST_TIMEOUT = get_env(
     "OSIDB_DEFAULT_REQUEST_TIMEOUT", default="30", is_int=True
 )
+
+CORS_ALLOWED_ORIGINS = [
+  "https://ui.osim-ui--runtime-int.apps.int.spoke.preprod.us-east-1.aws.paas.redhat.com"
```
costaconrado, line 263: This expose an internal url, I'd like to recommend using get_env("WEBUI_URL") or any other name, include it in your `.env` file and list it on `docs/developer/DEVELOP.md`. I can add it on the ops project and deploy it on stage environment.

costaconrado (dismissed):

osoukup (approved):
I agree with the point risen by Conrado but otherwise LGTM

config/settings.py
costaconrado: To be honest I'm not sure what else we would like to plug on OSIDB, but I believe one variable that represents the frontend and change with the environment would be enough, but I'm ok if we decide to go for a more robust solution that allows multiple URLs.

sfxworks: We will have, in the very near future, both a staging and prod url that correlates with the staging and prod MP+ clusters. I can make an adjustment to settings.py to read that env as an array.

Elkasitu: I would also highly suggest setting up at least the stage CNAME and going with that instead of the canonical URL, but in general we can make those changes on-the-fly without requiring a new OSIDB release if we use the JSON array as env var (see https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/helpers.py#L24)

Elkasitu (changes requested):
A couple of remarks:

- The required django library that provides CORS header functionality is missing from the requirements files, so this will not work as-is, see https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#using-pip-tools for how to add it
- Please squash your commits and make sure to follow the usual git commit message guidelines (imperative mood of what the commit does)

Thanks!

config/settings.py
```diff
@@ -259,6 +261,4 @@
     "OSIDB_DEFAULT_REQUEST_TIMEOUT", default="30", is_int=True
 )
 
-CORS_ALLOWED_ORIGINS = [
-  "https://ui.osim-ui--runtime-int.apps.int.spoke.preprod.us-east-1.aws.paas.redhat.com"
-]
+CORS_ALLOWED_ORIGINS = os.getenv('WEBUI_URLS', '').split(',')
```
Elkasitu, line 264: `OSIDB_CORS_ALLOWED_ORIGINS` is a better fit I think

sfxworks:
> * The required django library that provides CORS header functionality is missing from the requirements files, so this will not work as-is, see https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#using-pip-tools for how to add it

I run code-server in a container, so I can't use podman nor your `make dev-env` script since it assumes dnf is used. Can you or someone else add this?

Elkasitu (approved):

costaconrado (approved):
LGTM.

Elkasitu:
Closed in favor of #247 as we can't accept fork-based PRs as of yet (the fault being GitHub's lack of handling of secrets for public repositories)

PR #235, opened 2023-06-13
Update tests for FlawReference API

jobselko:
This PR updates tests for FlawReference API to use `embargo_access`.

Follow-up to https://github.com/RedHatProductSecurity/osidb/pull/215.

Related to "implement adding of external links to flaws" (OSIDB-71).

### DIFF
--- a/osidb/tests/test_endpoints.py
+++ b/osidb/tests/test_endpoints.py
@@ -1541,7 +1541,7 @@ class TestEndpoints(object):
         response = auth_client.get(affect_url)
         assert response.status_code == 404
 
-    def test_flawreference_create(self, auth_client, test_api_uri):
+    def test_flawreference_create(self, auth_client, embargo_access, test_api_uri):
         """
         Test the creation of FlawReference records via a REST API POST request.
         """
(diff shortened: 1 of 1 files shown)


osidb/tests/test_endpoints.py
```diff
@@ -1581,7 +1581,7 @@ def test_flawreference_update(self, auth_client, test_api_uri):
         """
         Test the update of FlawReference records via a REST API PUT request.
         """
-        flaw = FlawFactory()
+        flaw = FlawFactory(embargoed=False)
```
osoukup, line 1584: The issue here is that when the flaw is embargoed the user needs to have an embargoed access to be able to update it so the issue is with the user and not with the data. I would be for fixing the user ACLs and leaving the test general. You can do it by simply adding `embargo_access` fixture to the test method parameter

https://github.com/RedHatProductSecurity/osidb/blob/master/conftest.py#L167

osidb/tests/test_endpoints.py
```diff
@@ -1607,7 +1607,7 @@ def test_flawreference_delete(self, auth_client, test_api_uri):
         """
         Test the deletion of FlawReference records via a REST API DELETE request.
         """
-        flaw = FlawFactory()
+        flaw = FlawFactory(embargoed=False)
```
osoukup, line 1610: and same here

osoukup (changes requested):
I would like to have this fixed in a different way. If you are strongly for your solution I am however opened to a discussion.

osoukup (approved):
LGTM

PR #239, opened 2023-06-16
Create CVSS model

tciernik:
Related to OSIDB-377


osidb/models.py
```diff
@@ -1487,6 +1489,139 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
 
 
+class CVSSManager(
+    ACLMixinManager,
+    TrackingMixinManager,
+):
+    """CVSS manager"""
+
+    @staticmethod
+    def create_cvss(flaw, affect, vector, issuer, version, **extra_fields):
```
osoukup, line 1499: Based on my comment to the vector bellow I think this should actually be rather maybe `create_or_update` (the `_cvss` part in the name is unnecessary as this is all CVSS) and by used to potentially update the vector.

osoukup: I am not sure what are the `extra_fields` but if this is eventually used for the update too we should account for them appropriately whatever it means.

osidb/models.py
```diff
@@ -1487,6 +1489,139 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+        Flaw, on_delete=models.CASCADE, blank=True, related_name="cvss_scores"
+    )
+
+    affect = models.ForeignKey(
+        Affect,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+        related_name="cvss_scores",
+    )
```
osoukup, line 1562: I would deal with the relations in the child classes. The reason is that a single CVSS record cannot relate to a flaw and an affect at the same time. It is always one or the other. I would simply create a child classes `FlawCVSS` and `AffectCVSS` which would account for the relations so those relations would be removed from the base `CVSS`.

osidb/models.py
```diff
@@ -1487,6 +1489,139 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+
+    vector = models.CharField(max_length=100, blank=False)
+
+    version = models.CharField(
+        choices=CVSSVersion.choices,
+        default=CVSSVersion.VERSION3,
```
osoukup, line 1568: I am not sure the default value should be here. True that currently the CVSSv3 is the one we use the most but I think that the version should be always explicitly specified and not defaulted.

osidb/models.py
```diff
@@ -2110,7 +2245,6 @@ def validate(self, *args, **kwargs):
 
 
 class CVEv5PackageVersions(PackageVersions):
-
```
osoukup, line 2113: All these black line removals are unrelated to the task so should be either done in an independent commit with an explanation why we remove them or not done at all.

osidb/tests/factories.py
```diff
@@ -8,6 +8,7 @@
 
 from osidb.constants import AFFECTEDNESS_VALID_RESOLUTIONS, DATETIME_FMT
 from osidb.models import (
+    CVSS,
```
osoukup, line 11: This is not alphabetically ordered.

osidb/tests/factories.py
```diff
@@ -509,3 +510,22 @@ class Meta:
...
     unacked_to_ps_module = factory.LazyAttribute(lambda o: choice([o.ps_module, None]))
+
+
+class CVSSFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = CVSS
+
+    flaw = factory.SubFactory(FlawFactory)
+    affect = factory.SubFactory(AffectFactory)
+    vector = "CVSS:3.0/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N"
```
osoukup, line 521: It would be nice to have here at least a few choices.

osoukup (changes requested):
This is very good and I think close to be done but I am rising a few suggestions and questions. I am opened to a different opinion however.

JakubFrejlach:
Closed in favour of https://github.com/RedHatProductSecurity/osidb/pull/271

PR #240, opened 2023-06-16
make impact comparable

osoukup:
As the tracker priority/severity depends on the highest impacts of all the related affects and flaws so we would like to be easily able to do things like `max` with it.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -112,7 +112,72 @@ class FlawType(models.TextChoices):
     WEAKNESS = "WEAKNESS"
 
 
-class FlawImpact(models.TextChoices):
+class ComparableTextChoices(models.TextChoices):
+    """
+    extension of the models.TextChoices classes
+    making them comparable with the standard operators
+
+    the comparison order is defined simply by the
+    top-down order in which the choices are written
+    """
+
+    @classmethod
+    def get_choices(cls):
+        """
+        get processed choices
+        """
+        return [choice[0] for choice in cls.choices]
+
+    @property
+    def weight(self):
+        """
+        weight of the instance for the comparison
+        defined by the order of the definition of the choices
+        """
+        return self.get_choices().index(str(self))
+
+    def incomparable_with(self, other):
+        """
+        to ensure that that we are comparing the instances of the same type as
+        comparing different types (even two ComparableTextChoices) is undefined
+        """
+        return type(self) != type(other)
+
+    def __hash__(self):
+        return super().__hash__()
+
+    def __eq__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self.weight == other.weight
+
+    def __ne__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self.weight != other.weight
+
+    def __lt__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self.weight < other.weight
+
+    def __gt__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self.weight > other.weight
+
+    def __le__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self == other or self.__lt__(other)
+
+    def __ge__(self, other):
+        if self.incomparable_with(other):
+            return NotImplemented
+        return self == other or self.__gt__(other)
+
+
+class Impact(ComparableTextChoices):
     """allowable impact"""
 
     NOVALUE = ""
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -342,7 +342,7 @@ class FlawHistory(NullStrFieldsMixin, ValidateMixin, ACLMixin):
     )
 
     # flaw severity, from srtnotes "impact"
-    impact = models.CharField(choices=FlawImpact.choices, max_length=20, blank=True)
+    impact = models.CharField(choices=Impact.choices, max_length=20, blank=True)
 
     # from BZ summary
     title = models.TextField()
@@ -531,7 +531,7 @@ class Flaw(
     )
 
     # flaw severity, from srtnotes "impact"
-    impact = models.CharField(choices=FlawImpact.choices, max_length=20, blank=True)
+    impact = models.CharField(choices=Impact.choices, max_length=20, blank=True)
 
     # flaw component was originally a part of the Bugzilla sumary
     # so the value may depend on how successfully it was parsed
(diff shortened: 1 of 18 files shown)


osoukup:
@costaconrado this is another small step towards the trackers

osidb/mixins.py
```diff
@@ -462,3 +462,37 @@ def _validate_acl_identical_to_parent_flaw(self):
...
+        return self.weight < other.weight
+
+    def __gt__(self, other):
+        return self.weight > other.weight
+
+    def __le__(self, other):
+        return self == other or self.__lt__(other)
+
+    def __ge__(self, other):
+        return self == other or self.__gt__(other)
```
Elkasitu, line 498: Please see and use `IntegerChoices` https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types

Elkasitu (changes requested):

osidb/mixins.py
osoukup: This reasonable. I will have a look Thank you

osoukup: @Elkasitu this does not seem to work the way we would like to. It seems to only be a named integer enumeration while we really need strings. The string conversions then result in strings like `1` instead of the actual name. Even the last example in the documentation with the `Answer` class works the same way.

osidb/mixins.py
```diff
@@ -462,3 +462,37 @@ def _validate_acl_identical_to_parent_flaw(self):
...
+    @property
+    def weight(self):
+        """
+        weight of the instance for the comparison
+        defined by the order of the definition of the choices
+        """
+        return self.get_choices().index(str(self))
+
+    def __lt__(self, other):
+        return self.weight < other.weight
```
Elkasitu, line 489: you need to safeguard against `other` being of a different type, two things can happen here:

1. the other object is of a different enumeration that does implement this mixin, in which case the comparison doesn't make sense
2. the other object does not implement this mixin, in which case this will raise an error because `weight` is not a member of `other`

Elkasitu: for those reasons and potentially others I would prefer going with something other than overriding the rich comparison operators, e.g. something like `FlawImpact.max(FlawImpact.CRITICAL, FlawImpact.LOW)` but I leave it up to you as I don't know what kind of comparisons will be performed regularly with these enumerations.

osidb/mixins.py
osoukup: But anyway thank you for pointing me to this. I completely forgot that it exists :angel:

osidb/mixins.py
osoukup: I agree with the risk of using some other `other` but it seemed to me as to that it would be rather a programmer bug. But no problem to add it. On the other hand I was also thinking about allowing mutual comparison of `FlawImact` and `AffectImpact` (not have a use case but seems meaningful to me) which would make it more complicated to safeguard (the exact class restriction would prevent comparison between these two).

@Elkasitu do you think it could be probably/frequent to compare `FlawImpact` and `AffectImpact` with other classes using the comparison operators so the safeguard would be really needed? To me it seems like not very probably case.

Having it implemented as `FlawImpact.max` would make it syntactically harder to use with larger arrays or generators. Would be doable but I am not fan of this approach.

costaconrado (approved):
LGTM.

osidb/mixins.py
Elkasitu: We shouldn't allow comparisons between different enums because that can lead to an apples vs oranges type of error -- in your example it works because both `FlawImpact` and `AffectImpact` have the exact same enum values, but this just means that we should instead have an `Impact` enum that we use in both `Flaw` and `Affect`.

Yes, comparing `ComparableTextChoices` to `other` is a programming error, but the error being raised is not a correct one: it should tell the programmer that he's comparing apples to oranges, not that `attribute x is not a member of y` as that is a confusing error.

See https://docs.python.org/3/library/functools.html#functools.total_ordering for an example of this pattern

```python
def _check_apples_oranges(self, other):
    # type() is necessary to ensure that we're checking that we're comparing
    # apples to apples and oranges to oranges, checking for instances
    # of ComparableTextChoices could lead to such an error
    return type(self) == type(other)

def __eq__(self, other):
    if self._check_apples_oranges(other):
        return self.weight == other.weight
    return NotImplemented
```

osoukup: @Elkasitu thank you for the suggestion! Merging the `FlawImpact` and the `AffectImpact` actually makes a lot of change to me so I have added a commit addressing this. I also extended the `ComparableTextChoices` so it ensures that the compared objects are of the same type. It increases the complexity of `ComparableTextChoices` but should hopefully now be more acceptable. Please have one more look.

osidb/models.py
```diff
@@ -112,7 +112,71 @@ class FlawType(models.TextChoices):
...
+
+    def __hash__(self):
+        return super().__hash__()
+
+    def __eq__(self, other):
+        # let us only prevent equal comparison to different ComparableTextChoices
+        # as Django internally uses it heavily (probably together with some conversions)
+        if isinstance(other, ComparableTextChoices):
+            self._check_type(other)
+        return super().__eq__(other)
```
Elkasitu, line 157: we don't need to override `__eq__()` as the default is identity comparison (e.g. `self is other`), this means that you also don't have to override `__hash__()`

osidb/models.py
```diff
@@ -112,7 +112,71 @@ class FlawType(models.TextChoices):
...
+
+    def _check_type(self, other):
+        """
+        ensure that that we are comparing the instances of the same type as
+        comparing different types (even two ComparableTextChoices) is undefined
+        """
+        if type(self) != type(other):
+            raise NotImplementedError(
+                f"Comparison of {type(self)} and {type(other)} is undefined"
+            )
```
Elkasitu, line 147: raising an error is a breach of contract for rich comparison methods, you need to return a boolean value or `NotImplemented` (which is different from `NotImplementedError`

osoukup: I consider this as a programmer error and reacts on your comment

> Yes, comparing ComparableTextChoices to other is a programming error, but the error being raised is not a correct one: it should tell the programmer that he's comparing apples to oranges, not that attribute x is not a member of y as that is a confusing error.

telling the programmer what exactly (s)he did wrong. Why not to be vocal about it?

osidb/models.py
osoukup: Without overriding this the result is simply `False` which might leave the programmer in a false conviction that (s)he is actually not making a programming error comparing apples and oranges but the objects are simply comparable but not the same. I think this contradicts what you suggested before which was to prevent comparing apples and oranges. So I would like to understand whether your are OK with comparing apples and oranges or not.

Elkasitu: Yes, maybe I contradicted myself and took the usage of the word error liberally, but I showed a clear example of how it's supposed to be done.

If we compare Enum A to Enum B the result of that comparison should always be `False` or `NotImplemented` because they're not comparable even if they might have similar values defined.

We could raise an error but since we don't control the base object (`TextChoices`) nor the systems that interface with it (Django, as you point out in your comment) then we should stick to the base API, which is to return `NotImplemented` if the objects are not comparable. This is done so that we don't inadvertently break parts of Django, and it's also the reason why I was originally against this approach, because we just don't know how it might affect the rest of the application, especially if you're not careful about the implementation.

osoukup: I am OK with replace `raise ...` with `return NotImplemented` however then in all cases including `__eq__` so it is consistent. I will adjust the implementation.

osoukup:
@Elkasitu I hope I incorporated your suggestions so please have one more look

Elkasitu (approved):
LGTM!

JakubFrejlach (approved):

PR #242, opened 2023-06-19
Allow lint and test workflows to run on PRs from forks

Elkasitu:
Previously this was not possible as these workflows were set to run only on repository push, which worked for PRs from branches of the repository as those are considered push events, but not on PRs from community forks as those are *not* considered push events.

By adding `pull_request` to the list of events that trigger these workflows PRs from forks should be covered.


jobselko (approved):

osoukup (approved):
Nice!

PR #243, opened 2023-06-19
prepare 3.2.2

osoukup:
to account for the triage workflow


Elkasitu (commented):
How will we fix the existing data?

osoukup:
> How will we fix the existing data?

By the resync. I did some analysis of the possible migration but it seems to complicated to me.

> In OSIDB we currently do not store the complete original Bugzilla flaw summary but parse it into parts and strip off the unnecessary bits. TRIAGE keyword is one of those stripped off parts. Therefore we are not easily able to tell which flaws are in triage. We also do not store the tracker summaries/titles at all but the unexpected TRIAGE keyword is considered to be part of the parsed ps_component so we can find it there in the meta_attr. Such unlinked trackers however do not contain any information about the flaw they should be linked to. We can theoretically go through all the flaws updated during the last few days and parse their meta_attr for depends_on and jira_trackers and try to match it with the trackers with TRIAGE in the meta_attr ps_component and eventually pair them. We should keep the Bugzilla and Jira timestamps so the integrated services would have to resync too. Sure that the resync of OSIDB may be costly but not when it is a few days so here I see the migration implementation more costly.

(from the GChat)

The good thing is that SDEngine will not need the followup resync as it does not care for the flaws without a CVE which is the case of the triage flaws. I will only notify the IR Dashboard team with the timestamp.

PR #245, opened 2023-06-20
Update Flaw to contain major_incident_state

jobselko:
This PR introduces the `major_incident_state` field in `Flaw`. A similar approach was already discussed in #21.

The intention is to use both `is_major_incident` (bool) and `major_incident_state` (enum string) in `Flaw` so backward compatibility is ensured (fields will be set in bzimport). In the future, `is_major_incident` should be converted into the property.


osidb/models.py
```diff
@@ -558,6 +588,12 @@ class FlawState(models.TextChoices):
     # should be set True if MAJOR_INCIDENT or MAJOR_INCIDENT_LITE FlawMeta exists, from BZ flagsq
     is_major_incident = models.BooleanField(default=False)
 
+    major_incident_state = models.CharField(
+        choices=FlawMajorIncident.choices,
+        default=FlawMajorIncident.MI_UNKNOWN,
+        max_length=100,
```
osoukup, line 594: nitpick: sounds like a too much :thinking: not that it would matter much

osidb/models.py
```diff
@@ -696,6 +732,24 @@ def _validate_cvss3(self):
                 "CVSSv3 score is missing",
             )
 
+    def _validate_major_incident_state(self):
+        """
+        Checks that a flaw that is Major Incident has a valid state.
+        Valid states are MI_APPROVED or MI_CISA_APPROVED.
```
osoukup, line 738: I do not understand this validation. It seems to me that it checks whether **bzimport** works correctly and maps the Bugzilla flags to the boolean the right way. Or when should this alert trigger?

osidb/models.py
```diff
@@ -696,6 +732,24 @@ def _validate_cvss3(self):
...
+            self.FlawMajorIncident.MI_REQUESTED,
+            self.FlawMajorIncident.MI_REJECTED,
+            self.FlawMajorIncident.MI_INVALID,
+        ]
+
+        if self.is_major_incident and self.major_incident_state in invalid_states:
+            self.alert(
+                "mi_state_invalid",
+                "A flaw marked as Major Incident does not have a valid state.",
+            )
```
osoukup, line 751: I think that the correct validation would be simply like

`if self.major_incident_state == self.FlawMajorIncident.MI_INVALID: ...alert`

osidb/models.py
osoukup: I would alert on `MI_INVALID` only. Everything else is valid. The boolean should not play a role in it. It should be soon a property based on the MI state anyway so no need to check it.

osidb/models.py
```diff
@@ -463,6 +463,36 @@ class FlawState(models.TextChoices):
...
+        If a state does not match any of BZ combinations, MI_INVALID is set.
+        """
+
+        MI_UNKNOWN = "MI_UNKNOWN"
+        MI_REQUESTED = "MI_REQUESTED"
+        MI_REJECTED = "MI_REJECTED"
+        MI_APPROVED = "MI_APPROVED"
+        MI_CISA_APPROVED = "MI_CISA_APPROVED"
+
+        MI_INVALID = "MI_INVALID"
```
Elkasitu, line 494: There's no need to namespace each value like this, remove the `MI_`

osidb/models.py
```diff
@@ -696,6 +732,16 @@ def _validate_cvss3(self):
...
 
+    def _validate_major_incident_state(self):
+        """
+        Checks that a flaw has a valid Major Incident state.
+        """
+        if self.major_incident_state == self.FlawMajorIncident.MI_INVALID:
+            self.alert(
+                "mi_state_invalid",
+                "A flaw does not have a valid Major Incident state.",
+            )
```
Elkasitu, line 743: I think having an `INVALID` value is a mistake:

* This means clients can set `major_incident_state` to `INVALID` by hand through the OSIDB API, this should not happen and all values should be valid
* We could set this to be a hard error, i.e. we raise a `ValidationError` which will be non-blocking at ingestion but this sends mixed signals to clients, the schema tells them `INVALID` is a valid value to send in an API call but the server will refuse to accept it via REST API.

Instead, we could simply have invalid states be the default (`UNKNOWN`), clients will complain when their `major_incident_state` is not correct either way, regardless of it being `UNKNOWN` or `INVALID`.

Elkasitu (changes requested):
A couple of open questions:

* Should we set this at ingestion (bzimport)?
* Should we make `is_major_incident` a property based on the value of `major_incident_state`?
* Should we mark this part of the FlawMeta API as deprecated?
* Should we set this value based on the FlawMeta values?

I feel like without an answer to those questions, these changes won't bring much value

osidb/models.py
osoukup: Good points. Client should definitely not be able to set an invalid state. Therefore if we go with it there should be a blocking validation checking that this value is not present (could be set by `bzimport` only).

With @jobselko we were discussing how to approach the situation when there is a combination of the Bugzilla flag values (it is two independent flags) which does not make sense/is not clear what is meant by it. We came with two possible solutions.

- I was first suggesting to give priority to some values (generally to the regular Major Incident above the CISA one) and simply map all the combinations to the valid values only. The downside is that when synced back we would basically fix the combination possibly not in the way the user intended to.
- Therefore Jitka came with the solution of the `INVALID` value which explicitly means that the original Bugzilla data must be fixed (either in Bugzilla or by setting a valid value in OSIDB) before we can do anything with the flaw.

True that this is really Bugzilla related thing and should be temporary (and probably very rare) and it would appear in the schema as a possible values (though not settable).

Your suggestion to use `UNKNOWN` instead is interesting but I see an issue there. The `UNKNOWN` means that there are no Major Incident flags present in Bugzilla (not sure if the name is great then). Mapping the invalid combination to it would result in erasing the flags on flaw update which would be unexpected from the user point of view and possibly harmful - even though it was set incorrectly the Major Incident is very serious matter so we should rather force the user to fix ti instead of erasing it.

osoukup:
> A couple of open questions:
> 
>     * Should we set this at ingestion (bzimport)?

I think we should and eventually just throw away the `FlawMeta` part and it would keep working.
 
>     * Should we make `is_major_incident` a property based on the value of `major_incident_state`?

It makes sense to me so I say yes.

>     * Should we mark this part of the FlawMeta API as deprecated?

I am not sure what is the deprecated functionality granularity and whether we can deprecate specific values of the the enum. I would probably just leave it as it is (basically duplicate value for some time) and when everything is stripped off of the `FlawMeta` just deprecate it as a whole.

>     * Should we set this value based on the FlawMeta values?

I say no as that would prevent us from removing `FlawMeta` in the future.

> I feel like without an answer to those questions, these changes won't bring much value

@Elkasitu check my answers and whether they make sense to you. I am opened to different opinions.

osidb/models.py
Elkasitu: We could have `bbsync` set up so that if the state is `UNKNOWN` we simply write back whatever we have in `meta_attr`.

But this is a suggestion, having a `ValidationError` is acceptable.

Elkasitu:
@osoukup I am satisfied by your answers, thanks! 👍🏽

osidb/models.py
jobselko: I would use `ValidationError` because we should differentiate between `UNKNOWN` and `INVALID` (if something is wrong in BZ, we just cannot rewrite it by a different value). `UNKNOWN` could be rewritten as `NOT_SET` to better describe the current state. I will also change `is_major_incident` field to property.

All functionality should be independent on `FlawMeta`, and I will not be touching `FlawMeta` for now.

@Elkasitu Do you agree with this proposal?

osoukup: Renaming `UNKNOWN` to `NOT_SET` makes a lot of sense. It actually means that no flag was raised so the status is known. Then I think it becomes self-explanatory that no other status should be mapped to it.

Elkasitu: I believe `blank=True` might be a better choice then? I guess that it depends on whether we consider a `Flaw` complete without a `major_incident_state` or not

Elkasitu: In fact we have a lot of cases already in which the field is `blank=True` and we assign the empty value to an enum value like so: `NOVALUE = ""`, `FlawImpact` is one such case

jobselko: @Elkasitu All changes except the `is_major_incident` change to the property are done. I tried to make it a property, but many tests failed, and fixing them would take me some time. So can I leave it as it is now and make this change later?

osidb/tests/test_flaw.py
```diff
@@ -917,6 +917,49 @@ def test_validate_major_incident_summary(
...
+        FlawReferenceFactory(
+            flaw=flaw,
+            type=FlawReference.FlawReferenceType.ARTICLE,
+            url="https://access.redhat.com/link123",
+        )
+        FlawMetaFactory(
+            flaw=flaw,
+            type=FlawMeta.FlawMetaType.REQUIRES_SUMMARY,
+            meta_attr={"status": "+"},
+        )
```
osoukup, line 954: Why do we define affect, reference, and meta here? It seems unnecessary with respect to the `_validate_major_incident_state` validation. Maybe the affect is necessary to make the flaw generally valid but why the rest?

osoukup: I take this back... I realized that Major Incident has extra requirements. Ignore.

osoukup (approved):
LGTM. One minor comment.

osidb/models.py
Elkasitu: Sure

osidb/models.py
```diff
@@ -696,6 +728,16 @@ def _validate_cvss3(self):
                 "CVSSv3 score is missing",
             )
 
+    def _validate_major_incident_state(self):
+        """
+        Checks that a flaw has a valid Major Incident state.
+        """
+        if (
+            self.major_incident_state
```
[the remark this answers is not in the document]
jobselko: So this is a power of `NOVALUE`, thanks :D

osidb/models.py
jobselko: Thank you. I will create a Jira for it.

Elkasitu (approved):

PR #248, opened 2023-06-21
Add FlawAcknowledgment model

jsvob:
Related to "Implement new model for FlawAcknowledgment" (OSIDB-1001)


osidb/models.py
```diff
@@ -1952,6 +1952,83 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+    def create_flawacknowledgment(
+        flaw, name, affiliation, from_upstream, **extra_fields
+    ):
+        """return a new flaw acknowledgment or update an existing flaw acknowledgment without saving"""
+        try:
+            flawacknowledgment = FlawAcknowledgment.objects.get(
+                flaw=flaw,
+                name=name,
+                affiliation=affiliation,
+                from_upstream=from_upstream,
```
osoukup, line 1968: I am thinking whether we should add `from_upstream` to the `get` query or rather just always update it the same way as the `extra_fields` two lines below.

I am thinking about a situation when there is already an acknowledgment `flawX`, `nameY`, `affiliationZ`, `from_upstream==False` and someone will attempt to create another one  `flawX`, `nameY`, `affiliationZ`, `from_upstream==True` (same but different `from_upstream`). Then `get` will fail with `ObjectDoesNotExist` and the `except` block will fail violating the unique constrain.

osoukup (commented):
Almost LGTM. I have risen one comment and I would like it either be addressed or clarified so that the current solution is OK.

osidb/models.py
```diff
@@ -1952,6 +1952,83 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+    # internal primary key
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    name = models.CharField(
+        max_length=255
+    )  # max length seen in production as of 02/2023 == 122
+    affiliation = models.CharField(
+        max_length=255
+    )  # max length seen in production == 86
+    from_upstream = models.BooleanField()
```
Elkasitu, line 1996: can we document what these two are supposed to mean?

Elkasitu (approved):
LGTM but see that @osoukup's remark gets addressed

osidb/models.py
osoukup: Not sure if @jsvob is aware :smile: I myself have only a vague understanding. Affiliation is a company or an organization which the reporter is reporting the vulnerability on behalf (probably makes me think that it should be optional as some people may report on their own). I am not sure what exactly means from upstream. I guess that the person is part of RH consumed upstream project. I would ask either @mprpic or someone from IR team like Marian.

osidb/models.py
jsvob: Agreed. This will allow for an easy switch/fix of the `from_upstream` bool. Alternatives might be more "correct", but much more complicated, comparatively. Fixed in an amended commit.

osidb/models.py
jsvob: I thought there's no hidden complexity but I don't know any more details :D The [relevant part of the schema](https://github.com/RedHatProductSecurity/osidb/blob/fc93883d9e7d1a531b15eeb4e1f019dd2bd7f843/apps/bbsync/srtnotes-schema.json#L113) seemed very straightforward:

```
    "acknowledgment": {
      "type": "object",
      "description": "A single acknowledgment",
      "required": ["name", "affiliation", "from_upstream"],
      "properties": {
        "name": {
          "type": "string",
          "description": "The name of the person or entity being acknowledged."
        },
        "affiliation": {
          "type": ["string", "null"],
          "description": "The affiliation of the person being acknowledged."
        },
        "from_upstream": {
          "type": "boolean",
          "description": "Whether this acknowledgment comes an upstream source."
        }
      }
```

As for _upstream_, that's a term we've been using frequently and I also don't see any hidden complexity:

> [(...) upstream in open source is the source repository and project where contributions happen and releases are made (...)](https://www.redhat.com/en/blog/what-open-source-upstream)

So it all seems that the `from_upstream` bool labels `affiliation`. And I guess that if an upstream of project A discovers a flaw in otherwise-unrelated project B, then the given `affiliation` string would be labeled with a `False` `from_upstream` in that acknowledgement, whereas in an unrelated flaw in project A discovered by someone from the same project, there might be another acknowledgement with the same `affiliation` string but a `True` `from_upstream`. Right?

Schema explanation added as comments in an amended commit.

osidb/models.py
```diff
@@ -1952,6 +1952,82 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+        except ObjectDoesNotExist:
+            return FlawAcknowledgment(
+                flaw=flaw,
+                name=name,
+                affiliation=affiliation,
+                **extra_fields,
+            )
+
+
+class FlawAcknowledgment(ACLMixin, BugzillaSyncMixin, TrackingMixin):
```
jobselko, line 1980: Could you please describe in docstring how flaw acklowledgments (with public/private flaw source) work?

osidb/models.py
```diff
@@ -1952,6 +1952,82 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+    # non operational meta data
+    meta_attr = HStoreField(default=dict)
+
+    objects = FlawAcknowledgmentManager()
+
+    class Meta:
+        """define meta"""
+
+        # A flaw can't have multiple identical acknowledgments.
+        # Can't have identical acknowledgments simultaneously from upstream and not from upstream.
```
jobselko, line 2011: `not upstream` is probably downstream :)

osidb/tests/test_flaw.py
```diff
@@ -73,7 +75,7 @@ def test_create(self, datetime_with_tz, good_cve_id):
             title="title",
             description="description",
             impact=FlawImpact.CRITICAL,
-            source=FlawSource.INTERNET,
```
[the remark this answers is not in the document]
jsvob: Because INTERNET is a public source and shouldn't have acknowledgments. If it has, an otherwise-unrelated validation fails. Hence the change to a private source.

jobselko:
FYI, if there is no change in changelog, please add the `technical` label.

osidb/models.py
jsvob: I don't understand. Can you show me an example how that should look, please?

osidb/models.py
jsvob: What if it's from an otherwise-unrelated entity? :thinking:

jsvob:
> FYI, if there is no change in changelog, please add the `technical` label.

There is a change in changelog. Shall I remove the line from the changelog and add the label?

osidb/tests/test_flaw.py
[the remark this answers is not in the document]
jobselko: Then it should be in a different commit with some short description :)

osidb/models.py
jobselko: Then the line 2011 will still be confusing. I guess that `unique_together` is self-explanatory enough and you can remove that line.

osidb/tests/test_flaw.py
```diff
@@ -1598,6 +1616,13 @@ def test_validate_article_links_count_via_flaw(self):
         with pytest.raises(ValidationError, match=error_msg):
             flaw.save()
 
+    def test_validate_public_source_no_ack(self, both_source):
```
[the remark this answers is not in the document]
jsvob: Added.

osidb/models.py
jsvob: Added explanation in docstring.

osidb/models.py
jsvob: OK, removed the comment.

jsvob:
I don't understand the migrations check. Runs fine on my local machine.

```
(venv) [jsvoboda@jsvoboda-laptop osidb]$ ./manage.py makemigrations --check --dry-run --settings=config.settings_ci && echo ok
/home/jsvoboda/work-prodsec/osidb/venv/lib64/python3.9/site-packages/django/core/management/commands/makemigrations.py:105: RuntimeWarning: Got an error checking a consistent migration history performed for database connection 'default': could not translate host name "postgres" to address: Name or service not known

  warnings.warn(
No changes detected
ok
(venv) [jsvoboda@jsvoboda-laptop osidb]$ /usr/bin/bash -c ./scripts/migrations-check.sh && echo ok
>Checking local dev venv existence
>Checking for active venv
ok
(venv) [jsvoboda@jsvoboda-laptop osidb]$ 

```

osidb/models.py
mprpic: `from_upstream` means that the original acknowledgment came from an upstream project. It allows us to create statements such as: "Red Hat would like to thank Bob for reporting this issue to us; the upstream project XYZ acknowledges Alice as the original reporter". This would translate to two acks:

- Name: Bob
- Name: Alice; from_upstream: yes

I think it would be best to align to an existing standard for acknowledgment data, either:

https://github.com/CVEProject/cve-schema/blob/20a9e977d9020c12d7dce07eb7ef8de30bd61f64/schema/v5.0/CVE_JSON_5.0_schema.json#L971
OR
https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#3115-acknowledgments-type---example

We already track names and organizations, and the rest of the data can either be expressed as a defined list of types (as in the CVE v5 schema) or a free-form text summary with an optional URL. That would cover many more relationship types than just the "upstreamness" of an acknowledgment.

jsvob: Thank you, @mprpic !

- It seems neither of the existing standards allow for data that would produce "the upstream project XYZ acknowledges Alice as the original reporter". Right?
- In that case, whatever we go with, we'll still need the `from_upstream` bool, or to have an additional free-form field that captures who expresses the acknowledgment, right?
- Which standard shall we choose? @osoukup @Elkasitu
- Should that happen now? It seems to me it would have to be stored elsewhere than in the existing srtnotes["acknowledgments"].

mprpic: > * It seems neither of the existing standards allow for data that would produce "the upstream project XYZ acknowledges Alice as the original reporter". Right?
> 

Why not? If we chose CVEv5 types, the original reporter (upstream one) would be tagged with `finder` and whoever notified us as `reporter`. If we used the summary field from the CSAF spec, it could be represented there in textual form and left to the analyst (or if we want to handhold a bit more, using static values).

osidb/models.py
```diff
@@ -1952,6 +1952,83 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+            )
+
+
+class FlawAcknowledgment(ACLMixin, BugzillaSyncMixin, TrackingMixin):
+    """Model representing flaw acknowledgments"""
+
+    # internal primary key
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    name = models.CharField(
```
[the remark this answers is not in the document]
jobselko: Ok, it is up to you to decide :)

docs/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement adding new flaw comments (OSIDB-81)
 - Erratum advisory name to flaw filter (OSIDB-922)
 - CORS allow-list functionality (OSIDB-967, OSIDB-965)
+- Implement a new FlawAcknowledgment model (OSIDB-1001)
```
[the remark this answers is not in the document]
jsvob: Aha! Thanks, I'll fix it based on your instruction.

osidb/models.py
osoukup: >     * Which standard shall we choose? @osoukup @Elkasitu

Nothing new was requested. We have no requirements from the IR team other than do what SFM2 does so I would stick with it and just implement the counterpart in OSIDB. What is described above sounds interesting and might be the future but I would consider it out of scope of this task and also out of scope of SFM2 decommission work.
 
>     * Should that happen now? It seems to me it would have to be stored elsewhere than in the existing srtnotes["acknowledgments"].

Let us stay backwards compatible for now so SFM2 and OSIDB understand each other.

jobselko:
@jsvob CI is failing because there is a collision in migrations.

osidb/models.py
```diff
@@ -1952,6 +1952,83 @@ def bzsync(self, *args, bz_api_key=None, **kwargs):
...
+                name=name,
+                affiliation=affiliation,
+                **extra_fields,
+            )
+
+
+class FlawAcknowledgment(ACLMixin, BugzillaSyncMixin, TrackingMixin):
+    """
+    Model representing flaw acknowledgments.
+    Note that flaws with a public source can't have acknowledgments.
```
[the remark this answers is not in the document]
jsvob: Doesn't seem so based on my limited understanding. I was able to create a flaw without an acknowledgement (using the `master` version of osidb) and it didn't raise an alert.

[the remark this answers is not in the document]
jobselko: Ok, so you can leave it as it is. Now it is more apparent that only a flaw with a private source can have an acknowledgment.

PR #250, opened 2023-06-23
Include flaw acknowledgments when saving data via bzimport

jsvob:
Followup to https://github.com/RedHatProductSecurity/osidb/pull/248

Related to "Implement BZImport part for FlawAcknowledgment" (OSIDB-1003). Closes OSIDB-1003


jobselko:
You can write `Closes OSIDB-X` directly to trigger changes in the relevant Jira.

I used `Related to ...` only in cases where more PRs are connected to one Jira, so the Jira changes were not triggered.

jsvob:
I'm testing various configurations on my local machine to figure out what's wrong. Doing other work in between because the new env setup and tests take long. I'll be done before Tuesday.

collectors/bzimport/convertors.py
```diff
@@ -301,6 +305,15 @@ def clean_meta(self):
         for meta in to_delete:
             meta.delete()
 
+    def clean_acknowledgments(self):
+        """clean obsoleted flaw acknowledgments"""
+        old_acknowledgments = set(self.flaw.acknowledgments.all())
+        new_acknowledgments = set(self.acknowledgments)
+
+        to_delete = list(old_acknowledgments - new_acknowledgments)
```
osoukup, line 313: I am not sure this would catch the case when an acknowledgment changes the `from_upstream` boolean. Not saying that it will not be rewritten later just not sure so rather check it.

osoukup (approved):
LGTM

JakubFrejlach (approved):
One suggestion, otherwise LGTM.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -361,6 +393,107 @@ def test_meta_removed(self):
         assert flaw.meta.first() == self.get_meta(flaw)[1]
         assert flaw.meta.first().meta_attr["name"] == "Lone Wanderer"
 
+    def test_acknowledgment_removed(self):
```
jobselko, line 396: Could you please use `@pytest.mark.parametrize` for these tests?

collectors/bzimport/tests/test_convertors.py
```diff
@@ -124,6 +125,23 @@ def get_meta(self, flaw):
             ),
         ]
 
+    def get_acknowledgments(self, flaw):
+        """
+        minimal acknowledgments getter
+        """
+        return [
+            FlawAcknowledgment(
```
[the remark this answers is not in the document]
jsvob: The fix for addressing Ondrej's feedback required me to start using `create_flawacknowledgment()`, so I moved away from the solution you commented on here that just created new FlawAcknowledgment instances (which could have been created by `FlawAcknowledgment`).

Btw, the factory exists already.

collectors/bzimport/tests/test_convertors.py
jsvob: I'm afraid not.

Would the fixture scope be clear then? I'm afraid setting the fixture scope so that they survive the individual parametrized calls would in the end be more complicated than what is there now.

collectors/bzimport/convertors.py
jsvob: Right, that has not been tested and I enhanced `test_acknowledgment_removed` to test it.

collectors/bzimport/tests/test_convertors.py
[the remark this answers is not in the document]
JakubFrejlach: Ah, sorry, missed that.

collectors/bzimport/tests/test_convertors.py
jobselko: You do not need to deal with a scope. Create a flaw,  perform an operation on it and then check the result.

jsvob: Then the `*_orig` variables would overreach the scope of each parametrized test. A workaround for that would be more complex than the current test. And it would change the style of the test from the rest. Feel free to submit a follow-up MR.

PR #252, opened 2023-06-23
create a foundation of Bugzilla tracker filing functionality based on BBSync

osoukup:
This is a very foundation of Bugzilla tracker filing functionality. A lot of BBSync stuff is reused. There is a number of TODOs and the code itself is not integrated into OSIDB yet (not being called anywhere). The goal was to be able to create a Bugzilla tracker using OSIDB shell which was successful

https://bugzilla.stage.redhat.com/show_bug.cgi?id=2014466

and also checked in SFM2 that the tracker is valid enough (according to the current level of the implementation). There is still a lot of work but that is tracked in the followup Jiras.

Closes OSIDB-92

### DIFF
--- a/apps/bbsync/save.py
+++ b/apps/bbsync/save.py
@@ -9,7 +9,7 @@ from osidb.exceptions import DataInconsistencyException
 from osidb.models import Flaw
 
 from .exceptions import UnsaveableFlawError
-from .query import BugzillaQueryBuilder
+from .query import FlawBugzillaQueryBuilder
 
 
 class BugzillaSaver(BugzillaQuerier):
@@ -22,6 +22,14 @@ class BugzillaSaver(BugzillaQuerier):
     def model(self):
         """
         instance model class getter
+        needs to be defined in the subclasses
+        """
+        raise NotImplementedError
+
+    @property
+    def query_builder(self):
+        """
+        query builder class getter
         needs to be defined in the subclasses
         """
         raise NotImplementedError
@@ -48,7 +56,7 @@ class BugzillaSaver(BugzillaQuerier):
         """
         create a bug underlying the model instance in Bugilla
         """
-        bugzilla_query_builder = BugzillaQueryBuilder(self.instance)
+        bugzilla_query_builder = self.query_builder(self.instance)
         response = self.bz_conn.createbug(bugzilla_query_builder.query)
         self.instance.bz_id = response.id
         return self.instance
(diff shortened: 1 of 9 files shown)


apps/bbsync/save.py
```diff
@@ -129,10 +137,17 @@ def flaw(self):
     @property
     def model(self):
         """
-        Flaw model class getter
+        instance model class getter
```
JakubFrejlach, line 140: Nitpick: since this is already a concrete BugzillaSaver for Flaws, wording `Flaw model class getter` might stay.

osidb/models.py
```diff
@@ -1794,6 +1794,34 @@ def fix_state(self):
...
+
+        unacked_ps_update_stream = (
+            PsUpdateStream.objects.filter(name=self.ps_update_stream)
+            .get()
+            .ps_module.unacked_ps_update_stream.first()
+        )
+        if unacked_ps_update_stream is None:
+            return False
+
+        return self.ps_update_stream == unacked_ps_update_stream.name
```
JakubFrejlach, line 1823: I wonder whether this could be simplified to something like this:

```python
unacked_ps_update_stream = PsUpdateStream.objects.filter(name=self.ps_update_stream, unacked_to_ps_module__isnull=True).first()

if unacked_ps_update_stream is None:
    return False
return True

```

osidb/tests/test_tracker.py
```diff
@@ -56,6 +57,26 @@ def test_aggregeted_impact(
...
+        unacked_ps_update_stream = PsUpdateStreamFactory(
+            ps_module=ps_module,
+            unacked_to_ps_module=ps_module,
+        )
+
+        acked_tracker = TrackerFactory(ps_update_stream=acked_ps_update_stream.name)
+        unacked_tracker = TrackerFactory(ps_update_stream=unacked_ps_update_stream.name)
+
+        assert acked_tracker.is_acked and not acked_tracker.is_unacked
+        assert not unacked_tracker.is_acked and unacked_tracker.is_unacked
```
JakubFrejlach, line 78: Could this be simplified? From the logic of `is_acked` and `is_unacked` properties it kinda feels like testing `True and not False` statement :grinning:

JakubFrejlach (approved):
Couple of suggestions otherwise LGTM, but I did not dig much into the last commit since there are a lot of TODOs as you've mentioned.

apps/bbsync/save.py
osoukup: Good point, I will rename back

osidb/models.py
osoukup: Great hint! Actually it should be `isnull=False` :smile: but otherwise with this the method can be shortened to a single expression.

osidb/tests/test_tracker.py
osoukup: I know that the model methods are defined in the way that only one contains the logic and the other only reverts the first one but I think that the test should not be aware of it and really check whether both give the correct results.

osidb/models.py
JakubFrejlach: Ah, I always confuse this expression, `isnull=True` somehow translates in my head to just `True` :grin:

PR #253, opened 2023-06-26
Use alerts while validating that MI has summary

jobselko:
Closes OSIDB-951

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -809,11 +809,15 @@ class Flaw(
             return
 
         if not self.summary:
-            raise ValidationError("Flaw marked as Major Incident does not have Summary")
+            self.alert(
+                "mi_summary_missing",
+                "Flaw marked as Major Incident does not have Summary",
+            )
 
         if not req or req.meta_attr.get("status") == "?":
-            raise ValidationError(
-                "Flaw marked as Major Incident does not have Summary reviewed"
+            self.alert(
+                "mi_summary_not_reviewed",
+                "Flaw marked as Major Incident does not have Summary reviewed",
             )
 
         # XXX: In SFM2 we check that the REQUIRES_DOC_TEXT flag is set by
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1005,6 +1005,11 @@ class Flaw(
             if self.meta.filter(type=FlawMeta.FlawMetaType.ACKNOWLEDGMENT).exists():
                 return
 
+            # Acknowledgments are stored in both FlawMeta and FlawAcknowledgment.
+            # FlawMeta will be deprecated in the future, so this check should be kept.
+            if self.acknowledgments.count() > 0:
+                return
+
             if source.is_public():
                 alert_text = (
                     f"Flaw source of type {source} can be public or private, "
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -923,28 +923,28 @@ class TestFlawValidators:
         assert should_alert == ("cvss3_missing" in FlawFactory(cvss3=cvss3)._alerts)
 
     @pytest.mark.parametrize(
-        "summary,is_major_incident,req,should_raise",
+        "summary,is_major_incident,req,should_alert,alerts",
         [
-            ("", True, None, True),
-            ("", True, "+", True),
-            ("", True, "?", True),
-            ("", True, "-", False),
-            ("", False, None, False),
-            ("", False, "+", False),
-            ("", False, "?", False),
-            ("", False, "-", False),
-            ("foo", False, None, False),
-            ("foo", False, "+", False),
-            ("foo", False, "?", False),
-            ("foo", False, "-", False),
-            ("foo", True, None, True),
-            ("foo", True, "+", False),
-            ("foo", True, "?", True),
-            ("foo", True, "-", False),
+            ("", True, None, True, ["mi_summary_missing", "mi_summary_not_reviewed"]),
+            ("", True, "+", True, ["mi_summary_missing"]),
+            ("", True, "?", True, ["mi_summary_missing", "mi_summary_not_reviewed"]),
+            ("", True, "-", False, []),
+            ("", False, None, False, []),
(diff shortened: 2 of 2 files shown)


osidb/tests/test_flaw.py
```diff
@@ -923,55 +923,60 @@ def test_validate_cvss3_model(self, cvss3, should_alert):
...
-        self, summary, is_major_incident, req, should_raise
+        self, summary, is_major_incident, req, should_alert, alerts
     ):
         """
         Test that a Flaw that is Major Incident has a summary
         """
         flaw1 = FlawFactory.build(
             summary=summary,
             is_major_incident=is_major_incident,
+            embargoed=False,
```
osoukup, line 955: I wonder why is this important. If the reason is that the test was giving you an error that the user is missing some LDAP group it should rather be solved by adding `embargo_access` fixture.

osoukup (approved):
LGTM

Elkasitu (commented):
Can we document why this change is being made? It's not mentioned in the PR description, commit message or task description.

osidb/tests/test_flaw.py
jobselko: The `Embargoed` field in `FlawFactory` determines the `Source` field. If `Embargoed` is `True`, `Source` will be private (this is an original case). 

This source is then checked in `_validate_private_source_no_ack`. However, this checks an acknowledgement only in `FlawMeta`, not in `FlawAcknowledgement`. If I create a flaw with a private source and with an acknowledgement via `FlawAcknowledgement` (as `FlawMeta` will be deprecated), this test fails.
The problem is that we forgot to update `_validate_private_source_no_ack` to check also `FlawAcknowledgement.`

I see 3 possible solutions how to fix this problematic test:
1. Use `Embargoed=False`, so this problem does not happen (I did this one)
2. Use `FlawMetaFactory` of the `ACKNOWLEDGMENT` type so `_validate_private_source_no_ack` passes
3. Use `FlawAcknowledgement` and update `_validate_private_source_no_ack` to check also `FlawAcknowledgement` so `_validate_private_source_no_ack` passes.

@osoukup @Elkasitu Which solution do you prefer? I prefer option 3.

jobselko: I am sorry I did not point this out from the start.

osoukup: Thank you for the explanation! Then any solution sounds OK. Probably 1 or 2 as 3 sounds out of scope of this PR.

jobselko: I updated the test as described in option 3. Please check if it is ok.

osoukup: It is all OK. Thank you

jobselko:
@Elkasitu Should I more comment on the changes in the second commit in the PR description?

Elkasitu:
> @Elkasitu Should I more comment on the changes in the second commit in the PR description?

No, I want to know why we are changing these `ValidationError`s into alerts

osidb/models.py
```diff
@@ -1001,6 +1005,11 @@ def _validate_private_source_no_ack(self):
             if self.meta.filter(type=FlawMeta.FlawMetaType.ACKNOWLEDGMENT).exists():
                 return
 
+            # Acknowledgments are stored in both FlawMeta and FlawAcknowledgment.
+            # FlawMeta will be deprecated in the future, so this check should be kept.
+            if self.acknowledgments.count() > 0:
+                return
```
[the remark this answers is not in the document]
jobselko: So I can extend the test by adding also `FlawMeta` of the `ACKNOWLEDGMENT` type, so the test will have both `FlawMeta` and `FlawAcknowledgment.`

jobselko:
> > @Elkasitu Should I more comment on the changes in the second commit in the PR description?
> 
> No, I want to know why we are changing these `ValidationError`s into alerts

We discussed it with @osoukup a few weeks ago, and If I am right, we want to have it as non-blocking alerts when saving data.

osoukup:
> > > @Elkasitu Should I more comment on the changes in the second commit in the PR description?
> > 
> > 
> > No, I want to know why we are changing these `ValidationError`s into alerts
> 
> We discussed it with @osoukup a few weeks ago, and If I am right, we want to have it as non-blocking alerts when saving data.
> 
> EDIT: If somebody sets a flaw flag incorrectly, we do not want to immediately block any work on the flaw.

Let explain to prevent more confusion :smile: the word **incorrectly** is not correct. The motivation behind this change is in the first place to let people continue working on the flaw even when the MI flag is raised while the flaw was not fully populated with the data (eg. summary is not yet there). The MI is often being resolved under time pressure and blocking changes should be sensitive. Ideally we would have a state-full workflow and blocking would be done in some later state only but we do not have it (yet).

Elkasitu:
> > > > @Elkasitu Should I more comment on the changes in the second commit in the PR description?
> > > 
> > > 
> > > No, I want to know why we are changing these `ValidationError`s into alerts
> > 
> > 
> > We discussed it with @osoukup a few weeks ago, and If I am right, we want to have it as non-blocking alerts when saving data.
> > EDIT: If somebody sets a flaw flag incorrectly, we do not want to immediately block any work on the flaw.
> 
> Let explain to prevent more confusion smile the word **incorrectly** is not correct. The motivation behind this change is in the first place to let people continue working on the flaw even when the MI flag is raised while the flaw was not fully populated with the data (eg. summary is not yet there). The MI is often being resolved under time pressure and blocking changes should be sensitive. Ideally we would have a state-full workflow and blocking would be done in some later state only but we do not have it (yet).

We can merge this, however since I'm working on `FlawDraft` at the moment, I think that we can, in the future, use `FlawDraft` as the basis for this instead of making these alert-raising validations hard validations.

The goal of `FlawDraft` will be to match the same fields as `Flaw` very closely but without any or very minimal validations (e.g. field-level validations like making sure the CVE id format is correct), when the analyst is done with all the preliminary work it can then be promoted to an actual flaw (this will trigger all Flaw and related validations).

So what I'm saying is that we may come back to this and revert into a hard validation after that's done.

PR #255, opened 2023-06-27
resurrect OSIM

osoukup:
TODO


osoukup:
This at the end served rather as an example than a regular contribution. The functionality is now already implemented as part of a different pull request so I closing this one.

PR #256, opened 2023-06-27
FlawReference BBSync and API fixes

jobselko:
This PR fixes a few `FlawReference` issues:
- References from OSIDB are saved into Bugzzilla in incorrect format. This causes references to be missing in SFM2.
- `FlawReferenceView` does not propagate the deletion to Bugzilla. This causes references to be deleted in OSIDB but not in Bugzilla.
- `FlawReferenceView` does not return references for a specified flaw only. This causes all references to be returned.
- `FlawReferenceView` does not contain a flaw uuid in a serializer only.
- `FlawReferenceSerializer` does not inherit from `IncludeMetaAttrMixin.`

Fixes OSIDB-1035

### DIFF
--- a/apps/bbsync/srtnotes.py
+++ b/apps/bbsync/srtnotes.py
@@ -176,12 +176,17 @@ class SRTNotesBuilder:
     def generate_references(self):
         """
         generate array of references to SRT notes
+
+        OSIDB uses "ARTICLE" and "EXTERNAL",
+        but Bugzilla uses "vuln_response" and "external".
         """
+        references_mapping = {"ARTICLE": "vuln_response", "EXTERNAL": "external"}
+
         self.add_conditionally(
             "references",
             [
                 {
-                    "type": reference.type,
+                    "type": references_mapping[reference.type],
                     "url": reference.url,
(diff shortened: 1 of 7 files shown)


apps/bbsync/srtnotes.py
```diff
@@ -176,12 +176,17 @@ def generate_jira_trackers(self):
...
+        but Bugzilla uses "vuln_response" and "external".
         """
         self.add_conditionally(
             "references",
             [
                 {
-                    "type": reference.type,
+                    "type": "vuln_response"
+                    if reference.type == "ARTICLE"
+                    else "external",
```
osoukup, line 189: nitpick: I would rather see there a map to make it more future proof but so far with the two values it is not really necessary. You can ignore this comment.

osoukup (approved):
LGTM. Thank you for fixing it so quickly!

osoukup (approved):
LGTM

jobselko:
@JakubFrejlach @osoukup This should be the final version. Could you please check that everything looks ok?

JakubFrejlach (approved):

PR #257, opened 2023-06-27
Add scheduled closing for published tasks

costaconrado:
This PR introduces changes in task status update.

After 14 days without changes, a published task should close itself automatically.


apps/taskman/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+
+class ScheduledClosingTask(models.Model):
+    """This instance represents when should a Jira task should be closed if there is no changes"""
```
osoukup, line 5: nitpick: double `should`
nitpick: `is` -> `are`

apps/taskman/service.py
```diff
@@ -176,6 +186,9 @@ def update_task_status(
                         status=409,
                     )
 
+            if status == TaskStatus.REFINEMENT:
```
osoukup, line 189: If I understand the workflow correctly we should auto-close in either Refinement (Published) or In Progress (Review) - I mean in either of these two, not just Refinement.

osoukup (commented):
I think that the status condition should be expanded but otherwise LGTM

apps/taskman/service.py
costaconrado: That makes sense, I will submit a fix for that.

apps/taskman/models.py
costaconrado: Thank you, submitting a fix for that.

osoukup (approved):
LGTM

apps/taskman/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    today = timezone.now()
+    jira_tasks = ScheduledClosingTask.objects.filter(scheduled_dt__lt=today)
+
+    done_tasks = []
+
+    jira = JiraTaskmanQuerier(JIRA_SERVICE_TOKEN)
+    for jira_task in jira_tasks:
+        res = jira.update_task_status(
```
Elkasitu, line 27: What happens if it's already set to Close? Is it a no-op? Should it be?

Furthermore I think that calling this function from here is dangerous since `update_task_status` can queue a new `close_unchanged_published_tasks`. Can't you simply call `jira_conn.transition_issue` instead?

Elkasitu (changes requested):
I don't agree with the approach, however the main reason for blocking the PR is that there are still a lot of unknowns and untested paths, once those are corrected we can go ahead with whichever approach (pseudo-queue vs JQL)

JakubFrejlach:
As https://github.com/RedHatProductSecurity/osidb/pull/351 brings a new GH action check which checks for schema with osidb-bindings compatibility, make sure you rebase this PR on top of the master branch so mentioned GH action may be triggered if this PR makes changes to `openapi.yml`.

PR #258, opened 2023-06-28
Include raw summary from BZ in Flaw.meta_attr

Elkasitu:
The Flaw.title attribute is based off of the `summary` attribute from Bugzilla, however before setting Flaw.title the summary goes through various transformations which might not be perfect and the resulting title might be incorrect.

With this commit we include the raw BZ version of the title (summary in BZ) so that clients have the full context should they need it.

Closes OSIDB-1016

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -613,6 +613,7 @@ class FlawConvertor(BugzillaGroupsConvertorMixin):
         meta_attr["alias"] = self.alias
         meta_attr["depends_on"] = self.depends_on
         meta_attr["related_cves"] = [c for c in self.cve_ids if c != cve_id]
+        meta_attr["bz_summary"] = self.flaw_bug["summary"]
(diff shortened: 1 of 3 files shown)


collectors/bzimport/convertors.py
```diff
@@ -613,6 +613,7 @@ def get_meta_attr(self, cve_id):
         meta_attr["alias"] = self.alias
         meta_attr["depends_on"] = self.depends_on
         meta_attr["related_cves"] = [c for c in self.cve_ids if c != cve_id]
+        meta_attr["summary"] = self.flaw_bug["summary"]
```
osoukup, line 616: I would call this eg. `bz_summary` (or somehow differently) because the term summary in OSIDB actually means Bugzilla Doc Text. WDYT?

osoukup (approved):
I have one suggestion but generally LGTM

collectors/bzimport/convertors.py
Elkasitu: Makes sense, done!

PR #261, opened 2023-06-28
Convert hightouch and hightouch-lite into major_incident_state in BZImport

jobselko:

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -549,6 +549,39 @@ class FlawConvertor(BugzillaGroupsConvertorMixin):
         return [flag for flag in self.flaw_bug["flags"] if isinstance(flag, dict)]
 
     @cached_property
+    def major_incident_state(self):
+        """
+        A Major Incident state created from hightouch and hightouch-lite flags.
+        """
+        # Default values
+        hightouch = ""
+        hightouch_lite = ""
+
+        # Sets values from Bugzilla
+        for flag in self.flags:
+            if flag["name"] == "hightouch":
+                hightouch = flag["status"]
+            if flag["name"] == "hightouch-lite":
+                hightouch_lite = flag["status"]
+
+        valid_pairs = {
+            ("", ""): Flaw.FlawMajorIncident.NOVALUE,
+            ("?", "?"): Flaw.FlawMajorIncident.REQUESTED,
+            ("?", ""): Flaw.FlawMajorIncident.REQUESTED,
+            ("", "?"): Flaw.FlawMajorIncident.REQUESTED,
+            ("-", "-"): Flaw.FlawMajorIncident.REJECTED,
+            ("-", ""): Flaw.FlawMajorIncident.REJECTED,
+            ("", "-"): Flaw.FlawMajorIncident.REJECTED,
+            ("+", ""): Flaw.FlawMajorIncident.APPROVED,
+            ("+", "-"): Flaw.FlawMajorIncident.APPROVED,
+            ("", "+"): Flaw.FlawMajorIncident.CISA_APPROVED,
+            ("-", "+"): Flaw.FlawMajorIncident.CISA_APPROVED,
+        }
+        flags_from_bz = (hightouch, hightouch_lite)
+
+        return valid_pairs.get(flags_from_bz, Flaw.FlawMajorIncident.INVALID)
+
+    @cached_property
     def package_versions(self):
         """parse fixed_in to package versions"""
         fixed_in = self.flaw_bug["fixed_in"]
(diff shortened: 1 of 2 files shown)


collectors/bzimport/tests/test_convertors.py
```diff
@@ -562,6 +564,58 @@ def get_flaw_srtnotes(cls):
...
+        ],
+    )
+    def test_flags_major_incident(self, hightouch, hightouch_lite, result):
+        """
+        Tests that hightouch and hightouch-lite flags from Bugzilla are correctly
+        converted into major_incident_state in OSIDB.
+        """
+        flaw_bug = self.get_flaw_bug()
+        flaw_bug["flags"] = [
+            {"name": "hightouch", "status": hightouch},
```
jobselko, line 597: I am unsure if it is OK to write flags this way because Bugzilla returns flags in the following format:

`"flags": [{"name": "hightouch", "status": "+", "id": 100, "type_id": 5, "is_active": 1, "setter": "user@mail.com", "creation_date": "2020-09-30T08:00:00Z", "modification_date": "2020-09-30T09:00:00Z"}]`

collectors/bzimport/convertors.py
```diff
@@ -548,6 +548,45 @@ def flags(self):
...
+            [("?", ""), Flaw.FlawMajorIncident.REQUESTED],
+            [("", "?"), Flaw.FlawMajorIncident.REQUESTED],
+            [("-", "-"), Flaw.FlawMajorIncident.REJECTED],
+            [("-", ""), Flaw.FlawMajorIncident.REJECTED],
+            [("", "-"), Flaw.FlawMajorIncident.REJECTED],
+            [("+", ""), Flaw.FlawMajorIncident.APPROVED],
+            [("+", "-"), Flaw.FlawMajorIncident.APPROVED],
+            [("", "+"), Flaw.FlawMajorIncident.CISA_APPROVED],
+            [("-", "+"), Flaw.FlawMajorIncident.CISA_APPROVED],
+        ]
```
JakubFrejlach, line 580: I think `dict` mapping would be a more suitable in this case:

```python
valid_pais = {
    ("", "") : Flaw.FlawMajorIncident.NOVALUE,
    ...
}
```

Then you can simply check it for expected value and avoid iterating through the whole thing.

collectors/bzimport/convertors.py
```diff
@@ -548,6 +548,45 @@ def flags(self):
 
         return [flag for flag in self.flaw_bug["flags"] if isinstance(flag, dict)]
 
+    @staticmethod
```
jobselko, line 551: I do not like using `@staticmethod` but I did not want to change how the `flags` property works, and I was not able to come up with a better solution. So I would appreciate any recommendation.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -43,6 +43,7 @@ def get_flaw(self):
             title="title",
             description="description",
             impact=Impact.CRITICAL,
+            major_incident_state="REQUESTED",
```
JakubFrejlach, line 46: Use enums (eg. `Flaw.FlawMajorIncident.REQUESTED`) instead of string if possible.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -209,6 +210,7 @@ def test_basic(self):
         assert flaw.meta.first() == meta
         assert flaw.references.first() == reference
         assert flaw.package_versions.first() == package_versions
+        assert flaw.major_incident_state == "REQUESTED"
```
JakubFrejlach, line 213: Same as above.

collectors/bzimport/tests/test_convertors.py
JakubFrejlach: I would say that is OK, you are only interested in the `name` and `status` in this case, rest of the data is irrelevant and it doesn't matter whether it is present. 

On the other hand, are both `hightouch` and `hightouch_lite` always present in the `flags`? Not sure right now. If not, it might be a good thing to also supply some test scenario with one of those missing.

collectors/bzimport/convertors.py
JakubFrejlach: Since `flags` is attribute of the `FlawCollector`, why not simply just turn this into the `@property` and instead of passing the flags as an argument, access them through the `self`.  In fact, this can be a cached_property, since it is pointless to run the parsing logic again when calling the property for the second time.

JakubFrejlach (changes requested):
Couple of suggestions. Good job.

collectors/bzimport/convertors.py
jobselko: Thanks! This is what I was looking for. I was not able to realize that I could access a property from another property :)

collectors/bzimport/tests/test_convertors.py
jobselko: You are right. I will update it.

JakubFrejlach (approved):
LGTM now! Just squash it into 1 commit.

PR #262, opened 2023-06-29
Change Flaw endpoints to trigger taskman

costaconrado:
This PR changes Flaw endpoints in order to trigger a Task creation in Jira when a Flaw is created or updated.

Closes OSIDB-861.


osidb/api_views.py
```diff
@@ -166,7 +168,13 @@ def get_valid_http_methods(cls: ViewSet, excluded: list[str] = None) -> list[str
         "or the CVE number corresponding to a Flaw"
     ),
 )
-
+jira_token_param = OpenApiParameter(
+    name="JiraAuthentication",
```
osoukup, line 172: I am a bit late but I just noticed that for Bugzilla we use naming `Bugzilla-Api-Key` for the header and it would be nice to unify it. `Jira-Api-Key` would also make more sense to me, it is more clear what it is supposed to be.

osoukup (changes requested):
I am do not think that this generally a good approach. We would like also flaws created internally - eg. by Assembler collectors to have the underlying tasks created. Calling the API from within the app seems a bad approach to me. I think that we should rather integrate the task creation into a flaw `save` and eg. just turn it explicitly off when syncing the flaw from Bugzilla (so the old flaws do not get Jira tasks).

osidb/api_views.py
JakubFrejlach: I think neither of those are suitable names. In Bugzilla it is actually called API key, however in Jira it is called Personal Access Tokens and thus I would be for `Jira-Access-Token`.

osidb/api_views.py
```diff
@@ -166,7 +168,13 @@ def get_valid_http_methods(cls: ViewSet, excluded: list[str] = None) -> list[str
         "or the CVE number corresponding to a Flaw"
     ),
 )
-
+jira_token_param = OpenApiParameter(
```
JakubFrejlach, line 171: This is a great idea, I think we should also add this `OpenApiParameter` for `Bugzilla-Api-Key` as currently it is kinda obfuscated what actually is needed for the operations that syncs to Bugzilla.

JakubFrejlach:
> I am do not think that this generally a good approach. We would like also flaws created internally - eg. by Assembler collectors to have the underlying tasks created. Calling the API from within the app seems a bad approach to me. I think that we should rather integrate the task creation into a flaw `save` and eg. just turn it explicitly off when syncing the flaw from Bugzilla (so the old flaws do not get Jira tasks).

I agree that the calling API from within the app feels kinda cumbersome. +1 to integrating it into the flaw `save`.

osidb/api_views.py
costaconrado: I think either option would be good, one follows the current pattern and other uses jira naming, I have submitted as `Jira-Api-Key` but if we prefer I can change to `Jira-Access-Token`.

osidb/api_views.py
costaconrado: I have changed and submitted in this PR in a separated commit, please take a look if it will work.

apps/taskman/mixins.py
```diff
@@ -0,0 +1,35 @@
...
+        abstract = True
+
+    def save(self, *args, jira_token=None, **kwargs):
+        """
+        save the model by storing to Jira and then persisting in database
+
+        Jira sync is conditional based on environment variable so
+        development environment not enforces it
+        """
+        if JIRA_TASKMAN_AUTO_SYNC_FLAW:
```
osoukup, line 25: I am thinking about the condition here. It seems to me that even in a single environment with one value of `JIRA_TASKMAN_AUTO_SYNC_FLAW` we need to have two approaches here.

The reason is that we should not be creating the tasks for the historical flaws.

Probably for now we should not even create the tasks for flaws created originally in SFM2/Bugzilla but only for those created by OSIDB. If this makes sense to you I would suggest a similar behavior as it is in BBSync mixin in a way that you would additionally check the `jira_token` presence. The collectors would not provide it (so no task creation) and API queries which require the token would provide it. This would be a distinguisher of whether to create a task or not.

docs/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Erratum advisory name to flaw filter (OSIDB-922)
 - CORS allow-list functionality (OSIDB-967, OSIDB-965)
 
+### Changed
+- Flaw creation and update triggers taskman (OSIDB-861)
```
osoukup, line 15: I am not sure the users would know what `taskman` is :smile: so I would rather replace it with eg. `Jira task sync`

apps/taskman/mixins.py
osoukup: Actually the third option is going to be a flaw draft created by the integrated Assembler. I think that this would behave similarly to the API call in the sense of providing the Jira token but of course the service one would be used as it is all automation.

osidb/models.py
```diff
@@ -1125,6 +1144,27 @@ def bzsync(self, *args, bz_api_key, **kwargs):
         fc = FlawCollector()
         fc.sync_flaw(self.bz_id)
 
+    def tasksync(self, *args, jira_token, **kwargs):
+        """
+        Task sync of the Flaw instance in Jira
+        """
+        print("#########%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%")
```
osoukup, line 1151: Leftover debug message? :smile: (plus another below)

osoukup (approved):
I am suggesting a slightly different mixin behavior and a few minor changes but generally this LGTM and I agree with this approach

apps/taskman/tests/test_flaw_model_integration.py
```diff
@@ -0,0 +1,132 @@
...
+        sync_count = 0
+
+        def mock_create_or_update_task(self, flaw, fail_if_exists=False):
+            nonlocal sync_count
+            sync_count += 1
+
+        monkeypatch.setattr(
+            JiraTaskmanQuerier, "create_or_update_task", mock_create_or_update_task
+        )
+        monkeypatch.setenv("JIRA_TASKMAN_AUTO_SYNC_FLAW", 0)
```
Elkasitu, line 35: this test should work if the variable is simply unset, so no need to monkeypatch

apps/taskman/tests/test_flaw_model_integration.py
```diff
@@ -0,0 +1,132 @@
...
+            cve_id="CVE-2020-8001",
+            title="CVE-2020-8001 kernel: some description",
+            type=FlawType.VULNERABILITY,
+            acl_read=acl_read,
+            acl_write=acl_write,
+            description="Description",
+            component="component",
+            impact=Impact.LOW,
+            source=FlawSource.INTERNET,
+            unembargo_dt=make_aware(datetime.now()),
```
Elkasitu, line 55: you can use `django.utils.timezone.now` instead

osidb/models.py
```diff
@@ -472,6 +474,22 @@ def __str__(self):
...
 
+    def create(self, jira_token=None, **obj_data):
+        """
+        override create method in order to sync flaw created with Taskman
+        """
+        # imports here to prevent cycles
+        from apps.taskman.service import JiraTaskmanQuerier
+
+        # only creates in database if task were created in Jira
+        with transaction.atomic():
```
Elkasitu, line 485: all transactions are atomic in OSIDB

osidb/models.py
```diff
@@ -472,6 +474,22 @@ def __str__(self):
...
+            flaw = super().create(**obj_data)
+            if not JIRA_TASKMAN_AUTO_SYNC_FLAW or jira_token is None:
+                return flaw
+
+            JiraTaskmanQuerier(token=jira_token).create_or_update_task(flaw)
+            return flaw
```
Elkasitu, line 491: There's a couple of flaws with this approach:

* First, this doesn't cover all the ways that one can create a Flaw, for the sake of simplicity we'll assume that nobody will ever create a Flaw through pure SQL, but they can still do so through `create_flaw` or by using `Flaw`'s constructor and bypass this check.
* This shouldn't prevent the creation of flaws at all, we create flaws outside of the request-response lifecycle all the time (e.g. collectors), we should only block the creation through the API

osidb/models.py
```diff
@@ -1138,6 +1157,33 @@ def bzsync(self, *args, bz_api_key, **kwargs):
         fc = FlawCollector()
         fc.sync_flaw(self.bz_id)
 
+    def tasksync(self, *args, jira_token, **kwargs):
```
Elkasitu, line 1160: This method signature is weird, why is `*args` before `jira_token`? There's also no unit test for this method so I don't know what a call to it looks like or should look like.

osidb/models.py
```diff
@@ -1138,6 +1157,33 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+        Task sync of the Flaw instance in Jira
+
+        This function is responsible for synchronizing only existing flaws
+        that were authored in OSIDB which means they have a task already
+        created in Jira.
+        """
+        # imports here to prevent cycles
+        from apps.taskman.service import JiraTaskmanQuerier
+
+        # new instances are sync-ed in FlawManager.create
```
Elkasitu, line 1171: I actually think this method should handle new instances too, then you decouple the jira token check from the part that handles the actual creation/update of tasks, and it allows us to easily trigger task creation/update manually. Imagine that for x reason a JIRA task was not created during flaw creation, we'd have to recreate the flaw to create the JIRA task... (in practice we'd probably call `create_or_update_task` manually, but you get my point).

osidb/serializer.py
```diff
@@ -504,6 +505,41 @@ class Meta:
         abstract = True
 
 
+class JiraTaskSyncMixinSerializer(serializers.ModelSerializer):
```
Elkasitu, line 508: You don't seem to be using this mixin anywhere? If you did on the `FlawSerializer` you wouldn't need the check on `Flaw.create`, also this doesn't take into account the env variable you defined earlier

osidb/tests/conftest.py
```diff
@@ -136,3 +136,14 @@ def private_source():
...
+
+@pytest.fixture(autouse=True)
+def mock_taskman(monkeypatch):
+    def mock_create_or_update_task(*args, **kwargs):
+        pass
+
+    monkeypatch.setattr(
+        "apps.taskman.service.JiraTaskmanQuerier.create_or_update_task",
+        mock_create_or_update_task,
+    )
```
Elkasitu, line 149: This is dangerous and antithetical to testing, if someone makes a change in the osidb module that would somehow break the calls to `create_or_update_task` we might not see it (depends on a lot of factors).

osidb/models.py
costaconrado: I have opted to follow `bzsync` signature, but I agree, I will change `tasksync`.

osidb/serializer.py
```diff
@@ -505,6 +507,54 @@ class Meta:
...
+        """
+        perform the ordinary instance create
+        with providing Jira token while saving
+        """
+        flaw = super().create(validated_data)
+        if JIRA_TASKMAN_AUTO_SYNC_FLAW:
+            flaw.tasksync(
+                jira_token=self.__get_user_jira_token(), creation_required=True
+            )
+        return flaw
```
Elkasitu, line 534: Furthermore you should instead do like the `BugzillaSyncMixinSerializer` which creates an instance object and then saves it passing the jira token as an argument to `save()`, since the `save()` override will check for the `JIRA_TASKMAN_AUTO_SYNC_FLAW` feature flag.

Elkasitu (approved):

PR #263, opened 2023-06-30
Add major_incident_state to Flaw API and update filters

jobselko:

### DIFF
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -796,6 +796,7 @@ class FlawSerializer(
                 "cvss3_score",
                 "nvd_cvss3",
                 "is_major_incident",
+                "major_incident_state",
                 "affects",
                 "meta",
                 "comments",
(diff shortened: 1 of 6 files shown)


osidb/serializer.py
```diff
@@ -796,6 +796,7 @@ class Meta:
                 "cvss3_score",
                 "nvd_cvss3",
                 "is_major_incident",
+                "major_incident_state",
```
jobselko, line 799: I am still unsure if this is all that needs to be done.

osoukup: Can you perform an API query filtering flaws with a specific `major_incident_state` value?

jobselko: Yes. If I do `http://localhost:8000/osidb/api/v1/flaws?major_incident_state=REJECTED`, I get `count: 2`.

osoukup (approved):
LGTM

PR #265, opened 2023-06-30
Add new FlawAcknowledgment API

jsvob:
Follow-up to https://github.com/RedHatProductSecurity/osidb/pull/250.

Closes OSIDB-1002 ("Implement new API for FlawAcknowledgment")


openapi.yml
```diff
@@ -5619,6 +5904,84 @@ components:
...
+            it is used to detect mit-air collisions.
+      required:
+      - affiliation
+      - created_dt
+      - embargoed
+      - flaw
+      - from_upstream
+      - name
+      - updated_dt
+      - uuid
```
osoukup, line 6154: I may be wrong so take this just as something to check but is it correct to have `flaw` and `uuid` as required attributes when they are parts of the path? Just seems to be duplicite but I might miss something.

osoukup: Probably good to check with the bindings - FYI @JakubFrejlach

osoukup (approved):
LGTM

openapi.yml
JakubFrejlach: Agree with @osoukup, both PUT and POST should not require flaw ID inside the request body since it is already provided in the path.

osidb/api_views.py
```diff
@@ -345,6 +354,20 @@ def create(self, request, *args, **kwargs):
...
 
+@include_meta_attr_extend_schema_view
+@include_exclude_fields_extend_schema_view
+@extend_schema_view(
+    create=extend_schema(
+        request=FlawAcknowledgmentPostSerializer,
+    ),
+)
+class FlawAcknowledgmentView(ModelViewSet):
+    queryset = FlawAcknowledgment.objects.all()
```
JakubFrejlach, line 371: This queryset won't take into account the `flaw_id` path parameter and will include all acknowledgements. Similar handling as for Flaw Comments or Flaw References should be done.

osidb/api_views.py
```diff
@@ -345,6 +354,20 @@ def create(self, request, *args, **kwargs):
...
 
 
+@include_meta_attr_extend_schema_view
+@include_exclude_fields_extend_schema_view
+@extend_schema_view(
+    create=extend_schema(
+        request=FlawAcknowledgmentPostSerializer,
+    ),
+)
+class FlawAcknowledgmentView(ModelViewSet):
```
JakubFrejlach, line 370: This ViewSet does not have the destroy method overridden and thus the delete will be perform only on the OSIDB side and not on the BZ side. Similar handling as for Flaw Affects or Flaw References should be done.

Also I wonder whether it does make sense to create some common mixin which would already provide the overridden delete.

osidb/tests/test_endpoints.py
```diff
@@ -1591,6 +1617,89 @@ def test_affect_delete(self, auth_client, test_api_uri):
...
+        Test the creation of FlawAcknowledgment records via a REST API POST request.
+        """
+        # Source must be private in order for validation to pass.
+        flaw = FlawFactory(source=FlawSource.CUSTOMER)
+
+        flawacknowledgment_data = {
+            "name": "John Doe",
+            "affiliation": "Acme Corp.",
+            "from_upstream": False,
+            "flaw": str(flaw.uuid),
```
JakubFrejlach, line 1631: Should be already included in the path.

osidb/api_views.py
JakubFrejlach, line 370: Destroy method should also return code 200 instead of 204 because we inject some default info about OSIDB like:

```
  "dt": "2023-07-03T07:22:10.661Z",
  "env": "string",
  "revision": "string",
  "version": "string"
  ```

and 204 response by standard won't include any response body.

**Review by JakubFrejlach — DISMISSED** [review]
Couple of things which needs to be addressed.

File: osidb/serializer.py [hunk 5]
```diff
@@ -623,6 +624,32 @@ def to_representation(self, value):
         return list(trackers)
 
 
+class FlawAcknowledgmentSerializer(
```
jobselko, line 627: `FlawAcknowledgmentSerializer` should inherit also from `IncludeMetaAttrMixin`.

osidb/serializer.py
```diff
@@ -623,6 +624,32 @@ def to_representation(self, value):
...
+
+        model = FlawAcknowledgment
+        fields = (
+            ["name", "affiliation", "from_upstream", "flaw", "uuid"]
+            + ACLMixinSerializer.Meta.fields
+            + TrackingMixinSerializer.Meta.fields
+        )
+
+
+@extend_schema_serializer(exclude_fields=["updated_dt"])
```
jobselko, line 646: `flaw` should be also excluded

osidb/filters.py
```diff
@@ -194,6 +201,19 @@ class Meta:
             + LTE_GTE_LOOKUP_EXPRS
             + DATE_LOOKUP_EXPRS,
             "affects__trackers__errata__advisory_name": ["exact"],
+            # Acknowledgment fields
+            "acknowledgments__uuid": ["exact"],
```
jobselko, line 205: I am unsure whether we need all these filter fields (e.g. `"acknowledgments__created_dt": ["exact"]` does not make sense to me). @JakubFrejlach Could you please look at it?

jobselko (changes requested):

jobselko:
This PR should have an entry in CHANGELOG because it contains changes visible to customers.

osidb/filters.py
JakubFrejlach: Well, it is unlikely that someone will need to filter out exact datetime but no strong opinion to keep that or remove that.

jsvob: After some thought, the timestamps are unique to the given osidb instance and represent the time the osidb instance learned of the given acknowledgment, not the time when it was actually added into bugzilla. I'll try to reasonably trim the amount of filters in a separate commit. Thanks!

openapi.yml
jsvob: Hopefully addressed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4

The swagger page for POST and PUT (when generated **without** commit 4f068286685d90f3de2ea222b5a84348059af6d8 ) has the right `Example Value`s.

POST:

```
{
  "name": "string",
  "affiliation": "string",
  "from_upstream": true,
  "embargoed": true
}
```

and PUT:

```
{
  "name": "string",
  "affiliation": "string",
  "from_upstream": true,
  "embargoed": true,
  "updated_dt": "2023-07-04T18:04:44.202Z"
}
```

`flaw_id` and `id` are still marked as required, but also marked as belonging to `path` as opposed to `query`.

Does that do what you need @JakubFrejlach please? Thanks!

osidb/api_views.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4

osidb/api_views.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4 except for the new mixin. I'd prefer to do the mixin in a separate MR once we get an approval on this MR because of potential complexity, because it mostly depends on whether we agree on PR265, and because of the iterative approach we discussed two weeks ago.

osidb/tests/test_endpoints.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4 (hopefully, see discussion thread https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1250297359 )

osidb/api_views.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4

osidb/serializer.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4

osidb/serializer.py
jsvob: Fixed in commit 3c2d2dd147380fc8924fac4b9d539f85f86f36d4 (hopefully, see discussion thread https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1250297359 ). Thanks!

osidb/filters.py
jsvob: Fixed in commit 627341b808e39ae2bd298779e83eabf3aab3ca9a

jsvob:
> This PR should have an entry in CHANGELOG because it contains changes visible to customers.

Fixed in commit 4f068286685d90f3de2ea222b5a84348059af6d8 but perhaps in a way you might not approve of. IDK of a better way though. Should be addressed in OSIDB-1004 IMO. Open to suggestions. Thanks!

osidb/filters.py
osoukup: > After some thought, the timestamps are unique to the given osidb instance and represent the time the osidb instance learned of the given acknowledgment, not the time when it was actually added into bugzilla. I'll try to reasonably trim the amount of filters in a separate commit. Thanks!

Not sure if I understand this correctly but the timestamps should be inherited from the flaw. Bugzilla is still the authoritative source of the data so OSIDB should just mirror it.

jsvob: Aha, thank you for the correction. `convertors.py` literally contains `created_dt=self.flaw_bug["creation_time"]` when setting `created_dt` of a `FlawAcknowledgment`.

In that light, it's IMO still right to remove `FlawAcknowledgment`'s filtering based on `created_dt`/`updated_dt` because `Flaw` can already be filtered based on `created_dt`/`updated_dt`.

osoukup: It is not necessary now but in the future it might be useful. I would leave it there. I am not sure if exact match is useful for the timestamps but generally I see the timestamps filtering handy.

jsvob: In that case, reverted commit 627341b808e39ae2bd298779e83eabf3aab3ca9a in commit d0354a6c6b5652baaee4919d2a446f020b056bc2 and the end result of this conversation is no change.

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+        override the default behavior to proxy the delete to Bugzilla
+        """
+        flaw = instance.flaw
+        instance.delete()
+        flaw.save(bz_api_key=bz_api_key)
+
+    def get_flaw(self):
+        """
+        Get flaw object instance.
+        Similar to FlawView.get_object().
```
jobselko, line 397: Please be more descriptive in docstrings (but still brief). I mean - Which flaw should be gotten? The one associated with a given flaw acknowledgment. If a docstring is too general, it has no value for people who read it.

You can also omit `Similar to FlawView.get_object()` as it is useless.

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+        """
+        pk = self.kwargs["flaw_id"]
+        if CVE_RE_STR.match(pk):
+            obj = get_object_or_404(Flaw, cve_id=pk)
+        else:
+            obj = get_object_or_404(Flaw, uuid=pk)
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def get_queryset(self):
```
jobselko, line 407: Please add a docstring with a description why we need this (e.g. similar to `FlawReferenceView`'s `get_queryset`.

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+        if getattr(self, "swagger_fake_view", False):
```
jobselko, line 408: Remove this condition because we use `drf-spectacular` and not `drf-yasg`.

jobselko (changes requested):

osidb/api_views.py
jobselko: Add "common mixin" - it was not said that we should do it in this PR. I think it was only a suggestion for the future :).

osidb/api_views.py
Elkasitu: This is actually a common problem to [both libraries](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-get-queryset-depends-on-some-attributes-not-available-at-schema-generation-time) and it looks like `drf-spectacular` might've been inspired by `drf-yasg` and I assume the reason for it being needed here is because of the nesting.

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+    ),
+    update=extend_schema(
+        request=FlawAcknowledgmentPutSerializer,
+    ),
+)
+class FlawAcknowledgmentView(ModelViewSet):
+    serializer_class = FlawAcknowledgmentSerializer
+    # As long as SRTNotesBuilder.generate_acknowledgments() builds from FlawMeta,
+    # POST, DELETE and PUT don't work.
+    # TODO: Resolve in a follow-up to PR 265.
```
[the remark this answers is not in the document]
jobselko: If BBSync is done later, leaving the tests commented out makes sense.

osidb/tests/test_endpoints.py
```diff
@@ -1591,14 +1616,113 @@ def test_affect_delete(self, auth_client, test_api_uri):
...
+
+        # Tests "POST" on flaws/{uuid}/acknowledgments
+        response = auth_client.post(
+            f"{test_api_uri}/flaws/{str(flaw.uuid)}/acknowledgments",
+            flawacknowledgment_data,
+            format="json",
+            HTTP_BUGZILLA_API_KEY="SECRET",
+        )
+
+        # TODO: Resolve in a follow-up to PR 265.
```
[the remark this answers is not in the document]
jobselko: If BBSync is done later, leaving the tests commented out makes sense.

osidb/api_views.py
jobselko: Thanks for the clarification. In that case, we should keep that condition with the link you mentioned to explain why it is important.

jsvob: Will fix the comment.

The schema can be generated without this condition, but it generates a warning.

osidb/api_views.py
jsvob: Fixed in commits 09866d24b5b72af9f19ea30c518961a04686bd1b and eaaa28d2c2b1ea423c782b4745faa28b97bf3e7d

osidb/api_views.py
jsvob: Fixed in commits 09866d24b5b72af9f19ea30c518961a04686bd1b and eaaa28d2c2b1ea423c782b4745faa28b97bf3e7d

osidb/api_views.py
jsvob: Fixed in commits 09866d24b5b72af9f19ea30c518961a04686bd1b and eaaa28d2c2b1ea423c782b4745faa28b97bf3e7d

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def get_queryset(self):
+        if getattr(self, "swagger_fake_view", False):
+            # https://drf-yasg.readthedocs.io/en/stable/openapi.html
+            # Required for autogeneration of parameters to openapi.yml.
+            return self.serializer_class.Meta.model.objects.none()
+        flaw = self.get_flaw()
+        return self.serializer_class.Meta.model.objects.filter(flaw=flaw)
```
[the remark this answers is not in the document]
jsvob: Fixed in commit 09866d24b5b72af9f19ea30c518961a04686bd1b

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+
+    def get_serializer(self, *args, **kwargs):
+        if "data" in kwargs:
+            # Once upon a time, request.data used to be read-only, it seems.
```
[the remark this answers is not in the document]
jsvob: Fixed in commits 09866d24b5b72af9f19ea30c518961a04686bd1b and eaaa28d2c2b1ea423c782b4745faa28b97bf3e7d

osidb/api_views.py
```diff
@@ -345,6 +354,75 @@ def create(self, request, *args, **kwargs):
...
+
+    def get_queryset(self):
+        if getattr(self, "swagger_fake_view", False):
+            # https://drf-yasg.readthedocs.io/en/stable/openapi.html
+            # Required for autogeneration of parameters to openapi.yml.
+            return self.serializer_class.Meta.model.objects.none()
+        flaw = self.get_flaw()
+        return self.serializer_class.Meta.model.objects.filter(flaw=flaw)
+
+    def get_serializer(self, *args, **kwargs):
```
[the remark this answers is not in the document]
jsvob: Fixed in commits 09866d24b5b72af9f19ea30c518961a04686bd1b and eaaa28d2c2b1ea423c782b4745faa28b97bf3e7d

osidb/serializer.py
```diff
@@ -623,6 +624,41 @@ def to_representation(self, value):
...
+
+@extend_schema_serializer(exclude_fields=["flaw"])
+class FlawAcknowledgmentPutSerializer(FlawAcknowledgmentSerializer):
+    # Extra serializer for PUT request because flaw shouldn't be
+    # required in the body (already included in the path).
+    pass
+
+
+@extend_schema_serializer(exclude_fields=["flaw", "updated_dt"])
+class FlawAcknowledgmentPostSerializer(FlawAcknowledgmentPutSerializer):
```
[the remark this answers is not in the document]
jsvob: Fixed in commit 09866d24b5b72af9f19ea30c518961a04686bd1b

osidb/tests/test_endpoints.py
```diff
@@ -1591,14 +1616,113 @@ def test_affect_delete(self, auth_client, test_api_uri):
...
+        # TODO: assert FlawAcknowledgment.objects.count() == 0
+
     def test_flawreference_create(self, auth_client, embargo_access, test_api_uri):
         """
         Test the creation of FlawReference records via a REST API POST request.
         """
         flaw = FlawFactory()
 
         flawreference_data = {
-            "flaw": str(flaw.uuid),
```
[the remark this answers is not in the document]
jsvob: Fixed in commits c1ecc5d3479860137d8571768172bb620efd4360 and ccc82af3c0f5074bca8229906d82beeb427dcffe

jsvob:
There are many commits now, so that I could reply to feedback in a way that's IMO clearer than amend and force push. I'll squash the commits before merge so that there are 3 commits:

- The main part of the work
- Fix of comments in FlawCommentView
- Fix of request body in test_flawreference_create

osidb/api_views.py
jobselko, line 1: Could you please add a new commit (probably the 5th when you are squashing your final work) to fix [this problem](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-get-queryset-depends-on-some-attributes-not-available-at-schema-generation-time) with `get_queryset()`  for `FlawReference`?

I checked that spectacular reported this:
`Warning #9: FlawReferenceView: Failed to obtain model through view's queryset due to raised exception. Prevent this either by setting "queryset = Model.objects.none()" on the view, checking for "getattr(self, "swagger_fake_view", False)" in get_queryset() or by simply using @extend_schema. (Exception: 'flaw_id')`

I overlooked it when I was working on FlawReference. Thank you!

jobselko (changes requested):

osidb/api_views.py
jsvob: Interactively rebased, commits pruned, and your suggestion added as an extra commit. Thanks!

jsvob:
@JakubFrejlach There are backward compatibility errors in CI Meta checks:

> Backward compatibility errors (5):
> error at head/openapi.yml, in API GET /osidb/api/v1/flaws/{flaw_id}/references for the 'path' request parameter 'flaw_id', the type/format was changed from 'string'/'none' to 'string'/'uuid' [request-parameter-type-changed]. 
> ...

This is caused by commit 931021ce7351ddcfc2a0a1c172d5f82a1609a58c. How to resolve it, please?

osidb/api_views.py
jsvob: @jobselko Reverting the fix you requested. This is blocking other work (OSIDB-1004) and it's out of scope. Please file a separate task if necessary.

osidb/api_views.py
jsvob: This is blocking other work (OSIDB-1004) and it seems the concerns have been addressed. If a common mixin is still needed, please open a separate Jira, or I'll try to implement it as a part of future similar work (OSIDB-1006 potentially).

openapi.yml
jsvob: This is blocking other work (OSIDB-1004) and it seems the concerns have been addressed.

jsvob:
FTR, final changes requested by @jobselko are out of scope. Other feedback has been addressed. Also see these discussion threads for more context why I'm merging this:

- https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1265830310
- https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1265830183

jsvob:
@jobselko @JakubFrejlach Your reviews are blocking the merge. Please approve or provide actionable feedback.

jobselko:
Unfortunately, Jakub is on PTO now, so we will have to wait until he is back.

Regarding the issue with `get_queryset()` for `FlawReference,` I will look at it in the meantime. It could be an easy fix considering that it works for `FlawAcknowledgment.`

jsvob:
@jobselko I see nothing in team calendar. OK then. I dismissed @JakubFrejlach 's feedback. Please approve from your side. Your request is out of scope. I'm blocked.

jobselko (approved):
I approve this PR, but I would prefer @Elkasitu to review and approve it too (as we agreed on meeting a few weeks ago).

IMHO, this PR is not so urgent and can wait until @JakubFrejlach is back. If it was a necessary bug fix, then I understand. But in this situation, I find it rude to kick somebody off from the review because he is on PTO.

osidb/tests/test_endpoints.py
```diff
@@ -1720,7 +1720,6 @@ def test_flawreference_create(self, auth_client, embargo_access, test_api_uri):
...
```
Elkasitu, line 1598: I'm not sure what this change accomplishes, perhaps the bit that removes the flaw as required param is in another commit?

Elkasitu (approved):
Should be squashed, not sure if that is your intention so just a reminder.

In the future, should we do the bbsync part before the API implementation? This would avoid implementing a half-baked API that has only some working methods instead of all? Not sure of the feasibility, just a suggestion

osidb/tests/test_endpoints.py
jsvob: This is in a separate commit with the message "Do not require flaw value in body, because it is already provided in URL".

Jakub [requested](https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1250405741) that this change be done for FlawAcknowledgments.

Jitka [requested](https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1263797140) other FlawReference fixes in this PR to mirror how FlawAcknowledgments are handled.

Jitka [requested](https://github.com/RedHatProductSecurity/osidb/pull/265#discussion_r1262218968) that this change is in a separate commit.

jsvob:
> Should be squashed, not sure if that is your intention so just a reminder.

@Elkasitu My intention was **not** to squash these 4 commits. One of the commits will be reverted in a future PR (enabling write actions on the api). Two other commits are for changes in other code so that the approach is updated to match the new FlawAcknowledgments (comment updates mostly). @jobselko also requested multiple times that such changes should be in separate commits. Should the commits have different messages or is there any other concern?

> In the future, should we do the bbsync part before the API implementation? This would avoid implementing a half-baked API that has only some working methods instead of all? Not sure of the feasibility, just a suggestion

Yes, I agree and will do that next time.

The current approach is for 3 reasons

- I wasn't aware at this point that the data migration should have already been in place together with bbsync at this point. Will do better next time.
- This PR was a time sink for folks for long enough. I thought it would be better to get it merged and out of sight than reordering the other work.
- My original intention was not to have this as a user-facing issue and I wanted to just merge the API, then bbsync. Jitka didn't want the `technical` label and to add a CHANGELOG entry. That made it a user-facing issue in my opinion. And it's not ready for usage. That added the complication of disabling write actions (in a separate commit). If I get hit by a bus and its forgotten, with write actions enabled, we'd start seeing data loss in production bugzilla once prod bbsync was enabled, hence the caution and extra commit.

Our options with _this_ PR are IMO

- Postponing this, implementing bbsync first, and then removing commit 058f6c7ab89725dc9f8aec001c62294da917b258 from this PR. 
- Merging this, then continuing with bbsync, then enabling write api disabled in commit 058f6c7ab89725dc9f8aec001c62294da917b258.

What do you prefer?

osidb/tests/test_endpoints.py
Elkasitu: I mean that this commit does not do what it says it does "Do not require flaw value in body, because it is already provided in URL", the commit simply modifies a test, does not change anything about the actual code, given the commit message I would imagine some sort of change to the serializer.

Furthermore, shouldn't passing the flaw id parameter return a 400 response? That might be worthy of a bug ticket in JIRA

jobselko: @Elkasitu The code change happened in https://github.com/RedHatProductSecurity/osidb/pull/256/commits/15c9664494e00bd2a49143c97f14e6f6a70f1e61 in `osidb/serializer.py`. At that time, I forgot to update the test. And maybe, I forgot to update even more tests. I will check it and create another PR with FlawReference fixes, hopefully the last one :)

I agree that the used commit name ("Do not require flaw value in body, because it is already provided in URL".) is a bit confusing.

jsvob: Aha, I meant "Change the test to not provide the flaw value to be inside body, because it is already provided in the URL." I saw the test as something that effectively places requirements on how stuff should work (test should pass). Would the revised commit message in the first sentence work?

Theoretically, yes, it should return a 400, but that would IMO require changes I'm not sure are short enough for us to want. The reason the extra "flaw" in body is not detected (and in fact, always ignored) is that the viewset's `get_serializer()` injects the `flaw` from URL into the body, overwriting any potential value already in the body. The fix to return a 400 would probably be to have an extra condition in `get_serializer()` to detect the extra `flaw` in body and return a 400 response. Shall I file the ticket?

jsvob: Removing the commit removes the bad message.

PR #267, opened 2023-07-03
Convert is_major_incident field in Flaw into property

jobselko:
Closes OSIDB-266 (Improve Major Incident model and API)


osidb/models.py
jobselko, line 1: `FlawHistory` also contains the `is_major_incident` field. Should I remove it or leave it as it is?

openapi.yml
```diff
@@ -932,10 +932,6 @@ paths:
           - IMPORTANT
           - LOW
           - MODERATE
-      - in: query
-        name: flaw__is_major_incident
-        schema:
-          type: boolean
```
Elkasitu, line 938: this breaks API

osidb/filters.py
```diff
@@ -152,7 +152,6 @@ class Meta:
             "nvd_cvss2": ["exact"],
             "nvd_cvss3": ["exact"],
             "component": ["exact"],
-            "is_major_incident": ["exact"],
```
Elkasitu, line 155: you can just use a function for filtering over this attribute

osidb/migrations/0081_remove_flaw_is_major_incident.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ('osidb', '0080_alter_flawacknowledgment_affiliation'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='flaw',
+            name='is_major_incident',
+        ),
+    ]
```
Elkasitu, line 17: We try to keep migrations as N-1 compatible as possible, this means that in the scenario in which a deployment fails the database must be kept in a state that is usable both for version N and version N-1.

This change breaks such compatibility because version N-1 will find itself with a missing column in the database.

To amend this, what we do is [deprecate the field](https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#deprecate-fields) and then in version N+1 we can actually remove it from the database, because version N and version N-1 will no longer use the field from the database.

This means that is_major_incident can't be a property yet, but you can make a function that will be called when accessing the deprecated field with the `return_instead` argument (grep that in the code for some examples).

osidb/models.py
Elkasitu: You can ignore it for now since the model isn't being used

Elkasitu (changes requested):

osidb/migrations/0081_remove_flaw_is_major_incident.py
osoukup: `return_instead` could just do what your new property does for the time being and the API part can remain intact. The rest can be removed.

osidb/tests/factories.py
```diff
@@ -64,20 +64,35 @@ class Meta:
...
-        no_declaration=factory.Faker("random_element", elements=["", "foo"]),
+    summary = factory.LazyAttribute(
+        lambda f: "I am a spooky CVE"
+        if f.major_incident_state
+        in [
+            Flaw.FlawMajorIncident.REQUESTED,
+            Flaw.FlawMajorIncident.APPROVED,
+            Flaw.FlawMajorIncident.CISA_APPROVED,
+        ]
+        else factory.Faker("random_element", elements=["", "foo"])
```
JakubFrejlach, line 85: I did some digging I am not sure this will work, when the `factory.Faker` is used inside the `LazyAttribute` it won't evaluate itself and the faker object itself will be assigned instead https://github.com/FactoryBoy/factory_boy/issues/915.

I've found a way how to resolve that, you need to use `class Params` inside the factory like this:

```python
class Params:
        fallback_summary = factory.Faker(
            "random_element", elements=["","foo"]
        )
...
summary = factory.LazyAttribute(
        lambda f: "I am a spooky CVE"
        if f.major_incident_state
        in [
            Flaw.FlawMajorIncident.REQUESTED,
            Flaw.FlawMajorIncident.APPROVED,
        ]
        else f.fallback_summary
)          
```

or you can keep using the `factory.Maybe` and evaluating the complex condition in the `Params` class like this:

```python

class Params:
    complex_condition = factory.LazyAttribute(lambda f: True if f.major_incident_state in [...] else False)
```

and then use the `complex_condition` for the first argument of `factory.Maybe` and keep the `yes/no_declaration`.

osidb/tests/factories.py
```diff
@@ -64,20 +64,35 @@ class Meta:
...
-        no_declaration=factory.Faker("random_element", elements=["", "foo"]),
+    mitigation = factory.LazyAttribute(
+        lambda f: "CVE mitigation"
+        if f.major_incident_state
+        in [
+            Flaw.FlawMajorIncident.REQUESTED,
+            Flaw.FlawMajorIncident.APPROVED,
+            Flaw.FlawMajorIncident.CISA_APPROVED,
+        ]
+        else factory.Faker("random_element", elements=["", "foo"])
```
JakubFrejlach, line 95: Same as above.

JakubFrejlach (changes requested):
One problem with the test factories which needs to be address.

osidb/tests/factories.py
jobselko: Thanks! Updated.

openapi.yml
jobselko: These fields are back, but so far, I have not been able to mark them as deprecated.

openapi.yml
```diff
@@ -5479,7 +5476,9 @@ components:
           type: string
           maxLength: 100
         is_major_incident:
-          type: boolean
+          type: string
```
jobselko, line 5479: @JakubFrejlach I do not understand this change. Is it ok?

osidb/models.py
```diff
@@ -652,8 +651,25 @@ class FlawMajorIncident(models.TextChoices):
...
+            self.FlawMajorIncident.APPROVED,
+            self.FlawMajorIncident.CISA_APPROVED,
+        ]
+
     # should be set True if MAJOR_INCIDENT or MAJOR_INCIDENT_LITE FlawMeta exists, from BZ flagsq
-    is_major_incident = models.BooleanField(default=False)
+    is_major_incident = deprecate_field_custom(
+        models.BooleanField(default=False),
+        # required to keep backwards compatibility
+        return_instead=is_major_incident_temp,
```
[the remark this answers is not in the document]
Elkasitu: @jobselko yep, just gave it a whirl and turns out that DRF does some sort of introspection on the field and so the `self` argument that `is_major_incident_temp` receives might be `None`, so try doing something like `if self is None: return False` in `is_major_incident_temp`

JakubFrejlach (approved):
LGTM now! Good job.

osidb/filters.py
```diff
@@ -75,6 +75,8 @@ class FlawFilter(DistinctFilterSet):
     changed_before = DateTimeFilter(
         field_name="updated_dt", method="changed_before_filter"
     )
+    is_major_incident = BooleanFilter(method="is_major_incident_filter")
```
Elkasitu, line 78: Can't this filter be based on the `is_major_incident` field? Then you'd avoid a lot of duplication

osidb/serializer.py
```diff
@@ -717,6 +717,10 @@ class FlawSerializer(
     meta = serializers.SerializerMethodField()
     meta_attr = serializers.SerializerMethodField()
 
+    # This line forces the deprecated "is_major_incident" field NOT to change
+    # from boolean to string. Otherwise, some unknown logic turns it into string.
```
Elkasitu, line 721: Probably because it's a descriptor and it cannot infer what the return type will be, maybe with type hints it can?

Elkasitu (approved):
Some minor remarks, but mergeable nonetheless

osidb/filters.py
jobselko: Do you mean `is_major_incident = BooleanFilter(field_name="is_major_incident")`?

It seems it is working. Interesting that we did not think of it earlier. Could you please approve that I get you right? If so, I will update the code.

osidb/serializer.py
jobselko: You may be right, but this solution is also correct and not complicated, so I would prefer it for now. I will check later with @JakubFrejlach if it could be solved as you recommended.

osidb/filters.py
Elkasitu: Yes that is what I meant, I think the confusion comes from the deprecation docs which stated that all usage of the field *must* be removed -- but that is incorrect, write operations on the field should be removed but read operations can be kept since the database field won't be accessed in any way.

I amended the documentation in the PR that introduces the custom `deprecate_field`.

jobselko: In the end, I cannot use it.
When trying `flaw__is_major_incident = BooleanFilter(field_name="flaw__is_major_incident")` in `AffectFilter`, I got `FieldError at /osidb/api/v1/affects, Related Field got invalid lookup: is_major_incident` when accessing `http://localhost:8000/osidb/api/v1/affects?flaw__is_major_incident=true`.

I would prefer to leave the code as it is now, so we do not end up in a rabbit hole.

PR #268, opened 2023-07-04
Ensure that oasdiff returns exit code 1 on warnings

Elkasitu:
The oasdiff tool used in the CI pipeline does not return exit code 1 on warnings by default, yet the warnings can include backwards-incompatible changes even if these changes do not yield out an error for the client but are simply undefined behavior.

This commit adds the -fail-on-warns flag so that warnings are also considered breaking changes.


JakubFrejlach (approved):
LGTM.

PR #269, opened 2023-07-04
Update Django to latest security revision

Elkasitu:
This patch includes a fix to CVE-2023-36053 which can lead to ReDoS via URLValidator which we do use.


osoukup (approved):
LGTM. Thank you for keeping an eye on it!

JakubFrejlach:
Should we also bump the `cryptography` package alerted by the dependabot?

Elkasitu:
@JakubFrejlach I considered it but it's a major version change for a low impact vuln so I've decided against it, if you or anyone else has strong reason to believe we should upgrade it I'm willing to reconsider

PR #270, opened 2023-07-04
document no trailing slash in Bugzilla URL

osoukup:
The `python-bugzilla` lib logic will result in calling the endpoints without `/rest` and those do not exist.

FYI @jobselko


jobselko (approved):
Thanks :)

PR #271, opened 2023-07-04
Introduce CVSS model for Affects and Flaws

JakubFrejlach:
Replacement PR based on the https://github.com/RedHatProductSecurity/osidb/pull/239, hopefully all previous suggestions were addressed.

Related to OSIDB-377


osidb/models.py
```diff
@@ -1595,6 +1596,160 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+
+    class CVSSVersion(models.TextChoices):
+        VERSION2 = "V2", "version 2"
+        VERSION3 = "V3", "version 3"
+
+    class CVSSIssuer(models.TextChoices):
+        REDHAT = "RH", "Red Hat"
+        NIST = "NIST", "NIST"
+
+    cvss_handles = {CVSSVersion.VERSION2: CVSS2, CVSSVersion.VERSION3: CVSS3}
```
osoukup, line 1651: nitpick: the constants should be rather uppercase

osidb/models.py
```diff
@@ -1595,6 +1596,160 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+    )
+
+    objects = AffectCVSSManager()
+
+    class Meta:
+        """define meta"""
+
+        constraints = [
+            models.UniqueConstraint(
+                fields=["affect", "version", "issuer"], name="unique CVSS of a Affect"
```
osoukup, line 1748: nitpick: an Affect

osidb/tests/test_cvss.py
```diff
@@ -0,0 +1,125 @@
...
+        assert flaw_cvss2 in all_flaw_cvss_scores
+        assert flaw_cvss3 in all_flaw_cvss_scores
+
+        all_affect_cvss_scores = affect.cvss_scores.all()
+        assert len(all_affect_cvss_scores) == 2
+        assert affect_cvss2 in all_affect_cvss_scores
+        assert affect_cvss3 in all_affect_cvss_scores
+
+        # Test flaw/affect-version-issuer uniqueness constraint.
+        with pytest.raises(ValidationError) as e:
```
osoukup, line 45: nitpick: we try to standardize on `match` parameter instead of the later `assert`

osoukup: ...it is on multiple places here

docs/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Added
 - Implement major_incident_state in Flaw API (OSIDB-266)
+- Add CVSS model for Flaws and Affects (OSIDB-377)
```
osoukup, line 10: I am not sure we should add the changelog item when it is not yet visible to our users. I would probably add one item when all the support from `bzimport` through model and API to `BBSync` is done.

osoukup (commented):
Just a few minor comments but generally LGTM

osoukup (approved):

osoukup:
Just a few minor comments but LGTM

docs/CHANGELOG.md
JakubFrejlach: Makes sense, will remove for now

osidb/tests/test_cvss.py
JakubFrejlach: Changed

costaconrado (approved):
LGTM.

osidb/models.py
```diff
@@ -1595,6 +1596,160 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+        """
+        if self.comment and self.issuer != self.CVSSIssuer.REDHAT:
+            raise ValidationError(
+                "CVSS comment can be set only for CVSSs issued by Red Hat."
+            )
+
+    def __eq__(self, o):
+        if isinstance(o, CVSS) and self.version == o.version:
+            return self.cvss_object == o.cvss_object
+        return False
```
Elkasitu, line 1711: Unless you have a strong reason for this I'd rather not override `__eq__`, this also makes CVSS model object instances unhasheable.

osidb/models.py
```diff
@@ -1595,6 +1596,160 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+    def full_version(self):
+        """Full name of the CVSS version."""
+        return f"CVSS{self.version[1:]}"
+
+    @property
+    def cvss_handle(self):
+        """
+        Corresponding CVSS handle from CVSS library based on the version.
+        """
+        return self.CVSS_HANDLES[self.version]
```
Elkasitu, line 1679: You only use this in tests, and the instances in this model where you use it can be replaced by `cvss_object`, so I would just inline this into `cvss_object`.

osidb/tests/test_cvss.py
```diff
@@ -0,0 +1,127 @@
...
+        with pytest.raises(ValidationError, match="Invalid CVSS"):
+            FlawCVSSFactory(version=CVSS.CVSSVersion.VERSION2, vector=cvss2_vector)
+
+    @pytest.mark.enable_signals
+    def test_cvss_recalculation(self):
+        cvss = FlawCVSSFactory(
+            vector="CVSS:3.1/AV:A/AC:H/PR:L/UI:R/S:C/C:L/I:L/A:L",
+            version=CVSS.CVSSVersion.VERSION3,
+        )
+        assert cvss.cvss_handle(cvss.vector).scores()[0] == cvss.score
```
Elkasitu, line 103: This kind of test is quite masturbatory and doesn't actually test anything because you're using untested code to do the testing, what if the logic of `cvss_handle` is wrong? The test will pass but it will be wrong.

I would rather do

```python
def test_cvss_recalculation(self, vector):
    cvss = FlawCVSSFactory(vector=vector, CVSS.CVSSVersion.VERSION3)
    assert cvss.score = CVSS3(vector).scores()[0]
```

Elkasitu (changes requested):
Overall LGTM but there are some small issues.

Also I would suggest just squashing everything into a single commit, I don't see why we need separate commits for tests, secret updates, etc. It is all one unit of work.

osidb/tests/test_cvss.py
JakubFrejlach: But does that make a difference if I use CVSS3 from cvsslib directly or via cvss_handle? :thinking:

Elkasitu: Imagine that in the future someone makes a change to the `CVSS_HANDLES` class variable and maps CVSSv3 to CVSS4 by mistake, your test will still pass (wrongfully) because it depends on the implementation of `cvss_handle`.

By being explicit and using `CVSS3` directly you no longer depend on the implementation of `cvss_handle` and check for what you actually want which is the calculated score being correct.

JakubFrejlach: I see. Makes sense, will change that.

osidb/models.py
JakubFrejlach: :facepalm: Ah, this feels like sending an email with the attachment mention but without the actual attachment :grin:

PR #272, opened 2023-07-04
tracker BTS merge

osoukup:
First steps in attempt to merge the Bugzilla and Jira tracker functionality together so from the caller it does not matter what BTS and inside we share as much functionality as possible. There is of course lot more to do...

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -1,6 +1,5 @@
 """
-BTS Trackers model used to create in-memory trackers
-that will be persisted in the corresponding BTS
+common tracker functionality shared between BTSs
 """
 from typing import Any, Dict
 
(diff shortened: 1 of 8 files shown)


osoukup:
@costaconrado I unfortunately did not have enough time to merge this so just creating a draft PR in case you would work on the trackers while I am away (Wed, Thu) we can prevent some collisions. This PR is also opinionated so feel free to disagree with the approach/naming/structure etc.

osoukup:
> @costaconrado I unfortunately did not have enough time to merge this so just creating a draft PR in case you would work on the trackers while I am away (Wed, Thu) we can prevent some collisions. This PR is also opinionated so feel free to disagree with the approach/naming/structure etc.

@costaconrado so I have found a bit more time and at least fixed the errors. I was originally also moving the models and the migrations into the Jira subdirectory but it seems that it would have to be an independent app to make Django understand it which does not seem desirable. So I think this is ready for a review. It is mostly just moving furniture around and providing a common save API which would be part of the `tracker.save`.

PR #273, opened 2023-07-07
Copy FlawMeta ACKNOWLEDGMENT into FlawAcknowledgment.

jsvob:
Copy FlawMeta ACKNOWLEDGMENT into FlawAcknowledgment.

This makes the DB data consistent and ready for FlawAcknowledgment bbsync. Without this consistency, FlawAcknowledgment manipulation and bbsync would be either too complex or nonfunctional, because it wouldn't be clear whether to assemble srtnotes acknowledgments based on FlawMeta or FlawAcknowledgment, or would make acks disappear if a Flaw without populated FlawAcknowledgment was bbsynced.

Related to OSIDB-1004 ("Implement BBSync part for FlawAcknowledgment")

### DIFF
--- a/osidb/migrations/0082_migrate_flawacknowledgment_flawmeta.py
+++ b/osidb/migrations/0082_migrate_flawacknowledgment_flawmeta.py
@@ -1,0 +1,86 @@
+"""
+Written manually on 2023-07-14
+
+Copy FlawMeta ACKNOWLEDGMENT into FlawAcknowledgment.
+
+To prevent out of memory issues, FlawMeta are iterated using .iterator()
+so that FlawMeta instances are not cached. To prevent further out of
+memory issues and runtime efficiency, the migration uses a generator,
+islice and bulk_create as recommended by
+https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-create
+
+This makes the DB data consistent and ready for FlawAcknowledgment bbsync.
+Without this consistency, FlawAcknowledgment manipulation and bbsync would be
+either too complex or nonfunctional, because it wouldn't be clear whether to
+assemble srtnotes acknowledgments based on FlawMeta or FlawAcknowledgment, or
+would make acks disappear if a Flaw without populated FlawAcknowledgment was
+bbsynced.
+"""
+
+from django.db import migrations
+from ast import literal_eval
+from itertools import islice
+
+
+def generate_flaw_acknowledgments(apps):
+    FlawMeta = apps.get_model("osidb", "FlawMeta")
+    FlawAcknowledgment = apps.get_model("osidb", "FlawAcknowledgment")
+    for meta_ack in FlawMeta.objects.filter(type="ACKNOWLEDGMENT").iterator():
+        flaw = meta_ack.flaw
+        name = meta_ack.meta_attr["name"]
+        # Convert None to empty string.
+        affiliation = meta_ack.meta_attr.get("affiliation") or ""
+        # hstore holds bolean values as strings containing True|False
+        # so we need to explicitly convert it to the bolean value
+        from_upstream = literal_eval(meta_ack.meta_attr["from_upstream"])
+        # FlawAcknowledgment that already exist are skipped because they
+        # are exactly as up-to-date as FlawMeta ACKNOWLEDGMENT.
+        yield FlawAcknowledgment(
+            flaw=flaw,
+            name=name,
+            affiliation=affiliation,
+            from_upstream=from_upstream,
+            meta_attr=meta_ack.meta_attr,
+            acl_read=meta_ack.acl_read,
+            acl_write=meta_ack.acl_write,
+            created_dt=meta_ack.created_dt,
+            updated_dt=meta_ack.updated_dt,
+        )
+
+
+def forwards_func(apps, schema_editor):
+    batch_size = 1000
+
+    FlawAcknowledgment = apps.get_model("osidb", "FlawAcknowledgment")
+    generator = generate_flaw_acknowledgments(apps)
+    while True:
+        # Limiting memory usage by using a generator and slicing recommended in
+        # https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-create
+        batch = list(islice(generator, batch_size))
+        if not batch:
+            break
+
+        # There shouldn't be many existing FlawAcknowledgment instances. Those that
+        # do exist are identical to those that would be created by this migration,
+        # therefore ignoring conflicts.
+        FlawAcknowledgment.objects.bulk_create(batch, batch_size, ignore_conflicts=True)
+
+
+# About reversing the migration:
+# forwards_func() creates FlawAcknowledgment instances that mirror
+# FlawMeta ACKNOWLEDGMENT instances that are identical to FlawAcknowledgment
+# that would be created using Flaw bzimport, and that could have already
+# been created during Flaw bzimport before forwards_func() executed.
+# No way to revert, and no reason to revert.
+# You might consider reverting back to 0078 which is before FlawAcknowledgment
+# was created.
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("osidb", "0081_cvss"),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards_func, migrations.RunPython.noop, atomic=True),
+    ]


osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,54 @@
...
+        # Convert None to empty string.
+        affiliation = meta_ack.meta_attr.get("affiliation") or ""
+        # hstore holds bolean values as strings containing True|False
+        # so we need to explicitly convert it to the bolean value
+        from_upstream = literal_eval(meta_ack.meta_attr["from_upstream"])
+
+        # FlawAcknowledgment that already exist are skipped because they
+        # are exactly as up-to-date as FlawMeta ACKNOWLEDGMENT.
+
+        if not FlawAcknowledgment.objects.filter( flaw=flaw, name=name, affiliation=affiliation ).exists():
```
jsvob, line 22: Hmm, githooks black apparently ignores migrations. I'll fix it later after review.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,54 @@
...
+    pass
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('osidb', '0080_alter_flawacknowledgment_affiliation'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards_func, reverse_func),
```
costaconrado, line 53: I believe this operation should be a transaction, I think `atomic=True` would do the trick, what do you think?

costaconrado (changes requested):

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: Didn't know about that! Thanks! That would be bad, esp. in the hypothetical case where osidb runs past the partially failed migration and doesn't retry it / marks it as ok. I don't know enough about that, but I added `atomic=True` and locally tweaked the code to make the migration fail for some model instances and `make migrate` fails loudly and django never marks the migration as successful, so now it should be fine.

Tested on production data dump from February 2023.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: Fixed.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: (Fixed in a rebased commit. IMHO resolved.)

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("osidb", "0080_alter_flawacknowledgment_affiliation"),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards_func, reverse_func, atomic=True),
```
jobselko, line 56: You can use `migrations.RunPython.noop` instead of an empty reverse function.

jobselko (changes requested):
I am unsure if we want to do this change right now. If so, we should probably do it also for `FlawReferences` etc.
If I get it right, this could cause some other consequences.

costaconrado (approved):
LGTM.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,57 @@
...
+                meta_attr=meta_ack.meta_attr,
+                acl_read=meta_ack.acl_read,
+                acl_write=meta_ack.acl_write,
+                created_dt=meta_ack.created_dt,
+                updated_dt=meta_ack.updated_dt,
+            )
+
+
+def reverse_func(apps, schema_editor):
+    # forwards_func() creates FlawAcknowledgment instances that mirror
```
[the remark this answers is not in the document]
jsvob: There's nothing to update nor add IMO. Please explain.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: Fixed, thanks.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,57 @@
+# Written manually on 2023-07-07
+
+# https://docs.djangoproject.com/en/4.2/ref/migration-operations/
```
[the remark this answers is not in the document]
jsvob: Comment removed. It was a leftover note.

osoukup:
> I am unsure if we want to do this change right now. If so, we should probably do it also for `FlawReferences` etc. If I get it right, this could cause some other consequences.

I  would not mix this with the references or or other attributes. One thing is that they are out of scope of what @jsvob currently does but also not all of them need the same treatment - eg. I remember when we discussed Major Incidents we came to conclusion that they will need a resync anyway as the meta is missing some parts of the data (they are not too many). As this initiates the new model but does not change the old one, I do not see any breaking consequences.

jsvob:
> I am unsure if we want to do this change right now. If so, we should probably do it also for `FlawReferences` etc. If I get it right, this could cause some other consequences.

Please provide actionable feedback and/or file a respective Jira. Thank you.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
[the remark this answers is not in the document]
jobselko: Earlier, you described the behaviour of `forwards_func(...)` in `reverse_func(...)`. I wanted you to fix it.

jobselko (approved):
@jsvob I discussed it with Ondřej today, and this PR is fine. Since this data migration is not necessary, we do not need to make any changes to `FlawReference` in an upcoming PR.

osoukup:
> @jsvob I discussed it with Ondřej today, and this PR is fine. Since this data migration is not necessary, we do not need to make any changes to `FlawReference` in an upcoming PR.

Just to be precise :smile: it is necessary to migrate the data somehow. Somewhere it might be better by resync (eg. when the stored data are incomplete) somewhere we decided to postpone it for whatever reason etc. Generally we will need to migrate all the data to the new attributes at some point.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
```diff
@@ -0,0 +1,55 @@
...
+        # so we need to explicitly convert it to the bolean value
+        from_upstream = literal_eval(meta_ack.meta_attr["from_upstream"])
+
+        # FlawAcknowledgment that already exist are skipped because they
+        # are exactly as up-to-date as FlawMeta ACKNOWLEDGMENT.
+
+        if not FlawAcknowledgment.objects.filter(
+            flaw=flaw, name=name, affiliation=affiliation
+        ).exists():
+            FlawAcknowledgment.objects.create(
```
Elkasitu, line 25: Could we use [bulk_create](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create) instead? In production this will be a lot of data and the migration could take too long, potentially making the deployment fail

Elkasitu (approved):
Overall LGTM but please consider my suggestion.

In the future, we should create these data migrations along with the code that enables ingestion (in this case from bzimport), then you don't need to check whether the `FlawAcknowledgment` already exists or not.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: I did consider it, and I think it would require one to be careful about consumed memory. That quickly got me into a rabbit hole of using a generator instead of a list and setting `batch_size` for bulk_create. Could be done, but potentially complicated for me. I agree with your suggestion, but I'm not skilled enough to pull it off quickly and be otherwise productive.

So, I tested the current approach on production data and it took only a few seconds.

Theoretically, there can be an arbitrary number of FlawAcknowledgment instances, but I think in reality the number has a low upper bound.

IMO, the migration will be executed only in these two situations:

- On existing osidb instances in the year 2023 -> just a few seconds.
- On new osidb instances where an old database dump is being restored -> the data must be from 2023 at the latest -> just a few seconds.

Elkasitu: > So, I tested the current approach on production data and it took only a few seconds.

On *your machine*, remember that in production the pods running these workloads are constrained in both memory and CPU (but mostly CPU for the web pod).

There are roughly 9k `FlawMeta` records of type `ACKNOWLEDGMENT` in production, so we'll have to perform a little under 9k `FlawAcknowledgment` INSERTs while going through Django's ORM, I honestly don't know whether this would be an issue or not but I can tell you that we've had some issues with this in the past.

That said I just checked and the migration deadline is 30min, which should be plenty I think, but in the future we should always have performance in mind when developing data migrations, it's essential that both time and memory constraints are kept in check

jsvob: Aha! Thank you for the context and explanation!

In that case, I'll _try_ to rework it as per your suggestion and learn to make it use reasonably-sized batches. Seems this is a skill that will be necessary in the future. If I fail, we can merge _this_ migration once it becomes necessary, but we won't have that comfort in the future.

jsvob:
I previously said it would be merged after PR265, but after yesterday's talk with Jitka, I realized this is independent.

osidb/migrations/0081_migrate_flawacknowledgment_flawmeta.py
jsvob: Fixed by using a bulk approach recommended by django. Added explanation of that approach into the docstring.

Tested locally with millions of instances and expected memory consumption is roughly under 100 MB RAM and done within a minute for 10k objects. Likely my estimate is overshot by an order of magnitude based on testing of the migration file with actual production data (a few seconds, memory consumption not measurable).

PR #274, opened 2023-07-10
account for SRT email disabling Bugzilla data migration in bzimport

osoukup:
https://bugzilla.redhat.com/buglist.cgi?columnlist=product%2Ccomponent%2Cassigned_to%2Cstatus%2Csummary%2Clast_change_time%2Cseverity%2Cpriority&component=vulnerability&f1=delta_ts&f2=delta_ts&list_id=13258299&o1=greaterthaneq&o2=lessthaneq&order=status%2C%20priority%2C%20assigned_to%2C%20id%2C%20&product=Security%20Response&query_format=advanced&v1=2023-07-07T08%3A25%3A00&v2=2023-07-07T08%3A40%3A00

### DIFF
--- a/collectors/bzimport/collectors.py
+++ b/collectors/bzimport/collectors.py
@@ -304,6 +304,15 @@ class FlawCollector(Collector, BugzillaQuerier, JiraQuerier):
             "end": timezone.datetime(2023, 5, 13, 2, 3, tzinfo=TIMEZONE),
             "step": relativedelta(hours=5),
         },
+        # migration reassigning flaws to nobody@redhat.com
+        # after security-response-team@redhat.com disabling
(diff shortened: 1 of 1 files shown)


costaconrado (approved):
LGTM.

Elkasitu (approved):
should we have this be an env variable too? we need a more generic solution to such issues

osoukup:
> should we have this be an env variable too? we need a more generic solution to such issues

env variable would be better, we could account for a new migration without requiring a release. When this hardcoded approach was introduced the large data migrations were rather something historical we wanted to avoid in general but in this case the data migration was not even caused by us. I am not going to extend this PR but if this keeps happening we will have to do it more properly.

PR #275, opened 2023-07-11
Vendor deprecate_field to allow for more dynamic return_instead values

Elkasitu:
With this commit we extend django-deprecate-fields and provide our own version of the deprecate_field decorator, in this version the return_instead parameter can take a callable that accepts either 0 (like the original) or 1 argument.

If an argument can be passed, it shall be the Model object instance on which the field access was requested. This is useful in cases where the deprecation happens due to a rename or restructuring and can be extrapolated from another Model object instance attribute.

Hopefully I can get this change merged upstream and we can remove this from our code base (although I'm not holding my breath since it hasn't received much love for 3 years).


osoukup (approved):
LGTM and looks useful for a number of data migrations we do

PR #277, opened 2023-07-14
Assembler integration and other fixprovements

Elkasitu:
This PR introduces the following changes:

* The most significant change is the introduction of the `FlawDraft` model which is necessary to migrate Assembler into OSIDB
* A fix to the serialization of the `Profile` model
* Some extra shortcut methods have been added to `ACLMixin` that abstract the handling of acl_read/acl_write a bit more
* FlawSources have been reorganized into logical sets so that they can be easily combined to obtain all kinds of combinations

More information can be found on each individual commit's message.

Closes OSIDB-905


jsvob (commented):
LGTM, but I'm not knowledgeable enough to really provide useful review.

osidb/models.py
```diff
@@ -2733,6 +2747,83 @@ def __str__(self):
         return self.username
 
 
+class FlawDraft(FlawMixin, AlertMixin):
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    created_dt = models.DateTimeField(auto_now_add=True)
+    updated_dt = models.DateTimeField(auto_now=True)
+    origin = models.URLField(max_length=2048)
+    content = models.JSONField()
```
osoukup, line 2756: Could you describe this field?

osidb/models.py
```diff
@@ -2733,6 +2747,83 @@ def __str__(self):
         return self.username
 
 
+class FlawDraft(FlawMixin, AlertMixin):
```
osoukup, line 2750: I am thinking about having `Flaw` and `FlawDraft` modeled in this way. If we had one (probably draft) as a parent of the other it would result in a lot of flaws with number of empty attributes and other disadvantages. This is maybe better internally. However externally as a data consumer I should be able to query for eg. all flaws of some source and obtain both draft and regular flaws and probably even rejected if no specified otherwise. Draft is should be considered as a first state of the flaw rather then a different entity.

osidb/models.py
```diff
@@ -2733,6 +2747,83 @@ def __str__(self):
...
+        #   the case here
+        # 2) flaw is not null, which is not the case here
+        #
+        # the proper solution to this problem would be to decouple said
+        # validation from ACLMixin with e.g. a FlawChildACLMixin mixin
+        # which inherits from ACLMixin and applies the aforementioned
+        # validation and any others related to the Flaw parent relationship
+        ...
+
+    def _validate_embargoed_source(self):
```
osoukup, line 2775: I would rather move this to the base class to avoid duplicates.

osoukup (commented):
I would like to clarify a few things first but mostly this already looks well

osidb/models.py
```diff
@@ -588,10 +588,48 @@ def fts_search(q):
...
+# declared here
+# TODO: Move enums into their own namespace e.g. ./types.py
+class FlawMixin(ACLMixin, NullStrFieldsMixin):
+    """
+    Mixin for attributes in common between Flaw-like models
+    """
+
+    # short description of the Flaw
+    # TODO: should be a CharField?
+    title = models.TextField()
```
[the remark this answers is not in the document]
Elkasitu: This is the status quo, I suggest changing to a CharField because a TextField seems excessive, although I think postgres treats char columns with 255+ characters as text columns anyway so there wouldn't be a difference. I did not change it as it's an unnecessary migration IMO

osidb/models.py
```diff
@@ -2733,6 +2747,83 @@ def __str__(self):
         return self.username
 
 
+class FlawDraft(FlawMixin, AlertMixin):
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    created_dt = models.DateTimeField(auto_now_add=True)
+    updated_dt = models.DateTimeField(auto_now=True)
```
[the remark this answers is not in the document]
Elkasitu: Because this model lives in OSIDB only and `TrackingMixin` does a lot of specific logic for models for which the tracking is done externally.

I could create an `InternalTrackingMixin` that is simpler (simply use `auto_now_add` and `auto_now`) or something but I don't know how useful it'll be at this time.

osidb/models.py
Elkasitu: This is basically the equivalent of snippets, will update with a comment

osidb/models.py
Elkasitu: Going with a Polymorphic model was my first choice too but it complicates the "promotion" and "rejection" of drafts, the whole point of having a separate model is that you have the bare minimum amount of fields required in order to call it a proper Flaw, and promoting to an actual Flaw will trigger all of Flaw's validations.

Having it as a polymorphic model would imply, at the very least, a field to keep track of rejection/approval status in the polymorphic model itself and then a `save` override that would apply validations or not depending on the state.

I think in its current form it's sound from a modelling point of view, we can argue about the usefulness of displaying all draft flaws together with real flaws (there is a lot, and I mean a lot of vulnerabilities that will never materialize because they don't affect Red Hat products) but I don't see how including those together with actual flaws would be relevant.

osidb/models.py
Elkasitu: The problem is that `FlawMixin` does not inherit from `AlertMixin`, because if it did then we have a diamond inheritance problem because `Flaw` inherits from `BugzillaSyncMixin` which itself depends on `AlertMixin`, so it would not know which implementation of alert/save/etc. to use.

Elkasitu: But I can document this

osoukup: I see. No problem then. There should not be many validations for the drafts.

osidb/models.py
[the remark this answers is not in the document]
osoukup: It does all what is done here - introduces two `dt` attributes and implement `auto_now_add` and `auto_now` logic. Additionally it allows rewriting of this automation by explicit values which may or may not be useful but also mid-air collision detection. I see no reason why we should not care for the mid-air collisions in the case of drafts.

osidb/models.py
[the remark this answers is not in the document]
osoukup: I am OK with the text field. Only stating what is the Bugzilla limit.

osidb/models.py
Elkasitu: So I will leave it like this for now and think about this more in depth, but I think the current approach is correct:

* There are currently 97k assembler tickets and ~40k OSIDB flaws, merging them together into a single endpoint will not only make it a mess for users to navigate through, but will also impact our already poor performance.
* OSIDB's main mission is to display flaws that impact Red Hat products, this can be true for IR-created `FlawDraft`, but it is not the case for the vast majority of assembler tickets.
* Using a polymorphic model complicates things at the API level, because `FlawDraft` and `Flaw` will have different schemas (a draft can include very little information), we'll have to cater to lowest common denominator (`FlawDraft`) which would make the `/flaws` endpoint backwards incompatible.

Since we're only working at the model level right now and not the API level, we can see how it goes after a first collector implementation at the very least, and then we can always change the model before implementing the API.

osoukup: OK. Let us see. I still think that an option of querying all the flaws at once is going to be important but we can figure it out later. It should be no regression with regards to SFM2 where the ASM and Flaws are even further from each other. On the other hand the proper merge seems to me as on opportunity how to simplify this from the user point of view and the task should be to make the tools can handle it. I am not going to block this PR on this.

osoukup (approved):
LGTM. I have some conceptual doubts but I get the reasons so we will see later.

JakubFrejlach (approved):
Generally looks good. As far as the one-to-one instead of polymorphic approach is concerned I am not able to tell right now which one will work better so lets go with the current approach and we'll see.

PR #278, opened 2023-07-14
Use new model for flaw acknowledgments when saving data via bbsync

jsvob:
Follow up to https://github.com/RedHatProductSecurity/osidb/pull/273 and https://github.com/RedHatProductSecurity/osidb/pull/265

Related to OSIDB-1004 ("Implement BBSync part for FlawAcknowledgment")

### DIFF
--- a/apps/bbsync/tests/test_query.py
+++ b/apps/bbsync/tests/test_query.py
@@ -7,20 +7,12 @@ from freezegun import freeze_time
 
 from apps.bbsync.exceptions import SRTNotesValidationError
 from apps.bbsync.query import FlawBugzillaQueryBuilder, SRTNotesBuilder
-from osidb.models import (
-    Affect,
-    Flaw,
-    FlawComment,
-    FlawMeta,
-    FlawSource,
-    Impact,
-    Tracker,
-)
+from osidb.models import Affect, Flaw, FlawComment, FlawSource, Impact, Tracker
 from osidb.tests.factories import (
     AffectFactory,
+    FlawAcknowledgmentFactory,
     FlawCommentFactory,
     FlawFactory,
-    FlawMetaFactory,
     FlawReferenceFactory,
     PsModuleFactory,
     TrackerFactory,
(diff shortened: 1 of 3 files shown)


jsvob:
Blocked on https://github.com/RedHatProductSecurity/osidb/pull/265

apps/bbsync/tests/test_query.py
```diff
@@ -248,32 +241,23 @@ def test_generate_acknowledgments(self):
...
             flaw=flaw,
-            type=FlawMeta.FlawMetaType.ACKNOWLEDGMENT,
-            meta_attr={
-                "affiliation": "von Bahnhof",
-                "from_upstream": True,
-                "name": "Carl",
-            },
+            affiliation="von Bahnhof",
+            from_upstream=True,
+            name="Carl",
```
jobselko, line 254: `name` and `affiliation` are probably wrong. `name` can be e.g. "Carl von Bahnhof" and `affiliation` can be e.g. "Google Security Team". (I think "von Bahnhof" is a part of the name, not an organization. This problem occurs more times.)

apps/bbsync/tests/test_query.py
jobselko, line 1: Could you please update `test_schema`? It tests complex flaw SRT notes generation, and it still contains acknowledgment from `FlawMetaFactory`.

jobselko (changes requested):

apps/bbsync/tests/test_query.py
jsvob: Added names that should be more descriptive for demo purposes.

apps/bbsync/tests/test_query.py
jsvob: Updated the test function to use the new model and also added a test to make sure the test didn't run on empty data (before the fix it ran on empty data and didn't fail).

jobselko, line 1: You forgot to update `acknowledgments` in `srtnotes` (used in `FlawFactory`'s `meta_attr)` in `test_schema`.

jsvob: Fixed. Thanks.

jobselko: I am unsure if I understand the second part of your sentence. I would like to discuss it with you.

jsvob: I didn't notice the original mistake you noticed here. A part of the reason was that I didn't pay enough attention (not fixable by tests), and a part of the reason was that the test ran and didn't test FlawAcknowledgment srtnotes generation at all. So the newly added code at the end of `test_schema` tests that the main part of the test before it doesn't run on no data. Will explain more on 1:1.

jsvob: Aha, the sentence `# Additionally, ensure that the validations were not run on empty data.` was missing the word `not`. Sorry for the confusion.

Elkasitu (approved):
LGTM but I would squash these commits

jobselko (approved):

PR #280, opened 2023-07-18
Update Flaw to contain requires_summary

jobselko:
Related to OSIDB-1005

PR #281, opened 2023-07-20
Add requires_summary to Flaw API and update BZImport and BBSync

jobselko:
Follow-up to #280.

This PR updates the `Flaw` API to propagate the recently added `requires_summary` field in the `Flaw` model. The `requires_summary` value is read from the `requires_doc_text` flag via BZImport and synced backwards via BBSync.

Closes OSIDB-1005

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -596,6 +596,34 @@ class FlawConvertor(BugzillaGroupsConvertorMixin):
         return valid_pairs.get(flags_from_bz, Flaw.FlawMajorIncident.INVALID)
 
     @cached_property
+    def requires_summary(self):
+        """
+        A requires_summary state created from the requires_doc_text flag.
+        """
+        # Default values
+        status = ""
+        setter = ""
+
+        # Sets values from Bugzilla
+        for flag in self.flags:
+            if flag["name"] == "requires_doc_text":
+                setter = flag["setter"]
+                status = flag["status"]
+
+        # this combination is set automatically by BZ when Doc Text is added
+        if setter == "bugzilla@redhat.com" and status == "+":
+            return Flaw.FlawRequiresSummary.REQUESTED
+
+        pairs = {
+            "": Flaw.FlawRequiresSummary.NOVALUE,
+            "-": Flaw.FlawRequiresSummary.REJECTED,
+            "?": Flaw.FlawRequiresSummary.REQUESTED,
+            "+": Flaw.FlawRequiresSummary.APPROVED,
+        }
+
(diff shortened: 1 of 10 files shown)


osidb/tests/test_endpoints.py
```diff
@@ -1088,7 +1091,10 @@ def test_get_flaw_with_token(
         assert "refresh" in body
         token = body["access"]
 
-        flaw1 = FlawFactory.build(major_incident_state=Flaw.FlawMajorIncident.APPROVED)
+        flaw1 = FlawFactory.build(
+            major_incident_state=Flaw.FlawMajorIncident.APPROVED,
+            requires_summary=Flaw.FlawRequiresSummary.APPROVED,
+        )
```
Elkasitu, line 1097: You forgot to assert that the returned value matches this

Elkasitu (approved):
LGTM, just some minor nitpicks

collectors/bzimport/convertors.py
```diff
@@ -595,6 +595,34 @@ def major_incident_state(self):
...
+        for flag in self.flags:
+            if flag["name"] == "requires_doc_text":
+                setter = flag["setter"]
+                status = flag["status"]
+
+        # this combination is set automatically by BZ when Doc Text is added
+        if setter == "bugzilla@redhat.com" and status == "+":
+            return Flaw.FlawRequiresSummary.REQUESTED
+
+        pairs = {
```
jsvob, line 617: I suggest using `mapping` instead if `pairs`. I think it's more clear of what the dict is. `pairs` seem like a leftover from the dual-flag `major_incident_state()`'s `valid_pairs`.

collectors/bzimport/convertors.py
```diff
@@ -595,6 +595,34 @@ def major_incident_state(self):
...
+            "": Flaw.FlawRequiresSummary.NOVALUE,
+            "-": Flaw.FlawRequiresSummary.REJECTED,
+            "?": Flaw.FlawRequiresSummary.REQUESTED,
+            "+": Flaw.FlawRequiresSummary.APPROVED,
+        }
+
+        return pairs.get(status)
```
jsvob, line 624: I suggest using `pairs[status]`. That way, it crashes outright if there's an unexpected symbol. Shouldn't happen though, because BZ docs specify only ?+- in https://www.bugzilla.org/docs/2.18/html/flags-overview.html

jsvob (commented):
LGTM with nitpicks.

jsvob:
Hmm, I had the review open for the whole day and noticed "Merged" only after submitting.

PR #284, opened 2023-07-24
Use gevent worker_class when running through Gunicorn

Elkasitu:
The goal of this commit is to have Gunicorn workers that work with
cooperative multitasking in order to handle a lot more requests
concurrently without directly changing the amount of workers being
spawned by Gunicorn.

This is achieved in various steps:

* The psycogreen extension must be installed because by default psycopg2
  does not support async workloads
* In the WSGI entrypoint, both psycopg2 and Django/stdlib are patched by
  psycogreen and gevent respectively.
* Finally by changing the worker_class to be of type "gevent" we can
  leverage cooperative multitasking

This should result in an increase to the amount of requests/s that the
server can handle.

Closes OSIDB-964


jsvob (approved):
Based on 40 minutes of searching through the Internet, this approach seems sound and the PR LGTM.

I have no experience nor prior knowledge about wsgi & psycopg & greenlets & gevent and I didn't test this locally.

osoukup (approved):
LGTM. I have no experience with this but by a quick search it makes sense to me

PR #285, opened 2023-07-24
Change CORS policy to allow credentials

costaconrado:
This PR adds `Access-Control-Allow-Credentials: true` header in order to allow OSIM to authenticate.

Closes OSIDB-1115.


Elkasitu (approved):
Here's the documentation on how and why this works: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/access-control-allow-credentials

PR #287, opened 2023-07-27
Remove wrong __str__ in PsProduct

jobselko:
The `PsProduct` model contains `__str__` with `self.package`. However, this field does not exist in the model. This PR removes it with its `__str__`.


osoukup (approved):
LGTM

Elkasitu:
One of these is fixed in #277 and I think it might be best to change the field being used for serialization instead of removing the implementation of __str__ altogether

jobselko:
I reverted the changes in the `Profile` model as the model will be handled differently in #277.

PR #288, opened 2023-07-28
tracker validations

osoukup:
I went through the `Tracker` model and introduced data migrations which I could think about for now. It resulted in a refactoring of a huge number of tests as they used trackers without all the necessary context set and not all was easy to be done through the factories.

Closes OSIDB-787

### DIFF
--- a/apps/trackers/save.py
+++ b/apps/trackers/save.py
@@ -1,7 +1,7 @@
 """
 common tracker save funtionality
 """
-from osidb.models import PsModule, PsUpdateStream, Tracker
+from osidb.models import Tracker
 
 from .bugzilla.save import TrackerBugzillaSaver
 from .exceptions import BTSException
@@ -19,8 +19,6 @@ class TrackerSaver:
         detect and return the correct saver
         assuming that all prerequisites are met
         """
-        cls.assert_context(tracker)
-
         if tracker.type == Tracker.TrackerType.BUGZILLA:
             assert bz_api_key, "Bugzilla API key not provided"
             return TrackerBugzillaSaver(tracker, bz_api_key)
@@ -30,13 +28,3 @@ class TrackerSaver:
 
         # we should never get here
         raise BTSException("Unknown BTS")
-
-    @staticmethod
-    def assert_context(tracker):
-        """
-        the vital tracker related pieces of information need to exist
-        or otherwise we would not be able to construct the tracker query
-        """
-        assert tracker.affects.first()
-        assert PsModule.objects.filter(name=tracker.affects.first().ps_module)
-        assert PsUpdateStream.objects.filter(name=tracker.ps_update_stream)
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement ps_update_stream in Tracker API (OSIDB-1064)
 - Implement daily monitoring email for failed tasks
 - Implement nist_cvss_validation in Flaw API (OSIDB-1006)
+- Implement additional tracker validations (OSIDB-787)
 
 ### Changed
 - Change article link validation to be blocking (OSIDB-1060)
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1912,7 +1912,9 @@ class TrackerManager(ACLMixinManager, TrackingMixinManager):
     """tracker manager"""
 
     @staticmethod
-    def create_tracker(affect, external_system_id, _type, **extra_fields):
+    def create_tracker(
+        affect, external_system_id, _type, raise_validation_error=True, **extra_fields
+    ):
         """return a new tracker or update an existing tracker"""
         try:
             tracker = Tracker.objects.get(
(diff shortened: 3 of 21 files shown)


osidb/models.py
```diff
@@ -1894,6 +1894,36 @@ class Meta:
...
+
+    def _validate_tracker_ps_module(self):
+        """
+        check that the tracker is associated with a valid PS module
+        """
+        if not self.affects.first():
+            return
+
+        if not PsModule.objects.filter(name=self.affects.first().ps_module):
+            raise ValidationError("Tracker must be associated with a valid PS module")
```
Elkasitu, line 1931: This check is weird, `affects` is a many relationship, so how do we actually know whether a `Tracker` is associated with a valid PS module?

In most cases tracker<->affect is a 1:1 relationship but what happens if there's more than one affect per tracker? Should we only consider the first affect? What if the other affects DO have a valid `PsModule`?

Plus IMO this validation should be done at the affect level no? It would simplify the check.

osidb/models.py
```diff
@@ -1814,7 +1814,9 @@ class TrackerManager(ACLMixinManager, TrackingMixinManager):
     """tracker manager"""
 
     @staticmethod
-    def create_tracker(affect, external_system_id, _type, **extra_fields):
+    def create_tracker(
+        affect, external_system_id, _type, raise_validation_error=True, **extra_fields
```
Elkasitu, line 1831: not a big fan of bubbling up `raise_validation_error`... but don't see an alternative other than rewriting the convertor to not use `create_tracker`.

osidb/models.py
```diff
@@ -1484,6 +1484,19 @@ def _validate_notaffected_open_tracker(self):
                 "NOTAFFECTED but has open tracker(s).",
             )
 
+    def _validate_ooss_open_tracker(self):
+        """
+        Check whether out of support scope products have open trackers.
+        """
+        if (
+            self.resolution == Affect.AffectResolution.OOSS
+            and self.trackers.exclude(status__iexact="CLOSED").exists()
```
Elkasitu, line 1493: could you document why you're using `iexact` here? And are we sure `CLOSED` covers all? especially for JIRA trackers

osidb/models.py
```diff
@@ -1972,8 +1971,7 @@ def _validate_ooss_open_tracker(self):
         affect = self.affects.filter(resolution=Affect.AffectResolution.OOSS).first()
         if not self.is_closed and affect:
             raise ValidationError(
-                f"Affect ({affect.uuid}) for {affect.ps_module}/{affect.ps_component} is marked as "
-                "OOSS but has open tracker(s).",
+                f"The tracker is associated with an OOSS affect: {affect.uuid}",
```
Elkasitu, line 1974: this change you can just fixup into the previous commit

osidb/tests/test_flaw.py
```diff
@@ -1815,11 +1820,16 @@ def test_validate_ooss_open_tracker(
...
+            f"The tracker is associated with an OOSS affect: {affect.uuid}"
+            if entity == "tracker"
+            else f"{affect.uuid}.*is marked as.*but has open tracker"
+        )
         entity = tracker if entity == "tracker" else affect
         if should_raise:
             with pytest.raises(
                 ValidationError,
-                match=f"{affect.uuid}.*is marked as.*but has open tracker",
+                match=match,
```
Elkasitu, line 1832: ditto

apps/trackers/save.py
```diff
@@ -30,13 +28,3 @@ def __new__(cls, tracker, bz_api_key=None):
 
         # we should never get here
         raise BTSException("Unknown BTS")
-
-    @staticmethod
-    def assert_context(tracker):
```
Elkasitu, line 35: I would just merge this commit into commit a3cfe4f3081ec35b1ebc04d916bfc80b373653a2 and rename said commit to something like "Convert tracker context assertion to model validation"

docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement requires_summary in Flaw API (OSIDB-1005)
 - Implement ps_update_stream in Tracker API (OSIDB-1064)
 - Implement daily monitoring email for failed tasks
+- Implement additional tracker validations (OSIDB-787)
```
Elkasitu, line 14: this can go in the same commit that implements said validations

osidb/models.py
osoukup: @Elkasitu having multiple associated affects is a valid situation but then they all must have the same PS module and component which is being checked by `_validate_multi_flaw_tracker`. Therefore we can here simply take the first. 

There are validations on the `Affect` model checking whether it has a valid PS module. However, here it is necessary to be checked too as it is a requirement to be able to create/update a tracker and when doing it the affect validations are not run as the affect is not being changed. So here it is clearly a safeguard before the work is given to the `tracker` app.

osoukup:
@Elkasitu thank you for the review! All should be updated according to your feedback.

Elkasitu (approved):

PR #289, opened 2023-07-31
Move test_ps_module fixture into relevant tests

jobselko:
The `test_ps_module()` fixture with `autouse=True` is used only in `test_validate_affect_ps_module_alerts()` and `test_validate_affect_ps_module_errors()`. This PR moves the fixture's code directly into the tests and deletes the fixture.

### DIFF
--- a/osidb/tests/conftest.py
+++ b/osidb/tests/conftest.py
@@ -27,13 +27,6 @@ def enable_db_access_for_all_tests(db):
     pass
 
 
-@pytest.fixture(autouse=True)
-def test_ps_module():
-    from osidb.tests.factories import PsModuleFactory
-
-    PsModuleFactory(name="rhel-6")
-
-
 @pytest.fixture
 def root_url():
     return "http://osdib-service:8000"
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -831,6 +831,7 @@ class TestFlawValidators:
         ],
     )
     def test_validate_affect_ps_module_alerts(self, bz_id, ps_module, should_alert):
+        PsModuleFactory(name="rhel-6")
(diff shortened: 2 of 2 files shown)


JakubFrejlach (approved):
LGTM. It agree that the factories should be used directly in the tests or there should be some common fixture for the group of tests, but autouse global fixture seems like an overkill for me.

jobselko:
I do not know why flake8 is now complaining about https://www.flake8rules.com/rules/E721.html (`return type(self) != type(other)` in `ComparableTextChoices`). It was fine until now. Should I update it in a new PR and then rebase this one?

osoukup:
> I do not know why flake8 is now complaining about https://www.flake8rules.com/rules/E721.html (`return type(self) != type(other)` in `ComparableTextChoices`). It was fine until now. Should I update it in a new PR and then rebase this one?

I encountered the same in SFM2 today. My impression is that something got updated but in OSIDB we should have the versions pinned. Though I have not investigated further.

osoukup (approved):
LGTM

Elkasitu:
> I do not know why flake8 is now complaining about https://www.flake8rules.com/rules/E721.html (`return type(self) != type(other)` in `ComparableTextChoices`). It was fine until now. Should I update it in a new PR and then rebase this one?

You should be able to change it to `type(x) is type(y)`, the type comparison is valid and the check is overzealous, `isinstance` is not what we want here.

You can also pin the version of flake8 in `tox.ini` but the issue actually comes from `pycodestyle` [1], which I assume gets installed by default with flake8 but am actually unsure as to how to pin it, probably needs to be explicitly pinned in `tox.ini`.

You can do this either in this PR as separate commits or a new one as you prefer.

[1] https://github.com/PyCQA/pycodestyle/pull/1086

PR #291, opened 2023-07-31
Define DB config earlier for django-celery-results, fix INSTALLED_APPS

juspence:
@Elkasitu I found two possible issues, but I'm not sure if fixing them will actually make the monitoring email work:
1. INSTALLED_APPS ordering was different compared to Corgi, and django_celery_beat was not present so all the django_celery_beat.models couldn't be imported.
2. CELERY_RESULT_BACKEND was set to "django-db", but Django's DATABASES setting hadn't been defined yet, so maybe django-celery-results didn't know where to store each TaskResult. I'd expect this to raise an error instead of silently failing, but :shrug:

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -165,7 +165,7 @@
         "filename": "config/settings_ci.py",
         "hashed_secret": "efacc4001e857f7eba4ae781c2932dedf843865e",
         "is_verified": false,
-        "line_number": 44,
+        "line_number": 43,
         "is_secret": false
       },
       {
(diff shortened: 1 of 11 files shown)


Elkasitu (changes requested):
I am not convinced that either of these will change anything, as the task results seem to be working locally without `django-celery-beat`, with the current `INSTALLED_APPS` order and without the `DATABASES` setting being defined in the base `settings.py`file.

This seems to be a recurring error [1][2] and there doesn't seem to be a definitive solution, I don't have time for this right now either so let's leave it for Friday.

[1] https://github.com/celery/django-celery-results/issues/19
[2] https://github.com/celery/django-celery-results/issues/102

Elkasitu:
Closed by #296

PR #292, opened 2023-08-03
Validate MI and CISA MI separately

jobselko:
This PR separates validations for various MI states because they differ. Validations for approved `MI` and approved `CISA MI` were kept and updated, but requested `MI`/`CISA MI` is no longer validated.

Closes OSIDB-1104


osoukup (approved):
Good split, looks much better and clear now!

osidb/tests/test_flaw.py
```diff
@@ -970,6 +970,240 @@ def test_validate_major_incident_state(self, state, should_raise):
...
+        flaw.save()
+
+        FlawReferenceFactory(flaw=flaw, type=article[0], url=article[1])
+        AffectFactory(flaw=flaw)
+        flaw.save()
+
+        if should_alert:
+            assert len(alerts) == len(flaw._alerts)
+            for alert in alerts:
+                assert alert in flaw._alerts
```
jobselko, line 1117: As all test cases contain exactly one alert, I could change them to be of `str` type instead of `list` type, and then test directly `assert alert in flaw._alerts` without the `for` cycle.

Does anyone here agree with this and want me to change it? :D

osoukup: Makes sense to me but not insisting on it. If you want to improve it then great!

PR #295, opened 2023-08-07
refactor jira trackers to query generation

costaconrado:
This PR refactors Jira trackers behaviors making it looks like Bugzilla tracker's behavior.

This PR removes the code that mimics Jira actions and adds a query generation in order to create/edit a Jira tracker.
The intent of those changes is to make easier to pick between Jira or Bugzilla trackers depending on product definition.

Partially covers OSIDB-789.


apps/trackers/jira/query.py
```diff
@@ -1,107 +1,186 @@
...
+                "issuetype": {"name": "Bug"},
+                "project": {"key": self.ps_module.bts_key},
+            }
+        }
+        if self.tracker.external_system_id:
+            self._query["key"] = self.tracker.external_system_id
+
+    def generate_priority(self):
+        """
+        Convert OSIDB impact ot Jira Priority
```
osoukup, line 121: nitpick: `ot` -> `to`

osoukup (approved):
This is great! Thank you for the refactoring! Now we will be able to merge both BTS submodules on `tracker.save` and also reuse some functionality you already have here for the Bugzilla trackers. LGTM

PR #297, opened 2023-08-07
Store task args & kwargs into the DB

juspence:
Quick fix for `None: args=None, kwargs=None` in the monitoring email. That should say something like `task_name: args=(1, "a"), kwargs={2: 2, "b": "b"}` instead.


Elkasitu (approved):
LGTM

PR #294, opened 2023-08-03
tracker integration tests

osoukup:
This PR was aim to contain integration tests of the Bugzilla tracker functionality to conclude it but along the way I uncovered a number of small issues so it also brings various minor fixes. At the end I also added some missing unit tests.

Closes OSIDB-788

PR #298, opened 2023-08-08
tracker BTSs join

osoukup:
This PR introduces a shared base for the Bugzilla and Jira query building and integrates the Jira part to the joint saver class so the `Tracker` model does not have to care for the saver type when `tracker.save` (not yet integrated).

Based on https://github.com/RedHatProductSecurity/osidb/pull/294 (review first)

Closes OSIDB-1178

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -159,6 +159,16 @@
         "is_secret": false
       }
     ],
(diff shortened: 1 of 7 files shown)


costaconrado (approved):
LGTM.

PR #299, opened 2023-08-09
Validate NIST RH CVSS feedback loop

jobselko:
This PR adds validation to check if RH should send a request to NIST on flaw CVSS rescore when defined conditions are met.

Closes OSIDB-334

### DIFF
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -16,6 +16,7 @@ from osidb.models import (
     Flaw,
     FlawAcknowledgment,
     FlawComment,
+    FlawCVSS,
     FlawDraft,
     FlawMeta,
     FlawReference,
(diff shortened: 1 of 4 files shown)


osidb/tests/test_flaw.py
```diff
@@ -795,6 +797,138 @@ def test_cvss_rh_nvd_severity_diff_valid(self, vector_pair):
...
+            version=FlawCVSS.CVSSVersion.VERSION3,
+            issuer=FlawCVSS.CVSSIssuer.REDHAT,
+            vector=rh_cvss,
+            comment=rh_cvss_comment,
+        )
+        PsModuleFactory(
+            name=ps_module, ps_product=PsProductFactory(business_unit="Core RHEL")
+        )
+        AffectFactory(flaw=flaw, ps_module=ps_module)
+        flaw.save(raise_validation_error=False)
```
osoukup, line 924: I wonder why do we have to suppress the validation errors here. The tested validation does not raise so what is it we cannot just setup not by the factories not to raise any exceptions?

osoukup (approved):
LGTM

osidb/tests/test_flaw.py
jobselko: `raise_validation_error=False` should not be here, I will remove it.

PR #300, opened 2023-08-10
add additional index to Tracker model

cebarks:
Adds an additional index (`external_system_id`) to `Tracker` model to speed up main bottlenecking query for OSIDB-1113 (looking up trackers by bug system id).

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2007,7 +2007,7 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
             "uuid",
         )
         unique_together = ["type", "external_system_id"]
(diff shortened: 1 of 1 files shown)


Elkasitu:
Could you provide said query? How do you know that this is the main bottleneck of said query? Adding the index is probably inexpensive but I'm not sure you'll get the expected result

cebarks:
Sure, I'm doing lots and lots of lookups by `external_system_id`: ` session.trackers.retrieve_list(external_system_id=tracker_id, include_fields="created_dt")`

Indexing by this value should drastically speed up lookups when using that as the index. (Assuming I added the index right, never used django/python for API dev before)

Elkasitu:
Superseded by #333 (mostly due to CI not working with forks, but you keep commit authorship)

PR #302, opened 2023-08-10
Validate nist_cvss_validation and cvss_scores

jobselko:
This PR adds validation to check if the combination of `nist_cvss_validation` and `cvss_scores` is set correctly.

Closes OSIDB-1165

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -910,9 +910,6 @@ class Flaw(
         * it has no associated NIST feedback loop in progress (see nist_cvss_validation)
         * it has no RH CVSS3 explanation comment
         """
-        if self.cvss_scores.filter(version=FlawCVSS.CVSSVersion.VERSION3).count() != 2:
-            return
-
         nist_cvss = self.cvss_scores.filter(
             issuer=FlawCVSS.CVSSIssuer.NIST,
             version=FlawCVSS.CVSSVersion.VERSION3,
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -940,6 +940,27 @@ class Flaw(
                 f"an explanation comment for RH CVSSv3 score.",
             )
 
+    def _validate_cvss_scores_and_nist_cvss_validation(self):
+        """
+        Checks that if nist_cvss_validation is set, then both NIST CVSSv3 and RH CVSSv3
+        scores need to be present.
+        """
+        nist_cvss = self.cvss_scores.filter(
+            issuer=FlawCVSS.CVSSIssuer.NIST,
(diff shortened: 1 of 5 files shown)


osidb/models.py
jobselko, line 1: Should I add the same validation into the `FlawCVSS` class?

osidb/models.py
```diff
@@ -940,6 +940,21 @@ def _validate_nist_rh_cvss_feedback_loop(self):
                 f"an explanation comment for RH CVSSv3 score.",
             )
 
+    def _validate_cvss_scores_and_nist_cvss_validation(self):
+        """
+        Checks that if nist_cvss_validation is set, then both NIST CVSSv3 and RH CVSSv3
+        scores need to be present.
+        """
+        if (
```
jobselko, line 948: Should I add the same validation into the `FlawCVSS` class?

osidb/models.py
```diff
@@ -940,6 +940,21 @@ def _validate_nist_rh_cvss_feedback_loop(self):
                 f"an explanation comment for RH CVSSv3 score.",
             )
 
+    def _validate_cvss_scores_and_nist_cvss_validation(self):
```
jobselko, line 943: Should I add the same validation into the `FlawCVSS` class?

osidb/models.py
```diff
@@ -940,6 +940,21 @@ def _validate_nist_rh_cvss_feedback_loop(self):
...
+        """
+        Checks that if nist_cvss_validation is set, then both NIST CVSSv3 and RH CVSSv3
+        scores need to be present.
+        """
+        if (
+            self.nist_cvss_validation
+            and self.cvss_scores.filter(version=CVSS.CVSSVersion.VERSION3).count() != 2
```
osoukup, line 950: This will work for now but if we ever introduce another issuer it will break. I am not sure we will ever do it but I would make this specific to NIST and RH anyway or at least make there a note that it must be updated if we ever have a third issuer.

osidb/models.py
osoukup: Good question. I think that this validation could be typically triggered by assigning a flag to a flaw when there should be none and not by assigning a CVSS to it so I think this direction is enough. The other question is getting rid of this error as if one simply adds a CVSS it does not trigger `flaw.save` so the alert would stay there. I would suggest to make the alert blocking. There is no reason why we should allow an invalid flag. Then there will also be no dangling alert.

osoukup (changes requested):
I am for making this alert blocking.

osoukup (approved):
LGTM

PR #304, opened 2023-08-16
tracker description

osoukup:
This PR implements tracker description generation. There are still two missing pieces which are description generation based on the product definitions template as those templates are not yet in OSIDB (tracker in OSIDB-1191) and SLA description text as SLA computation is not yet implemented (tracked in OSIDB-96).

You may notice that the description composition is somewhat random - eg. Jira header goes before the triage text while the Bugzilla header goes afterwards etc. I do not think there is any explanation above that this is how it was generated in SFM2. The SFM2 code corresponding to this functionality is one big mess and a forest of if-else branching. I tried to decompose it a bit but there are limits resulting from that it was all stitched together over many years by many people.

Closes OSIDB-1173

### DIFF
--- a/apps/trackers/bugzilla/query.py
+++ b/apps/trackers/bugzilla/query.py
@@ -128,10 +128,10 @@ class TrackerBugzillaQueryBuilder(BugzillaQueryBuilder, TrackerQueryBuilder):
 
     def generate_description(self):
         """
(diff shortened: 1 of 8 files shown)


costaconrado (approved):
LGTM.

PR #307, opened 2023-08-17
API for Flaw's bugzilla fixed_in field

jsvob:
- Add API for manipulating the `fixed_in` bugzilla field through `/flaw/<flaw_id>/package_versions` endpoints.
- Add bbsync and update bzimport to allow the endpoints to work.
- Revise models used to store the data.
  - `CVEv5PackageVersions` and `CVEv5Version` models use polymorphism, but the features are not used in OSIDB. Operations are slower than on non-polymorphic models due to joins performed in the background. These models are deprecated by this PR and left untouched in case we need to rerun the migration in existing databases (until we decide to delete them). New models are `Package` and `PackageVer`.
  - The new models' names are chosen so that they are acceptable and we don't have to rename them later. It seems to me that renaming the models might turn out to be surprisingly complex (I tried and got too deep into a few rabbit holes), so better avoid it.
- The changes are so mutually-dependent that splitting to multiple PRs would be problematic and time-consuming, IMO.

Related to OSIDB-1066.


jasinner:
>       * `CVEv5PackageVersions` and `CVEv5Version` models use polymorphism, but the features are not used in OSIDB. Operations are slower than on non-polymorphic models due to joins performed in the background. These models are deprecated by this PR and left untouched in case we need to rerun the migration in existing databases (until we decide to delete them). New models are `CVEv5Package` and `CVEv5Ver`.
>       * The new models' names are chosen so that they are acceptable and we don't have to rename them later. It seems to me that renaming the models might turn out to be surprisingly complex (I tried and got too deep into a few rabbit holes), so better avoid it.
>       * `CVEv5Ver` are connected to `CVEv5Package` using many-to-many, so that one `CVEv5Ver` can be used by multiple unrelated or seemingly unrelated packages and/or flaws. The reason is that once OSIDB moves away from bugzilla, we might need to store substantially more data and types of the CVEv5 schema in these models and if a version happens to be shared across multiple packages, we'd end up storing duplicated data.
>         
>         * **TODO / open question: Is that true?**

To be honest on review of the many-to-many structure I think we probably don't need the many-to-many relationship here. I probably did it just to normalize the database, but I no longer think it makes sense to do so. In Component Registry (which this data will be compared to), we store both the package name and version in the same 'Component' table.

>         * **TODO / open question: Is the new structure appropriate for the envisioned future use / for enabling painless extensibility?** (Is it still too complicated? Or is it too simple?)

I originally designed it the way you have it now, I personally don't think it's too simple. @JimFuller-RedHat suggested I create the classes in this fashion to try and future proof the code in case a type of version tracking becomes popular.

>         * **TODO / open question: Is polymorphism unnecessary, or do I miss something?**

It's not necessary today, but could become useful in future if a new way of tracking versions becomes popular.

JimFuller-RedHat:
> I originally designed it the way you have it now, I personally don't think it's too simple. @JimFuller-RedHat suggested I create the classes in this fashion to try and future proof the code in case a type of version tracking becomes popular.

or a new rev of existing version tracking ;)

osoukup (commented):
So far LGTM. I went carefully through about the first half of the commits and less carefully through the other half after the status turned to Draft :smiley:

jsvob:
Thank you for the comments, @jasinner and @JimFuller-RedHat (and @osoukup in person)!

- Based on that, I'll rework the PR so that there's only 1 new model that stores both package names and versions in a single model. If it turns out we need a new way of tracking versions, it might be more expedient to enhance and/or migrate from a simple model than from comparatively more complex structure.
- I propose to leave the structure of the API the same as in this first PR draft, so that it's possible to enhance it in the future in a way compatible with the [CVE 5.0 schema](https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/CVE_JSON_5.0_schema.json), maintaining a portion of the spirit of the original work and of the first PR draft as discussed today and represented by soon-to-be removed commit b7494548579a994ea21720b9f5dd0b39457cfc0a ([models](https://github.com/RedHatProductSecurity/osidb/blob/b7494548579a994ea21720b9f5dd0b39457cfc0a/osidb/models.py#L2884), [viewset](https://github.com/RedHatProductSecurity/osidb/blob/b7494548579a994ea21720b9f5dd0b39457cfc0a/osidb/api_views.py#L568), [serializer](https://github.com/RedHatProductSecurity/osidb/blob/b7494548579a994ea21720b9f5dd0b39457cfc0a/osidb/serializer.py#L859)). (Already discussed with @osoukup , comments from others welcome @RedHatProductSecurity/osidb-devs )
    ```
    {
      "package": "string",
      "versions": [
        {
          "version": "string",
          "embargoed": true
        }
      ],
      "flaw": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "uuid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
      "embargoed": true,
      "created_dt": "2023-08-23T08:30:28.581Z",
      "updated_dt": "2023-08-23T08:30:28.581Z",
    }
    ```
- I propose that the new model be named simply `Package` and that `VersionStatus` is not used in the new model. For now, all versions would only be of the single type "unaffected" anyway. It can be added later if we need it. The new model can contain the following comment:
    ```
    The model's structure allows future extensibility with features from the CVE 5.0 schema.
    Currently, it tracks only versions corresponding to bugzilla fixed_in field, which correspond to
    CVE 5.0's `"status": "unaffected"`.
    References:
    - https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/docs/versions.md
    - https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/CVE_JSON_5.0_schema.json
    ```

jsvob:
The `technical` label is only temporary while the PR is a Draft, so that I can see CI results without the repeated CI shutdowns due to conflicts on CHANGELOG.md.

jsvob:
I [proposed](https://github.com/RedHatProductSecurity/osidb/pull/307#issuecomment-1689549002) to create only 1 new model, but ended up creating 2. The reasons are:

- Nested dicts for versions are necessary to keep the structure similar to the CVE 5.0 schema. That in turn allows future extensibility without too much API change and associated pain.
- Using the models for the nested version dicts is straightforward and enables automatic openapi.yml generation. Implementing the dicts with just 1 model would require more complex (de)serializers and potentially complicated manual schema declaration.

Both points briefly discussed with @osoukup in person last week.

osidb/models.py
```diff
@@ -2919,6 +2933,45 @@ def validate(self, *args, **kwargs):
...
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    flaw = models.ForeignKey(
+        Flaw,
+        related_name="package_versions",
+        on_delete=models.CASCADE,
+    )
+
+    package = models.CharField(max_length=2058)
```
osoukup, line 2982: This number caught my eye. If it was `2048` I would not ask but why `2058`? Also those numbers are something tied to the CVEv5 or our estimation of the worst case?

apps/bbsync/query.py
```diff
@@ -389,3 +389,85 @@ def generate_comment(self):
                 "body": pending_comments.first().text,
                 "is_private": False,
             }
+
+    # TODO: test this
```
osoukup, line 393: When do you plan to test this? I would be for adding at least some simple test together with this while you still remember how is this functionality supposed to work :smile:

osidb/models.py
```diff
@@ -2919,6 +2933,83 @@ def validate(self, *args, **kwargs):
...
+
+    objects = PackageManager()
+
+    # TODO: test this
```
osoukup, line 2986: One simple integration test would be nice.

osoukup (approved):
LGTM

osidb/models.py
```diff
@@ -2960,7 +2960,16 @@ def create_package(flaw, package, **extra_fields):
...
-    """Model representing a package with connected versions."""
+    """
+    Model representing a package with connected versions.
+
+    The model's structure allows future extensibility with features from the CVE 5.0 schema.
+    Currently, it tracks only versions corresponding to bugzilla fixed_in field, which correspond to
+    CVE 5.0's `"status": "unaffected"`.
+    References:
+    - https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/docs/versions.md
+    - https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/CVE_JSON_5.0_schema.json
```
Elkasitu, line 2971: Let's not base the model on something that we don't use, adhere to fully or can otherwise be arranged at the representation layer

Elkasitu (commented):
Some of my remarks may not fully apply as I review on a commit-by-commit basis, I assume some of these will be squashed (e.g. removing code added in commit n-1).

I would like to know or have documented somewhere (like in the commit message) why the refactoring of the PackageVersions and related models happened in the first place, I understand that it makes sync to bugzilla easier but I do not see why exactly.

I am not against the refactoring but I feel like if we're going to refactor something we might as well look at it from all angles.

osidb/models.py
jsvob: @jasinner Do you remember why it was `2058` in the original models, please? I just copied the value. The [schema](https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/CVE_JSON_5.0_schema.json) specifies `2048`.

apps/bbsync/query.py
jsvob: This must have slipped through during the rewriting of the commits. Definitely shouldn't be there. It was there as a reminder because the associated tests belonged to a different commit due to the size of the changes. I'll review the old commits and make sure that the accompanying test is present in the PR.

osidb/models.py
```diff
@@ -2896,14 +2906,18 @@ class PackageVersions(PolymorphicModel):
...
     def validate(self, *args, **kwargs):
         """validate package versions model"""
         super().clean_fields(*args, **kwargs)
 
 
 class CVEv5PackageVersions(PackageVersions):
+    """
+    This model is deprecated. Use Package instead. Delete when it's clear
+    that Package works correctly.
+    """
```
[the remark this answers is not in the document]
jsvob: Failing forwards might imply manual resync, right?

The old models ale left there temporarily. I guess we can remove them in a few weeks after a demo. This is so that we still have the old correct data in case there's a bug or unintended interaction with other changes that's discovered only in stage or production (as was the case a few weeks ago, discovered on the day of a demo). Once the old models are deleted, we can't revert the data migration from old to new models. Since fixed_in is in tens of thousands of flaws, a manual resync is too lengthy an endeavor. I propose to let the old models in there until the new API is demoed (performed read&write on stage and read-only on production) and then delete them in a separate PR.

osidb/models.py
```diff
@@ -2919,6 +2933,45 @@ def validate(self, *args, **kwargs):
...
+    )
+
+    class Meta:
+        verbose_name = "Version"
+
+    version = models.CharField(max_length=1024)
+
+    # TODO: This is temporarily present to make existing tests happy. Reconsider whether to delete.
+    def validate(self, *args, **kwargs):
+        super().clean_fields(*args, **kwargs)
```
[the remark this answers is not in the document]
jsvob: `validate` or `version`?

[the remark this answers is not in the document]
jobselko: I think he thought the whole `PackageVer` model.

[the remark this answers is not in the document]
jobselko: I do not see a reason for two models for a package and its version. I mean, the package without the version seems to be incomplete. I think this is a bad design.

osidb/models.py
jobselko: I agree with Adrian, and I would prefer to discuss it on the team call. Even if we seem to be compatible with the new schema now, something may happen in the future that will force new changes/redesign again.

osoukup: Would it still be compatible with the existing API or would it lead to a breaking changes? If the latter we should communicate it with the SDEngine guys. The naming change was done by @jasinner so he could comment on the reasoning and which one makes more sense.

osidb/models.py
```diff
@@ -2919,6 +2933,45 @@ def validate(self, *args, **kwargs):
...
+        Flaw,
+        related_name="package_versions",
+        on_delete=models.CASCADE,
+    )
+
+    package = models.CharField(max_length=2058)
+
+    # TODO: This is temporarily present to make existing tests happy. Reconsider whether to delete.
+    def validate(self, *args, **kwargs):
+        super().clean_fields(*args, **kwargs)
```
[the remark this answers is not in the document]
jsvob: Yes, it only says this. Since it is refactored away in follow-up commits, I don't see value in writing a fully-accurate explanation.

The short is that old models had `.validate()` and old tests ran the validation.

I made gradual commits/changes so as to make sure I didn't introduce a bug into both new models and new tests at the same time. So this is from a state where old tests (with changes as small as possible) ran against the new models.

osidb/models.py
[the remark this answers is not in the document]
jsvob: `PackageVer` exists so that serialization into the CVE 5.0-compatible structure is straightforward. Having ArrayField and the current API structure would require custom de/serializers and bending/duplicating the custom OSIDB mixins for serializers. If we end up deciding to drop CVE 5.0 compatibility, we can use ArrayField with only 1 model, just like the other models do.

osidb/models.py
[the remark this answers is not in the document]
jsvob: Also it seems this is from a state where the custom osidb mixins were not introduced to the new models yet. They have their own `validate()` IIRC.

osidb/models.py
jsvob: I guess we can do anything we want and just implement custom logic for the serializer behind [Flaw's `package_versions` read-only field (hyperlink to current master)](https://github.com/RedHatProductSecurity/osidb/blob/9533bb350265a73ca5d7c6bb5e5e70293ac91969/osidb/serializer.py#L808) to maintain compatibility with **existing** API and existing users.

As for the **_new_** API _added_ in this PR, that's not used by anyone else yet (obviously), so there (`/api/v1/flaws/CVE-xxx-xxx/package_versions`) we can do whatever we want and name it whatever we want (perhaps `/api/v1/flaws/CVE-xxx-xxx/fixed_in`?).

As to what's feasible and not overly complex to design and maintain, I propose choosing one of these two paths:
- Maintain CVE 5.0 compatibility and maintain current model structure, so that we can use standard serializer mixins, standard schema autogeneration, standard viewset and as small serializer changes as possible.
- Drop CVE 5.0 compatibility, make it into anything that makes sense, and have the default the API structure that results from the standard model structure (as used in other osidb models), standard osidb mixins, standard viewsets, standard auto-generated schema, etc, just as with any other OSIDB model.

I specifically advise to avoid trying to have one seemingly clear model structure and another seemingly clear API structure, because in between would be bespoke unmaintainable glue.

apps/bbsync/query.py
jsvob: Fixed in commit [1fa690c](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/1fa690cba505717ebc1353e8406395ea08d4a147)

osidb/models.py
jsvob: Fixed in commit [1fa690c](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/1fa690cba505717ebc1353e8406395ea08d4a147)

jsvob:
> Some of my remarks may not fully apply as I review on a commit-by-commit basis, I assume some of these will be squashed (e.g. removing code added in commit n-1).

Some will be squashed (esp. the newest commits), but I didn't really plan to squash those commits that were there at the time of submission for review. The reason is that IMO squashing some of the commits would make the changes large. At the end of the day, I don't have a preference. Please state your preferences (squash some, squash all?).

> I would like to know or have documented somewhere (like in the commit message) why the refactoring of the PackageVersions and related models happened in the first place, I understand that it makes sync to bugzilla easier but I do not see why exactly.

The commits have quite detailed messages. Do these answer your concerns?

```
Partially switch from using polymorphic models to non-polymorphic.
- This commit adds non-polymorphic counterparts to the polymorphic
  models.
- BZImport is switched to use the new non-polymorphic models.
- The original models are left untouched.
- Data NOT migrated in this commit.

Polymorphism is not needed and complicates further work and performance.
```

...

```
New API for Flaw package_versions.
- Endpoints under /flaw/<flaw_id>/package_versions.
- Standard set of URL path, api view, filter set.
- For now, these correspond to bugzilla's fixed_in field.
- The serializer performs nested serialization. The models
  Package and PackageVer are compatible with the structure
  of the CVE 5.0 schema, allowing potential expansion beyond
  fixed_in.
```

...

```
Migrate old models to new models.
- CVEv5PackageVersions and CVEv5Version models are old, contain valid
  data, are used as data sources for the migration, and are kept
  untouched by the migration.
- Package and PackageVer models are new, with simplified structure,
  added ACLs and other mixins, and are already used by revised bzimport
  and bbsync in closely-preceding commits. Data are migrated to these
  models.
```

osidb/serializer.py
```diff
@@ -471,16 +474,52 @@ def __get_bz_api_key(self):
...
+        the database.
+        """
+        try:
+            instance.save(bz_api_key=self.__get_bz_api_key())
+        except DataInconsistencyException as e:
+            # translate internal exception into Django serializable
+            raise BadRequest(
+                "Received model contains non-refreshed and outdated data! "
+                "It has been probably edited by someone else in the meantime"
+            ) from e
```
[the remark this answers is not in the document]
jsvob: Such a change would make `test_flaw_update_collision` fail with 

```
>       assert response.status_code == 400
E       assert 409 == 400
```

Perhaps should be done as a separate PR with its own analysis of the ramifications?

osidb/models.py
```diff
@@ -2933,7 +2933,33 @@ def validate(self, *args, **kwargs):
...
+        except ObjectDoesNotExist:
+            # The instance must be saved into the database, so that it's possible
+            # to connect it with PackageVer, but it must otherwise be as clean
+            # as an in-memory unsaved instance created with plain
+            # Package(flaw=flaw, package=package, **extra_fields). Calling
+            # objects.create() would involve osidb's custom logic that would
+            # interfere prematurely.
+            return Package._base_manager.create(
+                flaw=flaw, package=package, **extra_fields
+            )
```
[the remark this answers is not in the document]
jsvob: Whatever makes save fail on validation, IIRC. So yes, it does make sense to do it similarly to `TrackerManager`. I'll do that. Thanks!

osidb/models.py
osoukup: I agree. One or another. The first one is I think what SDEngine currently expects. What is the advantage of the second one? Simpler model I guess. That would however require SDEngine functionality rewrite, correct?

jsvob: > would however require SDEngine functionality rewrite

I don't think so. SDEngine uses only whatever was there before this PR. Which is the read-only field directly on Flaw. And this is the only serializer where manually creating an arbitrary data structure is without negative effects on maintainability. So we can go with the second option and still have this field return the structure expected by SDEngine.

osoukup:
Let us go with the model and resulting serializer structure currently introduced and move on.

osidb/models.py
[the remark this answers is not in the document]
jsvob: Changed in commit 66cd1c5118510b4105f23d0209adc64a840920be

Creating and then saving basically eliminates running `TrackingMixinManager.create()` that seems to be the "custom logic" I originally referred to.

jsvob:
I discovered a bug: PUT (update) requires `updated_dt` in the request, but the value is ignored. I can't figure out why `updated_dt` is checked for e.g. acknowledgments but not for this. Obviously, the culprit must be somewhere in the new code, but that still leaves me with debugging the long tangle of nested save() and bzsync() calls throughout the various classes in the inheritance tree.

osidb/models.py
jsvob: Closed as per https://github.com/RedHatProductSecurity/osidb/pull/307#issuecomment-1713798433

osidb/models.py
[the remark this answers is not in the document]
jsvob: Closed as per https://github.com/RedHatProductSecurity/osidb/pull/307#issuecomment-1713798433

osoukup:
> I discovered a bug: PUT (update) requires `updated_dt` in the request, but the value is ignored. I can't figure out why `updated_dt` is checked for e.g. acknowledgments but not for this. Obviously, the culprit must be somewhere in the new code, but that still leaves me with debugging the long tangle of nested save() and bzsync() calls throughout the various classes in the inheritance tree.

@jsvob not that I would see what is the issue but also run into a similar one a few times and usually this is caused by a bad order of the inheritance. Unfortunately for some of the mixins which implement `save` or `create`/`update` the order matters. It is not ideal but probably hard to fix as specifically the Bugzilla sync disrupts/changes the regular call order.

Elkasitu:
> > I would like to know or have documented somewhere (like in the commit message) why the refactoring of the PackageVersions and related models happened in the first place, I understand that it makes sync to bugzilla easier but I do not see why exactly.
> 
> The commits have quite detailed messages. Do these answer your concerns?
> 
> ```
> Partially switch from using polymorphic models to non-polymorphic.
> - This commit adds non-polymorphic counterparts to the polymorphic
>   models.
> - BZImport is switched to use the new non-polymorphic models.
> - The original models are left untouched.
> - Data NOT migrated in this commit.
> 
> Polymorphism is not needed and complicates further work and performance.
> ```

I read them but what I would like to know, specifically, is how or in what way does Polymorphism affect the sync to Bugzilla? Your answer won't change the outcome as the team has agreed to go with your implementation, but I'd like to have it documented in case we want to change this in the future.

jsvob:
> in what way does Polymorphism affect the sync to Bugzilla?

It doesn't. I didn't state that anywhere. That was your idea originating in [your comment](https://github.com/RedHatProductSecurity/osidb/pull/307#pullrequestreview-1612952913).

> why the refactoring of the PackageVersions and related models happened in the first place

Please see the [comment at the top](https://github.com/RedHatProductSecurity/osidb/pull/307#issue-1855463280).

> CVEv5PackageVersions and CVEv5Version models use polymorphism, but the features are not used in OSIDB. Operations are slower than on non-polymorphic models due to joins performed in the background.

The message of commit 5cf938968f1737d20d2fdd2597155bd2b8dd2eb5 says

> Polymorphism is not needed and complicates further work and performance.

This is IMO both self-evident and self-explanatory. Please provide a better explanation if this is not what you expect.

osidb/models.py
```diff
@@ -2882,12 +2887,17 @@ class CVEv5Version(Version):
...
+    This model is deprecated. Use Package instead. Delete when it's clear
+    that Package works correctly.
+    """
+
     # internal primary key
     uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
 
     flaw = models.ForeignKey(
         Flaw,
-        related_name="package_versions",
```
jobselko, line 2890: I am not sure about the future commits structure, but you should ensure that this change is done together (i.e in one commit) with creating the `Package` model and its `flaw = models.ForeignKey(Flaw, related_name="package_versions", on_delete=models.CASCADE,)`, so the `package_versions` does not disappear from the `Flaw` model.

osidb/models.py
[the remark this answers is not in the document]
jobselko: You added `meta_attr["fixed_in"] = self.flaw_bug["fixed_in"]` in BZImport in this PR, and this solves the data backup issue. Each flaw will have its original value here, so you can delete the old unused models. You just need to do the migrations for the change in `meta_attr`.

Also, I am not sure if we can delete model's fields without deprecating them first via `deprecate_field_custom` to keep N-1 compatibility. Or can we do it if these fields are unused in serializers/APIs? WDYT, @Elkasitu?

osidb/models.py
```diff
@@ -2919,6 +2933,76 @@ def validate(self, *args, **kwargs):
...
+        try:
+            package = Package.objects.get(flaw=flaw, package=package)
+            for attr, value in extra_fields.items():
+                setattr(package, attr, value)
+            return package
+
+        except ObjectDoesNotExist:
+            package = Package(flaw=flaw, package=package, **extra_fields)
+            package.save()
+            return package
```
jobselko, line 2953: You can put `return package` at the end of the method.

osidb/models.py
jobselko: I think this is just a typo.

jobselko: Would not be `name` or `pkg_name` better than `package`?

osidb/models.py
```diff
@@ -2919,6 +2933,76 @@ def validate(self, *args, **kwargs):
...
+    def bzsync(self, *args, bz_api_key, **kwargs):
+        """
+        Bugzilla sync of the Package instance and linked PackageVer instances.
+        """
+        self.save()
+        # needs to be synced through flaw
+        self.flaw.save(*args, bz_api_key=bz_api_key, **kwargs)
+
+
+class PackageVer(models.Model):
```
jobselko, line 2989: What about `PackageVersion`?

osidb/tests/test_package_versions.py
```diff
@@ -11,30 +11,29 @@ class TestPackageVersions(object):
     def test_create_cve_v5_version(self):
         """test raw verison range creation"""
 
-        vr_1 = CVEv5Version.objects.create(
-            version="3.2.1",
+        flaw = FlawFactory()
+        pkg = PackageFactory(
+            acl_read=flaw.acl_read,
+            acl_write=flaw.acl_write,
+            flaw=flaw,  # make _validate_acl_identical_to_parent_flaw happy
```
jobselko, line 18: `pkg = PackageFactory(flaw=flaw)` should be sufficient.

osidb/tests/test_package_versions.py
```diff
@@ -11,30 +11,29 @@ class TestPackageVersions(object):
...
 
-        assert vr_2.validate() is None
-        assert vr_2.save() is None
+        assert pkg.validate() is None
+        assert pkg.save() is None
 
     def test_create_cve_v5_package_versions(self):
 
-        vrs_1 = CVEv5PackageVersions()
+        vrs_1 = Package()
```
jobselko, line 31: This is most likely a useless test. A model without required fields will always raise an exception.

osidb/tests/test_package_versions.py
```diff
@@ -11,30 +11,29 @@ class TestPackageVersions(object):
...
             vrs_1.validate()
 
-        vr_1 = CVEv5VersionFactory()
-
-        vrs_2 = CVEv5PackageVersionsFactory.create(versions=(vr_1,))
+        vrs_2 = PackageFactory()
+        PackageVerFactory(package=vrs_2)
 
         assert vrs_2.validate() is None
         assert vrs_2.save() is None
```
jobselko, line 39: The original `test_create_cve_v5_version` can be removed as it tested `status`, which is no longer here. So we will end up with only one test, which could be moved to `test_flaw` (there is no need to have a separate file for it).

collectors/bzimport/convertors.py
```diff
@@ -750,6 +750,7 @@ def get_meta_attr(self, cve_id):
         meta_attr["original_srtnotes"] = self.flaw_bug["cf_srtnotes"]
         meta_attr["status"] = self.flaw_bug["status"]
         meta_attr["resolution"] = self.flaw_bug["resolution"]
+        meta_attr["fixed_in"] = self.flaw_bug["fixed_in"]
```
jobselko, line 753: Make migrations for this to keep the original data.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -37,6 +36,17 @@ def get_acls(self):
             )
         ]
 
+    def get_acls_write(self):
```
jobselko, line 39: Why do we need different acls?

collectors/bzimport/tests/test_convertors.py
```diff
@@ -1612,6 +1619,32 @@ def test_summary_meta_attr(self):
         assert "bz_summary" in flaw.meta_attr
         assert flaw.meta_attr["bz_summary"] == flaw_bug["summary"]
 
+    def test_fixed_in_meta_attr(self):
+        """
+        Test that fixed_in bz field is saved as-is in Flaw.meta_attr
+
+        The model structure of CVEv5Package and CVEv5Ver doesn't store the
```
jobselko, line 1626: There should be probably `Package` and `PackageVer`.

collectors/bzimport/tests/test_convertors.py
```diff
@@ -1612,6 +1619,32 @@ def test_summary_meta_attr(self):
...
+        flaw_bug["fixed_in"] = "foobar 1.2, foobaz-1.2"
+        fbc = FlawConvertor(
+            flaw_bug,
+            [],
+            None,
+            None,
+            [],
+            [],
+        )
+        [flaw] = fbc.bug2flaws()
```
jobselko, line 1641: What does the `[flaw]` mean?

collectors/bzimport/tests/test_convertors.py
```diff
@@ -258,16 +269,16 @@ def test_basic(self):
         reference = FlawReference.objects.first()
         cvss_score = FlawCVSS.objects.first()
         tracker = Tracker.objects.first()
-        package_versions = CVEv5PackageVersions.objects.first()
-        version = CVEv5Version.objects.first()
+        package_versions = Package.objects.first()
+        version = PackageVer.objects.first()
```
jobselko, line 273: Just a recommendation, but I would use better names, e.g. `package` for `Package` and `package_version` for `PackageVer`. The old names still implies the old models.

apps/bbsync/query.py
```diff
@@ -389,3 +389,84 @@ def generate_comment(self):
...
+
+    def generate_fixed_in(self):
+        """
+        Generates bugzilla query value for the bugzilla field fixed_in based
+        on the hierarchy of Flaw-associated models Package and PackageVer.
+
+        Except for flaws that have been stored in the database for a long time
+        and haven't been refreshed by bzimport since introducing this function
+        (and the associated save_packageversions in convertors.py). For such,
+        the bugzilla query for the flag is silently not generated.
```
jobselko, line 401: I am unsure if I understand this. My current understanding is that if BZImport is run for a flaw, then both `flaw.packages_versions` and `flaw.meta_attr["fixed_in"]` for that specific flaw should be updated (and there will be also migrations for all flaws).
Is this equal to what is written in your comment?

apps/bbsync/query.py
```diff
@@ -389,3 +389,84 @@ def generate_comment(self):
...
+            #   nontrivial, so intelligently deciding when to raise such an exception would also be
+            #   nontrivial.
+            # * The most proper way would be to bzimport the Flaw here and populate meta_attr.
+            #   However, that is potentially complicated and breaks decomposition for a probably
+            #   miniscule benefit.
+            # * Any Flaw bbsync triggers a follow-up bzimport that will load meta_attr["fixed_in"].
+            #   We can expect that if an old Flaw gets updated, fixed_in will not be the first
+            #   edit operation.
+            #
+            return
```
jobselko, line 432: I think this situation cannot happen. You added `meta_attr["fixed_in"] = self.flaw_bug["fixed_in"]` in BZImport, and even if `fixed_in` will be empty in BZ, `meta_attr` in OSIDB will contain `'fixed_in': ''`.

apps/bbsync/query.py
```diff
@@ -389,3 +389,84 @@ def generate_comment(self):
...
+        result = []
+        # Preserve order of existing versions in fixed_in, but remove those that were deleted.
+        for pkg, separator, ver in original_fixed_in:
+            if (pkg, ver) in proposed_fixed_in:
+                result.append(f"{pkg}{separator}{ver}")
+                proposed_fixed_in.remove((pkg, ver))
+        # And also add newly added versions to fixed_in.
+        result.extend(sorted([f"{pkg} {ver}" for (pkg, ver) in proposed_fixed_in]))
+
+        self._query["cf_fixed_in"] = ", ".join(result)
```
jobselko, line 472: Do I understand it correctly that all the original packages are removed, and only the new ones from `flaw.package_versions` are kept? If so, this could be simplified a lot.

apps/bbsync/tests/test_query.py
```diff
@@ -1252,3 +1254,58 @@ def test_generate_nist_cvss_validation(
             assert {"name": "nist_cvss_validation", "status": flag_value} in flags
         else:
             assert flags == []
+
+
+class TestGenerateFields:
```
jobselko, line 1259: What about `TestGenerateFixedIn`?

apps/bbsync/tests/test_query.py
```diff
@@ -1252,3 +1254,58 @@ def test_generate_nist_cvss_validation(
...
+        PackageVerFactory(version="3.4.5.6", package=pkg_c)
+
+        fbqb = FlawBugzillaQueryBuilder(flaw)
+        fixed_in = fbqb.query.get("cf_fixed_in")
+
+        # Test that
+        # - Versions already in fixed_in are left in fixed_in as they are, including order and
+        #   dash/space separator.
+        # - Versions no longer existing are removed from fixed_in.
+        # - New versions are added to fixed_in with space separator.
```
jobselko, line 1306: Expected behaviours should be in the docstring.

osidb/filters.py
jobselko, line 1: Do we want to also add these filters on the flaw endpoint?

jobselko (changes requested):
So far, I have not finished the review of `serializer` and `test_endpoints`, so this is a partial review for now.

jsvob:
> > I discovered a bug: PUT (update) requires `updated_dt` in the request, but the value is ignored. I can't figure out why `updated_dt` is checked for e.g. acknowledgments but not for this. Obviously, the culprit must be somewhere in the new code, but that still leaves me with debugging the long tangle of nested save() and bzsync() calls throughout the various classes in the inheritance tree.
> 
> @jsvob not that I would see what is the issue but also run into a similar one a few times and usually this is caused by a bad order of the inheritance. Unfortunately for some of the mixins which implement `save` or `create`/`update` the order matters. It is not ideal but probably hard to fix as specifically the Bugzilla sync disrupts/changes the regular call order.

Found the bug. `BugzillaSyncMixinSerializer.update()` iterates over `validated_data` and sets all fields from the client's request onto the in-memory instance representation, including `updated_dt`. All existing models except the new `Package` model use that mixin. That then enables `TrackingMixin.save()` to compare the instance's updated_dt with the in-database updated_dt (implicitly assuming that the instance has client-provided updated_dt). In `FlawPackageVersionSerializer`, the in-memory instance's `updated_dt` is never set to the value from the PUT request, so `TrackingMixin.save()` always compares the in-database value with the in-database value.

Tomorrow I'll think about how to fix `BugzillaBareSyncMixinSerializer` in such a way that simultaneously 
- is simple and adds as little additional code to the PR as possible
- still allows custom deserializer logic, such as `FlawPackageVersionSerializerMixin`
- generically works with existing and future mixins
- is obvious enough so that nobody spends a day hunting for this again when using `BugzillaBareSyncMixinSerializer` instead of `BugzillaSyncMixinSerializer`

collectors/bzimport/convertors.py
jsvob: How? This doesn't exist in the database yet. This information is currently thrown out in osidb master.

apps/bbsync/query.py
jsvob: Why can't it happen? The comment is detailed and explains what can happen, so I'd expect you to provide similarly detailed reasoning why it's wrong.

`meta_attr["fixed_in"] = self.flaw_bug["fixed_in"]` in bzimport saves the verbatim bugzilla `fixed_in` field for _newly bzimported flaws_. It doesn't add `meta_attr["fixed_in"]` to existing flaws in the database.

> and even if fixed_in will be empty in BZ, meta_attr in OSIDB will contain 'fixed_in': ''.

- Empty string value for an existing dictionary key is substantially different than a nonexistent dictionary key.
- "Detecting API-performed changes without flaw.meta_attr["fixed_in"] would be nontrivial(...)"
- We can't just create a fake empty string out of nothing and send that to bugzilla if we don't have `meta_attr["fixed_in"]`. That would destroy data in bugzilla.

apps/bbsync/query.py
jsvob: Can you rephrase the question please? I don't understand. The comment in the code specifically says `Preserve order of existing versions in fixed_in` and the code does that.

apps/bbsync/query.py
jsvob: The code/docstring you are commenting on is a part of bbsync, but you describe bzimport.

Your understanding of bzimport is correct.

Bbsync can be run on a flaw that has been sitting in the database so far untouched from times before this PR got merged and deployed. Such a flaw wouldn't have `flaw.meta_attr["fixed_in"]`. The docstring explains that.

osidb/tests/test_package_versions.py
jsvob: Good point. I'd prefer to add a commit (that won't be squashed at merge) that cleans up the tests. The reason is that the old tests were valuable during the transition from old models to new models to prevent regressions. Some additional context also in comment https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1320131359

osidb/tests/test_package_versions.py
jsvob: Similar to discussion https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1323017622

osidb/models.py
jsvob: That model name is already used. Renaming models is complicated. Additional context [in the first PR comment](https://github.com/RedHatProductSecurity/osidb/pull/307#issue-1855463280) and in [this comment](https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1319485140).

osidb/models.py
jsvob: `package` is the name mandated by the CVE 5.0 schema.

osidb/models.py
jsvob: But it doesn't disappear, does it? It's moved to the new model. Ctrl+f for `related_name="package_versions"` in the commit.

osidb/models.py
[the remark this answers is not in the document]
jsvob: > and this solves the data backup issue

No. Already explained why, for [example here](https://github.com/RedHatProductSecurity/osidb/blob/7bd1b89b7045e2f557cd198b2dd9c6895a4f63cf/apps/bbsync/query.py#L394).

> You just need to do the migrations for the change in meta_attr.

No, the data is not saved in the database in osidb master. Bzimport parses that data but throws the original string away. Only the original string contains the original separators. You also raised it [here, with a similar answer](https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1323363923).

> Also, I am not sure if we can delete model's fields without deprecating them first via deprecate_field_custom to keep N-1 compatibility.

I don't understand. Please search for `@extend_schema_serializer(deprecate_fields=["status"])` in the PR's diff.

osidb/filters.py
jsvob: A "no" from me. Let's add it only when requested. The flaw endpoint had the package_versions field for a long time and nobody ever requested it AFAIK, so it's out of scope for OSIDB-1066.

osoukup:
> > > I discovered a bug: PUT (update) requires `updated_dt` in the request, but the value is ignored. I can't figure out why `updated_dt` is checked for e.g. acknowledgments but not for this. Obviously, the culprit must be somewhere in the new code, but that still leaves me with debugging the long tangle of nested save() and bzsync() calls throughout the various classes in the inheritance tree.
> > 
> > 
> > @jsvob not that I would see what is the issue but also run into a similar one a few times and usually this is caused by a bad order of the inheritance. Unfortunately for some of the mixins which implement `save` or `create`/`update` the order matters. It is not ideal but probably hard to fix as specifically the Bugzilla sync disrupts/changes the regular call order.
> 
> Found the bug. `BugzillaSyncMixinSerializer.update()` iterates over `validated_data` and sets all fields from the client's request onto the in-memory instance representation, including `updated_dt`. All existing models except the new `Package` model use that mixin. That then enables `TrackingMixin.save()` to compare the instance's updated_dt with the in-database updated_dt (implicitly assuming that the instance has client-provided updated_dt). In `FlawPackageVersionSerializer`, the in-memory instance's `updated_dt` is never set to the value from the PUT request, so `TrackingMixin.save()` always compares the in-database value with the in-database value.
> 
> Tomorrow I'll think about how to fix `BugzillaBareSyncMixinSerializer` in such a way that simultaneously
> 
>     * is simple and adds as little additional code to the PR as possible
> 
>     * still allows custom deserializer logic, such as `FlawPackageVersionSerializerMixin`
> 
>     * generically works with existing and future mixins
> 
>     * is obvious enough so that nobody spends a day hunting for this again when using `BugzillaBareSyncMixinSerializer` instead of `BugzillaSyncMixinSerializer`

@jsvob is the issue that the old API does not provide (and the new accept) the **updated_dt**? If this was an issue then extending the API is no breaking change.

collectors/bzimport/tests/test_convertors.py
jsvob: Because of `_validate_acl_duplicite`

collectors/bzimport/tests/test_convertors.py
jsvob: [Sequence unpacking](https://docs.python.org/3/tutorial/datastructures.html#:~:text=%3E%3E%3E-,x%2C%20y%2C%20z%20%3D%20t,-This%20is%20called)

jsvob: Also somewhat relevant: [PEP 3132 - Extended Iterable Unpacking](https://peps.python.org/pep-3132/)

osidb/models.py
jsvob: Fixed in commit [c09a6b4](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/c09a6b4125ae40a2bf6f334ba66b9e0b5bc8bfac)

osidb/tests/test_package_versions.py
jsvob: Fixed in commit [c20061e](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/c20061e53092a9add1dd0a35563d31d53e07849e)

collectors/bzimport/tests/test_convertors.py
jsvob: Fixed in commit [266dca9](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/266dca9466f9c8a6c8283a42e86e532a5a98b558)

collectors/bzimport/tests/test_convertors.py
jsvob: Fixed in commit [0986a35](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/0986a3582c8c4dba8c830bdf8173d2f6f7b98c16)

apps/bbsync/tests/test_query.py
jsvob: Fixed in commit [49a61c3](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/49a61c3bba8cac77e241879831f422a0b39de6d1)

osidb/models.py
jobselko: It can disappear in commitX and reappear in commitY, which should not happen, as we discussed earlier in our google space.

collectors/bzimport/convertors.py
jobselko: `fixed_in` should be certainly in `flaw.package_versions`. It does not make sense to have `fixed_in` in `meta_attr` if it will be empty for each old flaw (so basically for all flaws now). I would make migrations for all flaws or remove `fixed_in` from `meta_attr`. Otherwise, this would be the only missing part in `meta_attr` for all flaws.

osidb/models.py
jsvob: Okay, but that's so close that I didn't even notice. That would be confusing IMO, even though the old models should be deleted in a month or so.

osidb/models.py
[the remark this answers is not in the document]
jobselko: Whether you deprecate the field in the model or API is different. See the discussion in #PR267

osidb/models.py
jsvob: I don't understand.

Please show commits where it disappears and commits where it reappears. I say that `related_name="package_versions"` is moved between models in commit 5cf938968f1737d20d2fdd2597155bd2b8dd2eb5

Reason for multiple commits is in the top-most commit in this PR.

osidb/models.py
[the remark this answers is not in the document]
jsvob: From the linked DEVELOP.md file:

> For the DB we need to ensure that OSIDB is N-1 compatible meaning that the app of version N can work with the DB of version N-1 and vice-versa.

This quality is met with this PR. Note that existing models are not touched at all (except for the `related_name`).

> For the API we need to ensure that the breaking changes are introduced only in major releases. The standard procedure is as follows.

This PR adds new API endpoints and doesn't change the existing field (except for marking one field as deprecated).

collectors/bzimport/convertors.py
jsvob: > fixed_in should be certainly in flaw.package_versions

This argument would gut `meta_attr` of most of its contents.

> It does not make sense to have fixed_in in meta_attr if it will be empty for each old flaw (so basically for all flaws now).

No. `meta_attr["fixed_in"]` **will not exist** for old flaws. Also please see the explanation in comment https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1323373897

> I would make migrations for all flaws or remove fixed_in from meta_attr.

Again, the original `fixed_in` string is currently thrown out in osidb master, so no migration possible. And it must be stored somewhere, because bbsync can't work without it (bbsync already contains a detailed explanation). Again please see the explanation in comment https://github.com/RedHatProductSecurity/osidb/pull/307#discussion_r1323373897

jsvob: > EDIT: If the original fixed_in in BZ contains wrong data (e.g. Upstream: kernel-1.3), then it could be problematic to recreate the original value, that is true. 

Wrong data are out of scope. We can't fix it here.

> In this situation, only resync would make sense for these problematic flaws.

That's still above the manual resync limit specified by Ondrej. Going back to the original issue, how would that be better than having this line in the code?

osidb/models.py
jobselko: I did not say `it disappeared`, but `it can disappear` and `you should ensure`. I think this was actionable enough.

osidb/tests/test_endpoints.py
```diff
@@ -1951,6 +1953,265 @@ def test_tracker_delete(self, auth_client, test_api_uri):
         # with the authoritative sources of the tracker data
         assert response.status_code == 405
 
+    def test_packageversions_filter(self, auth_client, embargo_access, test_api_uri):
+        """
+        Test the non-trivial parts of FlawPackageVersionFilter via REST API GET requests.
+        """
+        # Source must be private in order for validation to pass.
+        flaw = FlawFactory(source=FlawSource.CUSTOMER)
```
jobselko, line 1961: This should be written as `flaw = FlawFactory()` (more occurrences).

Having a flaw with a public source (i.e. where the flaw was first reported) is a valid option. `package_versions` does not require any restriction.

openapi.yml
```diff
@@ -5890,32 +6233,6 @@ components:
...
-      description: CVEv5 Package Version Serializer
-      properties:
-        version:
-          type: string
-          maxLength: 1024
-        status:
-          $ref: '#/components/schemas/StatusEnum'
-      required:
-      - status
-      - version
```
jobselko, line 5918: This will most likely break compatibility, but I am not 100% sure.

osidb/serializer.py
```diff
@@ -714,6 +754,167 @@ class FlawAcknowledgmentPostSerializer(FlawAcknowledgmentSerializer):
     pass
 
 
+class FlawVersionSerializer(
```
jobselko, line 757: This name does not make sense in the given context.

jsvob:
> @jsvob is the issue that the old API does not provide (and the new accept) the **updated_dt**? If this was an issue then extending the API is no breaking change.

@osoukup No. The issue was that the serializer `update()` didn't `setattr` the request-provided `updated_dt`. Fixed in commit [83dcb3e](https://github.com/RedHatProductSecurity/osidb/pull/307/commits/83dcb3e7ab31f3995f2d5d4c92e80b63782304d7)

osidb/models.py
jsvob: I will change `2058` to `2048` today or tomorrow morning while rebasing and squashing the commits.

openapi.yml
jsvob: `oasdiff` check is fine. We'll see for sure once it's rebased and partially squashed tomorrow.

osidb/serializer.py
jsvob: It does because this is for the `PackageVer` objects that are for tracking **versions**. It's used in the `FlawPackageVersionSerializer` that is for the nested combination of `Package` and `PackageVer` objects that together track the **packages** and **versions**.

osidb/tests/test_endpoints.py
jsvob: Thanks for spotting this. Resolved in commit 53f1e57a1461da2cc2a2cbdeceb114ad1389d482

osidb/tests/test_package_versions.py
jsvob: Resolved in rebased & partially squashed commits by adding explanatory commit messages into the three initial commits and deleting the file in the third commit. There's truly no need to test basic and obvious django functionality.

jsvob:
@osoukup @Elkasitu @jobselko Rebased and [partially squashed](https://github.com/RedHatProductSecurity/osidb/pull/307#issuecomment-1712035312:~:text=The%20reason%20is%20that%20IMO%20squashing%20some%20of%20the%20commits%20would%20make%20the%20changes%20large). I propose to merge this.

openapi.yml
jsvob: It's fine.

jobselko (dismissed):

PR #313, opened 2023-08-24
Validate summary and requires_summary

jobselko:
This PR adds validation to check if the combination of `summary` and `requires_summary` is set correctly.

Related to OSIDB-1164


osoukup (approved):
LGTM

PR #316, opened 2023-08-30
Trackers filing suggestion

costaconrado:
This PR adds a new endpoint to suggest trackers to file.
This PR also fixes small naming bugs for `trackers` modules which were named as `tracker` in some files.

Closes OSIDB-90.

### DIFF
--- a/apps/trackers/tests/test_file_offer.py
+++ b/apps/trackers/tests/test_file_offer.py
@@ -1,0 +1,314 @@
+"""
+Test cases for tracker suggestion generation
+"""
+
+import pytest
+
+from osidb.models import Affect, Flaw, Impact, PsUpdateStream, UbiPackage
+from osidb.tests.factories import (
+    AffectFactory,
+    FlawFactory,
+    PsModuleFactory,
+    PsUpdateStreamFactory,
+)
+
+pytestmark = pytest.mark.unit
+
+
+class TestTrackerSuggestions:
+    """
+    Test Tracker Suggestion generation
+    """
+
+    @pytest.mark.parametrize(
+        "affectedness,resolution,is_valid",
+        [
+            (Affect.AffectAffectedness.AFFECTED, Affect.AffectResolution.FIX, True),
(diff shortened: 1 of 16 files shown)


osidb/models.py
```diff
@@ -2927,6 +2928,22 @@ def is_rhscl(self) -> bool:
         """
         return self.bts_key == RHSCL_BTS_KEY
 
+    @property
+    def y_stream(self):
```
osoukup, line 2932: I am thinking whether it would be better to call the property `y_streams` when they can be more. I am not strongly for it but it would more naturally describe what can one await as a result.

apps/trackers/tests/test_file_offer.py
```diff
@@ -0,0 +1,314 @@
+"""
+Bugzilla specific tracker test cases
```
osoukup, line 2: I guess that the module description is just copy-pasted and not updated.

osoukup (approved):
LGTM. This must have been non-trivial indeed but the solution is nicely readable. Good job!

osidb/models.py
costaconrado: I think it makes sense, I will change to plural.

docs/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement collector for ps-constants project (OSIDB-1199)
 - Validate summary and requires_summary (OSIDB-1164)
 - Validate impact and summary (OSIDB-1164)
+- Implement endpoint for tracker auto filing suggestions (OSIDB-90)
```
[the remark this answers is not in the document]
costaconrado: rephrased both in commit message and changelog, thanks

apps/trackers/tests/test_file_offer.py
costaconrado: Updated for correct description.

PR #318, opened 2023-08-30
Set AFFECTED as highest precedence for delegated_resolution

Elkasitu:
For Affects that have a resolution of DELEGATED, we look at its trackers' own state/resolution to determine the actual delegated resolution of the Affect.

If there's more than a single Tracker for an Affect, we take the Tracker's state/resolution with highest precedence or "weight" and this precedence is determined by a list of possible resolutions.

Before this commit, said list was ordered in such a way that the NOTAFFECTED resolution had highest precedence however this is a mistake, as if we have 5 trackers, 4 with NOTAFFECTED and 1 with AFFECTED resolution we should determine that the Affect is AFFECTED, not NOTAFFECTED.

This commit solves the issue by swapping the order of the AFFECTED and NOTAFFECTED resolutions in the aforementioned list.

Closes OSIDB-1230

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Reduce the total amount of records per page when querying Bugzilla (OSIDB-1232)
+- Set AFFECTED as highest precedence resolution when calculating Affect.delegated_resolution (OSIDB-1230)
 
 ## [3.4.1] - 2023-08-21
 ### Changed
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1718,12 +1718,17 @@ class Affect(
             return Affect.AffectFix.AFFECTED
 
         statuses = [tracker.fix_state for tracker in trackers]
+        # order is **very** important here, if there are multiple trackers
+        # the order of these statuses determines which tracker status takes
+        # precedence over all the rest, meaning that if one tracker is affected
+        # and another is not affected, the overall affect delegated_resolution
+        # will be affected and not notaffected.
         for status in (
-            Affect.AffectFix.NOTAFFECTED,
             Affect.AffectFix.AFFECTED,
             Affect.AffectFix.WONTFIX,
             Affect.AffectFix.OOSS,
             Affect.AffectFix.DEFER,
+            Affect.AffectFix.NOTAFFECTED,
         ):
             if status in statuses:
                 return status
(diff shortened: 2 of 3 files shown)


docs/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [3.4.1] - 2023-08-21
+### Changed
+- Set AFFECTED as highest precedence resolution when calculating Affect.delegated_resolution (OSIDB-1230)
+
+## [3.4.1] - 2023-08-21
```
osoukup, line 11: This seems to be a duplicate of the above version

osoukup (approved):
LGTM

pmatouse:
I would argue that WONTFIX and OSS also need to be before NOTAFFECTED. basically all states that have a counterpart on CVE pages (I believe DEFER does not) have to go before NOTAFFECTED.

osoukup:
> I would argue that WONTFIX and OSS also need to be before NOTAFFECTED. basically all states that have a counterpart on CVE pages (I believe DEFER does not) have to go before NOTAFFECTED.

@pmatouse good to know as OSIDB-1230 does not seem to lead to any such conclusion. When we are in the middle of the change, what should be the exact order of all of the values?

Elkasitu:
I'm OK with doing the change but it's a bit surprising that this is a problem now after ~3y of it having been like this

mprpic:
> @mprpic WDYT? how DEFER would be handled on the sdengine side?

We don't really use the defer state any more so it's there mostly for historical reasons; it translates to "Fix deferred" on CVE pages. I think it's fine to have it before notaffected.

pmatouse:
> I'm OK with doing the change but it's a bit surprising that this is a problem now after ~3y of it having been like this

well, customers see NOTAFFECTED on CVE pages. that's the best possible resolution for them, because it translates to no action. it needs someone to start to wonder how that is possible, one way or another. this time it was someone internal who saw the state of the trackers and was also looking at customer facing "statements" because of Common Criteria Evaluation. for CCC purposes he considers only a subset of flaws affecting certain streams and meeting certain criteria.

pmatouse:
> > @mprpic WDYT? how DEFER would be handled on the sdengine side?
> 
> We don't really use the defer state any more so it's there mostly for historical reasons; it translates to "Fix deferred" on CVE pages. I think it's fine to have it before notaffected.

ack, thank you.

thanks @Elkasitu  for working on this issue!

PR #319, opened 2023-08-30
add shipped date to erratum

osoukup:
Closes OSIDB-1197


costaconrado (approved):
LGTM.

osidb/filters.py
```diff
@@ -222,6 +222,10 @@ class Meta:
             + LTE_GTE_LOOKUP_EXPRS
             + DATE_LOOKUP_EXPRS,
             "affects__trackers__errata__advisory_name": ["exact"],
+            "affects__trackers__errata__shipped_dt": ["exact"]
+            + LT_GT_LOOKUP_EXPRS
+            + LTE_GTE_LOOKUP_EXPRS
+            + DATE_LOOKUP_EXPRS,
```
Elkasitu, line 228: I would like to avoid these kind of filters since they are insane, instead what we can do is create a filter simply named `errata_shipped_dt` or similar and link it to a function, similar to what we do with `changed_after`, `changed_before` filters.

Elkasitu (approved):
Probably could use a data migration too

osidb/filters.py
osoukup: I am not against but that would be rather for a general filter functionality rework.

osoukup:
> Probably could use a data migration too

I am not sure how. The field is new and the values are not present anywhere in OSIDB.

PR #321, opened 2023-08-31
Update NVD collector to save data into FlawCVSS

jobselko:
This PR updates NVD collector to save cvss scores for flaws into the new `FlawCVSS` model. The original logic using fields in the `Flaw` model has been kept.

Related to OSIDB-1105

### DIFF
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -7,7 +7,7 @@ from django.utils import timezone
 
 from collectors.framework.models import Collector
 from osidb.core import set_user_acls
-from osidb.models import Flaw
+from osidb.models import Flaw, FlawCVSS
 
 logger = get_task_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


collectors/nvd/collectors.py
```diff
@@ -153,14 +153,59 @@ def collect(self, cve: Union[str, None] = None) -> str:
...
+
+                    cvss_score = FlawCVSS.objects.create_cvss(
+                        flaw=flaw,
+                        issuer=FlawCVSS.CVSSIssuer.NIST,
+                        version=version,
+                        vector=new_cvss,
+                        acl_write=flaw.acl_write,
+                        acl_read=flaw.acl_read,
+                    )
+                    cvss_score.save()
```
Elkasitu, line 203: why are we deleting the existing record and creating a new one with the same characteristics but the vector? why not just update the existing record with the new vector? I feel like in general a lot of the code here could be simplified, or am I missing something?

jobselko: `create_cvss(...)` performs either update or create, so line 193 with `delete()` is redundant - it will only force `create_cvss(...)` to always create a new cvss instead of updating the existing one if it is sufficient.

I forgot to remove line 193 when I was checking score updating. Thanks for noticing!

PR #324, opened 2023-09-05
Add gunicorn config for logging access

costaconrado:
This PR introduces access log files for stage and prod environment.

This config should be enough since `/var/log` is already configured and monitored.

Closes OSIDB-879.

### DIFF
--- a/scripts/run-service.sh
+++ b/scripts/run-service.sh
@@ -9,6 +9,15 @@ python3 manage.py collectstatic \
     --ignore 'tmp*' \
     --noinput
 
+# Defaults to stdout for local development
+ACCESS_LOG_FILE="-"
+
+if [ "$DJANGO_SETTINGS_MODULE" = "config.settings_stage" ]; then
+    ACCESS_LOG_FILE="/var/log/stage-access.log"
+elif [ "$DJANGO_SETTINGS_MODULE" = "config.settings_prod" ]; then
+    ACCESS_LOG_FILE="/var/log/prod-access.log"
+fi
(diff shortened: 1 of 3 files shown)


scripts/run-service.sh
```diff
@@ -9,6 +9,15 @@ python3 manage.py collectstatic \
     --ignore 'tmp*' \
     --noinput
 
+# Defaults to stdout for local development
+ACCESS_LOG_FILE="-"
+
+if [ "$DJANGO_SETTINGS_MODULE" = "config.settings_stage" ]; then
+    ACCESS_LOG_FILE="/var/log/stage-access.log"
+elif [ "$DJANGO_SETTINGS_MODULE" = "config.settings_prod" ]; then
+    ACCESS_LOG_FILE="/var/log/prod-access.log"
```
osoukup, line 18: A stupid question :smile: Does everything what goes to `log` directory goes automatically to Splunk or do the files need to be named in some specific way?

osoukup (approved):
LGTM

scripts/run-service.sh
osoukup: Thank you for the explanation!

Elkasitu (commented):
`gunicorn_config.py` is a python file so it should be perfectly fine to do something like:

```python

accesslog = settings.ACCESS_LOG_FILE_PATH
```

or something like that, which would simplify the changes done here.

Furthermore what do we do about the errorlogs? Both should be piped to the same file.
```

PR #325, opened 2023-09-05
Use cassettes when running tests against NVD

jobselko:
This PR fixes the situation when tests randomly fail with a 403 error.

Follow-up to #321


osoukup (approved):
That seems that those tests actually did HTTP calls over the network :thinking: Great to get fixed!

PR #327, opened 2023-09-06
Fix DELETE methods status code

JakubFrejlach:
This PR fixes discrepancy between OSIDB API schema and actual returned status code from DELETE methods.

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -418,6 +418,11 @@ class FlawView(ModelViewSet):
 
 
 class SubFlawViewDestroyMixin:
+    @extend_schema(
+        responses={
+            200: {},
+        }
+    )
     def destroy(self, request, *args, **kwargs):
         """
         Destroy the instance and proxy the delete to Bugzilla
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Link tracker to flaw(s) on create/update (OSIDB-1182)
 - Implement FlawCVSS and AffectCVSS APIs with filters (OSIDB-1105)
 
+### Fixed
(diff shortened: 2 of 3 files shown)


osoukup (approved):
LGTM

JakubFrejlach:
I see there is a backward compatibility check failed, howerver this is actually a false positive, as we simple advertised something which was not true (eg. schema advertising return code 204, but in fact return code is 200), should we bypass that in this case?

osoukup:
> @osoukup I see there is a backward compatibility check failed, howerver this is actually a false positive, as we simple advertised something which was not true (eg. schema advertising return code 204, but in fact return code is 200), should we bypass that in this case?

Makes sense to me. @Elkasitu WDYT?

osidb/api_views.py
```diff
@@ -410,6 +410,11 @@ def create(self, request, *args, **kwargs):
 
 
 class SubFlawViewDestroyMixin:
+    @extend_schema(
+        responses={
+            200: {},
```
Elkasitu, line 415: It can also return 400

JakubFrejlach: Well in that case we are missing the 400 and all possible error status codes in every endpoint for every method :laughing: There is a not a single response example for 4xx status code anywhere in the OSIDB currently

Elkasitu: Fair enough

Elkasitu (approved):

Elkasitu:
@JakubFrejlach the check is not marked as required so it should not block merge, are you sure it's not something else that's causing the failure? See #262 which was successfully merged in spite of the failing check

JakubFrejlach:
> @JakubFrejlach the check is not marked as required so it should not block merge, are you sure it's not something else that's causing the failure? See #262 which was successfully merged in spite of the failing check

Ahhh, thank you! It fails do to a schema check not schema backwards compatibility check. I mixed those two together.

PR #328, opened 2023-09-06
Deprecate comment in CVSS and add it to FlawCVSS

jobselko:
This PR deprecates the `comment` field in the `CVSS` model and adds it directly into the `FlawCVSS` model with its validation. This is because affects' CVSSs cannot have comments in BZ. 

Related to OSIDB-1105


osoukup (commented):
I am thinking why is this change necessary. I know that the old tooling only understands CVSSv3 flaw-level comment however why not to leave it general? We could just have one unified structure of the CVSS entity and Flaw/Affect-CVSS models would only differ in the relation.

jobselko:
My goal is to enable creation of only entities compatible with BZ, which is currently not true for `CVSSAffect`/`AffectCVSSFactory`. So we can leave it as it is and add a note with a description, or fix both model and factory, or fix only the factory (which is a partial solution).

Either way, I do not have a preference, and I am fine with anything you suggest.

osoukup:
I would prefer to leave both CVSS score models the same even though it will not have much effect in the Bugzilla world. The situation with no version 2 comment is similar. In the future we would not have to distinguish where the CVSS score is linked to, what version or issuer it is and just have the same behavior. I believe that the less exceptions the better :smile:

PR #329, opened 2023-09-07
Add APIs for FlawCVSS and AffectCVSS

jobselko:
This PR introduces APIs for `FlawCVSS` and `AffectCVSS`. They were tested both on localhost and in BZ.

Related to OSIDB-1105

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -29,6 +29,7 @@ from rest_framework.viewsets import (
 
 from .constants import OSIDB_API_VERSION, PYPI_URL, URL_REGEX
 from .filters import (
+    AffectCVSSFilter,
     AffectFilter,
     FlawAcknowledgmentFilter,
     FlawCommentFilter,
@@ -662,6 +663,7 @@ class AffectCVSSView(ModelViewSet):
     serializer_class = AffectCVSSSerializer
     http_method_names = get_valid_http_methods(ModelViewSet)
     permission_classes = [IsAuthenticatedOrReadOnly]
+    filterset_class = AffectCVSSFilter
 
     def get_affect(self):
         """
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -32,6 +32,7 @@ from .filters import (
     AffectFilter,
     FlawAcknowledgmentFilter,
     FlawCommentFilter,
+    FlawCVSSFilter,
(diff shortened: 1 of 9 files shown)


jobselko:
Should I add also filters?

osidb/api_views.py
```diff
@@ -517,6 +520,22 @@ class FlawReferenceView(
     permission_classes = [IsAuthenticatedOrReadOnly]
 
 
+@include_meta_attr_extend_schema_view
```
osoukup, line 523: CVSS does not seem to have any `meta_attr` attribute (and I think there is no reason it should have).

osoukup (approved):
LGTM. I would add some filters similarly the other embedded flaw models have them. Probably the most meaningful filter would be relative comparison on the score value but this is a different story.

osidb/filters.py
```diff
@@ -479,3 +480,25 @@ class Meta:
...
+            "flaw": ["exact"],
+            "issuer": ["exact"],
+            "score": ["exact"],
+            "updated_dt": ["exact"]
+            + LT_GT_LOOKUP_EXPRS
+            + LTE_GTE_LOOKUP_EXPRS
+            + DATE_LOOKUP_EXPRS,
+            "uuid": ["exact"],
+            "vector": ["exact"],
+            "version": ["exact"],
```
jobselko, line 503: I am unsure whether this line is correct, because `FlawCVSSSerializer` contains `cvss_version = serializers.CharField(source="version")`.  But when using `cvss_version` instead of `version`, I get `TypeError: 'Meta.fields' must not contain non-model field names: cvss_version`, so probably it is.

jobselko:
@osoukup I managed to finish both APIs yesterday. Could you please recheck it? Thanks!

osidb/filters.py
osoukup: That is quite undesirable because from the user point of view when the attribute is called differently in the filter than elsewhere. There should be some conversion or something. Otherwise the version filter could be useful.

osoukup (approved):
LGTM

No strong opinion but I would prefer the tests to be split rather than parametrize for two different APIs - though I agree they are very similar. Until the things go smoothly it does not matter but it would be harder to debug if the things go wrong. Feel free to ignore this.

PR #330, opened 2023-09-08
sync with Griffon on the new API entities before the release

osoukup:
Whenever there is a completely new API or we add some embedded entity (like acknowledgments, comments etc.) the Griffon needs to be adjusted to understand and manipulate it (not necessary when just adding an attribute to an existing entity). We should make this a process of some kind so we do not forget and make sure that with a new OSIDB there is a new Griffon already available and tested.

I am bit vague on where exactly to create the Jira issue etc. as this is not the OPS repo. However, the code freeze which is IMO a good (last) moment for this is described here and the int the OPS repo.

### DIFF
--- a/docs/developer/OPERATIONS.md
+++ b/docs/developer/OPERATIONS.md
@@ -20,6 +20,10 @@ the release branch so that the code can be reviewed.
 Make sure that the stage environment is set to pull from this branch at least until it is merged
 back into the main development branch.
 
+Additionally, if the API was extended with completely new entities, either independent
+or embeded under the others, make sure to raise the corresponding feature request to the
+[Griffon](https://github.com/RedHatProductSecurity/griffon/) project.
+
 ## Release
 
 Follow this procedure when performing OSIDB version X.Y.Z release, this step assumes that


osoukup:
Feel free to suggest a different location for this documentation etc.

JakubFrejlach (changes requested):
Might be worth to mention the coordination of the osidb-bindings release as well.

osoukup:
> Might be worth to mention the coordination of the osidb-bindings release as well.

@JakubFrejlach so what would be the right approach from the OSIDB side? Do the bindings need some underlying issue or just notification or should we describe the process (which I do not know :smile:) here?

osoukup:
> What is the core issue here? Does Griffon break when we introduce these new endpoints or is the problem simply that it becomes outdated?
> 
> We already communicate these changes to all clients via freeze/release emails and changelog, not sure why we need to target Griffon specifically.
> 
> What we can do instead is be more explicit in the changelog with things like "new endpoint `/foo` added for management of `Foo` objects".
> 
> GitHub also has a "watch repository" feature which can be set up to only notify you of new releases.

@Elkasitu no major issue. This is just an attempt to make this a process instead of ad-hoc actions when someone actually recalls that it should be also done. The bindings and Griffon are special in that it is us who is responsible for it. We are doing the release and at some point we should release the corresponding bindings and the Griffon. Then it makes sense to me to make it part of the release process. But I am opened to suggestions.

JakubFrejlach:
> What is the core issue here? Does Griffon break when we introduce these new endpoints or is the problem simply that it becomes outdated?
> 
> We already communicate these changes to all clients via freeze/release emails and changelog, not sure why we need to target Griffon specifically.
> 
> What we can do instead is be more explicit in the changelog with things like "new endpoint `/foo` added for management of `Foo` objects".
> 
> GitHub also has a "watch repository" feature which can be set up to only notify you of new releases.

Well from the griffon point of view, it just becomes outdated as the new resource API endpoint cannot be invoked within griffon. Being able to invoke griffon API endpoints was originally requested from OSIDB side, so I would expect request for new resources rather then need to watch for these.

For osidb-bindings, it is a lib which we ship with every major and minor version for OSIDB, so we are just trying to make it a standardized process so it is not forget.

PR #331, opened 2023-09-08
integrate TrackerSaver with the Tracker model

osoukup:
Similarly to the other models we should only call `Tracker.save` and everything else including the potential sync to Bugzilla or Jira should be done automatically. The other models are so far all stored in Bugzilla (as part of the flaw data). The trackers are bit different as they may be in either Bugzilla or Jira and therefore cannot use the `BugzillaSyncMixin` unlike the other models so I implemented `save` with the special handling directly on the model class.

OSIDB-1179

### DIFF
--- a/apps/trackers/constants.py
+++ b/apps/trackers/constants.py
@@ -1,8 +1,13 @@
 """
 common tracker constants
 """
+from osidb.helpers import get_env
 
 TRACKERS_API_VERSION = "v1"
+
+# switch to enable or disable the Jira tracker sync
+# - the one controlling the Bugzilla sync is defined in the BBSync app
+SYNC_TO_JIRA = get_env("TRACKERS_SYNC_TO_JIRA", default="False", is_bool=True)
 
 # tracker summary constants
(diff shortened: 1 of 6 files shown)


apps/trackers/constants.py
```diff
@@ -1,9 +1,14 @@
 """
 common tracker constants
 """
+from osidb.helpers import get_env
 
 TRACKERS_API_VERSION = "v1"
 
+# switch to enable or disable the Jira tracker sync
+# - the one controlling the Bugzilla sync is defined in the BBSync app
+SYNC_TO_JIRA = get_env("TRACKERS_SYNC_TO_JIRA", default="False", is_bool=True)
```
costaconrado, line 10: Please add `TRACKERS_SYNC_TO_JIRA` to `DEVELOP.md` in the .env section.

costaconrado (approved):
I have left a documentation improvement but otherwise LGTM.

apps/trackers/constants.py
osoukup: Thank you for the reminder!

PR #332, opened 2023-09-12
Remove django admin interface

Elkasitu:
The Django admin interface is not actively used by the OSIDB maintainers thus it has been decided to remove it in order to reduce a potential attack vector.

Closes OSIDB-1188


osoukup (approved):
LGTM

PR #333, opened 2023-09-12
Introduce index for Tracker.external_system_id

Elkasitu:
This should result in faster response times when filtering the /trackers API by the external_system_id filter, and since the external_system_id field is unlikely to change for most trackers it's probably an inexpensive index.


osoukup (approved):
Makes sense. LGTM

PR #335, opened 2023-09-14
Deprecate cvss fields in APIs and use FlawCVSS model for validations

jobselko:
This PR updates the following:
- `cvss2`, `cvss2_score`, `cvss3`, `cvss3_score`, `nvd_cvss2`, and `nvd_cvss3` fields are now deprecated in Flaw API
- `cvss2`, `cvss2_score`, `cvss3`, and `cvss3_score` fields are now deprecated in Affect API
- flaw validations now use the new `FlawCVSS` model instead of `cvss3_score`, `nvd_cvss3`, and `cvss3` fields from the `Flaw` model

Closes OSIDB-1105

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -345,6 +345,42 @@ def include_exclude_fields_extend_schema_view(
                 location=OpenApiParameter.QUERY,
                 deprecated=True,
             ),
+            OpenApiParameter(
+                "cvss2",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                deprecated=True,
+            ),
+            OpenApiParameter(
+                "cvss2_score",
+                type=OpenApiTypes.FLOAT,
+                location=OpenApiParameter.QUERY,
+                deprecated=True,
+            ),
+            OpenApiParameter(
+                "cvss3",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                deprecated=True,
+            ),
+            OpenApiParameter(
+                "cvss3_score",
+                type=OpenApiTypes.FLOAT,
+                location=OpenApiParameter.QUERY,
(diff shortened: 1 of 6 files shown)


osidb/models.py
```diff
@@ -846,48 +846,67 @@ def __str__(self):
         """convert to string"""
         return str(self.uuid)
 
-    def _validate_rh_nvd_cvss_score_diff(self):
+    def _validate_rh_nist_cvss_score_diff(self):
         """
-        Checks that the difference between the RH and NVD CVSS score is not >= 1.0
+        Checks that the difference between the RH and NIST CVSSv3 score is not >= 1.0.
```
jobselko, line 851: I changed `NVD` to `NIST` everywhere to keep the name consistent.

jobselko:
I intentionally did not deprecate cvss fields in `models.py`, so there is no extra work in `filters.py` now. Deprecation/removal can be done after a major release, when those fields are no longer needed.

osidb/api_views.py
```diff
@@ -345,6 +345,43 @@ def include_exclude_fields_extend_schema_view(
...
+                location=OpenApiParameter.QUERY,
+                deprecated=True,
+            ),
+            OpenApiParameter(
+                "nvd_cvss3",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                deprecated=True,
+            ),
+            # todo: cvss3 comment z meta attr?
```
osoukup, line 384: I think that there is no need to reflect the meta attribute changes in the API docs. There are no guarantees regarding the metadata.

osoukup (approved):
LGTM

osidb/api_views.py
jobselko: This does not belong here. It is just my old note.

PR #337, opened 2023-09-18
Add filters for flaw references in API and fix serializers

jobselko:
This PR updates several things around flaw references:
- `SubFlawViewNoneFormatQuerysetMixin` was removed from `FlawReferenceView` because its `get_queryset` method was incorrect. This method did not provide a fallback that allows the call of `get_queryset` in a situation when `get_queryset` depends on attributes not available at schema generation time. This also caused the `openapi.yml` file to contain the `flaw_id` parameter's incorrect format - unspecified `string` instead of `string($uuid)`.
- `FlawReferencePutSerializer` decorated by `@extend_schema_serializer(exclude_fields=["flaw"])` was added to `FlawReferenceView`. Without this serializer, the PUT request contained the `flaw` in two places - in its parameters and in its body, which might be confusing. The same problem was in the `openapi.yml` file.
- Filters for flaw references were added in API.

Closes OSIDB-1368

### DIFF
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -20,6 +20,7 @@ from .models import (
     FlawAcknowledgment,
     FlawComment,
     FlawCVSS,
+    FlawReference,
     Package,
     Tracker,
     search_helper,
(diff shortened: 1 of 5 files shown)


jsvob (approved):
LGTM. This follows a recent discussion between @jobselko and @JakubFrejlach which I joined. The gist is that **as long as** @RedHatProductSecurity/osidb-devs collectively agree about it, a number of changes that `oasdiff` flags as breaking are possible:

- Fix schema to match the actual behavior (formerly behavior advertised wrongly by the schema). This should be possible any time, since there was a bug in the schema in the first place.
- Change type to a compatible type, if the change fixes a bug in the schema. This should be possible any time. Based on my limited searching, there doesn't seem to be an openapi feature to mark type changes, similarly to the way openapi allows to mark fields as deprecated, so we have no better way to do that.
- Fix behavior and schema together. This should be possible with a new major release and we should sync about that, so the change goes in shortly before the major release and doesn't break any minor releases.

Now the question is, esp. for @osoukup and @Elkasitu, where this PR falls and whether we agree on the above.

If not, perhaps this PR should be split into two: Filters now, and type change planned for next major release.

osidb/filters.py
```diff
@@ -258,6 +259,19 @@ class Meta:
...
+            + LTE_GTE_LOOKUP_EXPRS
+            + DATE_LOOKUP_EXPRS,
+            "references__description": ["exact"],
+            "references__type": ["exact"],
+            "references__updated_dt": ["exact"]
+            + LT_GT_LOOKUP_EXPRS
+            + LTE_GTE_LOOKUP_EXPRS
+            + DATE_LOOKUP_EXPRS,
+            "references__url": ["exact"],
+            "references__uuid": ["exact"],
```
Elkasitu, line 274: Not really a remark for *you* specifically nor a blocker for merging this PR, but there is a clear pattern here for `datetime` fields, we could create a helper function for those and spare a lot of boilerplate code.

Elkasitu (approved):

PR #338, opened 2023-09-19
Workaround for `ImportError: cannot import name 'DEV_PKGS' from 'pip._internal.commands.freeze'`

jsvob:
Workaround for `ImportError: cannot import name 'DEV_PKGS' from 'pip._internal.commands.freeze'` when setting up a local dev environment or when applying requirements.txt.

### DIFF
--- a/devel-requirements.txt
+++ b/devel-requirements.txt
@@ -896,10 +896,12 @@ yarl==1.7.2 \
     # via vcrpy
 
 # The following packages are considered to be unsafe in a requirements file:
-pip==23.2.1 \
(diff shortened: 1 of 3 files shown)


jobselko:
I think we should update `pip-tools` instead of downgrading `pip`. See https://github.com/jazzband/pip-tools/releases/tag/7.0.0

jsvob:
Updated the branch https://github.com/RedHatProductSecurity/osidb/commits/pin_pip_23_1 with Jitka's suggestion and a makefile fix, but the pull request still shows the outdated now-removed commit. IDK why.

PR #342, opened 2023-09-19
Taskman osim tests update

costaconrado:
This PR reactivate unit tests in OSIM module.

This PR enabled all unit tests for OSIM module, in order to do that, unit tests were refactored using our current FactoryBoy where possible.
This PR also deleted curl commands for integration tests in OSIM.
This PR also deleted manual generators of workflows, states and workflow prerequisites in order to improve readability of the tests.

Closes OSIDB-1320.


apps/osim/checks.py
```diff
@@ -102,15 +104,17 @@ def decs2non_empty(cls, check_desc):
...
-                return (
-                    message,
-                    lambda instance: getattr(instance, attr) not in EMPTY_VALUES,
-                )
+                def has_element(instance):
+                    EMPTY_VALUES = [None, ""]
+                    field = getattr(instance, attr)
+
+                    if isinstance(field, models.manager.BaseManager):
+                        return field.count() > 0
```
Elkasitu, line 116: `field.all().exists()` and needs a test

apps/osim/tests/test_models.py
```diff
@@ -361,43 +372,30 @@ def test_satisfied_conditions(self, count):
...
-                type=FlawMeta.FlawMetaType.REQUIRES_SUMMARY,
-                meta_attr={"status": "-"},
-            )
-        flaw.save()
+        flaw = FlawFactory()
+        for condition in conditions:
+            if condition[:3] == "has":
+                setattr(flaw, condition[4:], "valid value")
+            else:
+                setattr(flaw, condition[4:], "")
```
Elkasitu, line 212: use something like `verb, attr = split(" ", maxsplit=1)`, what happens if after the verb we have more than a single word? e.g. `"has affects foo"`

apps/osim/tests/test_models.py
```diff
@@ -407,104 +405,58 @@ def test_unsatisfied_conditions(self, positive, negative):
...
+
+        # conditions partially satisfied
+        if len(conditions) > 1:
+            if conditions[0][:3] == "has":
+                setattr(flaw, condition[4:], "valid value")
+            else:
+                setattr(flaw, condition[4:], "")
+        assert not workflow.accepts(
+            flaw
+        ), f'flaw was wrongly accepted by workflow conditions "{conditions}"'
```
Elkasitu: same as before

apps/osim/tests/test_models.py
```diff
@@ -17,6 +17,16 @@
...
 
 
+def assert_state_equals(current, expected):
+    message = f'flaw classified as {current.name}, expected {expected["name"]}'
+    assert current.name == expected["name"], message
+
+
+def assert_workflow_equals(current, expected):
+    message = f"flaw classified in {current.name} workflow, expected {expected.name}"
+    assert current.name == expected.name, message
```
Elkasitu, line 23: unused in the commit that it's introduced

apps/osim/tests/test_models.py
```diff
@@ -668,37 +689,26 @@ def test_adjust(self):
...
-
-        assert flaw.classification == {
-            "workflow": "major incident workflow",
-            "state": "DRAFT",
-        }
 
-        flaw.is_major_incident = Flaw.FlawMajorIncident.NOVALUE
+        # TODO remove this 3 lines when major incident property is recreated
+        flaw.is_major_incident = True
+        flaw.save()
```
Elkasitu, line 533: why? IIRC `flaw.is_major_incident` is no longer used and `major_incident_state` is the canonical source of data

apps/osim/tests/conftest.py
```diff
@@ -58,15 +57,6 @@ def test_api_uri_osidb(test_scheme_host_osidb, api_version):
     return f"{test_scheme_host_osidb}/api/{api_version}"
 
 
-@pytest.fixture
-def command_curl():
-    """define path to curl"""
-    test_curl_path = get_env("TEST_CURL_PATH")
-    if test_curl_path is not None:
-        return test_curl_path
-    return "/usr/bin/curl"
```
Elkasitu, line 67: why? the whole point of integration tests is to test the application from outside, which we don't do in the CI yet btw so might as well not touch this

Elkasitu (changes requested):
LGTM overall but needs some minor changes I think, also probably requires a review by @osoukup

apps/osim/tests/test_models.py
```diff
@@ -287,87 +287,42 @@ def test_empty_requirements(self):
...
-            f"does not reject a flaw with {flaw_properties}"
-        )
-        assert not state.accepts(flaw), message
+        flaw.cwe_id = "CWE-1"
+        assert not state.accepts(
+            flaw
+        ), f'state accepted flaw without the requirements "{requirements}"'
+
+        flaw.impact = Impact.NOVALUE
+        assert not state.accepts(
```
osoukup, line 165: To make the second assert useful we should first set `cwe_id` back to an empty value or otherwise we cannot tell whether it fails because of the impact or again because of the CWE.

apps/osim/tests/test_models.py
```diff
@@ -407,104 +405,58 @@ def test_unsatisfied_conditions(self, positive, negative):
...
+        assert_state_equals(workflow.classify(flaw), state_second)
 
-        message = (
-            "flaw was not classified in any state despite the default state exists"
-        )
-        assert workflow.classify(flaw) is not None, message
+        # Test that a flaw will not bypass past states
+        bypass_flaw = Flaw()
+        bypass_flaw.cwe_id = "CWE-1"
+        assert_state_equals(workflow.classify(bypass_flaw), state_new)
```
osoukup, line 291: I do not understand what is tested by this last part of this test.

osoukup (approved):
I am raising a few comments but generally this LGTM

apps/osim/tests/test_models.py
costaconrado: I will reword it, the intention is to test that having all requirements for a state doesn't matter if previous state is not met.

apps/osim/tests/conftest.py
costaconrado: makes sense, I will leave it as-is

apps/osim/checks.py
costaconrado: Changed to `field.all().exists()` and added test to this commit.

apps/osim/tests/test_models.py
```diff
@@ -176,131 +182,97 @@ def generate(self, accepts=None, count=0, exclude=None):
...
+        [
+            ("affects", lambda flaw: AffectFactory(flaw=flaw)),
+            ("cvss_scores", lambda flaw: FlawCVSSFactory(flaw=flaw)),
+            ("package_versions", lambda flaw: PackageFactory(flaw=flaw)),
+        ],
+    )
+    def test_relational_property(self, field, factory):
+        """
+        test that properties from a relationship with flaw rejects
+        a empty list and accept while having at least one element
```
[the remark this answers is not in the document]
costaconrado: updated.

apps/osim/tests/test_models.py
costaconrado: makes sense, changed.

apps/osim/tests/test_models.py
costaconrado: changed to the correct commit

apps/osim/tests/test_models.py
costaconrado: OSIM currently only accepts properties and by the time I developed it, I didn't wanted to duplicate this function with an attribute since it has a comment saying it will be converted into attr in future... I believe the temporary solution for this would extend OSIM to accept the function `is_major_incident_temp` as a value. Please take a look in the new changes I made for this commit in `apps/osim/checks.py::CheckParser::desc2property` and see if it would be an acceptable solution.

apps/osim/tests/test_models.py
costaconrado: makes sense, I have replaced the spaces to underscore as framework does and now it should support multiple words

Elkasitu (approved):
LGTM, good job!

PR #347, opened 2023-09-27
Bypass row-level security when applying latest data migration

Elkasitu:
Because of the way that OSIDB handles access to records, database users
don't have access to any rows by default, this would lead to
ObjectDoesNotExist errors when querying certain records from certain
tables and would ultimately lead to the migration's failure.

To fix this, we simply set the user ACLs before executing the migration
and we grant ourselves every permission (r/w public/embargo).

This commit also updates the developer documentation so that this step
is a bit more clear.


Elkasitu:
I have manually deployed this fix and can confirm that the migration ran successfully (whether the resulting data is correct remains TBD)

PR #348, opened 2023-10-05
Configure CORS_ALLOW_HEADERS

d10n:
This variable is used by django-cors-headers
to set the Access-Control-Allow-Headers response header.

I haven't tested the change, but I think it should be correct


costaconrado (approved):
LGTM.

PR #349, opened 2023-10-05
Update changelog for CORS_ALLOW_HEADERS

costaconrado:
This PR merges #348 into master.
Also I've tested locally and it seems to be working.

Original message from @d10n:
This variable is used by django-cors-headers
to set the Access-Control-Allow-Headers response header.

I haven't tested the change, but I think it should be correct


config/settings.py
```diff
@@ -311,5 +312,12 @@
     "OSIDB_DEFAULT_REQUEST_TIMEOUT", default="30", is_int=True
 )
 
+# sets the Access-Control-Allow-Origin response header
+# example value: ["https://osidb.example.com", "https://osim.example.com"]
 CORS_ALLOWED_ORIGINS = get_env("OSIDB_CORS_ALLOWED_ORIGINS", default="[]", is_json=True)
+# sets the Access-Control-Allow-Headers response header; lowercase
+CORS_ALLOW_HEADERS = (
+    *default_headers,
+    *get_env("OSIDB_CORS_ALLOW_HEADERS", default='["bugzilla-api-key"]', is_json=True),
```
osoukup, line 321: Can we extend it with Jira one too? It is already required for the task management

https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/serializer.py#L558

and soon will be required for the tracker management too.

osoukup (approved):
LGTM. One comment.

PR #351, opened 2023-10-11
Fix schema

JakubFrejlach:
This PR:
* fixes OpenAPI schema, there was an incorrect type used (bool vs boolean)
* adds new GH action which checkouts the [osidb-bindings](https://github.com/RedHatProductSecurity/osidb-bindings) repository and tries to regenerate the bindings with the schema from branch which the GH action is run against, also it should only run when the `openapi.yml` file is changed.

Partially addresses OSIDB-230

### DIFF
--- a/collectors/framework/api.py
+++ b/collectors/framework/api.py
@@ -74,7 +74,7 @@ class status(APIView):
                                 },
                                 "error": {"type": "object", "nullable": True},
                                 "is_complete": {"type": "boolean"},
-                                "is_up2date": {"type": "bool"},
+                                "is_up2date": {"type": "boolean"},
                                 "data_models": {
                                     "type": "array",
                                     "items": {"type": "string"},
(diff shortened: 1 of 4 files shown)


collectors/framework/api.py
```diff
@@ -74,7 +74,7 @@ class status(APIView):
                                 },
                                 "error": {"type": "object", "nullable": True},
                                 "is_complete": {"type": "boolean"},
-                                "is_up2date": {"type": "bool"},
+                                "is_up2date": {"type": "boolean"},
```
osoukup, line 77: So this is on me :smile: how could I put `boolean` and `bool` on the very next line ...

osoukup (approved):
LGTM. Great improvement! This could decrease the number of headaches OSIDB does to the bindings (you :smile:)

collectors/framework/api.py
JakubFrejlach: It happens :smile:

PR #354, opened 2023-10-12
fix Celery broker URL

osoukup:
I believe that this change was done by mistake

https://gitlab.cee.redhat.com/prodsec-dev/osidb/-/commit/14b4dd06cf4160efcb1d04475cc57d8ed3bb6cdf#1376cb8c0673765bf96993bd680158d6cb41003c_82_82

I found that while looking at the Flower issue - not sure whether it is the cause or not.


skontar (approved):

osoukup:
So this was a stupid hasty "fix" ... https://www.iana.org/assignments/uri-schemes/prov/rediss

PR #355, opened 2023-10-17
Fix Erratum shipped_dt schema

JakubFrejlach:
Erratum `shipped_dt` in OpenAPI schema is marked as required, however it should also be nullable as we allow null values in the database for this field.


osoukup (approved):
LGTM

PR #356, opened 2023-10-18
Osim workflow

costaconrado:
This PR changes workflows to reflect current IR workflow.

In order to achieve that, this PR creates a new function in CheckParser to compare equal values. 
This PR also creates a model to store cached values from Jira used in new workflow classification.
This Pr also create properties in osidb.models.Flaw to map the new jira attributes.

Closes OSIDB-1319


apps/osim/workflows/rejected.yml
```diff
@@ -0,0 +1,14 @@
+---
+name: REJECTED
+description: >
+  vulnerability does not affect any Red Hat or community products
+  the flaw was analysed but all affects were resolved as notaffected
+  there is no follow-up action required after resolving the affects
```
osoukup, line 6: nitpick: I think that this block of text will compile into a single paragraph connecting the lines by spaces when read into memory and then it would make a weird sentence when the sub-sentences are not joint in human readable format (comma or conjunction). However I am not sure it is a real issue so feel free to ignore this comment.

apps/osim/workflows/rejected.yml
```diff
@@ -0,0 +1,14 @@
...
+priority: 1
+conditions:
+  - state equals REJECTED # check for manual change
+states:
+  - name: REJECTED
+    description: >
+      Task were rejected either by the queue captain or the analyst.
```
osoukup, line 13: nitpick: `Task was`

apps/taskman/models.py
```diff
@@ -40,14 +41,30 @@ class JiraResolution(models.TextChoices):
...
+        JiraStatus.REFINEMENT,
+        JiraResolution.NONE,
+    ),
+    WorkflowModel.OSIMState.SECOND_ASSESSMENT: (
+        JiraStatus.IN_PROGRESS,
+        JiraResolution.NONE,
+    ),
+    WorkflowModel.OSIMState.DONE: (JiraStatus.CLOSED, JiraResolution.DONE),
+    WorkflowModel.OSIMState.REJECTED: (JiraStatus.CLOSED, JiraResolution.WONT_DO),
+}
```
osoukup, line 57: I would suggest not to hard-code it but make it rather a part of the workflow definition as I drafted in
https://github.com/RedHatProductSecurity/osidb/pull/255/files
Or do you think there is reason why defining it in the code is better?

apps/taskman/models.py
```diff
@@ -58,12 +75,22 @@ class Task(NullStrFieldsMixin):
...
 
+    @property
+    def workflow_state(self):
+        for key, value in WORKFLOW_TO_JIRA_STATUS.items():
+
+            if value == (self.status, self.resolution):
+                return key
+        raise KeyError(
+            "Current combination of Jira status and resolution does not met any workflow state mapped."
+        )
```
osoukup, line 86: Conceptual note. I am not sure it will be always possible to clearly map the task to its workflow state only base on the Jira status:resolution. I can imaging eg. default workflow and Major Incident workflow being only different in the Major Incident status value but otherwise sharing the same Jira status:resolution values. Workflow state should be I think rather something what you get when you call `classify` on flaw-task combination (or `classification` if you want the stored one).

apps/taskman/models.py
```diff
@@ -40,14 +41,30 @@ class JiraResolution(models.TextChoices):
...
+
 class Task(NullStrFieldsMixin):
     """
     Data cached from Jira tasks
     """
 
     uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     jira_key = models.CharField(max_length=60)
-    jira_group_key = models.CharField(max_length=60)
+    jira_group_key = models.CharField(max_length=60, blank=True)
```
osoukup, line 67: If not to laborious it would be nice at the final stage of preparing the pull request to squash the changes which fixes the things previously introduced in the same pull request.

apps/taskman/models.py
```diff
@@ -58,12 +75,22 @@ class Task(NullStrFieldsMixin):
...
+            "Current combination of Jira status and resolution does not met any workflow state mapped."
+        )
+
     def update_in_memory(self, issue):
         """Deserializes a jira issue into task model"""
         fields = issue["fields"]
         self.jira_key = issue["key"]
         self.team_id = (
-            fields[JiraCustomFields.TEAM] if fields[JiraCustomFields.TEAM] else ""
+            fields[JiraCustomFields.TEAM]["id"] if fields[JiraCustomFields.TEAM] else ""
```
osoukup, line 93: This would also be better squashed.

osidb/models.py
```diff
@@ -1422,6 +1422,21 @@ def trackers_filed(self):
...
+
+    @property
+    def task_owner(self):
+        """return the current task owner in Jira"""
+        return self.task.owner if hasattr(self, "task") else ""
+
+    @property
+    def task_team_id(self):
+        """return the current team id in Jira"""
+        return self.task.team_id if hasattr(self, "task") else ""
```
osoukup, line 1438: It would be more natural to me to return `None` when there is really nothing. The empty string is still a valid string so I it is less significant nothing :smile: If there is however reason to return the empty string, I do not have a strong preference.

osoukup (commented):
I have a number of mostly smaller comments. It is not because I would not agree with this approach but rather that I think it is important piece to polish. Generally this is a great move forward! Some are actually discussion point. And sorry for a lot of nitpicking :smile:

apps/taskman/models.py
```diff
@@ -40,14 +41,30 @@ class JiraResolution(models.TextChoices):
     TEST_PENDING = "Test Pending"
 
 
+WORKFLOW_TO_JIRA_STATUS = {
+    WorkflowModel.OSIMState.NEW: (JiraStatus.NEW, JiraResolution.NONE),
+    WorkflowModel.OSIMState.TRIAGE: (JiraStatus.REFINEMENT, JiraResolution.NONE),
+    WorkflowModel.OSIMState.PRE_SECOND_ASSESSMENT: (
+        JiraStatus.REFINEMENT,
+        JiraResolution.NONE,
+    ),
```
osoukup, line 50: This is when the four states are maybe not enough. We cannot distinguish these two states here. We can do so based on other workflow conditions but important is that we will need to distinguish them in Jira too - so we can provide complete workflow boards. It can also be done base on other attributes of course so I would like to just clarify it. How do you plan to distinguish them?

apps/osim/checks.py
```diff
@@ -120,3 +122,35 @@ def has_element(instance):
...
+
+    @classmethod
+    def desc2not_equals(cls, check_desc):
+        """attribute value check"""
+        if "_not_equals_" in check_desc:
+            attr, value = check_desc.split("_not_equals_", maxsplit=1)
+            attr = cls.map_attribute(attr)
+
+            if hasattr(cls.model, attr):
+                message = f"check that {cls.model.__name__} attribute {attr} has a value not equals to {value}"
```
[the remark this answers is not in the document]
costaconrado: You're right, sorry about the poor grammar.

apps/taskman/models.py
```diff
@@ -0,0 +1,70 @@
...
+
+class JiraStatus(models.TextChoices):
+    """
+    list of allowable status for tasks in Jira
+    """
+
+    NEW = "New"
+    IN_PROGRESS = "In Progress"
+    REFINEMENT = "Refinement"
+    CLOSED = "Closed"
```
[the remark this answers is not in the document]
costaconrado: Makes sense, changed Jira workflow scheme from `OJA-WFS-010` to `OJA-WFS-012` in stage and prod to support 6 different status in jira.
new status will be `New, Refinement, To Do, In Progress, Review, Closed`.

apps/taskman/models.py
costaconrado: changed to the correct commit it was created.

apps/taskman/models.py
costaconrado: changed to the correct commit it was created.

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
+    description: >
+      A new ticket arrived from incoming queue without any validation
+      or analysis performed yet
     requirements: []
 
-  - name: ANALYSIS
+  - name: TRIAGE
+    description: >
+      Queue captain has taken a ticket off the incoming queue and will
+      evalute if it should be rejected or qualify for further steps in
```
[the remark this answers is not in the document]
costaconrado: my bad, just changed.

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
     requirements:
-      - affects resolved
-      - has cve
-      - has cwe
+      - has team
+      - has trackers
+      - has title
+      - has source
+      - has affects
+      - has impact
```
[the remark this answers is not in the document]
costaconrado: alphabetically sorted, thanks.

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
-      - has description
-      - has source
-      - has title
+      - state equals TRIAGE # check for manual change
 
-  - name: REVIEW
+  - name: PRE_SECONDARY_ASSESSMENT
+    description: >
+      Task qualified for further work and has initial data filled, triage
+      trackers has been filled and product team has been identified
```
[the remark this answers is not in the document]
costaconrado: fixed, thanks.

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
+      - has team
+      - has trackers
+      - has title
+      - has source
+      - has affects
+      - has impact
 
-  - name: FIX
+  - name: SECONDARY_ASSESSMENT
+    description: An analyst assigned the task to themselves.
```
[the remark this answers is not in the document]
costaconrado: added dots in all descriptions, thanks.

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
   - name: DONE
+    description: >
+      Task is done. We commonly expect that a CVE will be 
+      published, but it is not a requirement.
     requirements:
-      # TODO embargoed is not model attribute any more but annotation
-      # so this currently errors out and we need to accout for the change
-      #- not embargoed
-      - trackers resolved
+      - state equals DONE # check for manual change
```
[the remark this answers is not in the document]
costaconrado: The way OSIM is designed a flaw will progress in the workflow if all conditions are met (or if next state has no conditions), also `Check` does not hold any info about the workflow it is in (e.g the name of the current state using it), the way I thought to pass data to it was to have the data in the check text itself. 

If you believe it will look better I could during the `apps.osim.models.py::State::__init__`  change a word/phrase dynamically for the name of the state so we can have a more self-explanatory text in workflow (maybe `manually changed` or something else). WDYT?

apps/osim/workflows/default.yml
```diff
@@ -4,31 +4,41 @@ description: default vulnerability analysis
...
+    description: >
+      Queue captain has taken a ticket off the incoming queue and will
+      evalute if it should be rejected or qualify for further steps in
+      the workflow
     requirements:
-      - has affects
-      - has description
-      - has source
-      - has title
+      - state equals TRIAGE # check for manual change
```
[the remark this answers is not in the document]
costaconrado: I previously had not the understanding that QC would be assigned and then unassigned, I will change that and it will became a problem since as you mentioned it shouldn't impact past states, one way I can immediately fix this would be having a higher priority workflow that starts manually (like in rejection), I can even put in the API a validation to unset assignment when entering in pre-secondary assessment, what do you think? will this be too ugly for the workflow files? I have submitted what I think can solve this, please take a look if it makes it better or worse

apps/osim/workflows/rejected.yml
costaconrado: I will add some pauses (commas / breaks) in the text, those descriptions are not being used anywhere in the code right now, but I believe it will be better even as just "code comment", thank you.

osidb/models.py
costaconrado: I do not have a strong preference either, I have changed to `None`.

costaconrado:
@osoukup I have added a new commit to this PR to comply with the syntax used in #255, this impacts others commits, please take a look on everything again. I also believe now we can have the behavior we want that we talked (about assign / remove owner).

apps/osim/workflows/default.yml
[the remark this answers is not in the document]
osoukup: Adding a special workflow to solve it or doing some hack for specific states does not seem to me as a good solutions. I was thinking about how to address this by the workflow YAML definition. It complicates it of course. The following solutions came to my mind.

First is probably more dirty. It would be introducing **or** condition. Something like `has owner or has team` which would be in the triage status. Then having `has team` in the pre-secondary and `has owner` in secondary. That would be adding the conditions in a way that removing the owner while at the same time adding the team between triage and pre-secondary would not violate it and then it would correctly accumulate. It may however be confusing for the reader and is not well aligned with the definition where in the triage stage there is really no team.

Second is probably cleaner but maybe more extending the definition. It would be adding some flag/mark/something that this requirement **is** or **is not** cumulative meaning that it contributes (or not) to the next states. I am thinking how to make this human readable. Maybe something like adding a requirement rule prefix like **for now has owner** which would tell the cumulative check to skip this requirement for the future state checks.

WDYT?

osidb/models.py
```diff
@@ -1422,6 +1422,16 @@ def trackers_filed(self):
...
 
+    @property
+    def task_owner(self):
+        """return the current task owner in Jira"""
+        return self.task.owner if hasattr(self, "task") else ""
+
+    @property
+    def task_team_id(self):
+        """return the current team id in Jira"""
+        return self.task.team_id if hasattr(self, "task") else None
```
osoukup, line 1433: I would be for abstracting the task/Jira behind calling the properties just `owner` and `team`. WDYT? Also I think that this is not really a flaw functionality so it would be better to either create a mixin for it or eg. extend `WorkflowModel` (which flaw inherits) and its serializer https://github.com/RedHatProductSecurity/osidb/blob/master/apps/osim/workflow.py#L93 with it. That would keep the task management more logically separated from the flaw functionality.

osoukup (changes requested):
I do not think that the last change moved us forward very much. I maybe caused the confusion referencing that OSIM resurrection pull request. I was specifically referring to Jira status:resolution being defined there. It was not much reflected, only in the YAML itself but not in the YAML reading/processing logic. On the other hand you incorporated the automatic transitions which were already refused by IR. Only manual transitions are requested so this boolean is not useful now and only complicates the logic. It also seems that the terminology is being confused which makes impression that we are not aligned or do not understand each other. I am happy to discuss it.

apps/osim/tests/test_endpoints.py
```diff
@@ -211,7 +211,7 @@ def test_promote_endpoint(self, auth_client, test_api_uri_osidb, user_token):
         }
 
         state_first = {
-            "name": WorkflowModel.OSIMState.ANALYSIS,
+            "name": WorkflowModel.OSIMState.SECOND_ASSESSMENT,
```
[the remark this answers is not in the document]
osoukup: I see that you are probably using it as a shortcut. I am sure this is a good idea, people might start to confuse these two. I would be for using SECONDARY instead all the time.

apps/osim/workflows/default.yml
```diff
@@ -1,34 +1,59 @@
 ---
 name: DEFAULT
-description: default vulnerability analysis
+description: default vulnerability workflow.
```
osoukup, line 3: nitpick: I am sorry for being constantly nitpicky about this but it irritates me a bit when there is a dot and no starting uppercase letter :smile: I would either use both or none. This is no sentence anyway so really up to you. And feel free to ignore my unnecessary nitpicks...

osoukup (changes requested):
First of all let me say that this is a great improvement. The API looks well, the logic all seems to be nicely designed, the workflow is as it should be with one exception I think we should change. It is the REJECTED state.

Currently it is the last state of the default workflow. Basically in the DONE state you can do the promote action and continue to the REJECTED one. This is not correct. There should be a very simple rejected workflow for that purpose and some reject API endpoint similar to the promote one leading there.

I have some rather academic complains about the change in the workflow framework logic but I do not want to block this for such reason any more. The thing is that the workflow framework was designed the way that a future state in workflow can only be entered if all the conditions of the previous states are still fulfilled - and you can theoretically adjust back if you stop fulfilling something. However I know the **owner** condition which is TRIAGE requirement but there is no owner in PRE_SECONDARY_ASSESSMENT breaks this. We can eventually think about this later but let us move forward.

apps/osim/workflow.py
```diff
@@ -89,6 +95,28 @@ def classify(self, instance, state=True):
...
+    def validate_classification(self, instance, workflow, state):
+        """
+        This method will evaluate if it is possible to classify the current
+        instance as a target state
+
+        Raise exception if instance lacks requirements for the target state
+
+        This method does NOT update the flaw classification
+        """
+        for fw_workflow in self.workflows:
```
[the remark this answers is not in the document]
costaconrado: framework, I will rename the parameter of the function to target_workflow and use the framework one as workflow instead, I believe it will improve the readability

apps/osim/tests/test_endpoints.py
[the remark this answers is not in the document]
costaconrado: Makes sense, removed all short versions of it.

apps/osim/workflows/default.yml
costaconrado: Sorry about that, fixed.

apps/osim/tests/test_endpoints.py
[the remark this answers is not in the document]
osoukup: I see I missed an important **not ** in the second sentence but you understood what I mean :smile:

costaconrado:
@osoukup I've changed some commits:
1. new commit `fix flaw.affects_resolved filters`: I noticed that affects_resolved would present a weird behaviour saying a not affect is not resolved.
2. new commit `add manual state reject`: to cover manual changes to reject state.
3. chenged `reflect new IR workflow in osim`: added a workflow for rejected that would be used only if its manually changed to this state.

apps/osim/workflow.py
```diff
@@ -257,3 +257,24 @@ def promote(self, save=True, jira_token=None):
...
+
+    def reject(self, save=True, jira_token=None):
+        """
+        this is the cannonical way of rejecting a flaw / task
+
+        This method will change instance state to rejected if all conditions are met
+
+        Raise exception if instance lacks requirements for the rejected state
+        """
+        reject_workflow = "NOT-AFFECTED"
```
osoukup, line 273: I see that you are reusing the example NOT-AFFECTED workflow. I do not see that you would be changing it. There are a lot of issues with that. Simply starting with the name, it is clearly wrong. We should have some REJECTED workflow instead. Then description of the workflow is about something completely different. The conditions are unnecessary for the flaw draft just being rejected. There are three states while it is not clear why we should have more then one. There is no state REJECTED in that workflow. This workflow should be just dropped and not reused. It seems that the logic is already far from the original **osim** module aim and we will not have different workflows for different flaw contexts (like Major Incident, not affected, embargoed etc.) as originally intended. It seems that now it only makes sense to have DEFAULT and REJECTED.

apps/osim/workflows/not-affected.yml
```diff
@@ -1,25 +1,16 @@
...
+  Vulnerability does not affect any Red Hat or community products
+  the flaw was analysed but all affects were resolved as notaffected.
+priority: 0
 conditions:
-  - has affects
   - affects resolved
   - affects notaffected
 states:
-  - name: DRAFT
+  - name: REJECTED
```
osoukup, line 11: I see that my previous comment is not fully applicable after this commit (I am reviewing chronologically commit-wise). Still REJECTED does not mean not affected. Not affected means that we created affects and after analysis we closed all of them with **notaffected** affectedness value. Rejection may mean that we close the flaw draft right away without even creating the affects. The reasons may vary. Usually it would be because we do not ship that thing but also that we already know about this or it is some total spam.

osoukup (changes requested):
Almost LGTM. The logic seems OK to me. It is different from the original intents of the **osim** module but the IR/VM wants it in such way. I would like the NOT-AFFECTED workflow to be turned into a proper REJECTED workflow.

A side note, it is a bit confusing for the reviewer when you reference something but introduce it in the later commit only - REJECTED state. I am doing the reviews of large pull request going chronologically commit-wise so I can follow the implementation process.

osoukup (approved):
LGTM

PR #359, opened 2023-10-19
Bypass row-level security in CVSS scores migrations

jobselko:
This PR is the follow-up to #347. It updates migrations for CVSS scores to contain needed ACLs to bypass row-level security.

Fixes OSIDB-1455

### DIFF
--- a/osidb/migrations/0099_fix_flawcvss.py
+++ b/osidb/migrations/0099_fix_flawcvss.py
@@ -1,0 +1,98 @@
+"""
+Written manually on 2023-10-20.
+This fixes migration 0092, which was missing the correct ACLs.
+"""
+
+from osidb.core import set_user_acls
+
+from django.conf import settings
+from django.db import migrations
+from itertools import islice
+
+BATCH_SIZE = 1000
+
+
+def generate_flawcvss(apps):
+    Flaw = apps.get_model("osidb", "Flaw")
+    FlawCVSS = apps.get_model("osidb", "FlawCVSS")
+
+    for flaw in Flaw.objects.iterator():
+        # Default values
+        nist_cvss2_score, nist_cvss2_vector = 0.0, ""
+        nist_cvss3_score, nist_cvss3_vector = 0.0, ""
+        rh_cvss2_score, rh_cvss2_vector = 0.0, ""
+        rh_cvss3_score, rh_cvss3_vector, rh_cvss3_comment = 0.0, "", ""
+
+        other_fields = {
+            "acl_read": flaw.acl_read,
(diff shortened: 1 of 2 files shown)


osidb/migrations/0092_migrate_cvss_to_flawcvss.py
```diff
@@ -60,6 +63,15 @@ def generate_flawcvss(apps):
 
 
 def forwards_func(apps, schema_editor):
+    set_user_acls(
```
osoukup, line 66: @Elkasitu I am thinking that the pieces of code using this construction will soon miss the private stuff

osoukup (commented):
This seems to me to fix the migration. I am however not sure that this is the correct solution as this migration is already an applied one and not even the most recent one. One thing is that it is not going to be automatically run again. I am not sure it is possible to easily re-run it

jobselko:
@osoukup You are right. Updating old migrations when new ones are on top of them is not a good idea. A better solution would probably be to create new migrations with the same code as the original ones. I will wait for the opinion of others.

Elkasitu:
I think the easiest will be creating a new data migration in which we import the `forwards_func` from these two old migrations and re-run them in a new migration with ACLs bypassed

osidb/migrations/0092_migrate_cvss_to_flawcvss.py
Elkasitu: For migrations it doesn't matter that much now, but I'll create an `ALL_GROUPS` and replace calls like this with that

jobselko:
@Elkasitu It is not possible to import a file starting with numbers without using `__import__` or `importlib`, so I just copied the code. I think it should be okay, as this is just a migration.

osidb/migrations/0099_fix_flawcvss.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+
+
+def forwards_func(apps, schema_editor):
+    set_user_acls(
+        settings.PUBLIC_READ_GROUPS
+        + [
+            settings.PUBLIC_WRITE_GROUP,
+            settings.EMBARGO_READ_GROUP,
+            settings.EMBARGO_WRITE_GROUP,
```
jobselko, line 72: @Elkasitu `DEVELOP.md` now contains also `settings.INTERNAL_READ_GROUP, settings.INTERNAL_WRITE_GROUP`. Should I add them too?

Elkasitu: Not necessary at this stage since nothing uses those groups yet

Elkasitu (approved):

PR #361, opened 2023-10-20
Introduce ALL_GROUPS setting and simplify RLS bypass

Elkasitu:
Bypassing RLS in certain scenarios is a recurring pattern which requires developers to manually call `set_user_acls` and pass to it all necessary groups for their code to do its job.

As the number of groups grows this becomes a PITA, introducing an `ALL_GROUPS` setting which includes all user groups (admin/management groups are not included) makes this process easier and less error-prone.

While a `bypass_rls` helper or systematic RLS bypass within the collector framework were considered, these options either provide very little gain over calling `set_user_acls(settings.ALL_GROUPS)` or be too magic.

The correct long-term solution would be to have a database user which bypasses RLS and use that for both migrations and celery workers.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -185,7 +185,7 @@
         "filename": "config/settings_ci.py",
         "hashed_secret": "efacc4001e857f7eba4ae781c2932dedf843865e",
         "is_verified": false,
-        "line_number": 48,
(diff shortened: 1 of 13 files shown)


osoukup (approved):
LGTM

PR #363, opened 2023-10-23
Introduce the Snippet model

jobselko:
Related to [OSIDB-1487](https://issues.redhat.com/browse/OSIDB-1487)


jsvob (approved):

osidb/models.py
```diff
@@ -1501,6 +1501,37 @@ def tasksync(
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # set internal ACLs
+        self.set_internal()
+
+    # internal primary key
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    source = models.CharField(choices=Source.choices, max_length=100)
```
osoukup, line 1526: I know this is not completely straightforward to be done jointly for all the collectors but I would prefer more precise identification of the scraping above when (`TrackingMixin`) and the source. Each source is obviously the origin of a lot of snippets and having additionally only time may of course help to identify the exact piece in the source (maybe not) but it would not be were easy match. I would like to prevent identifying the snippet based on its data.

Ideally we should have an info on what exact entity (URL, ID, something) was scraped from that source. Completely OK if this more precise identification is some kind of union data type and the meaning depends on the source - I guess for NVD it can be CVE ID, for Bugzilla bug ID, not sure about the others. Then it would be very simple to reconstruct the exact URL or something and with the timestamp we have very exact information of the data the snippet was created from.

osoukup (changes requested):
I have one improvement (described in the comment) I would really like to see here but I am opened to discussion. Otherwise LGTM

osoukup (commented):
I do not request the change any more and leave it to the collector implementers to decide

osoukup (approved):

Elkasitu (approved):

PR #364, opened 2023-10-24
Osim taskman integration

costaconrado:
This PR introduces a query building based interaction with Jira.

This new query generator reduces the number of endpoints exposed by Taskman simplifying the interactions with the module.
This PR also disables tests from old service that will bi completely disabled as a following task (OSIDB-1490).
This PR also masks Jira Status/Resolution from queries, enabling us to freely map any status from Jira and exposing OSIM status for users that makes sense with new IR workflow.

Closes OSIDB-1321.


apps/taskman/jira_serializer.py
```diff
@@ -11,6 +11,89 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # flaw is required during creation
+        if self.context.get("request") and self.context["request"].method == "POST":
+            self.fields["flaw"].required = True
+        else:
+            self.fields["flaw"].required = False
+
+    def create(self, validated_data):
+        state = validated_data.pop("state", WorkflowModel.OSIMState.NEW)
```
osoukup, line 46: Does this mean that we can create eg. a closed task?

apps/taskman/jira_serializer.py
```diff
@@ -11,6 +11,89 @@
 """
 from rest_framework import serializers
 
+from apps.osim.workflow import WorkflowModel
+
+from .models import WORKFLOW_TO_JIRA_STATUS, JiraResolution, JiraStatus, Task
+
+
+class TaskSerializer(serializers.ModelSerializer):
```
osoukup, line 19: I am thinking about the task management use-cases and that they are very limited. The task creation should probably not eve be allowed through the API. It should be an automatic side-effect of the flaw creation. The possible task updates are also very limited - it is just promotion to the next state or rejection. Therefore I am not sure whether the full API makes much sense. The user of the API should IMO not even be aware of that there is a task in background. (s)he just works with the flaw and its status:workflow by the simple handles like promote.

osoukup (changes requested):
I am opened to discussion but I do not think this is the correct approach. I think that tasks should be more abstracted and in the background of the flaw management. Not directly operated through the task API.

osoukup (approved):
LGTM

osoukup (approved):
LGTM

PR #367, opened 2023-10-31
Feed collector

jsvob:
Closes OSIDB-958


collectors/feed/tests/test_collectors.py
```diff
@@ -0,0 +1,113 @@
...
+]
+
+
+@pytest.mark.django_db(True)
+class TestFeedCollector:
+    """
+    Feed collector tests
+    """
+
+    @pytest.mark.vcr
```
jsvob, line 95: Open question:

How to get it running with detect-secrets? The recorded VCR cassette contains responses that detect-secrets doesn't like.

```
ERROR: Potential secrets about to be committed to git repo!

Secret Type: Base64 High Entropy String
Location:    collectors/feed/tests/cassettes/test_collectors/TestFeedCollector.test_smoke_test.yaml:292

Possible mitigations:
  - For information about putting your secrets in a safer place, please ask in
    #security
  - Mark false positives with an inline `pragma: allowlist secret`
    comment

```

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+    FEEDS_FILE,
+    MAX_COLLECTOR_WORKERS,
+    REQUEST_TIMEOUT,
+)
+from .keywords import check_keywords
+from .models import FeedEntry, SkippedFeedEntry
+from .utils import find_cves, find_cvss, find_cwes, html_to_text
+
+log = logging.getLogger(__name__)
+
```
jsvob, line 25: Open problem (not related to this line, just putting it in a place that has discussion threading):

Bandit (as run via the githook) fails without an explanation.

Dockerfile
```diff
@@ -55,3 +55,9 @@ COPY . /opt/app-root/src
 
 RUN chgrp -R 0 /opt/app-root && \
     chmod -R g=u /opt/app-root
+
+# TODO remove the following block before merge!
```
jsvob, line 59: Open problem: elinks dependency

@Elkasitu You mentioned you are working on it?

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+# TODO: Refactor from the original source to a more efficient structure better matching OSIDB style
+#       - NOTE: If you diff this with the original, the changes are visible
```
jsvob, line 27: FYI, I'm working on refactoring it (what it inside a class, what is outside, what classes are there, etc). It won't be as directly comparable to the original code after the refactor.
For now, the value in non-refactored code is that it should be more obvious whether there are parts brought over that are either wrong or not needed anymore.

collectors/feed/config.yml
```diff
@@ -0,0 +1,89 @@
...
+  - chromium-browser
+  meta:
+  - description: "The Google Chrome Releases feed is a collection of
+      release notes that also include security advisories. It includes
+      advisories for various Google products; we are only interesting in
+      the Stable Channel Chrome updates."
+  - reason: "This feed is the best source of information to learn of new
+      Chrome vulnerabilities."
+
+## - url: https://nvd.nist.gov/feeds/json/cve/1.1/nvdcve-1.1-modified.json.gz
```
jsvob, line 68: Will be removed.

collectors/feed/models.py
```diff
@@ -0,0 +1,28 @@
...
+    Temporary container for feed entries before they are saved as new Snippet model instances.
+    """
+
+    url: str
+    cves: list
+    title: str
+    text: str
+    name: str
+    source: str
+    components: list
```
jsvob, line 15: Do we use that for anything?

collectors/feed/keywords.py
```diff
@@ -0,0 +1,523 @@
+# TODO tests
+
+import re
```
jsvob, line 3: FYI: I'm **not** going to refactor this file. Just add tests. If you have a different opinion, this is the place to voice it and change my plan. Thanks.

jsvob:
@jsvob @Elkasitu @RedHatProductSecurity/osidb-devs Please review despite the "Draft" in title. I'm especially interested in comparing thoughts on creating Snippets with Jitka and Adrian (is this the right way? what to put in contents?).

jsvob:
@jobselko @Elkasitu @RedHatProductSecurity/osidb-devs Please review despite the "Draft" in title. I'm especially interested in comparing thoughts on creating Snippets with Jitka and Adrian (is this the right way? what to put in contents?).

collectors/framework/models.py
```diff
@@ -570,3 +570,12 @@ def inner(*args, **kwargs):
         return inner
 
     return wrapper
+
+
+class KnownFeedEntry(models.Model):
```
osoukup, line 575: I do not like this approach as it is collector-specific (does not mean that I cannot be outvoted :smile:).

I would prefer creating a uniquely identifiable snippet instead (that is the reason why I was advocating for a snippet identifier). The following runs would look for the known scrape results among the snippets instead.

We could add a flag like **create_flaws** or something which could be set to **False** in the case that the snippet actually contains something we do not care for. The advantage would be that we would have a single place where to look for already know scrape results - both the ones resulting in flaws and the ones we ignored.

There would be a single storage for these, the **Snippet** model, and a single canonical way to identify them, **source:identifier** combination. No collector-specific functionality would be needed.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+    return feed_data
+
+
+def get_feed_entries():
+    """Concurrently scrape data from all feeds."""
+    log.info("Fetching new feeds")
+
+    feed_data = get_feed_data()
+
+    with ThreadPoolExecutor(MAX_COLLECTOR_WORKERS) as executor:
```
osoukup, line 238: In production the threading does not work for some reason. It seems that each pod runs on a single thread only. We encountered this when attempting to parallelize the flaw collector - it works locally but not in the production.

Also we should eventually think about what is the good way to do this. The collector framework prevents multiple concurrent runs of the same collector. We can however use Celery for the concurrency too so it is a question on which level we want to do it.

osoukup (changes requested):
I have not looked at every detail but mostly it LGTM. I made a few comments out of which the approach to the results storing is something I would really like to change unless the others are for the currently introduced approach.

collectors/feed/collectors.py
jsvob: There's currently no effective use of threading. This is a leftover from the original. I'll refactor this away. I prefer to gather feedback [like the one about Snippets](https://github.com/RedHatProductSecurity/osidb/pull/367#discussion_r1378609362) before actually doing the refactoring. See comment https://github.com/RedHatProductSecurity/osidb/pull/367#discussion_r1378022099 for additional explanation.

collectors/framework/models.py
jsvob: I agree this will be the correct approach. It should work with or without the identifiers but I slightly favor having that field in there because it will be used for lookups heavily (each collector can still decide how to use that field).

collectors/feed/models.py
jsvob: Aha. So I'll use the static per-feed list of components and give each snippet of that feed the same list of components, leaving it up to the analyst to decide whether all of them are relevant? This is the relevant part of the per-feed configuration: https://github.com/RedHatProductSecurity/osidb/blob/8b01f4604fbb6f2270e6fa4b5b75d49ba012333e/collectors/feed/config.yml#L45-L47

jsvob: And 2nd example: https://github.com/RedHatProductSecurity/osidb/blob/8b01f4604fbb6f2270e6fa4b5b75d49ba012333e/collectors/feed/config.yml#L82-L84

collectors/feed/utils.py
```diff
@@ -0,0 +1,204 @@
...
+    """Use elinks to convert HTML to plain text.
+
+    Other HTML->text solutions were either too heavy weight, returned ugly text, or failed to
+    parse certain parts of the HTML. (BeautifulSoup is great at scraping but not so good at
+    rendering plain text; html2text had terrible output if the parsed HTML included a table.)
+
+    Update 2018-08-27: Research was done in various options of rendering website contents as plain
+    text in bug https://bugzilla.redhat.com/show_bug.cgi?id=1558705.
+    """
+    with NamedTemporaryFile(prefix="html_to_text.") as tmpfile:
```
Elkasitu, line 186: can `elinks` not read from stdin? because if so this could be simplified

collectors/framework/models.py
Elkasitu: As already mentioned, this can just be an attribute in `Snippet.content`

osidb/management/commands/collectfeed.py
```diff
@@ -0,0 +1,51 @@
...
+                self.stdout.write(traceback.format_exc())
+        else:
+            self.stdout.write(self.style.SUCCESS("OK"))
+
+        # Revert the enhanced logging
+        if verbosity > 1:
+            logging.basicConfig(
+                level=old_level,
+            )
+            feed_log.setLevel(old_level)
```
Elkasitu, line 51: see https://docs.python.org/3/library/logging.html#logging.disable which is how `syncflaws` does it

collectors/feed/tests/test_collectors.py
Elkasitu: Per the documentation, call `make update-secrets` and avoid using the `pragma` directive

collectors/feed/tests/test_collectors.py
```diff
@@ -0,0 +1,113 @@
...
+            },
+            {
+                "reason": "This list is a single source of all vendor advisories. Rather than subscribing to feeds of all the vendors, we use this feed."
+            },
+        ],
+    },
+]
+
+
+@pytest.mark.django_db(True)
```
Elkasitu, line 89: The convention is to use the following autouse fixture, see osidb/tests/conftest.py:

```python
@pytest.fixture(autouse=True)
def enable_db_access_for_all_tests(db):
    pass
```

Elkasitu (changes requested):
Unsure which bits are from the original implementation and which aren't so I just commented where it made sense to me

osidb/models.py
```diff
@@ -1501,6 +1501,7 @@ class Source(models.TextChoices):
         """
 
         NVD = "NVD"
+        FEED = "FEED"
```
jobselko, line 1504: I do not think `Feed` is a good name for a source. It is only a general way to describe a subset of sources. I would prefer to use the names of parsed files as sources.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+                        skipped=True,
+                    )
+                )
+                continue
+
+            cves = find_cves(entry.title + text)
+            cwes = find_cwes(entry.title + text)
+            cvss = find_cvss(text)
+
+            blocklist, allowlist = check_keywords(entry.title + " " + text)
```
jobselko, line 160: I think we do not need `blocklist`. Instead of that, we can just easily ignore the non-matched items.

collectors/feed/config.yml
```diff
@@ -0,0 +1,89 @@
...
+## - url: https://www.us-cert.gov/ncas/all.xml
+##   name: US-CERT Advisories
+##   source: internet
+##   meta:
+##   - description: "The US-CERT Advisories feed includes vulnerability
+##       reports of the most frequent, high-impact security incidents
+##       currently being reported to the US-CERT."
+##   - reason: "This feed is watched mainly for redundancy to ensure we caught
+##       all security issues reported elsewhere. US-CERT rarely published an
+##       original advisories."
```
jobselko, line 30: Do we want to remove this source?

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+                "Packet Storm Advisory Files",
+            ]
+            text = None
+            skip_processing = False
+            try:
+                if self.feed_name in special_feeds:
+                    text = self.get_full_feed_text(entry.link)
+                else:
+                    if "summary" not in entry:
+                        log.info(
```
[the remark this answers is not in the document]
jsvob: My idea was to see these only when run on the command line (collectfeed.py reconfigures logging temporarily). With verbosity 2, info but not debug can be seen (which gives an overview of what it does), with verbosity 3, the debugs are visible as well.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+                )
+                skip_processing = True
+
+            if not text:
+                # TODO: Consider retrieving text using modified get_full_feed_text for node.js
+                log.info("No text for feed entry: %s", entry.link)
+                skip_processing = True
+
+            if skip_processing:
+                # Prevent reprocessing next time
```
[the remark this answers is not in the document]
jsvob: Next run of the collector. I'll fix the wording.

collectors/feed/utils.py
jsvob: Yes, it apparently can!

```
# echo "<html><body><b>hello</b> world</body></html><asdfsadf>" | elinks -no-home -dump -dump-width 80 -no-numbering -no-references
   hello world
```

collectors/feed/config.yml
jsvob: It seems it's defunct, so most probably yes.

collectors/feed/collectors.py
jsvob: I'll perform measurements, but even before that, I don't think that would work. For example, the allowlist contains the word "kernel". Reading through the blocklist, there's a number of things that interact with a kernel in an irrelevant way, have their own irrelevant version of the Linux kernel, or even have their own non-Linux kernel.

osidb/models.py
jsvob: Do you want as many sources as there are sources in the Feed collector's config file or did I misunderstand?

jsvob: That means that when `confix.yaml` is modified with a new feed, the Source class will have to be modified as well. Is that still what you meant?

jobselko: Yes.

jsvob: Clear now. Thank you.

collectors/feed/config.yml
```diff
@@ -0,0 +1,89 @@
+---
```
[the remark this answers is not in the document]
jsvob: How to solve the release-dependent changes potentially required in models then? See discussion https://github.com/RedHatProductSecurity/osidb/pull/367#discussion_r1383283922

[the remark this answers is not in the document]
jsvob: This is too complex to be in the scope of this task (I have no idea how to do that and if I did, I have no idea how to also make it work with local podman-compose). If you really want this, we can file a follow-up Task (I guess not a Story because I don't see end-user-visible value). Closing the conversation until then.

collectors/feed/collectors.py
jsvob: Without the blocklist, today's run is not affected. With creating Snippets only for things that explicitly trigger the allowlist (like you do in https://github.com/RedHatProductSecurity/osidb/pull/374 ) it skips everything.

Since the task is about porting the collector and not about reconsidering what exact data it fetches, and the original collector apparently took a lot of time to fine-tune w.r.t. the filtering, I think any functional change whatsoever in the filtering algorithm is out of scope. Not pursuing this further.

jobselko: > With creating Snippets only for things that explicitly trigger the allowlist (like you do in #374 ) it skips everything.

How? You want to create a snippet if a description contains any of the allowed keywords, and these are specified in the allowlist. Am I missing something?

osoukup: If having the blocklist and not having it gives the same results then throw it away. If not and it was in the original Assembler logic then do not change the logic and only migrate it. Any changes of the logic should be subjects of the future enhancements not to unnecessarily complicate the migration.

jsvob: > You want to create a snippet if a description contains any of the allowed keywords, and these are specified in the allowlist.

This is not how the feed collector in Assembler works. It generates snippets/tickets for everything by default, and skips feed entries with at least one blocklisted word iff that entry contains zero allowlisted words.

> Am I missing something?

Either of us is missing something, but IDK who. I _think_ the goal of the task is to port the collector and have the flaw drafts/snippets created directly instead of through Assembler so that the effective end result is similar from IR point of view, and I _think_ the goal is not changing how the collector works by e.g. eliminating the majority of what it generates.

jsvob: Saw Ondrej's message only after posting my reply to Jitka. Closing the thread then.

collectors/feed/tests/test_collectors.py
jsvob: Thank you, that helped.

collectors/feed/collectors.py
jsvob: Didn't end up adding such a comment, because I think reading the code is much more helpful, than any comment would be. It could also devolve into a screen-long risk analysis. If you have a better idea, please let me know.

Dockerfile
jsvob: Done

collectors/feed/models.py
jsvob: Done

collectors/feed/keywords.py
jsvob: Please file a task if that's necessary. IMO it wouldn't make sense because there was [another piece of feedback that locks the configuration to code releases](https://github.com/RedHatProductSecurity/osidb/pull/367/#discussion_r1383283922) anyway.

In any case, I think that is out of scope and would complicate and protract this story.

collectors/framework/models.py
jsvob: Done. Only `Snippet` model is used.

collectors/feed/collectors.py
jsvob: Threading removed.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+    CHROMIUM_REGEX,
+    FEEDS_FILE,
+    MAX_COLLECTOR_WORKERS,
+    REQUEST_TIMEOUT,
+)
+from .keywords import check_keywords
+from .models import FeedEntry, SkippedFeedEntry
+from .utils import find_cves, find_cvss, find_cwes, html_to_text
+
+log = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
jsvob: Done.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+        return new_feed_entries
+
+    @staticmethod
+    def filter_known_feed_entries(feed_entries):
+        """From a list of feed entries, return only those that are new (i.e. not stored in DB)."""
+        new_urls = [entry.link for entry in feed_entries]
+
+        known_feed_entries = KnownFeedEntry.objects.values_list(
+            "url", flat=True
+        ).filter(url__in=new_urls)
```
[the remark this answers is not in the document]
jsvob: Added an explanatory comment:

```python
        # Note that this limits the query to the fetched URLs, not the other way around.
        # This way, the "in" operator always works with a list of a limited size and
        # the resulting list always has limited size.
        # We can't control how long history a feed provides, hence no time-based filtering.
        known_feed_entries = (
            Snippet.objects.values_list("content__url", flat=True)
            .filter(source=feed_name)
            .filter(content__url__in=fetched_entry_urls)
        )
```

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+
+            # Google Chrome Releases: receives posts for all Chrome products and branches,
+            # but we're only interested in the Stable Channel updates. Skip the rest.
+            if "GoogleChromeReleases" in self.url and CHROMIUM_REGEX.match(entry.title):
+                continue
+
+            # Ignore RHSAs in the Packet Storm advisory feed
+            if (
+                "packetstormsecurity" in self.url
+                and "Red Hat Security Advisory" in entry.title
```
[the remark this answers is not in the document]
jsvob: Done

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+                response = requests.get(feed_url, timeout=REQUEST_TIMEOUT)
+                response.raise_for_status()
+            except RequestException as exc:
+                log.error("Error when requesting %s: %s", feed_url, exc)
+                return
+
+            return response.text
+
+        else:
+            log.error("Unknown URL to fetch full text for: %s", feed_url)
```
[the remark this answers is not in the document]
jsvob: Done

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+        """Get data for new entries for a particular feed and return FeedEntry instances."""
+        feed_entries = self.get_new_feed_entries()
+
+        feed_models = []
+        for entry in feed_entries:
+            # Special rules below to ignore certain posts.
+
+            # Google Chrome Releases: receives posts for all Chrome products and branches,
+            # but we're only interested in the Stable Channel updates. Skip the rest.
+            if "GoogleChromeReleases" in self.url and CHROMIUM_REGEX.match(entry.title):
```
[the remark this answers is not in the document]
jsvob: Done

osidb/management/commands/collectfeed.py
jsvob: Thanks. Used this feedback to make the logging work mostly the same as with syncflaws: The commandline util works the same and the collector logs roughly the same information and the logs can therefore be accessed roughly the same way. It's then up to the configuration whether the logs are visible.

collectors/feed/collectors.py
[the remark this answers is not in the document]
jsvob: Logging should now be resolved IMO - see the comment https://github.com/RedHatProductSecurity/osidb/pull/367/#discussion_r1404631657

collectors/feed/collectors.py
```diff
@@ -0,0 +1,281 @@
...
+    """Scrape a particular feed."""
+    feed_name = feed_data.get("name")
+    url = feed_data.get("url")
+    source = feed_data.get("source")
+
+    if not url or not feed_name or not source:
+        raise ValueError(
+            "Missing mandatory argument (url, feed name, source) in feed data: %s",
+            feed_data,
+        )
```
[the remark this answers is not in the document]
jsvob: Considered and ended up not using it. Including a schema would make the MR substantially larger. Instead of explicit error checking, used the square bracket operator / `__getitem__` so that an exception would be raised on some missing items. Since the config is stored in the exact same repo as the code, it doesn't make sense to have a double standard and protect from unskilled/uninformed/asinine changes in one file and not in the rest.

collectors/feed/collectors.py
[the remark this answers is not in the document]
jsvob: Done. The particular place in code disappeared, but in another place that has a similar meaning, a better comment is present: `# Prevent reprocessing (or fetching) the same entries in future Collector runs.`

collectors/feed/models.py
```diff
@@ -0,0 +1,28 @@
+from typing import NamedTuple
+
+
+class FeedEntry(NamedTuple):
```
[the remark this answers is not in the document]
jsvob: NamedTuple eliminated.

collectors/feed/utils.py
```diff
@@ -0,0 +1,204 @@
...
+    "local file": "CWE-552",
+    "out-of-bounds write": "CWE-787",
+    "csrf": "CWE-829",
+    "cross-origin": "CWE-829",
+    "same-origin": "CWE-829",
+    "permission checks bypass": "CWE-829",
+    "permission check bypass": "CWE-829",
+    "protections bypass": "CWE-829",
+    "protection bypass": "CWE-829",
+}
```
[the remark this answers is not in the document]
jsvob: Out of scope IMO. Also please see comment https://github.com/RedHatProductSecurity/osidb/pull/367/#discussion_r1404630174

collectors/feed/utils.py
jsvob: Done.

collectors/feed/tests/test_collectors.py
jsvob: Done. Thank you.

osidb/models.py
jsvob: Done.

collectors/feed/config.yml
jsvob: It seems this has been folded into CISA, which AFAIK we already process elsewhere. I consider this done.

collectors/feed/collectors.py
```diff
@@ -0,0 +1,359 @@
...
+                #   hence refactoring not worthwhile.
+                (
+                    snippet_format_cvss_score,
+                    snippet_format_cvss_vector,
+                ) = rh_format_cvss.split("/", 1)
+                content["cvss3"] = {
+                    "score": snippet_format_cvss_score,
+                    "vector": snippet_format_cvss_vector,
+                    # "issuer": unknown!
+                }
```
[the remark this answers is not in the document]
jsvob: I copied the structure from other Snippet uses in `master`. Shall I fix it only here based on your feedback and we'll fix the other uses in a different PR?

collectors/feed/collectors.py
```diff
@@ -0,0 +1,359 @@
...
+            logger.debug("new allowed Snippet data: %s", repr(new_snippet.__dict__))
+            # Save separately so that log shows the context if it fails.
+            new_snippet.save()
+
+        # Prevent reprocessing (or fetching) the same entries in future Collector runs.
+        new_blocked_snippet_urls = all_urls - allowed_snippet_urls
+        for url in new_blocked_snippet_urls:
+            content = {
+                "url": url,
+                "create_flaws": False,
```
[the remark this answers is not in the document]
jsvob: Makes sense. If we go without the switch as you describe, then it will be crucial for the collector itself to not hand over the would-be-`"create_flaws": False` Snippets to the Flaw creator, right?

If true, then I'll add a TODO to the code to send only `"create_flaws": True` to the Flaw creator, once it starts to exist, and to eliminate `create_flaws` if possible (the Flaw creator will ignore it anyway). Sound right?

The Snippet-to-Flaw creator doesn't exist yet AFAIK, so that is probably the next best thing.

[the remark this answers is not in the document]
osoukup: Actually you have a good point. We can leave it on the collector at this point when it is just RSS feed collector and eventually turn it into a proper attribute if we encounter another like this. For the time being your collector would just not call the flaw creation logic.

Elkasitu (approved):
LGTM, just some minor comments

dist/README
```diff
@@ -0,0 +1,18 @@
...
+
+curl -L --output gpm-1.20.7-17.el8.src.rpm http://vault.centos.org/centos/8/AppStream/Source/SPackages/gpm-1.20.7-17.el8.src.rpm
+
+curl -L --output elinks-0.12-0.58.pre6.el8.src.rpm http://vault.centos.org/centos/8/PowerTools/Source/SPackages/elinks-0.12-0.58.pre6.el8.src.rpm
+
+COPYING:
+
+The files gpm-libs-1.20.7-17.el8.x86_64.rpm and gpm-1.20.7-17.el8.src.rpm are distributed under the license stored in the file COPYING_gpm.
+
+The files elinks-0.12-0.58.pre6.el8.x86_64.rpm and elinks-0.12-0.58.pre6.el8.src.rpm are distributed under the license stored in the file COPYING_elinks.
```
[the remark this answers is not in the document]
Elkasitu: I see that there's both `COPYING_{gpm,elinks}` and `LICENSE_{gpm,elinks}`, the former seems to be added in the wrong commit, could you explain the difference in purpose between the two?

collectors/feed/collectors.py
```diff
@@ -0,0 +1,359 @@
...
+    def collect(self) -> str:
+        """
+        collector run handler
+
+        on every run the configured feed entries are fetched and saved to snippets
+        """
+
+        # set osidb.acl to be able to CRUD database properly and essentially bypass ACLs as
+        # celery workers should be able to read/write any information in order to fulfill their jobs
+        set_user_acls(settings.ALL_GROUPS)
```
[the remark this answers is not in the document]
jobselko: ACLs are set in `Snippet`'s `__init__`.

dist/README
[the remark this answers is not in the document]
jsvob: The explanation is that I made a mess without enough thinking and checking. Wanted to rename `LICENSE_` to `COPYING_` but amended the wrong commit **and** forgot to delete the old files on top of that.

dist/install_elinks.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+if [[ "$(sha256sum gpm-libs-1.20.7-17.el8.x86_64.rpm | cut -d' ' -f1)" == "9fe5b0c5fee372362d423e17b56f01d1c15cc0725bd989bdfc742f30c1119e0b" ]]
```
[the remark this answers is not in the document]
jsvob: I'd prefer to because if the files ever get re-downloaded or changed, this is an easily-retrievable record of change in git (probably harder to miss than a whole-file change) *and* a sanity check that the container image is really built the way we want. Both reasons are very weak, given the fact that it's all in the same repo. I can remove the check if you want.

jsvob:
Closing without merge. Not deleting branch for future reference. TODOs in code in sync with unresolved PR conversations.

PR #374, opened 2023-11-06
Update NVD collector and add Snippet creation

jobselko:
This PR refactors NVD collector and adds an ability to create snippets. If a snippet for a specific piece of information does not exist, it will be made. If a snippet already exists, it will be kept without any changes. This can be turned on or off via `SNIPPET_CREATION_ENABLED`. The default is off.

This PR does not yet enable the creation of new candidate flaws based on snippets or the connection of flaws and snippets. It also misses a functionality for checking whether a snippet's content is important to us. These things will be handled in upcoming work.

Closes OSIDB-1388


collectors/nvd/collectors.py
```diff
@@ -56,32 +58,68 @@ def response2result(self, vulnerabilities: list) -> list:
...
+        def get_description(descriptions: list) -> Union[str, None]:
             """
-            compose the result vector helper
+            Return a description only if it is written in English.
             """
-            for metric in metrics:
-                # we only care for NVD record
-                if metric.source == "nvd@nist.gov":
-                    return f"{metric.cvssData.baseScore}/{metric.cvssData.vectorString}"
+            return [d.value for d in descriptions if d.lang == "en"][0] or None
```
jsvob, line 65: Please match the language using the [official spec regex](https://github.com/CVEProject/cve-schema/blob/d2b6b34a527d18021209745d9ae19d966cc212f6/schema/v5.0/CVE_JSON_5.0_schema.json#L1030). Moreover, the spec requires that an English description is provided: 

So something like this:

```suggestion
            # Find the first English description
            # Ref: https://github.com/CVEProject/cve-schema/blob/d2b6b34a527d18021209745d9ae19d966cc212f6/schema/v5.0/CVE_JSON_5.0_schema.json#L744
            for d in descriptions:
                if re.match("^en([_-][A-Za-z]{4})?([_-]([A-Za-z]{2}|[0-9]{3}))?$", d.lang):
                    return d.value
```

jsvob (changes requested):
LGTM with minor nitpicks. Very nice. I'll use it for inspiration in refactoring https://github.com/RedHatProductSecurity/osidb/pull/367

collectors/keywords.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+WHITELIST = ["Spring", "dotnet", "kernel"]
```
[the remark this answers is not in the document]
osoukup: If we want to have this discussion in our team, we should probably start getting the support from the top. Just recently I was specifically asked to implement a **blacklist** at https://issues.redhat.com/browse/OSIDB-1453 so what is different here? I personally do not understand what could be disrespectful about using established technical terms but I understand that I am probably not the most sensitive person in the world. Anyway we should either decide about not using it at all or we should be free to use it at will. Restricting it only somewhere makes no sense.

collectors/nvd/collectors.py
```diff
@@ -210,6 +260,24 @@ def collect_updated(self) -> str:
...
+    @staticmethod
+    def create_snippet(item: dict) -> None:
+        """
+        Create a new snippet.
+        """
+        content = {
+            "cve_id": item["cve"],
+            "url": item["url"],
+            "description": item["description"],
+            "references": item["references"],
```
osoukup, line 272: I assume that we want to translate this into `FlawReference` list. Are we going to just default to external in the Snippet-Flaw translation? That might be OK. Taking the current options into account, it is the only meaningful. We could theoretically have an `NVD` reference type but I am not sure about the value and we can eventually retrieve it later if wanted.

collectors/keywords.py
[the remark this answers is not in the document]
jsvob: Asked our manager and CCd you.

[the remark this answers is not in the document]
JakubFrejlach: There is actually a lot to do if we would follow these guidelines. 36 occurrences of the word blacklist in OSIDB codebase, usage of the `master` branch instead of `main` and of course there would be definitely more.

collectors/nvd/collectors.py
```diff
@@ -210,6 +260,24 @@ def collect_updated(self) -> str:
             else ""
         )
 
+    @staticmethod
+    def create_snippet(item: dict) -> None:
+        """
+        Create a new snippet.
+        """
+        content = {
+            "cve_id": item["cve"],
```
osoukup, line 269: The feed collector uses `cve_ids`. We should maybe unify on one. It would be possible to use two but having one and create/link in `for` cycle would work for one or more items.

osoukup (approved):
LGTM. I would just be for some small changes in the `create_snippet`.

collectors/keywords.py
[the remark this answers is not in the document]
osoukup: The debate costs resources ATM.

collectors/nvd/collectors.py
jobselko: I prefer to use only one too, and the `Flaw` model already uses `cve_id`.

jsvob: If a single snippet is relevant to multiple CVEs, do we create multiple identical snippets (except each having their own `cve_id`)? I don't see another way to link one snippet to multiple CVEs.

collectors/nvd/collectors.py
jobselko: That regex is official only for the `cve-schema` project. And the result in nvd can look like this: 
```
'descriptions': [
    {'lang': 'en', 'value': 'The ip6_find_1stfragopt function in net/ipv6/output_core.c ...'}, 
    {'lang': 'es', 'value': 'La función ip6_find_1stfragopt en el archivo net/ipv6/output_core.c ...'}
]
```

So I think the regex will not ease the work here.

jsvob: Aha, sorry about the confusion!

I find it strange that they decided to use a different format for nvdlib, but you are correct. Tried with `CVE-2022-38014` which has `en-US` `descriptions` in `cvelistV5` and that CVE has two `descriptions` in `nvdlib`: `en` and `es`, no trace of `en-US`.

jsvob (approved):
LGTM

collectors/keywords.py
jsvob, line 3: Are these strings really the only things we're interested in?

collectors/nvd/collectors.py
osoukup: I am for using the same naming as the flaw attributes have wherever it makes sense BUT here it seems to me as a valid exception. The thing is that a snippet can commonly result in multiple flaws and then we need to potentially store more CVE IDs. Of course the case with only a single CVE ID is very common too. The thing is that singular `cve_id` string item can only hold one CVE ID so it covers only one of the two cases but plural list item `cve_ids` can contain both one or more and therefore it covers both cases. Having two possible attributes for CVE ID, one in singular for a single one and one in plural for multiple ones seems more complicated than just having one plural holding an arbitrary number of them.

collectors/keywords.py
jobselko: No, there are also a few `.net` combinations. These sources are from Assembler.

collectors/nvd/collectors.py
jobselko: Updated as we discussed.

collectors/nvd/collectors.py
jobselko: So as we discussed, I changed `cve_id` to `cve_ids` in a snippet's content. This does not change anything for NVD because every "piece of information" will always have only one CVE id.

collectors/nvd/collectors.py
```diff
@@ -56,32 +58,72 @@ def response2result(self, vulnerabilities: list) -> list:
...
+            return [d.value for d in descriptions if d.lang == "en"][0] or None
 
-        def get_metric(metrics: dict, version: str) -> Union[str, None]:
+        def get_references(data: CVE) -> dict:
             """
-            get metric record for the given version if exists
+            Return the source URL and all other URLs.
             """
-            if version not in metrics:
+            urls = [{"SOURCE": f"https://nvd.nist.gov/vuln/detail/{data.id}"}]
```
osoukup, line 71: The dictionary does not correspond to `FlawReference` model. It should be rather
```
{
    "type": "SOURCE",
    "url": f"https://nvd.nist.gov/vuln/detail/{data.id}"},
}
```

jobselko: Oh, my bad. Thanks :)

Elkasitu (changes requested):
Mostly nitpicks, otherwise LGTM

collectors/nvd/collectors.py
```diff
@@ -56,32 +57,37 @@ def response2result(self, vulnerabilities: list) -> list:
         filtering out everything unnecessary and simplifying
         """
 
-        def compose_cvss(metrics: list) -> Union[str, None]:
+        def get_cvss_metric(data: CVE, version: str) -> Union[dict, None]:
             """
-            compose the result vector helper
+            Return CVSS data for the given version.
             """
```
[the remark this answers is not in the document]
jobselko: Updated, please check. I think that creation of new `enum` is not necessary here.
I will probably add a proper description for all parameters at the end of the work.

collectors/nvd/collectors.py
```diff
@@ -56,32 +57,37 @@ def response2result(self, vulnerabilities: list) -> list:
...
-            for metric in metrics:
-                # we only care for NVD record
-                if metric.source == "nvd@nist.gov":
-                    return f"{metric.cvssData.baseScore}/{metric.cvssData.vectorString}"
+            if "metrics" not in data or (version not in data.metrics):
+                return None
```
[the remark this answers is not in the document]
jobselko: Would be sth like this sufficient?
```
            # `metrics` do not have to be present in `data`, and 
            # `version` do not have to be present in `data.metrics`
```

collectors/nvd/collectors.py
```diff
@@ -254,3 +209,100 @@ def collect_updated(self) -> str:
...
+            else False
+        )
+
+        # do the update if necessary
+        if were_updated:
+            if flaw.nvd_cvss2 != new_cvss2:
+                flaw.nvd_cvss2 = new_cvss2
+
+            if flaw.nvd_cvss3 != new_cvss3:
+                flaw.nvd_cvss3 = new_cvss3
```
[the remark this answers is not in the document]
jobselko: Yes. Unfortunately, I do. The reason for this is that `Flaw.nvd_cvss2` and `Flaw.nvd_cvss3` should be always equal to `Flaw.cvss_scores`. But if they would not be, I think it is better to trigger the update from the NVD page.

(It is possible that the problem I just mentioned is not a real problem, and you thought something different. If so, please let me know.)

collectors/nvd/collectors.py
```diff
@@ -148,26 +181,42 @@ def collect(self, cve: Union[str, None] = None) -> str:
...
+                updated_in_flawcvss = self.update_cvss_via_flawcvss(flaw, item)
 
-                # no automatic timestamps as those go from Bugzilla
-                # and no validation exceptions not to fail here
-                flaw.save(auto_timestamps=False, raise_validation_error=False)
+                if updated_in_flaw or updated_in_flawcvss:
+                    desync.append(item["cve"])
+                    # no automatic timestamps as those go from Bugzilla
+                    # and no validation exceptions not to fail here
+                    flaw.save(auto_timestamps=False, raise_validation_error=False)
```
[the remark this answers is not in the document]
jobselko: `Snippet.cve_ids` will always contain a list with one unique CVE, so multiple Snippet objects should not be a problem. Thank you!

collectors/nvd/collectors.py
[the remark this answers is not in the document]
Elkasitu: Then just write that, "depending on the data the following attributes might or might not be present" or something like that

collectors/nvd/collectors.py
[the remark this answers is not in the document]
Elkasitu: I meant that `were_updated` does the same check, no? so maybe you could do something like

```python
updated = False
if flaw.nvd_cvss2 != new_cvss2:
    ...
    updated = True
...
return updated
```

Elkasitu (approved):
Other than the remaining comments, LGTM, good job!

PR #379, opened 2023-11-13
Fix typo in NVD collector

jobselko:
This PR fixes incorrect `snippet_creation` to `snippet_creation_enabled` (introduced by mistake in #374).


osoukup (approved):
LGTM

Elkasitu:
A bit late to the party but we probably should not depend on an instance attribute for what are essentially feature flags, it is impossible for different instances of the collector to have different values (unless it's modified directly in the instance at runtime), we should use the env var / constant directly instead.

This could also use a test (we have a couple of tests that use the entire `FlawCollector` instance).

PR #380, opened 2023-11-14
Check keywords before creating Snippet

jobselko:
This PR adds a functionality for checking whether a new `Snippet` should be created. A new `Snippet` should be created only when data produced by a collector complies with specific rules.

Closes OSIDB-1558


osoukup (approved):
LGTM

jsvob (approved):

Elkasitu (commented):
While we're at it, use the `pytest.mark.default_cassette` decorator to prevent having two identical cassettes

PR #381, opened 2023-11-14
Extend FlawReferenceType to contain SOURCE

jobselko:
This PR extends `FlawReferenceType` to include the `SOURCE` option. This option should store the URL from which we obtained information about a new flaw.

Closes OSIDB-1556


jobselko:
Should I add an entry into `CHANGELOG` as this changes `openapi.yml`?

osoukup (approved):
LGTM. Changelog entry would not hurt, it is a user-facing change anyway.

PR #382, opened 2023-11-17
Add connection age limits to avoid connection reset by host

skontar:


osoukup (approved):
LGTM. Good job!

Elkasitu (changes requested):
I didn't finish the review because I think that while this'll work, there's a more fundamental issue:

Celery `Task` objects are instantiated only once [1], given that `Collector` objects are a subclass of `Task` and that the various collectors subclass `JiraQuerier` and `BugzillaQuerier` this means that the connection is created only once for the container's **entire lifecycle**.

If we change the `Collector` objects to **not** subclass from `JiraQuerier` and `BugzillaQuerier` and instead have them individually instantiate the querier objects when running the task, we'll have the best of both worlds:

* During the task's runtime, the connection will be reused as much as possible (with dumb caching, no need to cache per user/token/etc. as the user/token won't change during the task's runtime).
* After the task has finished running, the querier objects will be garbage collected as they are no longer in scope, once the task reruns a new querier object (=new connection) will be created

[1] https://docs.celeryq.dev/en/latest/userguide/tasks.html#instantiation

osoukup:
> I didn't finish the review because I think that while this'll work, there's a more fundamental issue:
> 
> Celery `Task` objects are instantiated only once [1], given that `Collector` objects are a subclass of `Task` and that the various collectors subclass `JiraQuerier` and `BugzillaQuerier` this means that the connection is created only once for the container's **entire lifecycle**.
> 
> If we change the `Collector` objects to **not** subclass from `JiraQuerier` and `BugzillaQuerier` and instead have them individually instantiate the querier objects when running the task, we'll have the best of both worlds:
> 
>     * During the task's runtime, the connection will be reused as much as possible (with dumb caching, no need to cache per user/token/etc. as the user/token won't change during the task's runtime).
> 
>     * After the task has finished running, the querier objects will be garbage collected as they are no longer in scope, once the task reruns a new querier object (=new connection) will be created
> 
> 
> [1] https://docs.celeryq.dev/en/latest/userguide/tasks.html#instantiation

@Elkasitu this assumes that a single task run is always under the connection time span limit. How do you know that it is the case? I mean how do you know that the scheduler under the resource limiting cannot hold the task idle too long and that this together with some potentially long running duty (like mega flaw) is always going to be under the connection life time limit?

skontar:
See https://github.com/RedHatProductSecurity/osidb/pull/389 for new iteration.

PR #383, opened 2023-11-22
update the tests to reflect the GitLab URL change

osoukup:
so they do not fail with .env with up-to-date GitLab URLs


osoukup:
I updated the product definitions and PS constants cassettes. The product definitions data also changed in the meantime so I had to slightly update the related tests too. With this you can have up-to-date CEE GitLab URLs in your **.env** file and the tests will not fail.

jobselko (approved):
LGTM. Do we know why CI tests are failing?

osoukup:
> LGTM. Do we know why CI tests are failing?
> 
> Edit: It seems we have to update vars also in CI.

@jobselko correct.

skontar:
All tests pass for me. Thanks!

skontar (approved):

osoukup:
I updated the URLs in the project secrets here in GitHub and the originally failing tests are not failing any more but some new and seemingly unrelated test fails now :smile: fun ahead.

PR #384, opened 2023-11-22
Set max tasks per worker in settings

skontar:
Set maximum tasks per worker in settings instead of commandline. Commandline option seem to be ignored in Celery / Django integration.

Closes OSIDB-1540.


osoukup (commented):
Based on the conversation we had this LGTM

osoukup (approved):

PR #390, opened 2023-11-27
Rename osim

costaconrado:
This PR renames OSIM module to Workflows.

In order to achieve that this PR is separated in 4 commits that respectively has the purpose of:
- Change the folder name modifying imports, settings and comments/documentations;
- Change classes, exceptions and models names and attributes to comply with new naming;
- Change constants names related to the module;
- Change the API path from `/osim/*` to `/workflows/*`.

Closes OSIDB-1395


apps/workflows/apps.py
```diff
@@ -0,0 +1,14 @@
+"""
+
+    Workflows - Issue Workflow Manager
```
osoukup, line 3: This is kinda nitpick but still. I know this naming is a result of the previous **OSIM - Open Security Issue Manager** adjustion. Before there was an abbreviation of the name chosen by Garth Mollett and it was in alignment with OSIDB name.

The think I not happy about :smile: is the word Issue. What does it mean in the context of Workflows? In OSIDB there are no issues. Is it mean for Jira issues? Even if so I would see it not obvious and confusing. My suggestion is just remove that word. The other two makes sense.

The line below is now not really precise. It is Python-like file path so the case must match. You can either fix it or remove it. Up to you.

osoukup (approved):
LGTM

This is obviously a breaking change so to be really precise we should release it as a new major version. We could argue that it was an experimental module up until now - no sure if it would be safe anyway @JakubFrejlach FYI.

Or maybe it is already a good time to do a major release? Maybe at point when we have all the task management done? Or maybe also trackers? (those are probably done API-wise) Waiting for the real GA would take more time.

WDYT @RedHatProductSecurity/osidb-devs ?

apps/workflows/apps.py
costaconrado: Fixed, thanks!

costaconrado:
@osoukup my opinion on that is should at least wait until all task management is done so we don't break it twice for task management, about trackers I think that we could wait until its done so we only do one major release in a short period of time, if OSIM team need task management before that I believe we could just deploy it to stage for development... Should we wait for more opinions on that before merging this branch or can I merge it and move this discussion to slack?

osoukup:
> @osoukup my opinion on that is should at least wait until all task management is done so we don't break it twice for task management, about trackers I think that we could wait until its done so we only do one major release in a short period of time, if OSIM team need task management before that I believe we could just deploy it to stage for development... Should we wait for more opinions on that before merging this branch or can I merge it and move this discussion to slack?

@costaconrado this discussion does not have to block the merge. I agree that it would be better to deliver this all together and ideally together with tracker workflow support to have only one major release.

osidb/migrations/0105_auto_20231127_2156.py
jobselko, line 1: Could you rename this file to something more readable? :)

costaconrado: I sure can, also noticed that I missed the last migration naming and added to this PR as a separated commit, please take a look.

jobselko: Thanks!
Regarding the last migration which is already in master - I am not sure whether it is safe to rename it. We discussed something similar in past with Ondrej and Adrian. @osoukup or @Elkasitu, WDYT?

jobselko: @costaconrado I think we can leave the 104 with old name and just fix 105 :).

osoukup: Changing the name of an existing migration can make some mess which is theoretically possible to clean manually but not trivial or save. I would leave the existing ones and just try to be more careful with the new ones. 

On one hand the migrations are auto-generated and theoretically people should not need to read them but sometimes it may be handy or at least good to get some understanding etc.

I prefer human readable migrations so eg. I try to process them by **black** but as a team we once decided to generally exclude them from this so it is up to every individual.

PR #391, opened 2023-12-01
Add external ID as a top-level attribute of the Snippet model

mprpic:
This allows easier operations on top of the Snippet model given a specific source since it removes the need to remember what identifier is used for a given source, i.e.:

Snippet.objects.get(source="NVD", content__cve_id="CVE-123")

becomes a more generic:

Snippet.objects.get(source="NVD", external_id="CVE-123")

The definition of what ID is used for what source should be noted for each source in Snippet.Source choices.

(This is technically a part of OSIDB-677, but I don't want it to close the issue once this PR is merged.)


osidb/migrations/0108_auto_20231130_2037.py
jobselko, line 1: Could you please rename this to something like `0108_snippet_external_id.py`?

osidb/models.py
```diff
@@ -1538,22 +1541,28 @@ def __init__(self, *args, **kwargs):
...
 
-    def convert_snippet_to_flaw(self) -> Union[Flaw, None]:
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                name="unique_snippets", fields=["source", "external_id"]
+            ),
+        ]
+
+    def convert_nvd_snippet_to_flaw(self) -> Union[Flaw, None]:
```
jobselko, line 1551: I would keep the original function name as we plan to use this function in all collectors.

osidb/tests/test_snippet.py
```diff
@@ -106,25 +106,24 @@ def test_create_flaw_from_snippet(self):
...
             ("CVE-2023-0001", False, False),
             ("CVE-2023-0001", True, False),
             ("CVE-2023-0001", True, True),
         ],
     )
-    def test_convert_snippet_to_flaw(self, cve_id, has_flaw, has_snippet):
+    def test_convert_nvd_snippet_to_flaw(self, cve_id, has_flaw, has_snippet):
         """
-        Tests the conversion of a snippet into a flaw and their linking.
+        Tests the conversion of a NVD snippet into a flaw and their linking.
```
jobselko, line 116: typo: `an NVD`

jobselko (commented):
Mostly LGTM. I am OK with the `external_id` field as we will have to keep identifiers for OSV snippets, and we cannot use `cve_id` for it, as in the case of NVD.

docs/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add new SOURCE option into FlawReferenceType (OSIDB-1556)
 - Add Celery worker concurrency
 - Add new NVD option into FlawSource
+- Add external_id attribute on Snippet model
```
osoukup, line 15: I know that we do not obey it always but once we agreed that the CHANGELOG should contain only user-facing changes. It is probably no issue to put there more, just that we were trying to behave like a service to be consumed and avoid changes which do not tell anything to the consumers (those would only go to commit messages). Adding **technical** GitHub label then turns of the CHANGELOG checking pipeline.

osidb/models.py
osoukup: Exactly. There should be nothing collector specific. If you see it differently, we should adjust the method and not the naming.

osidb/models.py
```diff
@@ -1538,22 +1541,28 @@ def __init__(self, *args, **kwargs):
...
         """
         Creates a new flaw from the snippet's content if a flaw with the given cve_id
         does not exist, and links them together. If a flaw already exists and does not
         contain the snippet of the current source, the snippet will be linked to it.
 
         Returns a flaw if it was newly created, None otherwise.
         """
-        cve_id = self.content["cve_id"]
+        cve_id = self.external_id
```
osoukup, line 1559: I do not think that this change is correct. The flaw is to be created based on the Snippet content. Are we sure that it will always contain a CVE ID for any future source and that it will be the unique Snippet identifier in the source? I think we are mixing two concepts together here.

On the other hand the original code should probably use `.get` so it counts on the case when the `cve_id` is actually not present and we should just create a CVE-less flaw. Sure that this is not going to be case for NVD but this method is supposed to be collector-independent and ideally never needed to be changed.

osidb/models.py
```diff
@@ -1538,22 +1541,28 @@ def __init__(self, *args, **kwargs):
...
         created_flaw = None
 
-        if not cve_id or not Flaw.objects.filter(cve_id=cve_id):
+        flaw = Flaw.objects.filter(cve_id=cve_id).first()
+        if not flaw:
```
osoukup, line 1563: The updated condition does not seem correct to me. In the case there is no CVE ID in the Snippet content we should not just blindly pair it with the first CVE-less flaw we have in the DB.

osidb/tests/test_snippet.py
```diff
@@ -10,7 +10,7 @@
 pytestmark = pytest.mark.unit
 
 
-def get_snippet(cve_id="CVE-2023-0001"):
+def get_nvd_snippet(cve_id="CVE-2023-0001"):
```
osoukup, line 13: I do not get why should we rename this and everything below. Sure we have a single Snippet type right now but this is supposed to test the general Snippet operations which should be type-independent.

osoukup (changes requested):
The changelog comment is rather a suggestion but I do not understand the rename and do not agree with the Snippet conversion logic changes.

osidb/models.py
jobselko: I overlooked this, @osoukup is right. The suggested change would work only for NVD.

mprpic:
@osoukup @jobselko After our discussion in today's meeting and with the comments here, I don't think the change to `external_id` makes sense any more. If the `content` field is supposed to contain mostly structured data then it might as well have a standard `content__cve_id` field that is the same across all snippets.

I'm going to close this without merging and continue my work on the OSV collector. Thanks!

osoukup:
> @osoukup @jobselko After our discussion in today's meeting and with the comments here, I don't think the change to `external_id` makes sense any more. If the `content` field is supposed to contain mostly structured data then it might as well have a standard `content__cve_id` field that is the same across all snippets.
> 
> I'm going to close this without merging and continue my work on the OSV collector. Thanks!

I am still in favor of having that identificator but we can leave it for now. CVE ID is IMO not general enough as there may exist Snippets with some CVE-less vulnerabilities - not in NVD but I guess elsewhere pretty regularly.

PR #393, opened 2023-12-01
Introduce usage of redbeat celery task scheduler

Elkasitu:
This commit adds celery-redbeat as a dependency of the project and sets it as the default scheduler for celery tasks.

Unlike celery-beat, redbeat uses a Redis instance for keeping track of task schedules and uses a distributed mutex lock in order to ensure that only one instance of the scheduler can actually schedule tasks.

This comes with the following benefits:

* The scheduler can have high availability without compromising scheduling, meaning that we can have e.g. 3 pods of the scheduler running but only one will be actively doing work at any given time, if two Availability Zones happen to go down, the remaining pod will pick up the slack.
* The schedule is stored in Redis, meaning that even if there are scheduler restarts, the work to be done can be picked up immediately without delays.
* Tasks can be dynamically created/modified.
* Faster startup.

This commit also simplifies the way that the celery-beat scripts work.

Closes OSIDB-1524

### DIFF
--- a/scripts/run-celery-beat.sh
+++ b/scripts/run-celery-beat.sh
@@ -15,12 +15,5 @@ echo
 # Postgresql makes the port available after the db is available, but wait some more to be sure. This can be removed if the real pg_isready is used.
 sleep 2
 
-echo "Waiting for http://osidb-service:8000/osidb/healthy to become available"
-while ! ( { curl -f http://osidb-service:8000/osidb/healthy >/dev/null 2>&1 || exit 1 ; } ) ; do echo -n "." ; sleep 2 ; done
-echo
-
-
-# custom run script for starting osidb celery-beat service in osidb-stage and osidb-prod environments.
-
-rm -f /tmp/celery_beat.pid
-exec celery -A config beat
+# Reuse existing standalone version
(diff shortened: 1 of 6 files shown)


osoukup (approved):
Thank you for the detailed explanation! Makes sense to me and LGTM

scripts/run-celery-beat.sh
```diff
@@ -15,12 +15,5 @@ echo
...
-echo "Waiting for http://osidb-service:8000/osidb/healthy to become available"
-while ! ( { curl -f http://osidb-service:8000/osidb/healthy >/dev/null 2>&1 || exit 1 ; } ) ; do echo -n "." ; sleep 2 ; done
-echo
-
-
-# custom run script for starting osidb celery-beat service in osidb-stage and osidb-prod environments.
-
-rm -f /tmp/celery_beat.pid
-exec celery -A config beat
+# Reuse existing standalone version
```
jsvob, line 18: Why was the healthy check removed? I don't see it in the standalone version either.

Elkasitu: I don't think it's necessary, the celery pods should be able to run independently of the web service as they don't depend on it.

PR #394, opened 2023-12-04
Enable row-level-security on local dev environment

Elkasitu:
This commit bridges the gap between local and stage/prod environments by enabling RLS by default on **new** local dev environment databases, this means that testing out things like performance, application of data migrations and general workflows should be closer to the aforementioned environments.

Closes OSIDB-264


osoukup (approved):
LGTM. Nice work!

osidb/tests/test_rls.py
```diff
@@ -0,0 +1,125 @@
...
+
+from osidb.core import set_user_acls
+from osidb.models import Flaw
+from osidb.tests.factories import FlawFactory
+
+pytestmark = pytest.mark.enable_rls
+
+
+class TestRLS:
+    # XXX: The following tests use Flaw for a model to test ACLs on,
```
jsvob, line 14: Leftover `XXX`?

Elkasitu: intended, it's just documentation

osoukup: Not sure whether it is a convention but at least my editor (Vim) highlights **XXX** string by default (the same ways is it does for **TODO**).

jobselko: Yeah, it is used very often and it means sth like `TODO`/`FIX`/`NOTE`.

PR #395, opened 2023-12-05
Add Snippet to Flaw conversion to NVD collector

jobselko:
This PR introduces `Snippet` to `Flaw` conversion in the NVD collector. The creation of the mentioned models is still disabled, and will be enabled later.

Closes OSIDB-1622

### DIFF
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -221,6 +221,7 @@ class NVDCollector(Collector, NVDQuerier):
         start_dt = timezone.now()
         updated_flaws = []
         new_snippets = []
+        new_flaws = []
 
         # fetch data
         # by default for the next batch but can be overridden by a given CVE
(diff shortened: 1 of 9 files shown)


collectors/nvd/collectors.py
```diff
@@ -232,41 +233,43 @@ def collect(self, cve: Union[str, None] = None) -> str:
...
-            if desync
-            else "No CVEs with desynced NVD CVSS."
-        )
-        logger.info(
-            f"New snippets were created for the following CVEs: {', '.join(new_snippets)}"
-            if new_snippets
-            else "No new snippets."
+            # create a new snippet with a flaw (if it does not exist) and link them
+            snippet = None
+            if self.snippet_creation_enabled:
```
jobselko, line 253: I discussed with @Elkasitu that `snippet_creation_enabled` should be rewritten into env variable.

osoukup (approved):
I am suggesting to remove the Flaw-Snippet ACLs validation. For the Snippet part maybe just turn it into validating that the Snippet is always internally visible only.

Otherwise LGTM

PR #396, opened 2023-12-05
hotfix for workflows tests.

costaconrado:
This PR changes tests using workflows APIs.

During #390 some tests were not proper changed for the new pattern of the API and this introduced faulty tests on master branch.


jobselko (approved):
Thanks! :)

PR #399, opened 2023-12-07
Update local venv creation to avoid a bug in pip

jobselko:
This PR fixes an error while creating a local venv due to a bug in pip. It is also a follow-up to the second commit in #393.
For more info, see https://github.com/pypa/pip/issues/9644.


osoukup (approved):
LGTM and works

PR #400, opened 2023-12-07
OSIDB-677: Add OSV collector

mprpic:
This PR adds a collector of OSV records from osv.dev. These records are saved as snippets and can be used to create a draft flaw (once that logic exists). OSV is an aggregator of various sources of vulnerabilities so this PR selects only a few of them. We can adjust the list of OSV ecosystems later on if needed.


mprpic:
@jobselko @osoukup This is sort of ready for initial review. Please have a look.

Each OSV record can have one or more CVE IDs, so I opted to create separate snippets for each CVE ID so we could keep the relationship of having one snippet for one CVE ID. We could also store CVE IDs as a list on the snippet, and then use a many-to-many relationship between Flaws and Snippets. This should also work with the current CVE ID-based snippet to flaw conversion method.

osoukup (approved):
LGTM

osidb/validators.py
```diff
@@ -11,7 +11,7 @@
 from django.utils import timezone
 
 # regex validation patterns
-CVE_RE_STR = re.compile(r"CVE-[12][0-9]{3}-\d{4,}")
+CVE_RE_STR = re.compile(r"CVE-(?:1999|2\d{3})-(?!0{4})(?:0\d{3}|[1-9]\d{3,})")
```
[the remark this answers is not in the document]
mprpic: Umm yes, I'll fix those tests as well.

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def __init__(self):
+        """initiate collector"""
+        super().__init__()
+        if self.snippet_creation_enabled is None:
+            self.snippet_creation_enabled = SNIPPET_CREATION_ENABLED
+
+    def fetch_osv_vulns_for_ecosystem(self, ecosystem: str) -> dict:
+        # Download the ZIP file from the URL
+        url = f"{self.OSV_DATA_SNAPSHOT_URL}/{ecosystem}/all.zip"
```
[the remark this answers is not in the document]
mprpic: No, the API only provides the ability to query by package identifiers (like name or purl) and version identifiers (version or commit): https://google.github.io/osv.dev/post-v1-query/

I might submit an issue to request fetching by timestamp (modified) but I doubt it would be implemented before this PR is merged. For now, fetching everything seems fine. The OSV records are relatively small (unlike our VEX files for example which are very verbose) so the processing of each ecosystem's ZIP is not too resource intensive.

osidb/models.py
```diff
@@ -1536,6 +1537,9 @@ def __init__(self, *args, **kwargs):
 
     # internal primary key
     uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    # A unique ID of this snippet as it is defined in the external data source where the
+    # snippet was collected from.
+    external_id = models.CharField(max_length=200)
```
jobselko, line 1542: The first commit should also contain the respective migration, which is currently included later. Also, I am not whether it is ok from the Django perspective to add a mandatory field into already existing model without a default value/null.

osidb/models.py
```diff
@@ -2158,6 +2169,7 @@ class CVSSVersion(models.TextChoices):
     class CVSSIssuer(models.TextChoices):
         REDHAT = "RH", "Red Hat"
         NIST = "NIST", "NIST"
+        OSV = "OSV", "OSV"
```
jobselko, line 2172: As we discussed - maybe a different name?

docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Implement tracker SLA start date setting (OSIDB-1393)
 - Implement tracker SLA end date setting (OSIDB-96)
 - Properly link Jira trackers to flaws on creation and update (OSIDB-1426)
+- Add OSV collector (OSIDB-677)
```
jobselko, line 20: We usually update README for user-visible changes only. I am not sure if this is that case.

openapi.yml
jobselko, line 1: Changes to `openapi.yml` should be done together with collector implementation (i.e. one commit).

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+                    try:
+                        osv_id, cve_ids, content = self.extract_content(osv_vuln)
+                    except Exception as exc:
+                        logger.error(
+                            f"Failed to parse data from {osv_vuln['id']} vulnerability: {exc}"
+                        )
+                        continue
+
+                    if any(
+                        osv_id.startswith(prefix)
```
jobselko, line 122: This could fail if `osv_id` is `None`.

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+                    ):
+                        continue
+                    try:
+                        created, updated = self.save_snippet(osv_id, cve_ids, content)
+                    except Exception as exc:
+                        logger.error(
+                            f"Failed to save snippet data for {osv_id} (error: {exc}): {content}"
+                        )
+                    new_count += created
+                    updated_count += updated
```
jobselko, line 133: If `try` fails, `created` and `updated` will be referenced before assignment.

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+            except requests.exceptions.RequestException as exc:
+                logger.error(f"Failed to fetch OSV vulns for {ecosystem}: {exc}")
+                continue
+
+        updated_until = timezone.now()
+        self.store(complete=True, updated_until_dt=updated_until)
```
jobselko, line 139: nitpick: `True` is default value here, so no need to rewrite it

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+            for cve_id in cve_ids:
+                snippet_content = content.copy()
+                snippet_content["cve_id"] = cve_id
+                # We need a unique ID and because we're creating a separate snippet for each
+                # CVE ID we can create a unique ID by appending that CVE ID to the OSV ID.
+                external_id = f"{osv_id}/{cve_id}"
+
+                _, created = Snippet.objects.update_or_create(
+                    source=Snippet.Source.OSV,
+                    external_id=external_id,
```
jobselko, line 175: `external_id` is `osv_id` and then `f"{osv_id}/{cve_id}"`. Is this correct?

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+                elif score["type"] == "CVSS_V2":
+                    cvss = CVSS2(score["score"])
+                    scores.append(
+                        {
+                            "issuer": FlawCVSS.CVSSIssuer.OSV,
+                            "score": float(cvss.base_score),
+                            "vector": cvss.vector,
+                            "version": FlawCVSS.CVSSVersion.VERSION2,
+                        }
+                    )
```
jobselko, line 246: nitpick: use `for` cycle (or a map) to avoid repeating

Also, it is not apparent from the code that `score["score"])` is a vector string, so I would prefer sth like:
```
vector = score["score"]
cvss = CVSS2(vector)
```

jobselko (commented):
LGTM, great work!

collectors/osv/collectors.py
```diff
@@ -0,0 +1,275 @@
...
+                            "vector": cvss.vector,
+                            "version": FlawCVSS.CVSSVersion.VERSION3,
+                        }
+                    )
+                elif score["type"] == "CVSS_V2":
+                    cvss = CVSS2(score["score"])
+                    scores.append(
+                        {
+                            "issuer": FlawCVSS.CVSSIssuer.OSV,
+                            "score": float(cvss.base_score),
```
jobselko, line 242: This could be simplified as the `CVSS` model in `osidb` uses `cvss` lib. The model takes `vector` and than calculates a score.

collectors/osv/collectors.py
jobselko:
```
            mapping = {
                "CVSS_V2": (FlawCVSS.CVSSVersion.VERSION2, CVSS2),
                "CVSS_V3": (FlawCVSS.CVSSVersion.VERSION3, CVSS3),
            }
            for cvss in data.get("severity", []):
                version, cvss_obj = mapping[cvss["type"]]
                vector = cvss["score"]
                scores.append(
                    {
                        "issuer": FlawCVSS.CVSSIssuer.OSV,
                        "version": version,
                        "vector": vector,
                        "score": cvss_obj(vector),
                    }
                )
```
Well, I am not sure if this better in the end :D

jobselko (approved):

docs/CHANGELOG.md
osoukup: Good point. I would probably leave this there however. It a significant change and should be eventually user-facing when they start processing the collected data collected from OSV.

collectors/osv/tests/test_collectors.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from collectors.osv.collectors import OSVCollector
+from osidb.models import Snippet
+
+pytestmark = pytest.mark.integration
+
+
+class TestNVDCollector:
```
jobselko, line 9: `NVD` instead of `OSV`

collectors/osv/tests/test_collectors.py
```diff
@@ -0,0 +1,35 @@
...
+        assert snippet.content["cve_id"] == "CVE-2023-50424"
+
+    @pytest.mark.vcr
+    def test_collect_multi_cve_osv_record(self):
+        """Test fetching a single OSV record that points to multiple CVEs."""
+        osvc = OSVCollector()
+        osvc.snippet_creation_enabled = True
+        osvc.collect(osv_id="PYSEC-2022-245")
+
+        snippet_1 = Snippet.objects.get(external_id="PYSEC-2022-245/CVE-2022-36359")
```
jobselko, line 29: Could you please add `assert Snippet.objects.count() == 2`?

collectors/osv/tests/test_collectors.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestNVDCollector:
+    @pytest.mark.vcr
+    def test_collect_osv_record(self):
+        """Test fetching a single OSV record."""
+        osvc = OSVCollector()
+        osvc.snippet_creation_enabled = True
+        osvc.collect(osv_id="GO-2023-2400")
+
+        snippet = Snippet.objects.last()
```
jobselko, line 17: Could you please add `assert Snippet.objects.count() == 1`?

collectors/osv/collectors.py
jobselko, line 1: Do we want to include also this check before creating a snippet? It is in the `NVD` collector.

https://github.com/RedHatProductSecurity/osidb/blob/d0b3185a635331c2b3d7c5611190a7e34afc2008/collectors/keywords.py#L519

collectors/osv/tests/test_collectors.py
mprpic: Sure!

jobselko:
@mprpic The `Snippet` to `Flaw` conversion is already in the `NVD` collector. Do you plan to create a new PR for it later? And I am sorry for removing this from the merge queue without letting you know :)

collectors/osv/tests/test_collectors.py
mprpic: Ah, nice catch.

collectors/osv/collectors.py
mprpic: I considered it but I think all of the ecosystems that we fetch from OSV are relevant to us and we should have a further filter on the components that get reported in each record. These components should ideally be paired down to just those that are present in Component Registry, but that's a much larger piece of work that will have to be done in a separate PR. I can certainly file an issue for it at least though, so thank you for reminding me of that.

mprpic:
> @mprpic The `Snippet` to `Flaw` conversion is already in the `NVD` collector. Do you plan to create a new PR for it later? And I am sorry for removing this from the merge queue without letting you know :)

No worries, at least I was able to fix that incorrect test name before it got merged :)

I think the conversion will have to be a bit more complicated than the NVD one since it would be a shame not to use the affects data that is a part of each OSV record and converting those to our Affects might require some changes to the Affects model itself. I'll file an issue for this (same one as for the `should_create_snippet()` usage you mentioned above) to track this work, but I guess someone else will have to pick this up instead of me :-/

mprpic:
Filed OSIDB-1741 to improve the snippet collection filtering.

jobselko (approved):
LGTM, just two minor things (which can be ignored if you want): 
- collector and its tests should be in one commit
- the migration name does not match its content, so maybe rename it to sth like `0110_snippet_miscellaneous.py`

mprpic:
> LGTM, just two minor things (which can be ignored if you want):
> 
>     * collector and its tests should be in one commit
> 
>     * the migration name does not match its content, so maybe rename it to sth like `0110_snippet_miscellaneous.py`

Both addressed.

PR #404, opened 2023-12-13
Update snippet creation bool to env var only

jobselko:
Obsoleted draft

### DIFF
--- a/collectors/constants.py
+++ b/collectors/constants.py
@@ -1,3 +1,5 @@
 from osidb.helpers import get_env
 
+# for visibility only (remove later)
+
 SNIPPET_CREATION_ENABLED = get_env("SNIPPET_CREATION", default="False", is_bool=True)
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -170,9 +170,6 @@ class NVDCollector(Collector, NVDQuerier):
     NVD CVSS collector
     """
 
-    # snippet creation is disabled by default for now
-    snippet_creation_enabled = None
-
     # the NIST NVD CVE project started in 1999
     # https://nvd.nist.gov/general/cve-process
     BEGINNING = timezone.datetime(1999, 1, 1, tzinfo=timezone.get_current_timezone())
@@ -184,9 +181,6 @@ class NVDCollector(Collector, NVDQuerier):
     def __init__(self):
         """initiate collector"""
         super().__init__()
-
-        if self.snippet_creation_enabled is None:
-            self.snippet_creation_enabled = SNIPPET_CREATION_ENABLED
 
     def get_batch(self) -> (dict, timezone.datetime):
         """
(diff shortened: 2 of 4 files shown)


collectors/nvd/collectors.py
```diff
@@ -250,7 +244,7 @@ def collect(self, cve: Union[str, None] = None) -> str:
 
             # create a new snippet with a flaw (if it does not exist) and link them
             snippet = None
-            if self.snippet_creation_enabled:
+            if SNIPPET_CREATION_ENABLED:
```
jobselko, line 247: @Elkasitu I am struggling with removing the `snippet_creation_enabled` instance attribute and replacing it with `SNIPPET_CREATION_ENABLED` env var only. The code works when testing it with `SNIPPET_CREATION=1`/`SNIPPET_CREATION=0` from the `.env` file, but it fails in `collectors/nvd/tests/test_collectors.py`.

It seems that the `snippet_creation_enabled` fixture is correctly set in `test_snippet_and_flaw_created()` in `collectors/nvd/tests/test_collectors.py`, but in `collectors/nvd/collectors.py`, this value is lost after the import from `collectors/constants.py`.

Do you have any idea how to solve this?

jobselko: So, while this change does not work in my local dev env, it works in CI :D

collectors/constants.py
```diff
@@ -1,3 +1,5 @@
 from osidb.helpers import get_env
 
+# todo?
```
osoukup, line 3: What is the meaning of this?

collectors/nvd/collectors.py
osoukup: I do not understand the changes in this file. Functionally they do nothing new.

osoukup (changes requested):
This seems to be a draft but it is not marked so.

collectors/nvd/collectors.py
jobselko: @osoukup It is a follow-up to Adrian's https://github.com/RedHatProductSecurity/osidb/pull/379#issuecomment-1810211075.

So far, I updated only one test, but its result differs depending on the environment (passing in CI / failing locally). This is not ready for merging but I would like to get some feedback regarding what might be a problem here.

collectors/constants.py
jobselko: I tried some `getenv` things here, but you can ignore it. It is mostly for visibility now.

collectors/nvd/collectors.py
Elkasitu: You have to monkeypatch the constant and not the env var, the constant's value remains static for the lifecycle of the test run regardless of changes to the env var

jobselko: The result is still same - not working locally. I tried to use both `1` and `True`.

jobselko: @Elkasitu I would prefer to leave the attr as it is now as this change seems to be not so important and hard to implement (or at least for me). WDYT?

Elkasitu: Sure

PR #406, opened 2023-12-15
minor Jira tracker related improvements

osoukup:
During the last sprint review preparations I had issues creating Jira trackers (so I used Bugzilla as an example). This is the followup fixing the stuff I had issues with. The first commit is actually not really a fix and it is opinionated. I am proposing the order but I am not pushing it. If you are against, we can drop it. Just in the original version it was hard to see the differences of my **.env** and the compose **environment** by human.

Closing OSIDB-1735


osoukup:
@RedHatProductSecurity/osidb-devs please review

docker-compose.yml
```diff
@@ -84,6 +84,7 @@ services:
         PRODUCT_DEF_URL: ${PRODUCT_DEF_URL}
         PS_CONSTANTS_URL: ${PS_CONSTANTS_URL}
         SNIPPET_CREATION: ${SNIPPET_CREATION}
+        TRACKERS_SYNC_TO_JIRA: ${TRACKERS_SYNC_TO_JIRA}
```
jobselko, line 87: Should this be added into `osidb-ops` too?

jobselko (approved):
LGTM

docker-compose.yml
osoukup: As long as the default value of `False` is OK I think it is not needed. Anyway this commit only allows the previously non-existing option to generally set it via the environment but does not account for any specific environment. When we think it should be defined in some we can propose that elsewhere.

PR #411, opened 2024-01-04
after-flaw-update tracker update

osoukup:
TODO ... not ready for a review


osoukup:
The issue OSIDB-97 will have another part of after-affect-update tracker updates which is however independent of this one so this one is ready for a review.

osidb/serializer.py
```diff
@@ -1359,6 +1362,95 @@ class Meta:
...
+        # TODO
+        # unembargo is more complicated
+        # and will be covered separately
+
+        # we only need to sync the trackers when crucial attributes change
+        # plus in the case of the impact we only care for the increase
+        #
+        # the crucial attributes are those influencing the SLA deadline plus the CVE ID
+        if not differs(
+            old_flaw, new_flaw, ["cve_id", "is_major_incident", "unembargo_dt"]
```
jobselko, line 1425: `is_major_incident` is deprecated, so you should check `major_incident_state` (I think that change from `REQUESTED` to `APPROVED` in test should be triggerring)

osidb/serializer.py
```diff
@@ -1359,6 +1362,95 @@ class Meta:
...
+        # perform the tracker updates
+        # could be done assync eventually
+        for tracker in trackers2update:
+            tracker.save(
+                # the serializer does not care for the backend system
+                # therefore at this point we simply require both secrets
+                bz_api_key=self.get_bz_api_key(),
+                jira_token=self.get_jira_token(),
+            )
+
```
jobselko, line 1453: Do we have to create and keep `trackers2update`? If no, we can save tracker directly if it is not closed:

        for affect in new_flaw.affects.all():
            # no tracker updates for community affects
            # because it was requested not to spam them
            if affect.is_community:
                continue

            for tracker in affect.trackers.all():
                # no tracker updates for the closed ones
                # because we consider these already done
                if not tracker.is_closed:
                    tracker.save(
                        # the serializer does not care for the backend system
                        # therefore at this point we simply require both secrets
                        bz_api_key=self.get_bz_api_key(),
                        jira_token=self.get_jira_token(),
                    )

jobselko (approved):
LGTM

PR #412, opened 2024-01-05
Performance setup and improvements

Elkasitu:
This PR:
* Sets up an entrypoint to launch osidb-service using gunicorn instead of django, providing better reproducibility when comparing with stage/prod
* Sets up a locust instance in the test compose file, for performance/load testing
* Changes the way the `changed_{before,after}` filter behaves, in order to increase performance

TODO:
- [x] Regression tests for the filters
- [x] Test running / improving data migration on a large database

Before this commit:
image

After this commit:
image

Closes OSIDB-1400
Closes OSIDB-196


docker-compose.test.yml
```diff
@@ -56,3 +56,12 @@ services:
         SNIPPET_CREATION: ${SNIPPET_CREATION}
       volumes:
         - ${PWD}:/opt/app-root/src:z
+
+    locust:
```
JimFuller-RedHat, line 60: nice!

mk/localdev.mk
```diff
@@ -83,6 +85,13 @@ restart_django_foreground:
 	$(podman) exec -it osidb-service pkill -f "python3 manage.py runserver"
 	$(podman) exec -it osidb-service python3 manage.py runserver 0.0.0.0:8000
 
+.PHONY : start_gunicorn_fg
+start_gunicorn_fg:
+	@echo ">Note that after CTRL+C, django server will restart in the background, as long as osidb-service is running."
+	@sleep 0.5
+	$(podman) exec -it osidb-service pkill -f "python3 manage.py runserver"
+	$(podman) exec -it osidb-service gunicorn config.wsgi --config gunicorn_config.py
```
JimFuller-RedHat, line 93: do we need to be explicit with using config.settings_local here ?

osidb/migrations/0111_flaw_local_updated_dt.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='flaw',
+            name='local_updated_dt',
+            field=models.DateTimeField(null=True, default=django.utils.timezone.now),
+        ),
+        migrations.AddIndex(
+            model_name='flaw',
+            index=models.Index(fields=['-local_updated_dt'], name='osidb_flaw_local_u_9d468b_idx'),
```
JimFuller-RedHat, line 21: just an observation - I have always disliked the autogenerated index names

osidb/migrations/0112_flaw_local_updated_dt_data.py
```diff
@@ -0,0 +1,73 @@
...
+from osidb.core import set_user_acls
+
+
+logger = logging.getLogger(__name__)
+
+
+BATCH_SIZE = 1000
+
+
+def forwards_func(apps, schema_editor):
```
JimFuller-RedHat, line 16: Have you had a chance to test this migration against a db dump of a non trivial size ? I would not want the first test to be data migration to prod ;)

perf/main.py
```diff
@@ -0,0 +1,19 @@
+from locust import HttpUser, constant_pacing, task
```
JimFuller-RedHat, line 1: I like locust though we are never going to have millions of users and our performance (where load is concerned) is easily loaded via a set of curl scripts (or using osidb bindings). I am not sure we want to embed perf tests directly in the repo itself ... I also think perf is heavily dependent on total env (eg. including network) and maybe such script would live in ops repo.

osidb/filters.py
```diff
@@ -116,21 +115,13 @@ def changed_after_filter(self, queryset, name, value):
...
     def changed_before_filter(self, queryset, name, value):
         """
         Returns a Flaw if it or any of its affects/trackers have been updated before `value`
         """
-        return queryset.filter(
-            Q(updated_dt__lte=value)
-            | Q(affects__updated_dt__lte=value)
-            | Q(affects__trackers__updated_dt__lte=value)
-        ).distinct()
+        return queryset.filter(local_updated_dt__lte=value)
```
JimFuller-RedHat, line 124: This is a time/space optimisation eg. push processing time at ingestion time to avoid spending resolving **OR** during runtime.

Were you able to look at _.query / .explain()_ for current **queryset.filter** and confirm it is the **OR** condition slowing this down. 

Does current queryset.filter hit indexes (they exist right ?) for resolving comparison on **affects__updated_dt__lte** ? If they do not then I would suggest doing that optimisation first (eg. ensure index exists and are used) before 'baking in' as proposed in this PR.

osidb/models.py
```diff
@@ -853,6 +855,7 @@ class Meta:
         )
         indexes = TrackingMixin.Meta.indexes + [
             models.Index(fields=["-cve_id"]),
+            models.Index(fields=["-local_updated_dt"]),
```
JimFuller-RedHat, line 858: As prev mentioned ... you might find adding indexes for existing **updated_dt** a better first approach (Flaw, Affect and Tracker).

osidb/signals.py
```diff
@@ -72,3 +72,25 @@ def auto_create_profile(sender, instance, created, **kwargs):
 @receiver(pre_save, sender=AffectCVSS)
 def populate_cvss_score(sender, instance, **kwargs):
     instance.score = instance.cvss_object.scores()[0]
+
+
+@receiver(post_save, sender=Flaw)
```
JimFuller-RedHat, line 77: What are the advantages of using django signals ? 

It is my understanding that this is just running in django/python eg. if a row is updated in the database outside of django then we need to manually update to ensure local_updated_dt is updated. Perhaps that is ok ... maybe comments saying so here would provide hints to our future selves.

mk/localdev.mk
```diff
@@ -9,6 +9,8 @@
 .PHONY : start-local
 start-local: check-venv generate_local_pg_tls_cert compose-up restart_django_foreground
 
+.PHONY : start-local-gunicorn
```
JimFuller-RedHat, line 12: Should we make this default ?

mk/localdev.mk
Elkasitu: nope, that is set at the container level already (notice how we don't do it for `start-local`)

osidb/migrations/0112_flaw_local_updated_dt_data.py
Elkasitu: I tested it against a db that is 1/4th the size and it takes ~1s per 1000 flaws, based off of production data, but as noted in the TODO in the PR description I might improve it further for faster deployment (but even in its current state it should not be too slow, definitely under the 30m deadline)

perf/main.py
Elkasitu: I don't see why not, dev dependencies are not installed in the production image, this provides a single way to define performance tests / benchmarking which reduces a lot of the risk of having hand-rolled solutions (different results from different people), and you can run these tests against stage/production directly and it's easy to integrate into various pipelines

osidb/filters.py
Elkasitu: I can take a look but it seemed pretty obvious to me, we're doing a lot of potentially useless joins and on top of it we add a distinct, just to determine whether the data has changed, seems like a clear case for denormalization.

I will take a look, but by "existing indexes are not being used" do you mean by the Django ORM / Programmer error or by the Postgres query planner?

perf/main.py
JimFuller-RedHat: Sure, but there is cognitive load adding any dependency... also I do not know what locusts gives us (or why we would need it) beyond what one can achieve using simpler methods ... for example with curl we can easily perform parallel loading with something like: 

```
curl "https://localhost:8010/osidb/api/v1/affects?offset=[1-100]"  -w "%{time_connect},%{time_total},%{speed_download},%{http_code},%{size_download},%{url_effective}\n" -o /dev/null -Z
```
this has the added benefit that the perf test is not executed in python ( variation is very useful in testing in general).

Alternately we might emulate how we have done this in [component-registry](https://github.com/RedHatProductSecurity/component-registry/blob/main/tests/test_performance.py) using plain ole pytest (which also uses python-request as well ... if u want that http1.1 constraint;))

The tests you have written seem to only scratch the surface of locust capability which makes me wonder if we need it.

osidb/filters.py
JimFuller-RedHat: To determine root cause of bad performance - its always good to view query plan ... SQL with a lot of joins can often 'boil down' to an efficient query plan ... but agree yes we should strive to write efficient SQL queries to begin with. 

In this specific case I feel like with the very little data we have that we should have good enough performance here (even with OR queries) using updated_dt indexes.

Using .explain() in django or via pgcli EXPLAIN ANALYZE on representative SQL query (using those updated_dt) should show if indexes are being used or not ... 

If pg_stat works we could do
```
SELECT schemaname, pg_stat_user_indexes.indexrelname, idx_scan, idx_tup_read, idx_tup_fetch                    
FROM pg_stat_user_indexes                                                                                      
WHERE pg_stat_user_indexes.indexrelname = '$1'                                                                 
```
where put index name for $1 to show usage of related indexes.

If indexes are not being used then that is a candidate for slow query and maybe we should understand why the query planner does not choose to use them ... if they are being used then your PR addresses it nicely ;)

I can also check tomorrow ....

osidb/signals.py
Elkasitu: As opposed to what? pg triggers?

You are correct in that this can be bypassed by using a few Django ORM methods (e.g. `update`, `bulk_update`, ...) but that also bypasses all the `save()` logic that we already have, so as a rule of thumb bypassing the ORM is a bad idea and in practice we only do it during hefty data migrations which are rare.

I can add comments, but it essentially applies to all signals

mk/localdev.mk
Elkasitu: I avoided it for two reasons:

* Didn't want to disrupt the status quo of local development
* Debugging code with multiple pods is harder

Eventually we could probably have it be the default with an nginx reverse-proxy in front and everything but for now this works

osidb/signals.py
JimFuller-RedHat: I do not see why we would use triggers here either eg. why not overload save() and call super.save() on the model itself ? It is quite possible I am django ignorant and missing some obvious benefit. Maybe I should learn how django signals are implemented as well ;)

As for bypassing ... was mostly thinking of raw SQL against the database (outside of django). I guess I am never going to feel comfortable about using an ORM if it always forces one to never use the database (out of safety concerns)!

Elkasitu: `save` contains a lot of custom logic itself already and doesn't help in cases that bypass the ORM anyways, this is a cleaner, self-contained approach although I admit it can be "hidden" or "magic", but I'm just using the ORM's features.

Furthermore, with the latest change the use of signals is required, as there is no way (to my knowledge) to hook into a ManyToMany field's add/remove/etc. methods, and using these does not call `save`

Elkasitu:
Migration run on a DB that is 1/2 the size of production completed in just about 2min, estimated time for production should therefore be 4min which is acceptable

jobselko (approved):
LGTM. I do not understand everything, but I believe Jim would notice any important issue :).

PR #413, opened 2024-01-05
Use HTTPS_TASKMAN_PROXY

jsvob:
Closes OSIDB-1781


jsvob:
Not necessary. Quick fix already possible by setting HTTPS_PROXY env var. Might be properly addressed in OSIDB-1574. Details in OSIDB-1781.

PR #415, opened 2024-01-09
Make VCR ignore base URLs

costaconrado:
This PR makes VCRs ignores base URL during recording and testing.

This PR is currently not ready for review, I still figuring out how to rewrite VCRs that touches Bugzilla.

This PR also rewrites existing VCRs in order make tests more flexible in different environments.

Closes OSIDB-1678.


conftest.py
```diff
@@ -69,14 +73,37 @@ def filter_response(response):
     return response
 
 
+def remove_host_request(request):
+    request.uri = re.sub(r"(?<=://)[^/]+", "example.com", request.uri)
```
Elkasitu, line 77: Could you perhaps document what this regex does? It boggles the mind :)

Wouldn't something like `re.sub(r"https?://(.+)", "https://example.com", request.uri` make more sense and be easier to read?

Also since this RE will be used for every request, might as well compile it

conftest.py
```diff
@@ -69,14 +73,37 @@ def filter_response(response):
...
+    return request
+
+
+def remove_host_response(response):
+    body_string = re.sub(
+        r"(?<=://)[^/]+", "example.com", response["body"]["string"].decode("utf-8")
+    )
```
Elkasitu, line 84: same as above

conftest.py
```diff
@@ -69,14 +73,37 @@ def filter_response(response):
...
+    response["body"]["string"] = body_string.encode("utf-8")
+
+    # redirected requests need Location header
+    original_locations = response["headers"].get("Location", [])
+    if original_locations:
+        locations = []
+        for location in original_locations:
+            locations.append(re.sub(r"(?<=://)[^/]+", "example.com", location))
```
Elkasitu, line 92: ditto

collectors/jiraffe/tests/test_collectors.py
```diff
@@ -85,12 +88,14 @@ def test_collect_complete(self):
         """
         test that Jira collector data status is changed to complete when the data are current
         """
-        collector = JiraTrackerCollector()
-        collector.metadata.updated_until_dt = timezone.now()
-        assert not collector.is_complete
+        with warnings.catch_warnings():
```
Elkasitu, line 91: could you explain why do we need to suppress these warnings? could you provide an example of such warnings?

Elkasitu (approved):
LGTM, mostly nitpicks

collectors/jiraffe/tests/test_collectors.py
costaconrado: although it is called warning, it implements `Exception` and prevents the test to do an HTTPS request... 
It fails the test and do not rewrite the cassette with the following message `urllib3.exceptions.SystemTimeWarning: System time is way off (before 2022-01-01). This will probably lead to SSL verification errors`.

Full log error if not suppressed:

```
collectors/jiraffe/tests/test_collectors.py:221: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
collectors/jiraffe/collectors.py:164: in collect
    res = self.jira_querier.jira_conn._get_json(
.tox/record-rewrite/lib/python3.9/site-packages/jira/client.py:3544: in _get_json
    r = self._session.get(url, params=params)
.tox/record-rewrite/lib/python3.9/site-packages/jira/resilientsession.py:195: in get
    return self.__verb("GET", str(url), **kwargs)
.tox/record-rewrite/lib/python3.9/site-packages/jira/resilientsession.py:165: in __verb
    response = method(url, timeout=self.timeout, **kwargs)
.tox/record-rewrite/lib/python3.9/site-packages/requests/sessions.py:602: in get
    return self.request("GET", url, **kwargs)
.tox/record-rewrite/lib/python3.9/site-packages/requests/sessions.py:589: in request
    resp = self.send(prep, **send_kwargs)
.tox/record-rewrite/lib/python3.9/site-packages/requests/sessions.py:703: in send
    r = adapter.send(request, **kwargs)
.tox/record-rewrite/lib/python3.9/site-packages/requests/adapters.py:486: in send
    resp = conn.urlopen(
.tox/record-rewrite/lib/python3.9/site-packages/urllib3/connectionpool.py:712: in urlopen
    self._prepare_proxy(conn)
.tox/record-rewrite/lib/python3.9/site-packages/urllib3/connectionpool.py:1012: in _prepare_proxy
    conn.connect()
.tox/record-rewrite/lib/python3.9/site-packages/vcr/stubs/__init__.py:286: in connect
    return self.real_connection.connect(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <urllib3.connection.HTTPSConnection object at 0x7f485dd8a040>

    def connect(self):
        # Add certificate verification
        self.sock = conn = self._new_conn()
        hostname = self.host
        tls_in_tls = False
    
        if self._is_using_tunnel():
            if self.tls_in_tls_required:
                self.sock = conn = self._connect_tls_proxy(hostname, conn)
                tls_in_tls = True
    
            # Calls self._set_hostport(), so self.host is
            # self._tunnel_host below.
            self._tunnel()
            # Mark this connection as not reusable
            self.auto_open = 0
    
            # Override the host with the one we're requesting data from.
            hostname = self._tunnel_host
    
        server_hostname = hostname
        if self.server_hostname is not None:
            server_hostname = self.server_hostname
    
        is_time_off = datetime.date.today() < RECENT_DATE
        if is_time_off:
>           warnings.warn(
                (
                    "System time is way off (before {0}). This will probably "
                    "lead to SSL verification errors"
                ).format(RECENT_DATE),
                SystemTimeWarning,
            )
E           urllib3.exceptions.SystemTimeWarning: System time is way off (before 2022-01-01). This will probably lead to SSL verification errors

.tox/record-rewrite/lib/python3.9/site-packages/urllib3/connection.py:387: SystemTimeWarning
```

Elkasitu: Why is this a problem now compared to before? we're still doing HTTPS requests 🤔 

I think the better solution is to raise awareness of this issue when freeze_time and https requests interact, and instead of decorating the entire method, we can just use the `freeze_time` context manager, set `updated_until_dt` within its context, and then do the actual collecting outside.

WDYT?

conftest.py
costaconrado: I'm replacing the current pattern with `(https?://)[^/]+` for better readability, the option of using `(?<=://)` (positive lookbehind) was to match and keep the same protocol used in the original request but I do't think its really needed so I've removed it... About the last part I think I should keep the `[^/]+` so it matches everything but stop matching after the 1st slash it find. Something like `(.+)` would also replace the context, e.g.:

```
>>> re.sub("(https?://)(.+)", "https://example.com", "https://github.com/RedHatProductSecurity/osidb/pull/415")
'https://example.com'

>>> re.sub("(https?://)[^/]+", "https://example.com", "https://github.com/RedHatProductSecurity/osidb/pull/415")
'https://example.com/RedHatProductSecurity/osidb/pull/415'
```

Please let me know if this is a valid solution or if there is still room for improvement.

collectors/jiraffe/tests/test_collectors.py
costaconrado: makes sense, I've recorded and removed the decoration on the method

PR #416, opened 2024-01-10
Make tests independent from env file

costaconrado:
This PR make tests independent from env file.

In order to achieve that this PR set default strings in constants set from environment variables and can't be blank.
This PR is the second step from #415 that ignores urls in VCRs cassettes.
This PR also set default value for Bugzilla during tracker tests in order to test description generation in a controlled environment.

Closes OSIDB-1679.

### DIFF
--- a/apps/trackers/tests/test_common.py
+++ b/apps/trackers/tests/test_common.py
@@ -496,7 +496,7 @@ until the embargo has lifted. Please post the patch only to the \
                 """\
 More information about this security flaw is available in the following bug:
 
-https://bugzilla.redhat.com/show_bug.cgi?id=0
+https://example.com/show_bug.cgi?id=0
 
 Disclaimer: Community trackers are created by Red Hat Product Security team on a \
 best effort basis. Package maintainers are required to ascertain if the flaw indeed \
@@ -507,8 +507,8 @@ affects their package, before starting the update process."""
                 """\
 More information about these security flaws is available in the following bugs:
 
-https://bugzilla.redhat.com/show_bug.cgi?id=0
-https://bugzilla.redhat.com/show_bug.cgi?id=1
+https://example.com/show_bug.cgi?id=0
+https://example.com/show_bug.cgi?id=1
 
 Disclaimer: Community trackers are created by Red Hat Product Security team on a \
 best effort basis. Package maintainers are required to ascertain if the flaw indeed \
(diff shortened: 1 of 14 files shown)


apps/trackers/tests/test_common.py
```diff
@@ -3,6 +3,7 @@
 """
 import pytest
 
+import apps.trackers.common as commom
```
Elkasitu, line 6: typo?

apps/trackers/tests/test_common.py
```diff
@@ -448,10 +449,12 @@ def test_bugzilla_virtualizaiton(self, ps_component):
         )
 
     @pytest.mark.parametrize("flaw_count", [1, 2, 3])
-    def test_community(self, flaw_count):
+    def test_community(self, monkeypatch, flaw_count):
         """
         test community tracker description generation
         """
+        monkeypatch.setattr(commom, "BZ_URL", "https://example.com")
```
Elkasitu, line 456: maybe we should use an auto-use fixture instead? that way any test/code that depends on these vars will already be monkey patched implicitly

collectors/ps_constants/tests/cassettes/test_core/TestPsConstantsCollection.test_fetch_ps_constants.yaml
```diff
@@ -1358,14 +1344,13 @@ interactions:
...
       - python-requests/2.31.0
     method: GET
-    uri: https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/raw/master/data/ubi_packages.yml?job=build
+    uri: https://example.com/prodsec-dev/ps-constants/-/raw/master/data/ubi_packages.yml?job=build
   response:
     body:
       string: "# This list of packages is used to indicate the UBI flag in the SFM2.
         \n# This list is NOT used to calculate SLA.\n\n\n# list of packages / components
         in UBI images\n# separate list provided for each RHEL version, keys need to
-        match ps module\n# names\n# https://projects.engineering.redhat.com/browse/PSIRT-730\n#
```
jsvob, line 1367: Nitpick: I'd prefer the cassettes to retain their contents and have just the URLs replaced because the cassettes capture files that might already look different. I think it's ok to try to rerecord the cassettes and then check that bugs are filed for updating the relevant functionality, but I also think that doing it in this PR would duplicate work, for example work done in OSIDB-1876 related to changes in compliance-priority.yml.

jsvob (approved):

osoukup (changes requested):
This almost LGTM. I would like to see one thing changed. The URL constants should IMO not have a test value as their defaults. I would consider this as undesired intermingling of the app and its tests. I am suggesting to monkey-patch the values in the tests only instead by some auto-use fixture.

costaconrado:
I have reworked this PR so it uses conftest to set environments instead of setting constants as suggested, I've also removed the code that would ignore the warning about the freeze gun (this were suggested outside this PR discussion) and fixed the typo pointed.
For `collectors.product_definitions.constants.PRODUCT_DEFINITIONS_REPO_URL` I had to add an empty string as default, monkeypatching it would not be enough since it would break the import of `collectors.product_definitions.core` if its `None`.
 Please take a second look on this when you have time @osoukup

dmonzonis (approved):
LGTM

osoukup (approved):
LGTM

PR #419, opened 2024-01-15
Include osv and ps_constants in settings.py

jobselko:
This PR updates `settings.py` to include `osv` and `ps_constants` in `INSTALLED_APPS` and `LOGGING`. `osv` was accidentally omitted in #400 and `ps_constants` was probably never used. Also, it updates a few minor things.

Related to OSIDB-1764


config/settings.py
```diff
@@ -34,36 +34,36 @@
...
+    "django.contrib.sessions",
+    "django.contrib.staticfiles",
+    "django_extensions",
+    "django_filters",
     "drf_spectacular",
+    "osidb",
     "polymorphic",
+    "psqlextra",
+    "rest_framework",
     "rest_framework_simplejwt",
```
Elkasitu: note that while it most likely won't be an issue at this point in time, the order of installed apps *does* matter: https://docs.djangoproject.com/en/3.2/ref/settings/#installed-apps

config/settings.py
```diff
@@ -48,7 +48,9 @@
     "collectors.framework",
     "collectors.jiraffe",
     "collectors.nvd",
+    "collectors.osv",
     "collectors.product_definitions",
+    "collectors.ps_constants",
```
Elkasitu, line 53: does it make sense to have it if nobody is using it? what was the rationale for this collector?

config/settings.py
jobselko: Thanks, I was not aware of this. So I should I revert it or leave it in the original order just to be sure?

config/settings.py
jobselko: While I was checking snippets in OCP pre-prod, I found out that there is nothing from `osv` because `osv` was not in `settings.py`. I noticed that `ps_constants` was also missing,  so I asked @osoukup whether it should be there, and he said yes (but I do not know how `ps_constants` exactly works).

osoukup (approved):
LGTM

config/settings.py
osoukup: PS constants are definitely needed but they are needed by the tracker filing functionality so nobody could realize the missing data yet. This is my mistake not to set the app to start the periodic sync. I probably did the sync one time locally when developing it and then did not realize that the data is not being updated.

Elkasitu (approved):

PR #420, opened 2024-01-15
Add label verification-requested to New jira trackers with NEW affects.

jsvob:
Add label verification-requested to New jira trackers with NEW affects.

Closes OSIDB-1185

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -8,7 +8,7 @@ from apps.sla.framework import SLAFramework
 from apps.trackers.common import TrackerQueryBuilder
 from apps.trackers.exceptions import NoPriorityAvailableError
 from apps.trackers.models import JiraProjectFields
-from osidb.models import Impact
+from osidb.models import Affect, Impact
 
 logger = logging.getLogger(__name__)
 
@@ -127,6 +127,12 @@ class TrackerJiraQueryBuilder(TrackerQueryBuilder):
                 "flaw__cve_id", flat=True
             )
         )
+
+        # If all affects are NEW, add label validation-requested.
+        if set(self.tracker.affects.all().values_list("affectedness", flat=True)) == {
+            Affect.AffectAffectedness.NEW
+        }:
+            self._query["fields"]["labels"].append("validation-requested")
 
     def generate_sla(self):
         """
--- a/apps/trackers/tests/test_jira.py
+++ b/apps/trackers/tests/test_jira.py
@@ -150,6 +150,109 @@ class TestTrackerJiraQueryBuilder:
         assert "CVE-2000-2000" in labels
         assert len(labels) == 4
 
+    @pytest.mark.parametrize(
+        "external_system_id, affectedness, preexisting_val_req_lbl, result_val_req_lbl",
+        [
+            # New tracker with NEW affect gets the label added on save.
+            (None, Affect.AffectAffectedness.NEW, None, True),
+            # New tracker with non-NEW affect doesn't get the label added on save.
+            (None, Affect.AffectAffectedness.AFFECTED, None, False),
+            # Existing tracker with the label already present keeps it after update
+            # when its affects are NEW.
+            ("JIRA-123", Affect.AffectAffectedness.NEW, True, True),
+            # Existing tracker with the label already present doesn't keep it after update
+            # when its affects are not NEW.
+            ("JIRA-123", Affect.AffectAffectedness.AFFECTED, True, False),
+            # Existing tracker without the label present gets the label added if all
+            # its affects are NEW.
+            ("JIRA-123", Affect.AffectAffectedness.NEW, False, True),
+            # Existing tracker without the label already present doesn't get the label added
+            # if all its affects are not NEW.
+            ("JIRA-123", Affect.AffectAffectedness.AFFECTED, False, False),
+        ],
+    )
+    def test_generate_label_validation_requested(
+        self,
+        external_system_id,
+        affectedness,
+        preexisting_val_req_lbl,
+        result_val_req_lbl,
+    ):
+        """
+        test that the validation-requested label in the Jira query is generated correctly
+        """
+        flaw1 = FlawFactory(cve_id="CVE-2000-2000")
+        flaw2 = FlawFactory(embargoed=flaw1.embargoed, cve_id=None)
+        ps_module = PsModuleFactory(bts_name="jboss")
+        affect1 = AffectFactory(
+            flaw=flaw1,
+            ps_module=ps_module.name,
+            ps_component="component",
+            affectedness=affectedness,
+            resolution=Affect.AffectResolution.DEFER,
+        )
+        affect2 = AffectFactory(
+            flaw=flaw2,
+            ps_module=ps_module.name,
+            ps_component="component",
+            affectedness=affectedness,
+            resolution=Affect.AffectResolution.DEFER,
+        )
+        ps_update_stream = PsUpdateStreamFactory(ps_module=ps_module)
+
+        if not external_system_id:
+            # New tracker being created
+            meta_attr = {"test": 1}
+        else:
+            # Existing tracker supposedly being updated
+            if preexisting_val_req_lbl:
+                meta_attr = {
+                    "labels": [
+                        "CVE-2000-0000",
+                        "Security",
+                        "SecurityTracking",
+                        "pscomponent:component",
+                        "validation-requested",
+                    ]
+                }
+            else:
+                meta_attr = {
+                    "labels": [
+                        "CVE-2000-0000",
+                        "Security",
+                        "SecurityTracking",
+                        "pscomponent:component",
+                    ]
+                }
+        tracker = TrackerFactory(
+            affects=[affect1, affect2],
+            type=Tracker.TrackerType.JIRA,
+            ps_update_stream=ps_update_stream.name,
+            embargoed=flaw1.is_embargoed,
+            external_system_id=external_system_id,
+            meta_attr=meta_attr,
+        )
+
+        query_builder = TrackerJiraQueryBuilder(tracker)
+        query_builder._query = {"fields": {}}
+        query_builder.generate_labels()
+
+        labels = query_builder.query["fields"]["labels"]
+        assert "SecurityTracking" in labels
+        assert "Security" in labels
+        assert "pscomponent:component" in labels
+        assert "CVE-2000-2000" in labels
+        if result_val_req_lbl:
+            assert "validation-requested" in labels
+            assert len(labels) == 5
+        else:
+            assert "validation-requested" not in labels
+            assert len(labels) == 4
+
+        # NOTE: In real usage, this query is sent to Jira and it overwrites the list of
+        # labels stored in Jira, so if the label is not generated anymore, it effectively
+        # deletes it from Jira.
+
     def test_generate_sla(self, clean_policies):
         """
         test that the query for the Jira SLA timestamps is generated correctly
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -783,6 +783,7 @@ class AffectCVSSPutSerializer(AffectCVSSSerializer):
 class AffectSerializer(
     ACLMixinSerializer,
     BugzillaSyncMixinSerializer,
+    JiraAPIKeyMixin,
     TrackingMixinSerializer,
     IncludeExcludeFieldsMixin,
     IncludeMetaAttrMixin,
@@ -861,6 +862,66 @@ class AffectSerializer(
             + ACLMixinSerializer.Meta.fields
             + TrackingMixinSerializer.Meta.fields
         )
+
+    def update(self, affect, validated_data):
+        """
+        perform the tracker instance update
+        """
+
+        # NOTE: This overrides update() behavior provided by mixins, but mostly
+        #       mirrors the mixins. If mixins change, this method must be updated.
+        # NOTE: Look at TrackerSerializer.update() and
+        #       FlawPackageVersionSerializerMixin.update() for similar ways of
+        #       interposing actions within the mixins' update() MRO & flow.
+
+        ############################
+        # 1) prepare prerequisites #
+        ############################
+
+        # From ACLMixinSerializer.update()
+        validated_data = self.embargoed2acls(validated_data)
+
+        #########################
+        # 2) pre-update actions #
+        #########################
+
+        old_affectedness = affect.affectedness
+
+        #####################
+        # 3) update actions #
+        #####################
+
+        # From BugzillaSyncMixinSerializer.update()
+        for attr, value in validated_data.items():
+            setattr(affect, attr, value)
+
+        # From BugzillaBareSyncMixinSerializer.update()
+        affect.save(bz_api_key=self.get_bz_api_key())
+
+        ##########################
+        # 4) post-update actions #
+        ##########################
+
+        # When an affect goes from NEW affectedness to something else, any related tracker
+        # must have its validation-requested label removed, if it had one.
+        # Since the label is not stored in osidb, removal is done by regenerating the label
+        # by re-saving the tracker.
+
+        if old_affectedness == Affect.AffectAffectedness.NEW != affect.affectedness:
+            for tracker in affect.trackers.filter(type=Tracker.TrackerType.JIRA):
+                # do not raise validation errors here as the tracker is not what the user touches
+                # which would make the errors hard to understand
+                tracker.save(
+                    bz_api_key=self.get_bz_api_key(),
+                    jira_token=self.get_jira_token(),
+                    raise_validation_error=False,
+                )
+
+        #####################
+        # 5) return updated #
+        #####################
+
+        return affect
 
 
 @extend_schema_serializer(exclude_fields=["updated_dt"])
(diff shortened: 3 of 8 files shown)


.secrets.baseline
```diff
@@ -312,5 +312,5 @@
       }
     ]
   },
-  "generated_at": "2024-01-10T09:34:31Z"
+  "generated_at": "2024-01-16T08:14:55Z"
```
osoukup, line 315: This change seems to be unnecessary when no secrets were updated in the file.

apps/trackers/jira/query.py
```diff
@@ -128,6 +128,16 @@ def generate_labels(self):
             )
         )
 
+        # If all affects are NEW and if this tracker is New, add label validation-requested.
+        # Status values governed by Jira project workflow, not accessible through JiraProjectFields.
+        # Most probably, New will always be something like NEW, New, new.
+        if self.tracker.status.lower() == "new":
```
osoukup, line 134: I am not sure this condition should be there. I think we do not care for the tracker status very much. That is rather engineering think. Mostly the label will only be added on creation and as we do not set the tracker status I assume it will become some kind of **new** by default. Theoretically if someone removes the label while from our point of view it should still be there it would be re-added - which I think should happen regardless of the tracker status. I do no see any other option. WDYT?

osidb/serializer.py
```diff
@@ -861,6 +862,73 @@ class Meta:
...
+        # changes to/from NEW.
+
+        if (
+            old_affectedness != affect.affectedness
+            and Affect.AffectAffectedness.NEW in [old_affectedness, affect.affectedness]
+        ):
+            # Status values governed by Jira project workflow, not accessible through JiraProjectFields.
+            # Most probably, New will always be something like NEW, New, new.
+            for tracker in affect.trackers.filter(
+                status__in=["NEW", "New", "new"]
```
osoukup, line 916: Again I do not think that we should care for the tracker status.

apps/trackers/tests/test_jira.py
```diff
@@ -148,7 +148,57 @@ def test_generate_labels(self):
...
+        query_builder._query = {"fields": {}}
+        query_builder.generate_labels()
+
+        labels = query_builder.query["fields"]["labels"]
+        assert "SecurityTracking" in labels
+        assert "Security" in labels
+        assert "pscomponent:component" in labels
+        assert "CVE-2000-2000" in labels
+        assert "validation-requested" in labels
+        assert len(labels) == 5
```
osoukup, line 201: You should also test that when the affect gets updated to AFFECTED the label is removed.

osoukup (commented):
I have a few comments to address but mostly it looks good.

.secrets.baseline
jsvob: I was also surprised by this. It was required by CI. I will try submitting without this yet again to see if that was a fluke.

apps/trackers/tests/test_jira.py
jsvob: How? I thought about it for hours. The label is generated only when sent to Jira and not processed by OSIDB in any other way.

osidb/serializer.py
jsvob: Removed in today's commit. Please re-review.

apps/trackers/jira/query.py
jsvob: > I am not sure this condition should be there. I think we do not care for the tracker status very much. That is rather engineering think. Mostly the label will only be added on creation and as we do not set the tracker status I assume it will become some kind of new by default.

Changed the code in today's commit to only ever add the label when the tracker is being created, i.e. if the tracker already exists and moves to New from any other state, the label is not added. Re-reading the Description of the story, it seems today's commit brings the PR to what has been asked in Description (sorry for the confusion).

> Theoretically if someone removes the label while from our point of view it should still be there it would be re-added - which I think should happen regardless of the tracker status. I do no see any other option. WDYT?

I agree on no other options existing, but probably disagree on what these options are.

With today's commit, the label would **not** be regenerated in such a scenario. Before today's commit, it would be regenerated. Which do you want, then? Without today's commit or with it? I don't see a viable algorithm that would fulfill all requirements in any other way (ignoring tracker state but re-adding the label if state is New is logically inconsistent, unless I'm missing something).

.secrets.baseline
jsvob: Interesting, it didn't complain today. Fixed.

apps/trackers/jira/query.py
```diff
@@ -128,6 +128,14 @@ def generate_labels(self):
...
         )
 
+        # If all affects are NEW and if this tracker is new, add label validation-requested.
+        if not self.tracker.external_system_id:
+            affect_states = list(
+                self.tracker.affects.all().values_list("affectedness", flat=True)
+            )
+            if set(affect_states) == set([Affect.AffectAffectedness.NEW]):
+                self._query["fields"]["labels"] += ["validation-requested"]
+
```
Elkasitu, line 138: I think this could be more efficient and less verbose if we leverage querysets:

```suggestion
        # If all affects are NEW and if this tracker is new, add label validation-requested.
        if self.tracker.external_system_id or self.tracker.affects.exclude(affectedness=Affect.Affectedness.NEW).exists():
            return
        self._query["fields"]["labels"] += ["validation-requested"]
```

osidb/serializer.py
```diff
@@ -861,6 +862,69 @@ class Meta:
...
+                    jira_token=self.get_jira_token(),
+                    raise_validation_error=False,
+                )
+
+        #####################
+        # 5) return updated #
+        #####################
+
+        return affect
+
```
Elkasitu, line 927: I think this is a bad idea for two reasons:

1. The obvious one being that it's brittle, instead of depending on an abstraction's API (the mixins) we depend on its implementation details, I know we do this already elsewhere but that doesn't mean it should be the norm.
2. Business logic shouldn't really live this far up the stack (visual layer), I presume the reason for doing so is to have the user's JIRA API key, but do we really care about recording the person that changed the affectedness as the person setting/removing the label? I feel like an automated action like this one can be simply marked as "done by OSIDB" using OSIDB's JIRA API key, then this logic could live much more comfortably as a signal or within a save override, and in turn be less brittle.

WDYT?

apps/trackers/tests/test_jira.py
osoukup: This is a very good question and I see I was writing faster than thinking :smile: So we should test this but it will not be possible right now. It intersects with https://issues.redhat.com/browse/OSIDB-1440 in the sense that for both we need to store the labels on fetch from Jira which we do not do now. Basically we need to add the labels to https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/jiraffe/convertors.py#L115 and then the integration test could verify what we need. I see two options and if you agree feel free to proceed however you prefer. One is to add the labels on Jira tracker fetch as part of this PR and add the test verifying the label presence or removal as expected. Or note it down somewhere (new Jira?) that this test(s) should be written and leave it for later after OSIDB-1440 is done.

osidb/serializer.py
osoukup: I agree that it does not seem right to have it there and it was me who started it. However I think that the audit is important so if we can do it on some other layer etc. then we can shift this logic there. Another reason is that on the serializer layer we have both the previous and the new object. If it is also simple to have it on the object level, then no problem to shift it there. Otherwise the comparison of what has changed would not be possible.

jsvob: I'll let you sort it out on the design level.

Low-level resolution suggestion: If Affect.save() always also requires jira key besides bz key, the custom logic could be moved there IMO. Would that be good enough for this PR?

Elkasitu: Let me be a bit more clear:

I am *against* having any api key logic further down the stack (even if it already is the case).
I am *for* having the label-setting logic further down the stack.

The latter implies that we don't track or audit _who_ triggered the setting of the label, because it is **unimportant**, setting the label is not something a user can do deliberately, the action of setting the label is not transparent to them, they are not aware of how or what triggers it, so why should they be liable for it? It's an automation from the system that may be triggered by a user's action, in this case changing an affect's affectedness for which we already have an audit trail.

TL;DR: unless we agree not to keep an audit of who set/unset the label, keep the current implementation.

apps/trackers/jira/query.py
```diff
@@ -128,6 +128,17 @@ def generate_labels(self):
...
+        if (
+            not self.tracker.external_system_id
+            or "validation-requested" in self.tracker.meta_attr.get("labels", [])
+        ):
+            affect_states = list(
+                self.tracker.affects.all().values_list("affectedness", flat=True)
+            )
+            if set(affect_states) == set([Affect.AffectAffectedness.NEW]):
+                self._query["fields"]["labels"] += ["validation-requested"]
+
```
jobselko, line 141: To avoid unnecessary nested `if` and `set(list())` constructions for better readability:

```
        # Add "validation-requested" if a tracker is new and all affects are new
        if (
            not self.tracker.external_system_id
            or "validation-requested" in self.tracker.meta_attr.get("labels", [])
        ) and (
            set(self.tracker.affects.all().values_list("affectedness", flat=True))
            == {Affect.AffectAffectedness.NEW}
        ):
            self._query["fields"]["labels"] += ["validation-requested"]
```

apps/trackers/tests/test_jira.py
```diff
@@ -150,6 +150,178 @@ def test_generate_labels(self):
         assert "CVE-2000-2000" in labels
         assert len(labels) == 4
 
+    @pytest.mark.parametrize(
+        "external_system_id, affectedness, resolution, num_labels",
```
jobselko, line 154: Maybe a stupid question, but why you are parametrizing `resolution`? It does not seem to be related to `validation-requested`.

apps/trackers/tests/test_jira.py
```diff
@@ -150,6 +150,178 @@ def test_generate_labels(self):
...
+            ),
+            (
+                "JIRA-123",
+                Affect.AffectAffectedness.AFFECTED,
+                Affect.AffectResolution.FIX,
+                4,
+            ),
+        ],
+    )
+    @pytest.mark.parametrize(
```
jobselko, line 172: I would recommend using only one `@pytest.mark.parametrize` instead of two (easier readability and maintenance in this case).

apps/trackers/tests/test_jira.py
jsvob: `Affect.AffectResolution.NOVALUE` is not allowed with `Affect.AffectAffectedness.AFFECTED`. Wanted to have NEW affects in their most probable configuration.

apps/trackers/tests/test_jira.py
jsvob: Please suggest a code change to that effect.

AFAIK the test needs a cartesian product of possible initial states and possible new states. For example, I had a bug in a single resulting test and having just a few scenarios without the cartesian product would have probably missed it.

apps/trackers/jira/query.py
jsvob: Fixed, thank you.

apps/trackers/tests/test_jira.py
jobselko: In `AFFECTEDNESS_VALID_RESOLUTIONS`, you can see that `Affect.AffectAffectedness.NEW` and `Affect.AffectAffectedness.AFFECTED` have more options in common, so you can use e.g. `Affect.AffectResolution.DEFER`.

`AffectResolution` does not affect the test, so it is better not to parametrize non-required values because this makes a test harder to understand, but I will leave the final decision up to you.

apps/trackers/tests/test_jira.py
jobselko: I am not talking about removing test cases. It is possible to keep all necessary tests, but inside one `parametrize`. In most cases, when you are using multiple `parametrize` in one test, it is wrong (e.g. you are testing too many unrelated situations). In your case, you are testing that when `tracker is new` and `all affects are new`, the `validation-requested` label should be added (otherwise, the `validation-requested` label should be absent/removed).

You could either test everything at once:
```
@pytest.mark.parametrize("tracker_id, affectedness, is_validation_requested, new_tracker_id, new_affectedness, new_is_validation_requested")
def test()
```
or create two separate tests (one for creation and second for update):
```
@pytest.mark.parametrize("tracker_id, affectedness, is_validation_requested")
def test_1()

@pytest.mark.parametrize("new_tracker_id, new_affectedness, is_validation_requested")
def test_2()
```

But again, I will leave the final decision up to you.

apps/trackers/tests/test_jira.py
jsvob: Removed `Affect.AffectResolution` from parametrization and used static `Affect.AffectResolution.DEFER`.

apps/trackers/tests/test_jira.py
jsvob: Thank you for the explanation, I think I understand the issue now. The part of the test where updating a tracker was tested was overlapping with the beginning of the test that also tested some of the scenarios with updating a tracker. I used your suggestions (not verbatim) to remove the duplication and flatten the parametrization without eliminating tested scenarios.

jobselko: Good job, your solution is even better than what I had in my mind :)

osidb/serializer.py
```diff
@@ -861,6 +862,69 @@ class Meta:
...
+        # When an affect goes from NEW affectedness to something else, any related tracker
+        # must have its validation-requested label removed, if it had one.
+        # Since the label is not stored in osidb, removal is done by regenerating the label
+        # by re-saving the tracker.
+
+        if (
+            old_affectedness != affect.affectedness
+            and old_affectedness == Affect.AffectAffectedness.NEW
+        ):
```
jobselko, line 912: can be changed to `if old_affectedness == Affect.AffectAffectedness.NEW != affect.affectedness:`

jsvob: It can, but I'd rather not. It's technically correct, but its correctness seems subjectively so unintuitive seeing such an expression stops me for some time thinking about it. Python is AFAIK an outlier language in [doing that correctly](https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-or_expr:~:text=Formally%2C%20if%20a%2C%20b%2C%20c%2C%20%E2%80%A6%2C%20y%2C%20z%20are%20expressions%20and%20op1%2C%20op2%2C%20%E2%80%A6%2C%20opN%20are%20comparison%20operators%2C%20then%20a%20op1%20b%20op2%20c%20...%20y%20opN%20z%20is%20equivalent%20to%20a%20op1%20b%20and%20b%20op2%20c%20and%20...%20y%20opN%20z). In e.g. C that would be illegal and in Javascript, it is a left-to-right series of individual comparisons that results in expressions like `3 == 3 == 1` and `3 === 3 === true` being true.

jobselko: Chained comparisons are pretty common in Python, and you can read this one directly the way it is written:
Do something if `old affectedness is NEW and this is different from the current affectedness` (i.e. `A == B and B =! C` is simplified to `A == B != C`).

I think that the argument regarding invalidity in other languages is not fair. For example, somebody can use increment operator ++ in Java even though this is invalid in Python. But again, I will leave it up to you.

jsvob: Changed based on your feedback.

osidb/serializer.py
jsvob: Can we then merge this PR? I think having the label-setting logic further down the stack would require reengineering that would be out of the scope of this task.

apps/trackers/jira/query.py
jsvob: Can we merge this PR? I think this discussion became somewhat out-of-date. If someone doesn't like the current behavior (the label is never re-added, once removed), they can file a bug.

osoukup:
I submitted a review to make the pending comments not pending any more. Sorry for not noticing that they were actually not a real comments...

apps/trackers/jira/query.py
jsvob: Thank you! We seem to agree on leaving it as it is now then.

(for posterity, the messages from Jan 18 and Jan 22 happened in parallel universes without knowing about each other, then the universes merged today)

apps/trackers/jira/query.py
```diff
@@ -129,7 +129,10 @@ def generate_labels(self):
         )
 
         # If all affects are NEW and if this tracker is new, add label validation-requested.
-        if not self.tracker.external_system_id:
+        if (
+            not self.tracker.external_system_id
+            or "validation-requested" in self.tracker.meta_attr.get("labels", [])
+        ):
```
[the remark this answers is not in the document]
jsvob: OK, I'll make it stateless - add the label whenever all affects are NEW, drop the label whenever they aren't.

Relevant discussion: https://github.com/RedHatProductSecurity/osidb/pull/420#discussion_r1463393381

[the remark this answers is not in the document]
jsvob: Please review commit [6012a47](https://github.com/RedHatProductSecurity/osidb/pull/420/commits/6012a478a616279dbb3998a878c116c898e64b29)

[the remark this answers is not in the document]
jsvob: I'm still leaving the storage of labels in meta_attr because it makes the test more descriptive and close to reality and the tracking would have to be added in OSIDB-1440 anyway (i.e. removing it in PR420 would add extra work).

PR #422, opened 2024-01-18
Refactor internal ACLs and change their using in Snippet

jobselko:
This PR updates the setting of internal ACLs for the `Snippet` model to avoid overriding the `__init__` method, as mentioned in https://docs.djangoproject.com/en/3.2/ref/models/instances/#creating-objects. 
It also reworks the usage of ACLs and fixes a typo in the `root_url` fixture.

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -493,7 +493,7 @@ class ACLMixin(models.Model):
         """
         get set of all ACLs
         """
-        return self.acls_read | self.acls_write | self.acls_internal
+        return self.acls_read | self.acls_write
 
     def _validate_acls_known(self):
         """
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1535,10 +1535,10 @@ class Snippet(ACLMixin, AlertMixin, TrackingMixin):
         NVD = "NVD"
         OSV = "OSV"
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def save(self, *args, **kwargs):
         # set internal ACLs
         self.set_internal()
+        super().save(*args, **kwargs)
 
     # internal primary key
     uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
@@ -1627,11 +1627,10 @@ class Snippet(ACLMixin, AlertMixin, TrackingMixin):
 
     def _validate_acl_identical_to_parent_flaw(self) -> None:
         """
-        No validations are run for snippet's flaw as its ACLs can be different.
+        No ACL validations are run for snippet's flaw as its ACLs can be different.
         However, snippet should always have internal ACLs.
         """
-        if {self.acl_read[0], self.acl_write[0]} != self.acls_internal:
-            raise ValidationError("Snippet must have internal ACLs.")
+        pass
 
 
 class AffectManager(ACLMixinManager, TrackingMixinManager):
--- a/osidb/tests/conftest.py
+++ b/osidb/tests/conftest.py
@@ -57,43 +57,33 @@ def good_cve_id2():
 
 
 @pytest.fixture
-def public_read_group():
-    return generate_acls(settings.PUBLIC_READ_GROUPS)
+def public_read_groups():
+    return [uuid.UUID(acl) for acl in generate_acls(settings.PUBLIC_READ_GROUPS)]
 
 
 @pytest.fixture
-def embargo_read_group():
-    return generate_acls([settings.EMBARGO_READ_GROUP])
+def embargoed_read_groups():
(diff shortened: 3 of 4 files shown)


osidb/tests/conftest.py
```diff
@@ -66,6 +66,11 @@ def embargo_read_group():
     return generate_acls([settings.EMBARGO_READ_GROUP])
 
 
+@pytest.fixture
```
osoukup, line 69: I know this is not your fault but this is a weird set of fixtures. I do not understand the purpose of having both the singular and plural once. The singular once does not even seem to be used for anything than creating the plural once. Also one can easily make a type omitting **s** and will get an existing but different thing. When we are refactoring this I would be for merging the singular and plural ones together.

osidb/models.py
```diff
@@ -1535,10 +1535,10 @@ class Source(models.TextChoices):
         NVD = "NVD"
         OSV = "OSV"
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def save(self, *args, **kwargs):
         # set internal ACLs
         self.set_internal()
+        super(Snippet, self).save(*args, **kwargs)
```
osoukup, line 1541: Is there a reason to use the arguments of **super** here?

osidb/mixins.py
```diff
@@ -493,7 +493,7 @@ def acls_all(self):
         """
         get set of all ACLs
         """
-        return self.acls_read | self.acls_write | self.acls_internal
+        return self.acls_read | self.acls_write
```
osoukup, line 496: Why do you remove the internal ACLs from here? It clearly changes the method behavior in an unexpected way. All means all. If we need a method to get all without internal, we should rather write a new one.

osoukup (changes requested):
I do not understand the **acls_all** change. Otherwise it looks OK.

osidb/mixins.py
jobselko: I am sorry for unclear commit message, all information is still here :)

`self.acls_read` contains `return self.acls_public_read | self.acls_embargo_read | self.acls_internal_read`, and `self.acls_write` contains `return(self.acls_public_write | self.acls_embargo_write | self.acls_internal_write)`.

osoukup: Oh I see! Then it makes sense to remove.

osidb/tests/conftest.py
jobselko: I agree with you. In my case, I just want to avoid writing `[uuid.UUID(acl) for acl in generate_acls([settings.INTERNAL_WRITE_GROUP])]` every time I need it.

I was thinking about refactoring but was unsure whether I understood all ACLs usage correctly, so I left it for later.

osidb/models.py
jobselko, line 1: I am unsure how to handle `_validate_acl_identical_to_parent_flaw()` in `Snippet`. This method must be overridden here because otherwise, it would fail in `ACLMixin` (parent flaw's ACL can be different from snippet's ACL).

I see two options:

- keep the current body (but then it seems like the body is required, which is not true)
- change the body to simple `pass` (this is sufficient because internal ACLs will be set in `save()`).

@osoukup @Elkasitu Any ideas how to approach this?

Elkasitu: https://github.com/RedHatProductSecurity/osidb/pull/277/files#diff-8d7ea87df8ca210c5df403a76b8228d0d1e1d2ae7fc99f54b5c6f7e02d396deeR2870-R2881

osoukup: I would be for the second option with a comment why we do it.

osidb/tests/test_snippet.py
```diff
@@ -51,20 +50,6 @@ def test_create_snippet(self, internal_read_groups, internal_write_groups):
...
-        Tests that a snippet with non-internal ACLs raises an error.
-        """
-        snippet = get_snippet()
-
-        snippet.acl_read = embargoed_read_groups
-        snippet.acl_write = embargoed_write_groups
-
-        with pytest.raises(ValidationError, match="Snippet must have internal ACLs."):
-            snippet.save()
-
```
Elkasitu, line 77: Instead of removing the test, you could simply check that after save the ACLs are internal

osidb/tests/conftest.py
```diff
@@ -29,7 +29,7 @@ def enable_db_access_for_all_tests(db):
 
 @pytest.fixture
 def root_url():
-    return "http://osdib-service:8000"
```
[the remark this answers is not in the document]
osoukup: I assume that for the purpose of the tests it can be any random string. If not I do not get why it did not break.

osidb/tests/test_snippet.py
jobselko: Internal ACLs are tested in `test_create_snippet()` and `test_create_flaw_from_snippet()`.

PR #423, opened 2024-01-22
Improve performance of include_fields filter

Elkasitu:
This PR introduces changes to increase the performance of the `include_fields` filter by leveraging `QuerySet.prefetch_related()` and `QuerySet.only()`.

Before this commit, a query such as `/flaws?include_fields=cve_id` with 10 concurrent users would yield about 5.3 requests per second and have an average response time of 2180ms

image

After this commit, the same query with 10 concurrent users yields 10 requests per second and has an average response time of 179ms.

image

In practice this means that for 10 concurrent users, there is no bottleneck for this particular query, on top of that it means that the performance gains are of around an order of magnitude in the best-case scenario.

The max throughput was also tested, and with 100 concurrent users (unrealistic scenario) the average requests per second is about 30 and the average response time 1200ms, which is manageable.

image

I believe that at 100 users, the bottleneck is mostly at the CPU/RAM level and could be further improved by scaling horizontally, meaning that code-wise the bottlenecks are unlikely.

### DIFF
--- a/etc/pg/postgresql.conf
+++ b/etc/pg/postgresql.conf
@@ -61,7 +61,7 @@ listen_addresses = '*'		# what IP address(es) to listen on;
 					# defaults to 'localhost'; use '*' for all
 					# (change requires restart)
 #port = 5432				# (change requires restart)
-max_connections = 100			# (change requires restart)
+max_connections = 200			# (change requires restart)
 #superuser_reserved_connections = 3	# (change requires restart)
 #unix_socket_directories = '/var/run/postgresql, /tmp'	# comma-separated list of directories
 					# (change requires restart)
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -2,6 +2,7 @@
 Implement filters for OSIDB REST API results
 """
 
+from django.core.exceptions import FieldDoesNotExist
 from django_filters.rest_framework import (
     BaseInFilter,
     BooleanFilter,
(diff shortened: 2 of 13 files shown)


osoukup (approved):
Great improvements! Well documented too. LGTM

etc/pg/postgresql.conf
```diff
@@ -61,7 +61,7 @@ listen_addresses = '*'		# what IP address(es) to listen on;
 					# defaults to 'localhost'; use '*' for all
 					# (change requires restart)
 #port = 5432				# (change requires restart)
-max_connections = 100			# (change requires restart)
+max_connections = 200			# (change requires restart)
```
JimFuller-RedHat, line 64: Informational - osidb stage/prod are either 3000/5000 max conns ... though it is good to constrain this further for local development

JimFuller-RedHat (approved):
nice and LGTM

osidb/tests/endpoints/flaws/test_flaws.py
```diff
@@ -778,6 +778,32 @@ def test_list_flaws_include_and_exclude_fields(self, auth_client, test_api_uri):
...
+        assert body["count"] == 0
+
+        affect = AffectFactory()
+        AffectFactory(flaw=affect.flaw)
+
+        response = auth_client().get(f"{test_api_uri}/flaws?include_fields=affects")
+        assert response.status_code == 200
+        body = response.json()
+        assert body["count"] == 1
+        # only one key should be present, affects"
```
jobselko, line 803: Redundant `"`

jobselko (approved):
LGTM, great job!

osidb/filters.py
```diff
@@ -70,7 +71,94 @@ def __init__(self, data=None, queryset=None, *, request=None, prefix=None):
...
+                continue
+            field_set.add(fname)
+        return prefetch_set, field_set
+
+    def include_fields_filter(self, queryset, name, value):
+        """
+        Optimizes a view's QuerySet based on user input.
+
+        This filter will attempt to optimize a given view's queryset based on an
+        allowlist of fields provided by the user in order to improve performance.
```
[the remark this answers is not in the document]
Elkasitu: I will but this is essentially knowledge of the django-filters API

PR #424, opened 2024-01-22
after-affect-update tracker updates

osoukup:
The second part of this

Closes OSIDB-97

TODO ... not ready for a review


osidb/serializer.py
```diff
@@ -864,6 +864,73 @@ class Meta:
...
+        if old_affect.is_community:
+            return
+
+        # we only need to sync the trackers when crucial attributes change
+        # plus in the case of the impact we only care for the increase
+        #
+        # the crucial attributes are those influencing the SLA deadline or linking
+        # however in the case of PS module we cannot auto-adjust the trackers
+        # since we simply cannot guess which PS update stream is correct
+        if not differ(old_affect, new_affect, ["flaw", "ps_component"]) and Impact(
```
jobselko, line 920: `flaw` is not tested

jobselko (approved):
LGTM, discussed in person

osidb/serializer.py
jobselko: Oh, I overlooked it. Sorry :)

osoukup:
I am going to merge this despite the API breaking change warning. The reason is that as the production OSIDB instance is in fact still read-only and the breaking change is in POST and PUT methods. It is on purpose as now on the affect endpoint query may trigger a tracker write action which would not be possible without the API key and also at the point of sending the query the user cannot not know whether it actually triggers it and if so in which BTS.

PR #425, opened 2024-01-23
Update cryptography to latest security revision

Elkasitu:
Versions below 41.0.6 are vulnerable to CVE-2023-49083


jobselko (approved):

jobselko:
Will you add also `jinja2` which was reported by dependabot?

osoukup (approved):
LGTM. Thank you for keeping an eye on this!

Elkasitu:
> Will you add also `jinja2` which was reported by dependabot?

It's a dev-only dependency and we don't use the impacted codepath so I've just dismissed it

PR #426, opened 2024-01-24
Change Bugzilla collector and Flaw model to allow multi-components in bz_summary

costaconrado:
This PR changes the way Bugzilla collector parse summary.

This PR changes flaw model so component can support multiple components.
This PR also changes the parser of bugzilla collector to bz_summary be parsed as a list instead of a single component.
This PR introduces a major change in API (component became components) and the filter for flaws changed in order to support it.
This PR also adds a property called `component_str` for checks that uses component as a condition.

Closes OSIDB-1420.


costaconrado:
To reviewer: Sorry for the single commit PR with multiple changes, the migration should be the 1st and it makes the API fail the tests making it unable to be in a different commit.

Elkasitu (changes requested):
This change is not backwards compatible and contains non-safe migrations, the easiest solution at this point in time would probably be:

* Keep `component` field as-is
* Deprecate `component` field in the API
* Create new field `components`, prefill with existing `component` field
* Expose `components` field in API

osoukup (changes requested):
The new functionality looks OK to me but I agree with @Elkasitu that we should for now keep the old **component** field for the backward compatibility. It should be deprecated and maybe just return the first item from **components** list.

costaconrado:
Makes sense, I've deprecated the `component` field and added the `components`. @osoukup do you think it would be better to return the 1st value of `components` or split the value by colon and do a contains? (e.g. if user pass `golang: archive/zip` I return all flaws that contains both `golang` and `archive/zip`)

osoukup:
> Makes sense, I've deprecated the `component` field and added the `components`. @osoukup do you think it would be better to return the 1st value of `components` or split the value by colon and do a contains? (e.g. if user pass `golang: archive/zip` I return all flaws that contains both `golang` and `archive/zip`)

@costaconrado I am not sure I understand the question :angel: I mean that the value of the old attribute **component** should simply be the first item in the new **components** list attribute. It was just mean for the backwards compatibility so there is some meaningful value until we deprecate it full. It is probably not very important which item from **components** would be returned as the **component** - first/last/random/whatever. Just some meaningful value. It will be temporary anyway.

osidb/serializer.py
```diff
@@ -1453,6 +1456,16 @@ class Meta:
             + WorkflowModelSerializer.Meta.fields
         )
 
+    def create(self, validated_data):
+        component = validated_data.pop("component", None)
+        components = validated_data.pop("components", None)
+        if component and not components:
+            # Split the component and update the components to provide retro-compatibility
+            validated_data["components"] = [
+                value.strip() for value in component.split(":")
```
osoukup, line 1465: Even though the split would have no impact (so would not be an issue) when there is only a single component, I would not expect the user to put multiple components where one is expected.

osoukup (changes requested):
I propose slight changes in how the old `component` is treated - it is rather a single component even though there are multiple of them in the original Bugzilla summary and I think it should be still consider so before we remove it.

osidb/filters.py
```diff
@@ -301,7 +314,7 @@ class Meta:
             "cvss3_score": ["exact", "lt"] + LT_GT_LOOKUP_EXPRS + LTE_GTE_LOOKUP_EXPRS,
             "nvd_cvss2": ["exact"],
             "nvd_cvss3": ["exact"],
-            "component": ["exact"],
+            "components": ["exact"],
```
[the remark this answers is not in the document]
costaconrado: good question: user send `a,b,c` and the `flaw.components` should contain all of them if we have:
```
flaw1 = FlawFactory(components=["a", "b"])
flaw2 = FlawFactory(components=["a", "b", "c"])
flaw3 = FlawFactory(components=["a", "b", "c", "d"])
```

the return of `/flaws?components=a,b,c` is the content of `flaw2` and `flaw3`. I'm adding this scenario in the tests.

osidb/serializer.py
costaconrado: makes sense, I will add the input string as a single value in `components`

osidb/models.py
```diff
@@ -745,8 +745,14 @@ class FlawRequiresSummary(models.TextChoices):
 
     # flaw component was originally a part of the Bugzilla summary
     # so the value may depend on how successfully it was parsed
-    component = models.CharField(max_length=100, blank=True)
-
+    component = deprecate_field_custom(
+        models.CharField(max_length=100, blank=True),
+        # required to keep backwards compatibility
+        return_instead=lambda self: ": ".join(self.components),
```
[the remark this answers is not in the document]
osoukup: I see. My view is that the fix is introducing `components` and deprecating the `component`. I would rather consider the change of the format of `component` attribute as not backward compatible.

I have no strong preference as `component` is going to go away anyway.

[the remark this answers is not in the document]
costaconrado: Makes sense, I will leave `component` returning `net/http` in this example.

costaconrado:
I have changed the behavior so `component` returns the last value of `components` returning the same value it would return without this PR. Also if user pass something like `golang: net/http` to the old field it will be not parsed, just accepted as a single component.
I have also removed the field `component` from tests that uses `apps.workflows.checks`, the previous solution created a `component_str` property in flaw model but as discussed before it currently not is being used anywhere but in the test.

Please take a second look in this PR when you have some time @osoukup. Thanks!

osidb/filters.py
```diff
@@ -268,6 +271,16 @@ def is_major_incident_filter(self, queryset, name, value):
             ]
         )
 
+    def component_filter(self, queryset, name, value):
+        """
+        Based on the `value`, returns all Flaws which contains all components listed separated by colon.
+        This enables retro compactibility with deprecated field `component`.
+        """
+        if value:
+            components = [value.strip() for value in value[0].split(":")]
```
osoukup, line 280: I am not sure I understand this correctly but is the **colon** meant to be the separator of the component list in the query string? I think that comma is more common separator. Why colon?

osidb/tests/endpoints/test_endpoints.py
```diff
@@ -382,10 +382,12 @@ def test_custom_exception_serialization(self):
         assert response.status_code == 409
         assert response.json()["detail"] == "This was a big failure"
 
+
```
osoukup, line 385: nitpick: these formatting changes should be part of the previous commit (no need to change it)

osoukup (approved):
LGTM

osidb/filters.py
costaconrado: this function is used only on the old field `component`, if user try to filter with something like `/flaws?component=golang: net/http` it would convert the filter for the new field that stores `golang` and `net/http` as different components

costaconrado: a second thought on that: since we changed to not parse the input of the user directly in the `component` (old) field, I believe I should just search as is, submitting the change.

PR #428, opened 2024-01-24
Preserve Jira tracker labels.

jsvob:
Closes OSIDB-1440

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -9,6 +9,7 @@ from apps.trackers.common import TrackerQueryBuilder
 from apps.trackers.exceptions import NoPriorityAvailableError
 from apps.trackers.models import JiraProjectFields
 from osidb.models import Affect, Impact
+from osidb.validators import CVE_RE_STR
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM

apps/trackers/jira/query.py
```diff
@@ -112,19 +119,40 @@ def generate_description(self):
...
+                or lbl in ["SecurityTracking", "Security", "validation-requested"]
+                or CVE_NAME_RE.match(lbl)
+            ):
+                labels_allowed_to_edit_here.add(lbl)
+
+        # These labels are from elsewhere than this method and preserved with their ordering intact.
+        # Because the engineering may use them.
+        labels_to_preserve = [
+            x for x in all_existing_labels if x not in labels_allowed_to_edit_here
+        ]
```
jobselko, line 143: `labels_allowed_to_edit_here` are not necessary for the rest of the code, so there is no need to have them.
We are interested only in `labels_to_preserve`, which can be created directly from `all_existing_labels` in one loop and condition (if I am right, then `all_existing_labels` = `labels_allowed_to_edit_here` + `labels_to_preserve`).

apps/trackers/jira/query.py
```diff
@@ -112,19 +119,40 @@ def generate_description(self):
...
+        self._query["fields"]["labels"].extend(
+            [
+                "SecurityTracking",
+                "Security",
+                f"pscomponent:{self.ps_component}",
+            ]
+            + list(  # add all linked non-empty CVE IDs
+                self.tracker.affects.exclude(flaw__cve_id__isnull=True).values_list(
+                    "flaw__cve_id", flat=True
+                )
```
jobselko, line 155: You can build `self._query["fields"]["labels"]` at once, i.e.:

```
        self._query["fields"]["labels"] = [
            "SecurityTracking",
            ...,
            *list1,
            *list2,
        ]
```

jobselko (changes requested):

apps/trackers/jira/query.py
jsvob: You are correct in all points. I don't understand how that is better or more understandable, but did as you said. (I presume you didn't just want to state truthful statements, but wanted me to change it, right?) Please review commit [2c8f57f](https://github.com/RedHatProductSecurity/osidb/pull/428/commits/2c8f57f136f522f3cffc42df037bee5920c849ae)

apps/trackers/jira/query.py
jsvob: I don't understand how that is better or more understandable, but did as you said. (I presume you didn't just mean that _I can_ but that _I should_, otherwise why write it, right?) Please review commit [2c8f57f](https://github.com/RedHatProductSecurity/osidb/pull/428/commits/2c8f57f136f522f3cffc42df037bee5920c849ae)

jobselko: Why not create a variable directly if we can :)? I think this is more compact and understandable. Thank you for the update.

apps/trackers/jira/query.py
jobselko: Initially, there was unclear sorting, two loops and two variables. Now, there is one loop, one variable and straightforward logic (with the code of half-length). I think the updated version is definitely better. Thank you!

jobselko (approved):

apps/trackers/jira/query.py
```diff
@@ -12,6 +13,12 @@
 
 logger = logging.getLogger(__name__)
 
+# From sdengine/models.py (constant NAME_RE).
+# Restrict CVE ID to valid years (first one assigned was in 1999; hopefully we'll transcend
+# software vulnerabilities in year 3000+), and disallow ID of 0000 and leading zeros for 4+
+# IDs (e.g. -00001 is an invalid ID).
+CVE_NAME_RE = re.compile(r"CVE-(?:1999|2\d{3})-(?!0{4})(?:0\d{3}|[1-9]\d{3,})")
```
jobselko, line 20: You can import this from `osidb/validators.py`

jsvob: Good catch, thanks.

PR #429, opened 2024-01-25
Tracker status field not writable.

jsvob:
Closes OSIDB-1780


jsvob:
@RedHatProductSecurity/osidb-devs I'm not sure about the changes in openapi.yml and its potential knock-on effects in projects that use the API definition. In this PR I tried going the usual route of marking the field as readonly in the serializer and as blankable in the model (without `blank=True` no POST/PUT can succeed because the field can't be blank because well it's read only and thus not provided in the request). If it's a problem, I think we can make it `blank=False` and just fake-fill it in `TrackerSerializer.create()` and `TrackerSerializer.update()` I guess (but that didn't seem right unless we agree it is)?

osoukup (approved):
LGTM. Even this is shown as breaking it is in fact not with the still-read-only production. It makes sense to me to set the **status** field with **black=True** to enable omitting it on OSIDB creation. It will anyway immediately propagate to Jira and back and the value will be filled from there. In OSIDB we basically only care for what is in Jira to display it and we do not want to touch it at all.

PR #432, opened 2024-01-31
Fix preserving jira labels. Deal with HStoreField storing list as str.

jsvob:
Bug introduced in PR428.

Closes OSIDB-2042

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -1,6 +1,7 @@
 """
 Jira tracker query generation module
 """
+import json
 import logging
 from functools import cached_property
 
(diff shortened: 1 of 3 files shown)


apps/trackers/jira/query.py
```diff
@@ -119,6 +120,9 @@ def generate_labels(self):
         """
 
         all_existing_labels = self.tracker.meta_attr.get("labels", [])
+        if isinstance(all_existing_labels, str):
+            # HStoreField's values are strings, so convert back to list.
+            all_existing_labels = literal_eval(all_existing_labels)
```
osoukup, line 125: There are already some similar pieces of code
(eg. https://github.com/RedHatProductSecurity/osidb/blob/master/apps/bbsync/query.py#L342)
and they use **json** library instead which seems to me as a more appropriate solution - **literal_eval** is a general evaluation while **json.loads** is specifically for loading JSON data. WDYT about using it instead?

jsvob: fixed

osoukup (approved):
LGTM

PR #439, opened 2024-02-06
WIP: Stream pre-selection for fedramp

jsvob:
Closes OSIDB-1876


osidb/models.py
```diff
@@ -2088,15 +2088,21 @@ def is_community(self) -> bool:
             name=self.ps_module, ps_product__business_unit="Community"
         ).exists()
 
+    # TODO unlike SFM2 this is mostly unused. Is that intentional or did something get missed?
```
jsvob, line 2091: @osoukup Unlike SFM2, is_compliance_priority is mostly unused in OSIDB. Is that intentional or did something get missed?

(This is sth I identified during work on stream preselection for fedramp, but should be filed as a separate Jira if a valid concern)

osidb/models.py
```diff
@@ -2088,15 +2088,21 @@ def is_community(self) -> bool:
...
+    #      - see e.g. backends/jira/rest/request.py in SFM2
+    # TODO unlike SFM2, SLA can't (?) evaluate the new streams part as long as this is a property
```
jsvob, line 2093: @osoukup unlike SFM2, SLA can't (?) evaluate the new streams part of is_compliance_priority as long as this is a property on affect (since affect doesn't have streams). Should we file a new Jira to fix it?

(This is sth I identified during work on stream preselection for fedramp, but should be filed as a separate Jira if a valid concern)

apps/trackers/product_definition_handlers/fedramp_handler.py
```diff
@@ -0,0 +1,84 @@
...
+            #             ^ TODO/NOTE: I prefer the last option, but don't understand implications.
+            #       TODO/NOTE: Also, I *think* that it doesn't matter much because all streams
+            #                  higher than selection are then selected anyway, so it seems the
+            #                  intent is to select whole ranges starting with the oldest in the
+            #                  respective policy files (ubi_packages.yml and compliance_priority.yml).
+
+            # TODO: Why is this in SFM2? Why can't we just preselect all, given the fact that
+            #       the block after that tries to select anything newer than preselected anyway?!
+            # If multiple fedramp streams are configured, only seed the
+            # streams_to_preselect with the first oldest one.
```
jsvob, line 62: @osoukup A bunch of questions re. stream preselection.

First intro: In SFM2 all stream preselection is done in a single function. In OSIDB, the functions are mutually independent, but contributing to a shared dict of selected streams. In SFM2, fedramp handling cancels whatever ubi handling did (overwrites the streams_to_preselect list). In OSIDB, this would be non-trivial.

Questions:

1. Is it a problem? It seems to me that the ubi and fedramp policies just identify some specific streams and then that and all newer streams are selected anyway. Trying to not handle UBI if handling fedramp for a given affect(?) would probably be difficult but probably not too different? (I don't understand the consequences though.)
2. Can we depart in OSIDB from SFM2 in that UBI and Fedramp are independent? That would force the policies behind UBI handling and fedramp handling to be clear. If yes, it will allow us to keep the current structure of OSIDB's product_definition_handlers.
3. Why does SFM2 contain __If multiple fedramp streams are configured, only seed the streams_to_preselect with the first oldest one._? That seems superfluous and prone to a mistake by omission because the following code selects all newer streams anyway. I don't understand the implications/consequences though.

jsvob: Re. 3.: I think it might be because of code complexity and computational complexity. Best solution probably: Update compliance_priority.yml to explicity state the first stream must be the oldest one.

osidb/models.py
```diff
@@ -2092,6 +2092,10 @@ def is_community(self) -> bool:
     #      - see e.g. backends/jira/rest/request.py in SFM2
     # TODO unlike SFM2, SLA can't (?) evaluate the new streams part as long as this is a property
     #      on affect (since affect doesn't have streams). Should we file a new Jira to fix it?
+    # TODO: There's probably an omission in not creating a compliance-priority tracker label IN OSIDB.
```
jsvob, line 2095: @osoukup There's probably an omission in not creating a compliance-priority tracker label IN OSIDB. See backends/jira/rest/request.py in SFM2 and nonexistence of adding the label in OSIDB. Am I misunderstanding something or shall we file a Jira for OSIDB?

osidb/models.py
```diff
@@ -2092,6 +2092,10 @@ def is_community(self) -> bool:
...
+    #       See backends/jira/rest/request.py in SFM2. If that's the case, file a Jira for OSIDB.
+    # TODO: There's probably an omission in not creating a contract-priority tracker label IN SFM2.
```
jsvob, line 2097: @osoukup There's probably an omission in not creating a contract-priority tracker label **_IN SFM2_**. OSIDB creates that label, but can't find it in SFM2. If that's the case, shall we file a Jira for SFM2?
(Found this while looking at the SFM2 changes that are the blueprint for this PR.)

apps/trackers/product_definition_handlers/ubi_handler.py
```diff
@@ -16,6 +16,7 @@ def __init__(self) -> None:
 
     def has_ubi_packages(ps_module: PsModule, affect: Affect) -> bool:
         """check weheter a ps_module has ubi packages given a target PsUpdateStream name"""
+        # TODO: I can't find any code that sets special_handling_features. Is this missing and should we file a Jira?
         if "ubi_packages" not in ps_module.special_handling_features:
```
jsvob, line 20: @osoukup I can't find any code that sets `special_handling_features`. Is this missing and should we file a Jira?

osoukup: Looks like a bug :lady_beetle:

osidb/models.py
jsvob: Yes. Made a comment so that whoever takes it takes it into account.

osidb/models.py
jsvob: Yes, most probably. Added a comment to that Jira.

osidb/models.py
jsvob: Yes. Wrote a comment about that to the Jira.

osidb/models.py
jsvob: Facepalm. I looked at an outdated version of SFM2. Thanks!

apps/trackers/product_definition_handlers/ubi_handler.py
jsvob: Filed as OSIDB-2161

apps/trackers/product_definition_handlers/fedramp_handler.py
jsvob: :+1: 

I now have tests robust enough and enough understanding (thanks for your 1:1 explanations!) to try to replicate SFM2's algorithm. I'm just starting to try that.

jsvob: It seems I found a sane way to replicate SFM2's algorithm and give FedrampHandler the ability to effectively undo what UnackedHandler or UbiHandler does.

`offer = []` is not very feasible because it is pre-filled before any handler runs, but I also missed some understanding of the algorithm, which I tried to rectify for posterity by creating more educative tests.

docker-compose.yml
```diff
@@ -73,6 +73,7 @@ services:
         BZIMPORT_BZ_URL: ${BZIMPORT_BZ_URL}
         DJANGO_SETTINGS_MODULE: "config.settings_local"
         ET_URL: ${ET_URL}
+        HTTPS_PROXY: ${HTTPS_PROXY}
```
jsvob, line 76: NOTE: This is from a commit that's already in master. I didn't rebase yet.

jsvob:
@osoukup Please review. Commit history is a mess and it's not rebased. 

My aim here is for us to agree on the algorithm and the basics of the tests before a final review with whole team invited. 

- If the algorithm and tests are fine, I will just extend the tests [past the TODO on the last line of the tests](https://github.com/RedHatProductSecurity/osidb/pull/439/files#diff-3398886cd3481ed0eddb4840d8b52003f4e92c2ac7ea0c0682b8fe6b85706a41R464) and rebase and clean up the commit history. 
- If I'm misunderstanding something in the algorithm, the current tests should allow enough description and specification of the wrong&intended behaviors without incurring the days of work of polishing.

A part of the changes here are also in https://github.com/RedHatProductSecurity/osidb/pull/458. The two topics of PR458 (creating a more realistic stream structure in tests and querying the unacked stream properly) are also applied here in PR439.

Thank you!

apps/trackers/api.py
```diff
@@ -48,6 +48,7 @@ def post(self, request, *args, **kwargs):
             affects = selected_affects.filter(
                 affectedness=Affect.AffectAffectedness.AFFECTED,
                 resolution__in=[
+                    # TODO: Most probably, only DELEGATED should be here, see OSIDB-2143
```
osoukup, line 51: Good point. I thinking about it and even though there should be no new affects with FIX resolution (which is on validations) and it is probably not very common to file trackers against old affects, there is no issue in letting the users to do so.

apps/trackers/product_definition_handlers/fedramp_handler.py
```diff
@@ -0,0 +1,73 @@
...
+from osidb.models import Affect, CompliancePriority, Impact, PsModule, PsUpdateStream
+
+from .base import ProductDefinitionHandler
+
+
+class FedrampHandler(ProductDefinitionHandler):
+    """
+    Fedramp product definition handles
+
+    This handler should run anytime after UnackedHandler
```
osoukup, line 11: The order between this and `UBIHandler` does not matter?

apps/trackers/product_definition_handlers/fedramp_handler.py
```diff
@@ -0,0 +1,73 @@
...
+        by other handlers to avoid doing offers edits that would have to be reverted
+        by FedrampHandler.
+        """
+        return FedrampHandler.compliance_priority_fedramp_streams(affect, ps_module) and impact in FedrampHandler.FEDRAMP_OVERRIDES
+
+    @staticmethod
+    def compliance_priority_fedramp_streams(affect: Affect, ps_module: PsModule) -> list[PsUpdateStream]:
+        if affect.is_compliance_priority:
+            stream_names = CompliancePriority.objects.get(ps_module=ps_module.name).streams
+            streams = [PsUpdateStream.objects.get(name=name) for name in stream_names]
```
osoukup, line 28: Maybe something like
```
PsUpdateStreams.objects.filter(name__in=stream_names).all()
```
could save us some DB calls?

apps/trackers/product_definition_handlers/fedramp_handler.py
```diff
@@ -0,0 +1,73 @@
...
+                    offers[stream.name] = {
+                        "ps_update_stream": stream.name,
+                        "selected": True,
+                        "eus": bool(
+                            ps_module.eus_ps_update_streams.filter(name=stream.name)
+                        ),
+                        "aus": bool(
+                            ps_module.aus_ps_update_streams.filter(name=stream.name)
+                        ),
+                        "acked": not ps_module.unacked_ps_update_stream,
```
osoukup, line 70: I am not sure about this line or I do not understand it. It seems that it sets this stream as unacked whenever there exists an unacked stream for this module - which does not have to be the same stream.

apps/trackers/product_definition_handlers/ubi_handler.py
```diff
@@ -22,12 +23,17 @@ def has_ubi_packages(ps_module: PsModule, affect: Affect) -> bool:
         return bool(packages)
 
     def get_offer(self, affect: Affect, impact: Impact, ps_module: PsModule, offers):
+        if FedrampHandler.will_modify_offers(affect, impact, ps_module):
+            # FedrampHandler is next and it would have to undo everything done by UbiHandler.
+            return offers
```
osoukup, line 28: I would be for some more general approach (considering there may be more handlers in the future) but I am not strongly against this.

collectors/ps_constants/core.py
```diff
@@ -34,9 +34,13 @@ def sync_compliance_priority(source_dict):
     sync compliance priority data
     """
     CompliancePriority.objects.all().delete()
-    for ps_module, ps_component_list in source_dict.items():
-        for ps_component in ps_component_list:
-            CompliancePriority(ps_module=ps_module, ps_component=ps_component).save()
+    for ps_module, json_field in source_dict.items():
+        components = json_field.get("components", [])
+        # Streams must not be missing, crash if missing.
+        streams = json_field["streams"]
```
osoukup, line 40: The approach is basically correct but I would be for some more readable error message at least including the name of the module which has the empty list of streams.

osidb/migrations/0117_auto_20240213_1815.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2.23 on 2024-02-13 18:15
```
osoukup, line 1: Use some human readable migration file name.

requirements.txt
```diff
@@ -601,6 +601,10 @@ pyyaml==6.0 \
     # via
     #   -r requirements.in
     #   drf-spectacular
+q==2.7 \
```
osoukup, line 604: What is this used for?

osoukup (commented):
@jsvob I left a few comments but generally this goes in the right direction

requirements.txt
jsvob: For debugging. I'll remove it before submitting a proper PR for the final review.

apps/trackers/product_definition_handlers/fedramp_handler.py
jsvob: Good catch! The order absolutely **does** matter.

apps/trackers/api.py
jsvob: OK, I'll remove the TODO line then.

apps/trackers/product_definition_handlers/ubi_handler.py
jsvob: In SFM2, the function `unacked_target_allowed` does something similar to `.will_modify_offers` (given the structural differences in SFM2's vs. OSIDB's versions of the algorithm) and duplicates some conditions from the would-be handlers to predict the future and decide whether to select the unacked stream.

osoukup: Maybe just leave it as it is. I agree that there is a few handlers and it is still all pretty clear. Definitely better than in SFM2. If the number of handlers increases, we can think about something more general.

apps/trackers/product_definition_handlers/fedramp_handler.py
jsvob: If I understand SFM2 right, `acked` is `True` for everything but the unacked stream. The unacted stream has it always `False` no matter whether it's selected or not. So perhaps a bug in OSIDB?

osoukup: Exactly as you describe. And it seems that it is probably bug in OSIDB.

apps/trackers/product_definition_handlers/ubi_handler.py
jsvob: Got it. I made some changes (basically, ubi_handler gets .will_modify_offers) that I can absolutely revert to the state you reviewed if anyone so wishes. It makes it more straightforward in my opinion, but I'm aware that most people don't share my tastes. I'll present the changes with the final review request.

apps/trackers/product_definition_handlers/fedramp_handler.py
jsvob: Fixed.

apps/trackers/product_definition_handlers/fedramp_handler.py
jsvob: Fixed.

apps/trackers/product_definition_handlers/ubi_handler.py
jsvob: All handlers now try to predict the future (what the higher-priority == later running handlers will do) in a consistent way.

collectors/ps_constants/core.py
jsvob: Fixed.

osidb/migrations/0117_auto_20240213_1815.py
jsvob: Fixed.

apps/trackers/product_definition_handlers/unacked_handler.py
```diff
@@ -1,10 +1,20 @@
...
 
 from .base import ProductDefinitionHandler
+from .ubi_handler import UBIHandler
 
 
 class UnackedHandler(ProductDefinitionHandler):
+    """
+    Unacked stream handler
+
+    This handler should run before UnackedHandler and before FedrampHandler
```
osoukup, line 12: `UnackedHandler` should be actually `UBIHandler` ?

osoukup (approved):
LGTM

apps/trackers/product_definition_handlers/unacked_handler.py
jsvob: Fixed, thanks!

apps/trackers/tests/test_file_offer.py
```diff
@@ -339,3 +346,305 @@ def test_trackers_file_offer_unacked(
...
+        assert (
+            "regular-stream-1.3.z" in streams_dict
+            and streams_dict["regular-stream-1.3.z"]["selected"]
+        )
+
+        # The unacked stream should NOT be selected when fedramp_handler handles the situation
+        assert (
+            "regular-stream-1" in streams_dict
+            and not streams_dict["regular-stream-1"]["selected"]
+        )
```
jobselko, line 521: Similarly:
```
        streams = res["modules_components"][0]["streams"]
        assert [stream["ps_update_stream"] for stream in streams] == all_streams
        
        streams_dict = {stream["ps_update_stream"]: stream for stream in streams}
        for stream in all_streams:
            if stream in ["regular-stream-1", "regular-stream-1.0.z"]:
                # todo: describe why this is different
                assert stream in streams_dict and not streams_dict[stream]["selected"]
            else:
                assert stream in streams_dict and streams_dict[stream]["selected"]
```

jobselko (approved):
LGTM, good job!

I struggled to understand one test and added a few comments, but nothing significant (you can ignore them if you want). Also, changelog and migration commits should not be standalone but joined with the commit to which they belong.

apps/trackers/tests/test_file_offer.py
```diff
@@ -339,3 +346,305 @@ def test_trackers_file_offer_unacked(
...
+        PsUpdateStream(
+            name="regular-stream-1.3",
+            ps_module=ps_module_regular,
+        ).save()
+        PsUpdateStream(
+            name="regular-stream-1.3.z",
+            ps_module=ps_module_regular,
+            active_to_ps_module=ps_module_regular,
+            default_to_ps_module=ps_module_regular,
+        ).save()
```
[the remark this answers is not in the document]
jsvob: Thank you, but that doesn't work for me.

PR #441, opened 2024-02-07
Faster exclude fields

Elkasitu:
Closes OSIDB-1871

### DIFF
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -71,9 +71,7 @@ class DistinctFilterSet(FilterSet):
                     filter_.distinct = True
 
 
-class IncludeFieldsFilterSet(FilterSet):
-    include_fields = CharFilter(method="include_fields_filter")
-
+class SparseFieldsFilterSet(FilterSet):
     def _preprocess_fields(self, value):
         """
         Converts a comma-separated list of fields into an ORM-friendly format.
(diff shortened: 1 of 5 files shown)


osidb/filters.py
```diff
@@ -170,6 +172,31 @@ def include_fields_filter(self, queryset, name, value):
         )
 
 
+class ExcludeFieldsFilterSet(SparseFieldsFilterSet):
```
jobselko, line 175: This class seems to be unused (see `FlawFilter(DistinctFilterSet, IncludeFieldsFilterSet)` and similarly)

Elkasitu: Good catch, thanks!

jobselko: If everything that inherits from `IncludeFieldsFilterSet` will also inherit from  `ExcludeFieldsFilterSet`, could we join these two together? Or would it be a bad practice?

osoukup (approved):
LGTM and nice fixes on top!

jobselko (approved):

osidb/filters.py
Elkasitu: It wouldn't be bad practice, in fact we could probably invert the dependency to just have every class just use `SparseFieldsFilterSet`, but having them separate gives us the flexibility of having new views with either/or, exclude_fields is less useful than include_fields and less performant in general, so I can see a future in which new views only use `include_fields`

PR #442, opened 2024-02-08
Fix ContractPriority model using component instead of update stream.

jsvob:
Closes OSIDB-2099

### DIFF
--- a/collectors/ps_constants/core.py
+++ b/collectors/ps_constants/core.py
@@ -45,9 +45,12 @@ def sync_contract_priority(source_dict):
     sync contract priority data
     """
     ContractPriority.objects.all().delete()
-    for ps_module, ps_component_list in source_dict.items():
-        for ps_component in ps_component_list:
-            ContractPriority(ps_module=ps_module, ps_component=ps_component).save()
+    # ps_update_stream is unique across all ps_modules in the hierarchy of product
+    # definitions so there's no need to store ps_module, although it is in the
+    # source yaml for human readability.
+    for ps_update_streams in source_dict.values():
+        for ps_update_stream in ps_update_streams:
+            ContractPriority(ps_update_stream=ps_update_stream).save()
 
 
 @transaction.atomic
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2099,16 +2099,6 @@ class Affect(
         ).exists()
 
     @property
-    def is_contract_priority(self) -> bool:
-        """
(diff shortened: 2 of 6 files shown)


osidb/models.py
```diff
@@ -2665,6 +2655,18 @@ def is_triage(self):
...
+    def is_contract_priority(self) -> bool:
+        """
+        check and return whether this affect module and update stream combination is contract
+        priority which is defined in PS constants repo in contract_priority.yml
+        """
+        # A multi-flaw tracker is assumed to be related only to a single module.
+        ps_module = self.affects.first().ps_module
+        return ContractPriority.objects.filter(
+            ps_module=ps_module, ps_update_stream=self.ps_update_stream
+        ).exists()
```
osoukup, line 2668: I am thinking about the definition of the contract priority and I tent to suggest a simplification.

The thing is that when I give you an update stream I am also giving you the module implicitly without even mentioning it. There is a clear hierarchy in product definitions so that each update stream belongs to one specific module. Then however the module here is obviously an unnecessary extra information. The update stream is enough to determine the contract priority.

I know that in the PS constants there are both and I was suggesting to avoid the modules there as unnecessary from the tooling perspective but the opinion that it is better for readability from the human perspective won. Here in our code/DB however we could prefer machine over man :smile: WDYT?

jsvob: Fixed.

collectors/ps_constants/core.py
```diff
@@ -45,9 +45,11 @@ def sync_contract_priority(source_dict):
     sync contract priority data
     """
     ContractPriority.objects.all().delete()
-    for ps_module, ps_component_list in source_dict.items():
-        for ps_component in ps_component_list:
-            ContractPriority(ps_module=ps_module, ps_component=ps_component).save()
+    for ps_module, ps_update_streams in source_dict.items():
```
osoukup, line 48: The **ps_module** seems to be unused so it could be just **_**.

osoukup: Or maybe even better you could use **values** instead of **items**.

osoukup (approved):
LGTM

collectors/ps_constants/core.py
jsvob: Expanded the comment so as to maintain the level of clarity (IMO) re. the difference in the source yaml structure and what's being stored.

jsvob: Fixed.

PR #445, opened 2024-02-09
fix tracker convertor

osoukup:
Closes OSIDB-2118


Elkasitu (approved):
LGTM but should probably explain why `tracker_data` shouldn't be used and we probably should have unit tests for the convertors that cover these cases

PR #446, opened 2024-02-13
Introduce SNIPPET_CREATION_START_DATE to NVD and OSV

jobselko:
This PR introduces the `SNIPPET_CREATION_START_DATE` constant, which sets the date from which the collectors should create snippets (and their respective flaws if they do not exist). This date will be adjusted again in the future to reflect the time when we switch from Assembler to OSIDB.
It also updates and extends tests in the NVD collector, so the snippet and flaw creation are additionally tested.

Closes OSIDB-2085

### DIFF
--- a/collectors/constants.py
+++ b/collectors/constants.py
@@ -1,3 +1,4 @@
-from osidb.helpers import get_env
+from osidb.helpers import get_env, get_env_date
 
 SNIPPET_CREATION_ENABLED = get_env("SNIPPET_CREATION", default="False", is_bool=True)
+SNIPPET_CREATION_START_DATE = get_env_date("SNIPPET_CREATION_START")
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -3,10 +3,10 @@ from typing import Union
 import nvdlib
 from celery.utils.log import get_task_logger
 from django.conf import settings
-from django.utils import timezone
+from django.utils import dateparse, timezone
 from nvdlib.classes import CVE
 
-from collectors.constants import SNIPPET_CREATION_ENABLED
+from collectors.constants import SNIPPET_CREATION_ENABLED, SNIPPET_CREATION_START_DATE
 from collectors.framework.models import Collector
 from collectors.nvd.keywords import should_create_snippet
 from osidb.core import set_user_acls
@@ -159,6 +159,8 @@ class NVDQuerier:
                     "references": get_references(vulnerability),
                     "source": Snippet.Source.NVD,
                     "title": "placeholder only, see description",
+                    # required for ignoring historical data
+                    "published_in_nvd": f"{vulnerability.published}Z",
                 }
             )
 
@@ -172,6 +174,10 @@ class NVDCollector(Collector, NVDQuerier):
 
     # snippet creation is disabled by default for now
     snippet_creation_enabled = None
+
+    # when the start date is set to None, all snippets are collected
+    # when set to a datetime object, only snippets created after that date are collected
+    snippet_creation_start_date = SNIPPET_CREATION_START_DATE
 
     # the NIST NVD CVE project started in 1999
     # https://nvd.nist.gov/general/cve-process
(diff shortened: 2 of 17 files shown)


collectors/constants.py
```diff
@@ -1,3 +1,6 @@
+from django.utils.timezone import datetime, make_aware
+
 from osidb.helpers import get_env
 
 SNIPPET_CREATION_ENABLED = get_env("SNIPPET_CREATION", default="False", is_bool=True)
+SNIPPET_CREATION_START_DATE = make_aware(datetime(2024, 1, 1))
```
osoukup, line 6: This date should be taken from an environment variable so it is configurable.

collectors/nvd/collectors.py
```diff
@@ -251,6 +253,13 @@ def collect(self, cve: Union[str, None] = None) -> str:
             # create a new snippet with a flaw (if it does not exist) and link them
             snippet = None
             if self.snippet_creation_enabled:
+                # we do not care about historical data
+                published = timezone.make_aware(
+                    dateparse.parse_datetime(item["published_in_nvd"])
+                )
```
osoukup, line 259: Is there a way to do the timezone specification on storing instead of on reading?

collectors/osv/collectors.py
```diff
@@ -100,8 +100,11 @@ def collect(self, osv_id: Union[str, None] = None) -> str:
             # Surface an exception when collecting an individual OSV vulnerability
             osv_vuln = self.fetch_osv_vuln_by_id(osv_id)
             osv_id, cve_ids, content = self.extract_content(osv_vuln)
-            self.save_snippet(osv_id, cve_ids, content)
-            return f"OSV collection for {osv_id} was successful."
+            created, updated = self.save_snippet(osv_id, cve_ids, content)
+            if (created, updated) == (0, 0):
+                return f"OSV collection for {osv_id} was not successful due to historical data."
```
osoukup, line 105: nitpick: I would probably rephrase the sentence so it is clear that the fetched data is the historical data and not that the reason is in some unrelated historical data :smile:

collectors/nvd/collectors.py
```diff
@@ -251,6 +253,13 @@ def collect(self, cve: Union[str, None] = None) -> str:
...
+                if published < SNIPPET_CREATION_START_DATE:
```
osoukup, line 260: I would be for an option to leave that constant empty (no environment variable set) meaning that we collect everything. Otherwise if you really want everything you have the only option to set it really in history but that sounds not very convenient.

collectors/constants.py
osoukup: And some reasonable defaults should be specified in the OPS repo.

osoukup (changes requested):

collectors/nvd/collectors.py
jobselko: `Snippet.content` is `JSONField`, so having a `datetime` object here would probably be a problem.

collectors/osv/collectors.py
jobselko: I was thinking about putting it here or not. It is not currently in NVD; it contains only the following:
```
        if cve is not None:
            return f"NVD CVSS collection for {cve} completed"
```
So should I add a similar part to NVD too?

jobselko:
@osoukup Could you please check again? Thanks :)

osoukup (changes requested):

collectors/nvd/collectors.py
osoukup: asdf

osidb/helpers.py
```diff
@@ -80,6 +81,15 @@ def get_env(
     return value
 
 
+def get_env_date(key: str) -> Union[datetime, bool]:
+    """get a date environment variable of the ISO format (YYYY-MM-DD)"""
+    value = getenv(key)
+    if not value:
+        return False
```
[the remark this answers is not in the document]
jobselko: test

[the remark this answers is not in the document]
osoukup: asqdf

jobselko:
@osoukup Updated, please check again.

osidb/helpers.py
```diff
@@ -80,6 +81,15 @@ def get_env(
     return value
 
 
+def get_env_date(key: str) -> Union[datetime, None]:
+    """get a date environment variable of the ISO format (YYYY-MM-DD)"""
+    value = getenv(key)
+    if not value:
+        return None
+
+    return make_aware(datetime.fromisoformat(value))
```
osoukup, line 90: Could be simplified to something like
```
if value := getenv(key):
    return make_aware(datetime.fromisoformat(value))
```
but not strong preference.

osoukup (approved):
LGTM

PR #447, opened 2024-02-13
Update packages to the latest security version

jobselko:
This PR updates `django` and `cryptography` packages to the latest security version to fix findings from Dependabot.


osoukup (approved):
LGTM

PR #448, opened 2024-02-14
added tests for filtering by workflow fields

costaconrado:
This PR add test for filtering by workflow fields.

Closes OSIDB-2083.


osidb/tests/test_search.py
```diff
@@ -227,3 +229,13 @@ def test_search_flaws_on_particular_columns(self, auth_client, test_api_uri):
         assert response.status_code == 200
         body = response.json()
         assert body["count"] == 1
+
+        response = auth_client().get(f"{test_api_uri}/flaws?team_id=1234")
+        assert response.status_code == 200
+        body = response.json()
+        assert body["count"] == 1
+
+        response = auth_client().get(f"{test_api_uri}/flaws?owner=example@redhat.com")
```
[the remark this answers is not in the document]
costaconrado: @superbuggy that's right, DRF client will handle both `example%40redhat.com` or `example%40redhat.com`

https://github.com/encode/django-rest-framework/blob/df89f32b88022828ca5d1b5e04be36d54f506e6c/rest_framework/test.py#L203

osidb/tests/test_search.py
```diff
@@ -227,3 +229,13 @@ def test_search_flaws_on_particular_columns(self, auth_client, test_api_uri):
         assert response.status_code == 200
         body = response.json()
         assert body["count"] == 1
+
+        response = auth_client().get(f"{test_api_uri}/flaws?team_id=1234")
+        assert response.status_code == 200
+        body = response.json()
+        assert body["count"] == 1
```
Elkasitu, line 236: this doesn't really test the filter, if the filter is not valid this will also be `True`, you need more than one flaw and to check that the team_id corresponds

osidb/filters.py
```diff
@@ -305,6 +305,12 @@ class Meta:
             "major_incident_state": ["exact"],
             "requires_summary": ["exact"],
             "nist_cvss_validation": ["exact"],
+            # Workflow fields
+            "workflow_state": ["exact"],
+            "owner": ["exact"],
+            "group_key": ["exact"],
+            "task_key": ["exact"],
+            "team_id": ["exact"],
```
Elkasitu, line 313: some of these filters are untested

Elkasitu (commented):
Do we need all of these filters? owner and team_id probably make the most sense, workflow_state makes sense too, not sure what `{group,task}_key` are about nor whether they make sense to filter by

costaconrado:
@Elkasitu as suggested I have remove the filters that does not contain use case yet and added a second flaw to make sure the filters are working, please take a second look at this PR.

osidb/tests/test_search.py
Elkasitu: you need to check that `body["team_id"]` matches what you expect, if someone modifies the filter down the line and the filter returns the same number of records but the incorrect ones, you won't know it's broken, same for all the filters below

PR #451, opened 2024-02-21
Fix db users in db_backup.mk

jobselko:
This PR fixes `db-restore` in `db_backup.mk` to use the same user as `db-backup`.

The original problem was the following: `db-backup` created a db with the default `osidb_admin_user` user. Then, `db-restore` tried to remove that db before creating a new one, but with the incorrect `osidb_app_user` user. This was not a problem because `osidb_app_user` used to be a `SUPERUSER`. But in #394, we removed `SUPERUSER` from `osidb_app_user`, and `db-restore` thus stopped working.


osoukup (approved):
Makes sense. LGTM

PR #452, opened 2024-02-21
Update valid affectedness-resolution combinations

dmonzonis:
Some affectedness-resolution combinations have been removed to comply with the combinations defined in SFM2.

However, existing data is not to be modified as old combinations may still be pulled from a collector without running the validation, so the original enumerations have not been changed.

Closes OSIDB-2143

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1,6 +1,3 @@
-"""
-draft model for end to end testing
-"""
 import json
 import logging
 import re
@@ -1879,30 +1876,47 @@ class Affect(
                 "does not match any valid collection",
             )
 
-    def _validate_exceptional_affectedness_resolution(self):
+    def _validate_historical_affectedness_resolution(self):
         """
-        Alerts that an old flaw have empty affectedness.
-        (Only accepts WONTFIX and DEFER as resolution, validated by other validation)
+        Alerts that an old flaw has an affectedness/resolution combination that is now invalid,
+        but was valid in the past.
         """
-        valid_resolutions = [
-            Affect.AffectResolution.WONTFIX,
-            Affect.AffectResolution.DEFER,
-        ]
         if (
-            self.affectedness == Affect.AffectAffectedness.NOVALUE
-            and self.resolution in valid_resolutions
+            self.resolution
+            in AFFECTEDNESS_HISTORICAL_VALID_RESOLUTIONS[self.affectedness]
         ):
+            # Don't allow new records with historical combinations, or changing old records
+            # to an invalid combination
+            old_affect = (
+                Affect.objects.get(uuid=self.uuid) if not self._state.adding else None
+            )
+            if (
+                self._state.adding
+                or self.affectedness != old_affect.affectedness
+                or self.resolution != old_affect.resolution
+            ):
+                raise ValidationError(
+                    f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                )
+
+            # If modifying something else from a record with an invalid combination, e.g. the
+            # impact, throw an alert
             self.alert(
-                "flaw_exceptional_affect_status",
-                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} is in "
-                "a exceptional state having no affectedness.",
+                "flaw_historical_affect_status",
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has a "
+                "historical affectedness/resolution combination which is not valid anymore.",
             )
 
     def _validate_affect_status_resolution(self):
         """
-        Validates that affected product have a valid combination of affectedness and resolution
+        Validates that affected products have a valid combination (currently or historically)
+        of affectedness and resolution.
         """
-        if self.resolution not in AFFECTEDNESS_VALID_RESOLUTIONS[self.affectedness]:
+        if (
+            self.resolution not in AFFECTEDNESS_VALID_RESOLUTIONS[self.affectedness]
+            and self.resolution
+            not in AFFECTEDNESS_HISTORICAL_VALID_RESOLUTIONS[self.affectedness]
+        ):
             raise ValidationError(
                 f"{self.resolution} is not a valid resolution for {self.affectedness}."
             )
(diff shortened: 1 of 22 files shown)


osoukup:
@dmonzonis great to see your first OSIDB PR! I am going to do the review right now but first I two things to improve.

First is that you should set some reviewer(s) or otherwise nobody gets notification that there is something new to review. The best way is to set [osidb-devs](https://github.com/orgs/RedHatProductSecurity/teams/osidb-devs) group as a reviewer which will notify all of us.

Then I see that **Meta checks** are failing. This means that you forgot to update the **docs/CHANGELOG.md** with your change. Usually you should put something there - something like the title of the Jira or PR or whatever shortly describes the change and Jira issue ID in brackets. We usually do not add small rather technical changes which are no visible to user in any way there. In such case here in the PR you just add label **technical** and the check will be happy.

osoukup (changes requested):
The changes look great and definitely LGTM! I will however complicate it a bit :smile: and I am sorry I did not realize it when describing the requirements in the PR.

The thing is that with this change we would effectively block any changes to the old flaws having no more valid combinations. We should allow them to be still editable. We already had a similar situation in the past which is the reason of affectedness NOVALUE with options DEFER and WONTFIX also present in the AFFECTEDNESS_VALID_RESOLUTIONS. There is a more relaxed validation **_validate_exceptional_affectedness_resolution** to only warn but not raise an exception in the case of these two historical combinations. It however does not seem to be enough as it does not prevent creating new affects with these combinations.

I suggest the following extension of this PR. In `osidb/constants.py` split the dictionary into two where one would have eg. HISTORICAL in the name. Move there all the combinations being removed here plus also the ones with NOVALUE affectedness. Make **_validate_affect_status_resolution** to just ignore the historical combinations. Then adjust **_validate_exceptional_affectedness_resolution** (and probably rename exceptional to historical so it is more clear) to check the presence of the historical combinations and if so then on creation (self._state.adding) or update leading to that combination (load the model state from db and compare whether the values of affectedness and resolution change) raise an exception. Otherwise it means that the historical combination was already there and something else is being updated and we should permit it - append an alert instead of raising an exception. It would be good to create a test case for this.

Elkasitu:
> The thing is that with this change we would effectively block any changes to the old flaws having no more valid combinations.

From OSIDB only, no? Flaws modified in BZ and collected by bzimport should bypass the validation anyway, if edited through OSIDB then shouldn't we have whoever is editing it from OSIDB correct the affectedness/resolution to a valid state?

osoukup:
> > The thing is that with this change we would effectively block any changes to the old flaws having no more valid combinations.
> 
> From OSIDB only, no? Flaws modified in BZ and collected by bzimport should bypass the validation anyway, if edited through OSIDB then shouldn't we have whoever is editing it from OSIDB correct the affectedness/resolution to a valid state?

@Elkasitu you are correct that from Bugzilla it would not block anything. I rather thinking about a situation when there is an older flaw you need to modify a bit. As an example you need to correct an acknowledgment which in Bugzilla is in SRT notes JSON and so quite inconvenient for manual changes (or maybe there is no Bugzilla any more). Then, the old flaw has 50 affects you would have to update. Finally, the FIX was a valid value before so the affects are actually correct in the context of their age - changing them would be more like changing the history than fixing the invalid data. In this case I think we should not block the change and just warn that there are some outdated values. Does it make sense?

osidb/models.py
```diff
@@ -1,6 +1,7 @@
 """
 draft model for end to end testing
 """
+
```
osoukup, line 4: I wonder why you add the new line here. Is it based on some linter? Anyway I see I have not read the module description for very long probably :smile: and it is totally outdated - most probably from the very beginning of the project. When touching this piece you can remove the whole comment.

osidb/models.py
```diff
@@ -1855,30 +1856,41 @@ def _validate_sofware_collection(self):
...
+                need_alert = (
+                    self.affectedness != old_affect.affectedness
+                    or self.resolution != old_affect.resolution
+                )
+            if need_alert:
+                self.alert(
+                    "flaw_historical_affect_status",
+                    f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an "
+                    "historical affectedness/resolution combination which is not valid anymore.",
+                )
```
osoukup, line 1882: The logic should be slightly more restrictive. There should always be some kind of warning. In the case of `need_alert == False` we should only store the alert by `self.alert` but when `need_alert == True` we should really `raise ValidationError`. The idea is that we want to prevent any new values like this and `raise` will really interrupt the whole query processing while for the old data we still want to inform the user but softly.

osoukup (changes requested):
This is much better and almost there. I would just like to make the historical data validation more restrictive so it prevents creating new data with the historical combinations.

osidb/models.py
dmonzonis: Apparently the newline comes from my local version of `black` (24.1.1) which differs from the one in the pre-commit hooks (22.12.0). The new version [enforces this newline](https://black.readthedocs.io/en/latest/change_log.html#id10), which is why my editor must've added it, whereas the old version doesn't care.

I will remove this docstring anyway as it seems pretty useless.

osidb/models.py
dmonzonis: I modified it so that it raises an error when creating a new record instead, and throws an alert when modifying existing data

osidb/models.py
```diff
@@ -1855,30 +1852,43 @@ def _validate_sofware_collection(self):
...
+            if self._state.adding:
+                raise ValidationError(
+                    f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                )
+
+            old_affect = Affect.objects.get(uuid=self.uuid)
+            if (
+                self.affectedness != old_affect.affectedness
+                or self.resolution != old_affect.resolution
+            ):
```
osoukup, line 1875: This is one step closer but I probably did not explain myself well enough :smile:

Great that we now cannot create a new affect with the invalid values but still it is possible to assign invalid values to an existing one. We should prevent all new invalid combinations, either on creation or on update.

Only when the affectedness and resolution combination was already invalid and we do not touch it but eg. change the affect impact, we should create an alert but permit the change.

osoukup (changes requested):

osidb/models.py
```diff
@@ -1879,30 +1876,47 @@ def _validate_sofware_collection(self):
...
+                )
+
+            # If modifying something else from a record with an invalid combination, e.g. the
+            # impact, throw an alert
             self.alert(
-                "flaw_exceptional_affect_status",
-                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} is in "
-                "a exceptional state having no affectedness.",
+                "flaw_historical_affect_status",
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an "
```
osoukup, line 1906: nitpick: `a historical`

osoukup (approved):
Now it is perfect! Thank you for addressing all my feedback!

PR #454, opened 2024-02-26
Fix jira integration

costaconrado:
This PR fix integration with Jira when Bugzilla token is present.

This PR changes the order of mixins and moves the responsibility to save the flaw model to the mixin instead of the integration service, this assures that during flaw creation will run validations and not save multiple times during creation through REST API.

Closes OSIDB-2171


Elkasitu:
>This PR changes the order of mixins

but why? mixins should work regardless of order

jsvob (approved):
LGTM. IMO the order of the mixins matters and should matter. If we want them to be order-independent, it would require a major reengineering of the mixins (not saying it's not possible, just saying it might be so large as to be out of scope of our current work).

docs/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Ignore hosts on VCR recording (OSIDB-1678)
 - Included workflow fields in OpenAPI document for filtering (OSIDB-2083)
+### Fixed
```
osoukup, line 14: nitpick: So far we put an empty line between the sections.

osoukup (approved):
LGTM. I agree that it is not expected that the Mixin order would matter and also that in our case it is not easy to rewrite the existing code so the order does not matter. Specifically Bugzilla mixin cuts the natural `save` chain and goes to Bugzilla instead and then fetches the flaw from there. This will go away with the Bugzilla decommission.

docker-compose.test.yml
```diff
@@ -47,6 +47,7 @@ services:
         HTTPS_TASKMAN_PROXY: ${HTTPS_TASKMAN_PROXY}
         JIRA_AUTH_TOKEN: ${JIRA_AUTH_TOKEN:?Variable JIRA_AUTH_TOKEN must be set.}
         JIRA_TASKMAN_PROJECT_KEY: ${JIRA_TASKMAN_PROJECT_KEY}
+        JIRA_TASKMAN_AUTO_SYNC_FLAW: ${JIRA_TASKMAN_AUTO_SYNC_FLAW}
```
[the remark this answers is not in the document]
costaconrado: thanks, it is sorted now.

docs/CHANGELOG.md
costaconrado: thanks, there is a empty line now.

PR #456, opened 2024-02-27
Set migrated/duplicated delegated resolution to not affected

dmonzonis:
Before this commit, the MIGRATED and DUPLICATE resolutions which can come from BZ were not contemplated in osidb, and in these cases, they get assigned a state of AFFECTED by default. Since AFFECTED has the highest precedence, even if there is another tracker that says it's not a bug, the delegated resolution is AFFECTED which is wrong.

With this commit, trackers with MIGRATED/DUPLICATE resolutions are set a NOTAFFECTED state, since it is assumed that if there is such a tracker, then there is at least another one with actual information about the state, and since NOTAFFECTED has the lowest precedence, the final delegated resolution will be that of the other tracker with a valid state.

Closes OSIDB-1406

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2615,7 +2615,13 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
             if self.resolution in ("won't do", "won't fix", "wontfix", "obsolete"):
                 return Affect.AffectFix.WONTFIX
             # Added rejected to code inherited from SDEngine because samples such as MGDSTRM-4153
-            elif self.resolution in ("notabug", "not a bug", "rejected"):
+            elif self.resolution in (
+                "duplicate",
+                "migrated",
+                "notabug",
+                "not a bug",
+                "rejected",
+            ):
                 return Affect.AffectFix.NOTAFFECTED
             elif self.resolution in ("eol", "out of date"):
                 return Affect.AffectFix.OOSS
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -399,6 +399,29 @@ class TestFlaw:
         )
         assert undelegated_affect.delegated_resolution is None
 
+        # Migrated/duplicate trackers should be ignored
+        affect = AffectFactory(
+            affectedness=Affect.AffectAffectedness.AFFECTED,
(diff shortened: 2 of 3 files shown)


osidb/tests/test_flaw.py
```diff
@@ -427,6 +427,20 @@ def test_tracker_fix_state(self):
...
+            type=Tracker.TrackerType.BUGZILLA,
+        )
+        assert migrated_tracker.fix_state == Affect.AffectFix.NOTAFFECTED
+        duplicate_tracker = TrackerFactory(
+            status="closed",
+            resolution="duplicate",
+            affects=[affect],
+            type=Tracker.TrackerType.BUGZILLA,
+        )
+        assert duplicate_tracker.fix_state == Affect.AffectFix.NOTAFFECTED
```
skontar, line 443: Add some tests for `affect.delegated_resolution` based on some real cases, such as https://osidb.prodsec.redhat.com/osidb/api/v1/flaws?cve_id=CVE-2021-39537 or https://osidb.prodsec.redhat.com/osidb/api/v1/flaws?cve_id=CVE-2020-19188 , because that is where the problem mostly manifests. In those cases there are multiple trackers and some are migrated, some are won't do. We expect won't fix resolution for the affect, but we get affected. Thanks!

dmonzonis: I have added a test case in `test_delegated_affects` which tests this scenario

osoukup (approved):
LGTM

jsvob (approved):

osidb/models.py
```diff
@@ -2615,7 +2615,13 @@ def fix_state(self):
             if self.resolution in ("won't do", "won't fix", "wontfix", "obsolete"):
                 return Affect.AffectFix.WONTFIX
             # Added rejected to code inherited from SDEngine because samples such as MGDSTRM-4153
-            elif self.resolution in ("notabug", "not a bug", "rejected"):
+            elif self.resolution in (
+                "notabug",
+                "not a bug",
+                "rejected",
+                "migrated",
+                "duplicate",
```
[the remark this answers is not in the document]
dmonzonis: Sure, I changed it. I will keep this in mind in the future.

PR #459, opened 2024-02-29
Fix applying requirements.txt on celery hosts

jsvob:
Bug introduced in commit d1e4804b553ea0f32e248dfc1f4a8e73af1d1893

### DIFF
--- a/mk/localdev.mk
+++ b/mk/localdev.mk
@@ -30,7 +30,7 @@ start-local-psql: check-reg check-venv generate_local_pg_tls_cert
 .PHONY : stop-local
 stop-local:
 	@echo ">stopping local env without deleting containers"
-	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner phpldapadmin testldap flower redis celery_beat celery osidb-service osidb-data
+	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner testldap flower redis celery_beat celery osidb-service osidb-data
 
 
 
--- a/mk/localdev.mk
+++ b/mk/localdev.mk
@@ -30,7 +30,7 @@ start-local-psql: check-reg check-venv generate_local_pg_tls_cert
 .PHONY : stop-local
 stop-local:
 	@echo ">stopping local env without deleting containers"
-	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner phpldapadmin testldap flower redis celery_beat celery_host1 celery_host2 osidb-service osidb-data
+	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner phpldapadmin testldap flower redis celery_beat celery osidb-service osidb-data
 
 
 
--- a/mk/setup.mk
+++ b/mk/setup.mk
@@ -85,10 +85,10 @@ upgrade-dep: check-venv-active
 apply-requirements-txt: check-reg check-venv sync-deps compose-up
 	@echo ">appyling requirements.txt on osidb-service"
 	$(podman) exec -it osidb-service pip3 install -r /opt/app-root/src/requirements.txt
-	@echo ">appyling requirements.txt on celery_host1"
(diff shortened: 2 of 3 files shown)


mk/localdev.mk
```diff
@@ -30,7 +30,7 @@ start-local-psql: check-reg check-venv generate_local_pg_tls_cert
 .PHONY : stop-local
 stop-local:
 	@echo ">stopping local env without deleting containers"
-	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner phpldapadmin testldap flower redis celery_beat celery_host1 celery_host2 osidb-service osidb-data
+	$(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop || $(podmancompose) -f docker-compose.yml -f docker-compose.test.yml stop testrunner phpldapadmin testldap flower redis celery_beat osidb_celery_1 osidb_celery_2 osidb-service osidb-data
```
osoukup, line 33: `osidb_celery_X` (with X being a number) is the container or host name the automatic replicas will get? It would then be clear that the old names in `setup.mk` do not correspond any more. I am just making sure I understand the context of this change correctly. If so, by `grep` I also get the old name at
https://github.com/RedHatProductSecurity/osidb/blob/master/docker-compose.yml#L51
which has of course no influence on anything be might be later confusing if anyone reads the comment so maybe could you update it accordingly?

jobselko:
@jsvob Which bug was introduced in [d1e4804](https://github.com/RedHatProductSecurity/osidb/commit/d1e4804b553ea0f32e248dfc1f4a8e73af1d1893)? That commit changes `docker-compose.yml`, but you are changing `localdev.mk` and `setup.mk`, so it is not clear to me. Thanks for the clarification.

mk/setup.mk
```diff
@@ -85,10 +85,10 @@ upgrade-dep: check-venv-active
...
 	@echo ">appyling requirements.txt on osidb-service"
 	$(podman) exec -it osidb-service pip3 install -r /opt/app-root/src/requirements.txt
-	@echo ">appyling requirements.txt on celery_host1"
-	$(podman) exec -it celery_host1 pip3 install -r /opt/app-root/src/requirements.txt
-	@echo ">appyling requirements.txt on celery_host2"
-	$(podman) exec -it celery_host2 pip3 install -r /opt/app-root/src/requirements.txt
+	@echo ">appyling requirements.txt on osidb_celery_1"
+	$(podman) exec -it osidb_celery_1 pip3 install -r /opt/app-root/src/requirements.txt
+	@echo ">appyling requirements.txt on osidb_celery_2" # if you have more celery replicas, you're on your own
+	$(podman) exec -it osidb_celery_2 pip3 install -r /opt/app-root/src/requirements.txt || true
```
jobselko, line 91: Why is there `|| true`, but line 89 does not have it?

mk/localdev.mk
jsvob: Turns out I was lazy so I fixed it now.

When the celery containers are running, they are indeed named `osidb_celery_X` as you say.

When podman-compose is invoked, it requires the `celery` name as written in `docker-compose.yml`, not `osidb_celery_X`. That then led to fixing https://github.com/RedHatProductSecurity/osidb/blob/master/docker-compose.yml#L51 properly.

jsvob:
> @jsvob Which bug was introduced in [d1e4804](https://github.com/RedHatProductSecurity/osidb/commit/d1e4804b553ea0f32e248dfc1f4a8e73af1d1893)? That commit changes `docker-compose.yml`, but you are changing `localdev.mk` and `setup.mk`, so it is not clear to me. Thanks for the clarification.

`make apply-requirements-txt` doesn't work since d1e4804b553ea0f32e248dfc1f4a8e73af1d1893

mk/setup.mk
jsvob: So that the rest of the make target still gets executed if only 1 replica is running. Added a comment.

mk/localdev.mk
osoukup: Thank you for the explanation!

osoukup (approved):
LGTM.

jobselko:
> > @jsvob Which bug was introduced in [d1e4804](https://github.com/RedHatProductSecurity/osidb/commit/d1e4804b553ea0f32e248dfc1f4a8e73af1d1893)? That commit changes `docker-compose.yml`, but you are changing `localdev.mk` and `setup.mk`, so it is not clear to me. Thanks for the clarification.
> 
> `make apply-requirements-txt` doesn't work since [d1e4804](https://github.com/RedHatProductSecurity/osidb/commit/d1e4804b553ea0f32e248dfc1f4a8e73af1d1893)

So I will answer myself - it seems the issue is that the name got changed :)

PR #460, opened 2024-03-04
increase log level for bugzilla in stage

costaconrado:
This PR increases the log level for bugzilla library in stage for debug purpose.


osoukup (approved):
I would even turn it on all environments but as the production is not yet writable this is enough. LGTM

PR #462, opened 2024-03-05
Add Snippet to Flaw conversion to OSV collector

jobselko:
This PR introduces `Snippet` to `Flaw` conversion in the OSV collector and snippet refactoring in tests.

Closes OSIDB-2198


collectors/osv/collectors.py
```diff
@@ -166,11 +168,14 @@ def save_snippet(
             return 0, 0
 
         if not cve_ids:
-            _, created = Snippet.objects.update_or_create(
+            # This keeps the structure consistent and ease snippets filtering without cve_id
+            content["cve_id"] = None
```
jobselko, line 172: This is open to discussion.

collectors/osv/collectors.py
```diff
@@ -217,12 +223,25 @@ def get_refs(data: dict) -> list:
...
+                    validate(ref["url"])
+                    refs.append(
+                        {
+                            "type": FlawReference.FlawReferenceType.EXTERNAL,
+                            "url": ref["url"],
+                        }
+                    )
+                except ValidationError:
+                    # Ignore the URL if it is invalid (e.g. due to a typo)
+                    pass
```
jobselko, line 244: And this one too. We have to handle invalid links somehow. One option is to ignore them, and the second is to try to fix them (as I did). Some form of handling should also be added to NVD.

jsvob: It's debatable based on what the underlying business process needs. I'd argue that keeping as much original information from the original report (snippet) as possible will help the analyst more than changing or hiding information. Even an invalid URL might be useful for the analyst when used with [Google, Kagi, DuckDuckGo, Bing], and arguably more so than trying to be clever and changing/hiding information.

jsvob (approved):

collectors/osv/collectors.py
jobselko: The issue is that `url` in `FlawReference` is `URLField`, so a validation must pass. We are not able to save an invalid link (we could disable validations, but I would be against this option).

collectors/osv/collectors.py
```diff
@@ -166,11 +168,14 @@ def save_snippet(
...
         if not cve_ids:
-            _, created = Snippet.objects.update_or_create(
+            # This keeps the structure consistent and ease snippets filtering without cve_id
+            content["cve_id"] = None
+            snippet, created = Snippet.objects.update_or_create(
                 source=Snippet.Source.OSV,
                 external_id=osv_id,
                 defaults={"content": content},
             )
+            snippet.convert_snippet_to_flaw()
```
osoukup, line 178: This seems to be creating new flaws also on snippet update.

collectors/osv/collectors.py
osoukup: I agree too.

collectors/osv/collectors.py
jobselko: Thanks, I will check it.

collectors/osv/collectors.py
```diff
@@ -166,12 +168,15 @@ def save_snippet(
...
-            _, created = Snippet.objects.update_or_create(
+            # This keeps the structure consistent and ease snippets filtering without cve_id
+            content["cve_id"] = None
+            snippet, created = Snippet.objects.update_or_create(
                 source=Snippet.Source.OSV,
                 external_id=osv_id,
                 defaults={"content": content},
             )
             if created:
+                snippet.convert_snippet_to_flaw()
```
jobselko, line 179: There is probably an issue when a flaw without `cve_id` is already created in BZ, but we lose the OSIDB data. In this case, even though flaws are downloaded before the OSV collector runs, we do not have a way to say that this newly created snippet should be connected to an already existing flaw in BZ (as there is no `cve_id`). Thus, we will create a new (duplicate) flaw in BZ. We can discuss this tomorrow.

apps/bbsync/srtnotes.py
jobselko, line 1: FYI, I will add more thorough tests as a part of upcoming Jira task.

apps/bbsync/tests/test_query.py
```diff
@@ -409,6 +411,19 @@ def test_generate_affects(self):
...
+        test generating of SRT external_ids attribute array
+        """
+        flaw = FlawFactory()
+        SnippetFactory(source=Snippet.Source.NVD, flaw=flaw)
+
+        bqb = FlawBugzillaQueryBuilder(flaw)
+        cf_srtnotes = bqb.query.get("cf_srtnotes")
+        cf_srtnotes_json = json.loads(cf_srtnotes)
+        external_ids = cf_srtnotes_json.get("external_ids", [])
+        assert external_ids == ["CVE-2023-0001"]
```
osoukup, line 425: This test is dependent on the current factory implementation so it can break if anyone decides to improve the factory. Could it be adjusted in a way that the CVE ID is provided as a parameter to the `SnippetFactory` so it is ensured that this exact CVE ID is there despite of the future changes to the factory? Or the other way around you can check that `external_ids == [snippet.external_id]`.

osoukup (approved):
LGTM

apps/bbsync/tests/test_query.py
jobselko: Thanks for noticing, I overlooked this one.

PR #465, opened 2024-03-06
Fix Bugzilla summary for first flaw creation

costaconrado:
This PR changes Bugzilla query builder to consider the first flaw creation into summary.

Closes OSIDB-2190.


osoukup (approved):
LGTM. I would just point out that we are actually fixing a correct code :smile: which just assumed that every flaw must have Bugzilla ID or it is currently being created which may be only a single flaw thanks to the API request atomicity nature. It of course did not could on an existence of 142K flaws without Bugzilla IDs :smile:

PR #466, opened 2024-03-07
Fix several issues around collectors

jobselko:
This PR fixes the following:
- BZ code ignored the `SOURCE` reference and failed to write a new flaw
- NVD ans OSV collectors created references that may be invalid and could not be stored in `FlawReference`
- NVD collector had an inappropriate `title`
- OSV collector did not always create snippets with `cve_id`
- `.env` section in `DEVELOP.md` did not contain new "snippet" variables

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -268,7 +268,7 @@
         "filename": "docs/developer/DEVELOP.md",
         "hashed_secret": "7c6a61c68ef8b9b6b061b28c348bc1ed7921cb53",
         "is_verified": false,
-        "line_number": 80,
+        "line_number": 84,
         "is_secret": false
       }
     ],
@@ -431,5 +431,5 @@
       }
     ]
   },
-  "generated_at": "2024-02-26T20:16:29Z"
+  "generated_at": "2024-03-13T16:13:36Z"
 }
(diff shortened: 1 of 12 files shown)


osoukup (commented):
Is this tested? I mean do you have a reference to a staging Bugzilla flaw draft(s) which was created by an independent run of the NVD collector (locally of course)? Calling API or playing with shell does not count here - we do not want to repeat the stage issues. It should be also verified (first locally) that we only create flaws we want and we do not create duplicates etc. so do not pollute the stage once this is merged.

jobselko:
@osoukup Yes, I tested it. You can check the BZ stage; new flaws look like `CVE-2023-7244 placeholder only, see description`. I just updated their title to sth more clean. Please check.

osoukup:
> @osoukup Yes, I tested it. You can check the BZ stage; new flaws look like `CVE-2023-7244 placeholder only, see description`. I just updated `title` in code to sth more clean. Please check.

There should be then some integration test to ensure that we do not break the functionality in the future.

jobselko:
@osoukup I updated the whole PR to fix the known issues and nothing else. The rest will be handled in the future.

osoukup (approved):
I would like the URL fixing code to be deduplicated but otherwise LGTM

PR #467, opened 2024-03-08
Add tracker label compliance-priority based on ps-constants

jsvob:
Add tracker label compliance-priority based on ps-constants
Closes OSIDB-2062

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -156,6 +156,11 @@ class TrackerJiraQueryBuilder(TrackerQueryBuilder):
         if self.tracker.is_contract_priority:
             self._query["fields"]["labels"].append("contract-priority")
 
+        if self.tracker.is_compliance_priority:
+            # excluding the low priority ones
+            if Impact.LOW != self.tracker.aggregated_impact:
+                self._query["fields"]["labels"].append("compliance-priority")
+
     def generate_sla(self):
         """
         generate query for Jira SLA timestamps
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -157,9 +157,7 @@ class TrackerJiraQueryBuilder(TrackerQueryBuilder):
             self._query["fields"]["labels"].append("contract-priority")
 
         if self.tracker.is_compliance_priority:
-            # excluding the low priority ones
-            if Impact.LOW != self.tracker.aggregated_impact:
-                self._query["fields"]["labels"].append("compliance-priority")
+            self._query["fields"]["labels"].append("compliance-priority")
 
     def generate_sla(self):
         """
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2136,10 +2136,14 @@ class Affect(
         Check and return whether this affect module and component combination is compliance
         priority which is defined in PS constants repo in compliance_priority.yml.
 
+        Compliance priority is always False for LOW impact.
+
         Note that stream is not evaluated by this property. Therefore, this property
         is not useful for evaluating SLA, but useful for computing which streams to
         file trackers for. Use Tracker.is_compliance_priority for SLA calculations.
         """
+        if self.aggregated_impact == Impact.LOW:
+            return False
         compliance_priority = CompliancePriority.objects.filter(
             ps_module=self.ps_module
         )
@@ -2738,9 +2742,14 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
         """
         Check and return whether this affect module, component and update stream combination is
         compliance priority which is defined in PS constants repo in compliance_priority.yml.
+
+        Compliance priority is always False for LOW impact.
         """
         affect = self.affects.first()
         if not affect:
+            return False
+
+        if self.aggregated_impact == Impact.LOW:
             return False
 
         compliance_priority = CompliancePriority.objects.filter(
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2078,10 +2078,13 @@ class Affect(
     @property
     def aggregated_impact(self):
         """
-        this property simply gives the maximum impact
-        aggregated from the affect and the related flaw
+        this property equals Flaw's impact if the Affect's impact is blank, or
+        equals the Affect's impact if the Affect's impact is not blank
         """
-        return max(Impact(self.impact), Impact(self.flaw.impact))
+        if not self.impact:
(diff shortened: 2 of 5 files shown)


apps/trackers/jira/query.py
```diff
@@ -156,6 +156,18 @@ def generate_labels(self):
...
 
+        if self.tracker.is_compliance_priority:
+            # excluding the low priority ones
+            if Impact.LOW != max(
+                [
+                    Impact(i)
+                    for i in self.tracker.affects.all().values_list(
+                        "flaw__impact", flat=True
+                    )
+                ]
```
osoukup, line 167: There is a `Tracker` property `aggregated_impact` which does this.

osidb/models.py
```diff
@@ -2723,6 +2726,30 @@ def is_triage(self):
...
+    def is_compliance_priority(self) -> bool:
+        """
+        Check and return whether this affect module, component and update stream combination is
+        compliance priority which is defined in PS constance repo in compliance_priority.yml.
+        """
+        affect = self.affects.first()
+        if not affect:
+            return False
+
+        ps_module = PsModule.objects.filter(name=affect.ps_module).first()
```
osoukup, line 2739: I am not sure about the necessity of this DB query. The PS module seems to be only used to get its name and the name is already present as part of the affect data. I actually think that this should cause troubles instead if we call at on some old/bogus affect with non-existing PS module as next `ps_module.name` would throw an `AttributeError`.

osoukup (changes requested):
I have to comments to address/answer but overall it looks well!

apps/trackers/jira/query.py
jsvob: It can't be queried like this and doesn't (IMO) correspond to SFM2's algorithm.

osoukup: Oh I see. You mean that the affect impact override does not count here? That is something I am not really sure about. I will ask Przemyslaw.

osidb/models.py
jsvob: You are right. It's a useless and slightly wrong query. Fixed.

apps/trackers/jira/query.py
jsvob: Hmmm, I apologize. After going deeper in SFM2 (the condition that decides whether to add `compliance-priority` label), it seems to do effectively what `aggregated_impact` would do. I'll fix it to use `aggregated_impact`. The query will have to be different, because the database doesn't know what `aggregated_impact` is.

jsvob: @osoukup Fixed, sorry for the noise.

osoukup: We discuss this with Przemyslaw and my first impression is that it is not really clear :smile: I will update you here when I have the final statement.

osoukup: @jsvob so this is actually not so trivial. I mean the answer is trivial, you should us the impact override. I used https://access.redhat.com/security/cve/cve-2021-45940 as an example (though the components are probably not under any special handling but just to denote the impact differences between the flaw and affects) and Przemyslav stated that **in this case compliance priority rules should be applied because the pair Product&Component matches the compliance priority list and the severity is Moderate** (again assume a hypothetical situation that it is really in the PS constants). The tricky part is that this situation should actually never happen as the flaw impact should always be greater of equal than any related affect impact. There even was an automation on the SFM2 central to fix this - but the central is obviously broken. So we should consider the case that flaw impact is LOW but affect impact is MODERATE and then the override (MODERATE) is what we should look for here. There can be even an opposite situation when a flaw is MODERATE and an affect is LOW. Then again the override (LOW) is the important here. I see that `Affect.aggregated_impact` unfortunately implements this wrong and takes the maximum while it should be take the flaw one when affect one is empty or take the affect one instead. Would you fix it as part of this or should I create a separate Jira?

osoukup: I see I had the tab opened when discussing with Przemyslaw and your comments did not show to me :smile: even if SFM2 implements it differently, we have no clear clarification how it is supposed to work (and makes sense to me). The thing is that these cases are super rare so if there was a bug in SFM2 no one might notice.

apps/trackers/jira/query.py
```diff
@@ -156,6 +156,11 @@ def generate_labels(self):
         if self.tracker.is_contract_priority:
             self._query["fields"]["labels"].append("contract-priority")
 
+        if self.tracker.is_compliance_priority:
+            # excluding the low priority ones
+            if Impact.LOW != self.tracker.aggregated_impact:
```
osoukup, line 161: I am thinking about one more thing. Should the impact check be really here or rather in the `is_compliance_priority` method? This way we split the business logic instead of keeping it all at one place. Or do you see a reason why the split is better/necessary?

osoukup (changes requested):
I have one last comment. If it is irrelevant or I misunderstand something, please explain it to me and I will approve :smile_cat:

apps/trackers/jira/query.py
jsvob: Looking at SFM2, is_compliance_priority is used for SLA calculations and for compliance-priority label. The latter has this LOW impact check. SLA calculations for compliance_priority are done differently for each impact. So moving it to the `is_compliance_priority` method would break any sane attempt at using it for SLA calculations, IMO.

As for _why_, I have no idea why in SFM2 the SLAs are that way nor why the LOW check is there when adding the compliance-priority label.

osoukup: I agree it is confusing there as the SLA is different for each impact by its nature. But SFM2 is actually a good example. In the SLA definition table at
https://gitlab.cee.redhat.com/prodsec-dev/sfm2/-/blob/master/sfm2/sla.py?ref_type=heads#L55
you can see that unlike the other impact values the LOW has the same SLA for both compliance and non-compliance priority. That basically means that the impact is part of the compliance priority check and LOW means that there is no special compliance priority handling. I would therefore rather put the impact check directly to the compliance priority check.

osoukup: I think that splitting the logic in SFM2 is actually a bad idea. I am for merging the impact and the product definitions checks under `is_compliance_priority`

jsvob: Fixed. Could you please do a final review on the last commit?

Thank you for the explanation. I have seen this code but didn't _actually see_ what you pointed out. Of course your suggestion makes perfect sense.

osoukup (approved):
LGTM

PR #468, opened 2024-03-11
Introduce CVSS v4 in OSIDB

jobselko:
This PR introduces CVSS v4 in OSIDB (tested in BZ stage). The NVD collector was not updated because NVD does not support v4 yet.

Closes OSIDB-528


osoukup (changes requested):
I see that the Bugzilla integration is being extended/updated but I do not see any integration test. We are quite constantly having bugs in the integration so we should be more mindful about testing it.

jobselko:
@osoukup I tested the code against the BZ stage, even though this change only extends the `CVSSVersion` enum by a new string and nothing more. I was able to test it only manually via curl. For some reason, none of the integration tests worked for me. I think it would be better to discuss it tomorrow.

osoukup:
> @osoukup I tested the code against the BZ stage, even though this change only extends the `CVSSVersion` enum by a new string and nothing more. I was able to test it only manually via curl. For some reason, none of the integration tests worked for me. I think it would be better to discuss it tomorrow.

Stating that only the enum is being changed is an underrating of this PR. I see both BBSync and bzimport are being modified too. Based on a number of recent negative experience I would like us to be more mindful about the integration testing.

apps/bbsync/srtnotes.py
```diff
@@ -96,6 +96,7 @@ def generate_affects(self):
         """
         generate array of affects to SRT notes
         """
+
```
osoukup, line 99: nitpick: should be in the previous commit (but you can leave it as it is)

osoukup (approved):
Good job. LGTM

And based on our previous troubles I am introducing this change
https://github.com/RedHatProductSecurity/osidb/pull/474

PR #469, opened 2024-03-12
allow list of states in flaw filter

costaconrado:
This PR changes flaw API to allow filtering by a list of `workflow_state`.

Closes OSIDB-2208.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -3442,14 +3442,19 @@ paths:
       - in: query
         name: workflow_state
         schema:
-          type: string
-          enum:
-          - DONE
-          - NEW
-          - PRE_SECONDARY_ASSESSMENT
-          - REJECTED
-          - SECONDARY_ASSESSMENT
-          - TRIAGE
+          type: array
+          items:
+            type: string
+            enum:
+            - DONE
+            - NEW
+            - PRE_SECONDARY_ASSESSMENT
+            - REJECTED
+            - SECONDARY_ASSESSMENT
+            - TRIAGE
+        description: Multiple values may be separated by commas.
+        explode: false
+        style: form
       tags:
       - osidb
       security:
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Included workflow fields in OpenAPI document for filtering (OSIDB-2083)
 - Set migrated/duplicated delegated resolution to not affected (OSIDB-1406)
 - Update valid affectedness-resolution combinations (OSIDB-2143)
+- Change Flaw API filter to allow a list of workflow_state (OSIDB-2208)
 
 ### Fixed
 - Fix Jira sync when bugzilla token is present (OSIDB-2171)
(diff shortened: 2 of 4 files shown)


openapi.yml
```diff
@@ -3442,14 +3442,12 @@ paths:
...
-          - DONE
-          - NEW
-          - PRE_SECONDARY_ASSESSMENT
-          - REJECTED
-          - SECONDARY_ASSESSMENT
-          - TRIAGE
+          type: array
+          items:
+            type: string
+        description: Multiple values may be separated by commas.
```
jsvob, line 3448: Nitpick: The enum in openapi.yml seems kind of useful for knowing which values are valid. We are losing that usefulness here. I see that Flaw's `cve_id` has the same drawback in openapi.yml, so this PR already has a precedent that it is OK. I'd just like to discuss (even after merging this) whether we want to somehow improve openapi.yml for fields that use `CharInFilter`, and whether anyone would even care. @RedHatProductSecurity/osidb-devs thoughts?

jsvob (approved):
LGTM, I'm for merging this and discussing the nitpick asynchronously.

openapi.yml
```diff
@@ -3442,14 +3442,12 @@ paths:
...
         name: workflow_state
         schema:
-          type: string
-          enum:
-          - DONE
-          - NEW
-          - PRE_SECONDARY_ASSESSMENT
-          - REJECTED
-          - SECONDARY_ASSESSMENT
-          - TRIAGE
```
JakubFrejlach, line 3452: I think this is not entirely correct as we lost information about possible values of the enum, this could be easily solved if you use a different filter for the workflow_state:
```python
from django_filters.rest_framework import ChoiceFilter, BaseInFilter

class ChoiceInFilter(BaseInFilter, ChoiceFilter):
    """
    Filter for choice csv
    """

    pass
    
workflow_state = ChoiceInFilter(field_name="workflow_state")
```

JakubFrejlach (commented):
Sorry for sticking my nose into this but I've noticed during the standup that this is about filters so I might give a quick opinion on this. :smile:

openapi.yml
JakubFrejlach: Actually this might not work correctly, there also exist `MultipleChoiceFilter`, not sure how that works, just pointing out that enum values should be exposed in the schema

openapi.yml
JakubFrejlach: For `cve_id` I think it is kinda different as it is not an enum and the field is pretty self explanatory, but loosing and enum values might be problematic for anyone who would like to work with the API. I posted some possible solutions below, not tested them thoroughly or sure about them, just something to think about.

osoukup (changes requested):
See my comment above. I forgot to request the change.

costaconrado:
Makes sense, submitted the change suggested by @JakubFrejlach, thanks you all

JakubFrejlach:
@costaconrado Actually I am not sure, but it might be necessary to explicitly list the choices via `choices` argument, I tried that quickly on my local instance and without the choices I was getting "Invalid option selected...." for valid options.

costaconrado:
@JakubFrejlach I believe I can use something like:
`workflow_state = ChoiceInFilter(field_name="workflow_state", choices=WorkflowModel.WorkflowState.choices)`

let me test it

osoukup (approved):
Much better :smiley_cat: thank you for addressing the feedback!

PR #470, opened 2024-03-12
Make workflows API RESTful

dmonzonis:
The workflows API does not follow REST priciples. Right now, the way to access a workflow is
```${SERVICE_URL}/workflows/v1/api/<workflow_uuid>```
but for it to be RESTful, the name of the resource being requested should go before the ID, so it should look like
```${SERVICE_URL}/workflows/v1/api/workflows/<workflow_uuid>```

This was the case before the endpoint was renamed from *osim* to *workflows*, and the *osim* endpoint deprecated.
With this commit, the URL will behave as before, making it comply with REST principles again. The endpoints that are nto RESTful have been deprecated.

Closes OSIDB-1716

### CODE REVIEW

**Comment by costaconrado** [comment]
Note for reviewers, IMO `api sunset date can't be parsed for deprecated API` error in openapi check can be safely ignored for this PR. The lib `oasdiff` suggests us to first add `x-sunset` with a date and in a future release add the `deprecated` flag, but I believe only having `deprecated: true` is enough.

**Review by costaconrado — APPROVED** [review]
LGTM. Nice job.

File: apps/workflows/urls.py [hunk 1]
```diff
@@ -16,12 +17,14 @@
...
+    path(f"api/{WORKFLOWS_API_VERSION}/workflows/<str:pk>", classification.as_view()),
+    path(f"api/{WORKFLOWS_API_VERSION}/workflows/<str:pk>/adjust", adjust.as_view()),
     path(f"api/{WORKFLOWS_API_VERSION}/graph/workflows", graph_workflows.as_view()),
     path(
         f"api/{WORKFLOWS_API_VERSION}/graph/workflows/<str:pk>",
         graph_classification.as_view(),
     ),
 ]
+
+urlpatterns += deprecated_urlpatterns
```
osoukup, line 30: I wonder whether there is any value in preserving the deprecated paths. The whole workflows (former osim) API is rather an experimental thing and whoever would integrate with it already is probably doomed anyway. @costaconrado WDYT?

osoukup: I know that anything red in openapi.yml is normally allowed in the major versions only but we already agreed before (with **taskman**) that we will exclude the experimental stuff from the guarantees.

osoukup (approved):
LGTM. I would be for just removing the old API as it was just experimental but I do not insist on it and would first like to ask you @dmonzonis and @costaconrado for your views.

apps/workflows/urls.py
costaconrado: I agree. The workflows we had before the changes did not reflect any real process we had in place, as you mentions it was experimental and I don't think anyone had integrated+used it before (probably only griffon through automation which would not be a problem), I believe it is safe to just remove this without deprecation.

dmonzonis: I preserved them mainly because it was a breaking change. But if they are safe to remove, I'll happily remove them.

osoukup: So I say these endpoints I experimental and safe to be removed so the responsibility is mine :smile: and you can proceed with the removal. It will simplify the code and I really see no drawbacks.

apps/workflows/tests/test_endpoints.py
```diff
@@ -208,7 +210,7 @@ def test_workflows_uuid_adjust_no_auth(self, client, test_api_uri):
         test authenticated workflow classification adjusting API endpoint without authenticating
         """
         flaw = FlawFactory()
-        response = client.post(f"{test_api_uri}/{flaw.uuid}/adjust")
+        response = client.post(f"{test_api_uri}/workflows/{flaw.uuid}/adjust")
```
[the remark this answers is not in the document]
dmonzonis: I'm just being consistent with what's done in the rest of the codebase. In other tests `test_api_uri` is also defined up to the version and then `/flaws`, `/trackers`, `/affects` etc are appended to it each time.
I don't mind doing it either way but if we change it here I'd change it everywhere else for consistency, but I don't think it's worth it.

dmonzonis:
@costaconrado @osoukup I went ahead and removed the old workflows API as well as the osim one assuming it's not in use yet. This cleans things up a bit and we don't have to carry over the old API for no reason.

apps/workflows/tests/test_endpoints.py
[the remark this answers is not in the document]
osoukup: Then, leave it as it is. Good to be consistent.

osoukup (approved):
Better!

osoukup:
You can ignore the failed meta check. It is a safety one not to introduce breaking changes by mistake but we introduce them intentionally.

PR #471, opened 2024-03-13
Fix integration tests to use HTTP_JIRA_API_KEY

jobselko:
This PR fixes the issue with tests mentioned in #468.


osoukup (approved):
LGTM

osoukup:
@costaconrado just FYI I think this is something you might be generally looking into as part of https://github.com/RedHatProductSecurity/osidb/pull/416 as I can reproduce these issue in my staging setup but my prod one is all green which (without any further analysis what is different) shows that the tests are environment-dependent

PR #473, opened 2024-03-14
adjust the delegated resolution handling

osoukup:
I first understood OSIDB-1406 wrong and this fixes it - the duplicate and migrated resolutions are now ignored

### DIFF
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -346,6 +346,24 @@ class TestFlaw:
         # no trackers = affected
         assert delegated_affect.delegated_resolution == Affect.AffectFix.AFFECTED
 
+        # Migrated/duplicate trackers should be ignored
+        # regardless whether there are other trackers or not
+        TrackerFactory(
+            affects=(delegated_affect,),
+            status="closed",
+            resolution="migrated",
+            embargoed=delegated_affect.flaw.is_embargoed,
+            type=Tracker.TrackerType.BUGZILLA,
+        )
+        TrackerFactory(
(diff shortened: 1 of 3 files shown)


osoukup:
@dmonzonis this reshuffles a bit your code created in https://github.com/RedHatProductSecurity/osidb/pull/456/files to address the comment from Przemyslaw in OSIDB-1406. Could you review the change please? Sorry for the confusion.

osidb/tests/test_flaw.py
```diff
@@ -404,29 +422,6 @@ def test_delegated_affects(self):
...
-            type=Tracker.TrackerType.BUGZILLA,
-        )
-        TrackerFactory(
-            affects=(affect,),
-            status="closed",
-            resolution="won't do",
-            embargoed=affect.flaw.is_embargoed,
-            type=Tracker.TrackerType.BUGZILLA,
-        )
-        assert affect.delegated_resolution == Affect.AffectFix.WONTFIX
```
dmonzonis, line 428: We could still leave this test, as it still checks that migrated/duplicate resolutions are ignored when in conjunction with another valid resolution

dmonzonis (approved):
LGTM, just one comment

osidb/tests/test_flaw.py
osoukup: @dmonzonis I am not removing it, I just moved the trackers to be ignored at the very top of the method to make sure that they get ignored all the time till the end. Does it make sense?

dmonzonis: You're right, my bad, all good then!

PR #476, opened 2024-03-18
Update sla.yml to match SFM2 w.r.t. tracker compliance-priority.

jsvob:
Closes OSIDB-2257

### DIFF
--- a/apps/sla/sla.yml
+++ b/apps/sla/sla.yml
@@ -44,14 +44,15 @@ sla:
 name: Critical Compliance Priority
 description: >
   SLA policy applied to critical impact on
-  compliance priority module and component
+  compliance priority module, component and stream
 conditions:
   affect:
     - aggregated impact is critical
-    - is compliance priority
     - is not community
   flaw:
     - is not embargoed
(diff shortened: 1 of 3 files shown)


apps/sla/sla.yml
```diff
@@ -44,11 +44,12 @@ sla:
 name: Critical Compliance Priority
 description: >
   SLA policy applied to critical impact on
-  compliance priority module and component
+  compliance priority module, component and stream
 conditions:
+  tracker:
+    - is compliance priority
```
osoukup, line 50: I would be for alphabetical ordering of the dictionary attributes from top to bottom if there is no reason not to do it that way - eg. that `tracker` would be some kind of title etc.

osoukup (approved):
I have one rather cosmetic comment but generally LGTM. I am glad that the SLA framework is clear enough to understand :smile:

apps/sla/sla.yml
jsvob: Fixed (put `tracker:` below `flaw:`)

PR #477, opened 2024-03-18
fix READ_ONLY operations

costaconrado:
This PR removes writing endpoints when OSIDB_READONLY_MODE=1.


apps/exploits/api.py
```diff
@@ -38,6 +40,8 @@ class ExploitsCollect(APIView):
     **NOTE:** Currently for CISA data only, which is very small and collection is fast.
     """
 
+    http_method_names = get_valid_http_methods(ModelViewSet)
```
osoukup, line 43: Was this consulted with @skontar ?

osoukup (changes requested):
I would like a confirmation from @skontar that it is OK to remove `/exploits/api/v1/collect` PUT method. Otherwise LGTM

apps/exploits/api.py
costaconrado: I have talked with @skontar and he believes it should stay and allow user to start the collector in read mode. I have removed the changes and kept the endpoint.

osoukup (approved):
LGTM

PR #480, opened 2024-03-21
Add support for additional_fields for Jira BTS

dmonzonis:
In product definitions, there can be optional additional fields for Jira, which at the time of this commit are `fixVersions` or `release_blocker`. This commit translates these fields to fields that the Jira tracker understands, and adds them to the query when creating/updating a tracker.

Closes OSIDB-696


jsvob (approved):

osoukup (approved):
LGTM

PR #481, opened 2024-03-22
Migrate data with outdated workflow_state values to the current ones

dmonzonis:
Commit d5e70ef6467d7ae739251eb31f9adc7357e76040 introduced a change in the available workflow states, but existing data with the old fields was not migrated. This leads to flaws with invalid values. This commit adds a migration to migrate the old data to the new states to fix these inconsistencies.

Closes OSIDB-1718


jsvob (approved):
LGTM, but someone else should critically evaluate the mapping

osoukup (approved):
LGTM. Thank you for handling this!

PR #482, opened 2024-03-25
OSIDB-2347: Readonly CVSS score

JakubFrejlach:
This PR:
* makes AffectCVSS and FlawCVSS scores readonly
* minor - changes AffectCVSS DELETE response code from 204 to 200 (just a schema adjustment, in really we are already sending 200)

Closes OSIDB-2347


JakubFrejlach:
`Meta checks / openapi (pull_request)` fails because the DELETE return code change so it is potentially a breaking changes however we are only changed the description, in really 200 instead of 204 is returned for some time so this is technically not a breaking change.

osoukup (approved):
LGTM

PR #483, opened 2024-03-25
OSIDB-2269: implement history on Flaw and Affects

JimFuller-RedHat:
This PR implements the audit history tracking feature ... and surfaces up a _history_ key on **Flaws** and **Affects** in REST API.

By default history is not included and one must use the **include_history** url param to enable:
```
osidb/api/v1/flaws?include_history=True
osidb/api/v1/affects?include_history=True
```

In addition added a primitive audit endpoint which lists all events:
```
osidb/api/v1/audit
```

**NOTE** - The batch loading of all BZ historical data into these new event tables will be performed as a one off data migration (for stage/uat/prod).


JimFuller-RedHat:
got a weird django migration check error

_/home/jfuller/src/osidb/venv/lib64/python3.9/site-packages/django/core/management/commands/makemigrations.py:105: RuntimeWarning: Got an error checking a consistent migration history performed for database connection 'default': could not translate host name "postgres" to address: Name or service not known_

will check when I come back

osidb/migrations/0123_history_audit.py
```diff
@@ -0,0 +1,869 @@
...
+         AND   acl_write::uuid[] && string_to_array(current_setting('osidb.acl'), ',')::uuid[]);
+    --policy for entity delete
+    DROP policy if exists acl_policy_meta_delete on osidb_flawmetaaudit;
+    create policy acl_policy_meta_delete
+    on osidb_flawmetaaudit
+    for delete
+    USING (acl_write::uuid[] && string_to_array(current_setting('osidb.acl'),',')::uuid[]);
+    """,
+        ),
+    ]
```
osoukup, line 869: I think I will trust that this migration does all it should :smile: this is above my knowledge.

osidb/models.py
```diff
@@ -606,6 +607,13 @@ def fts_search(q):
         # If search has no results, this will now return an empty queryset
 
 
+@pghistory.track(
+    pghistory.InsertEvent(),
+    pghistory.UpdateEvent(),
+    pghistory.DeleteEvent(),
+    exclude="local_updated_dt,meta_attr,_alerts",
```
osoukup, line 614: The `_alerts` attribute will unfortunately soon cause some conflict as Daniel is just reworking it into a proper model. That future model can also be excluded from the history as we can recreate the alerts if needed.

osidb/models.py
```diff
@@ -2851,6 +2866,13 @@ def create_flawmeta(flaw, _type, meta, **extra_fields):
             )
 
 
+@pghistory.track(
+    pghistory.InsertEvent(),
+    pghistory.UpdateEvent(),
+    pghistory.DeleteEvent(),
+    exclude="meta_attr,_alerts",
+    model_name="FlawMetaAudit",
+)
```
osoukup, line 2875: `FlawMeta` is a relic model and a subject for a removal (with the next major release). All of its parts were already migrated into other places (like `FlawReference` etc.). You can simply exclude it from the audit.

osoukup (commented):
I am submitting a partial review. Will continue tomorrow.

osidb/serializer.py
```diff
@@ -890,6 +972,7 @@ class Meta:
             + ACLMixinSerializer.Meta.fields
             + AlertMixinSerializer.Meta.fields
             + TrackingMixinSerializer.Meta.fields
+            + HistoryMixinSerializer.Meta.fields
```
osoukup, line 975: nitpick: I am sure that it is not always possible but when it is I would prefer ordering the list-like structures alphabetically so a human can easily see what is there and what is not

osidb/tests/test_audit.py
```diff
@@ -0,0 +1,176 @@
...
+            uuid.uuid5(
+                uuid.NAMESPACE_URL,
+                "https://osidb.prod.redhat.com/ns/acls#data-prodsec-write",
+            )
+        ]
+
+    def test_flawevent(self):
+        """ """
+        # create a flaw with an affect
+        Flaw.objects.all().delete()
```
osoukup, line 38: I wonder why the flaws are deleted first. Looking at the event count later it seems that it generates no event so I guess it has no effect - which is what I would assume at the beginning of a test (no flaws present).

osoukup (commented):
Submitting the second part of my review.

osoukup (changes requested):
So overall this LGTM with one exception which is `FlawMeta`. It should be removed soon so we should not add more code using it so the removal is not getting complicated. Otherwise a lot of great work done!

osidb/serializer.py
```diff
@@ -402,6 +404,79 @@ class Meta:
...
+
+    def __init__(self, *args, **kwargs):
+        # Instantiate the superclass normally
+        super().__init__(*args, **kwargs)
+
+        request = self.context.get("request")
+        include_history_param = None
+        # Get include history from request
+        if request:
+            include_history_param = request.query_params.get("include_history")
```
JakubFrejlach, line 456: Suggestion: You might want to describe this query parameter in OpenAPI schema the same way as we did with the `include_fields` or `exclude_fields`, see api_views.py for that.

osidb/serializer.py
```diff
@@ -402,6 +404,79 @@ class Meta:
...
+        else:
+            include_history_param = self.context.get("include_history")
+
+        if include_history_param is None:
+            self.fields.pop("history", None)
```
JakubFrejlach, line 461: Putting basically `include_fields=<anything>` will surface the history?

JakubFrejlach (commented):
I have same smaller remarks or question but otherwise looking good!

JakubFrejlach (approved):

osidb/api_views.py
```diff
@@ -1046,3 +1048,16 @@ def options(self, request, *args, **kwargs):
...
+    """basic view of audit history events"""
+
+    queryset = pghistory.models.Events.objects.all().order_by("-pgh_created_at")
+    serializer_class = AuditSerializer
+    filter_backends = (DjangoFilterBackend,)
+    filterset_fields = ["pgh_slug", "pgh_label", "pgh_obj_model", "pgh_created_at"]
+    http_method_names = get_valid_http_methods(
+        ModelViewSet, excluded=["delete", "post", "update"]
+    )
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
osoukup, line 1063: I have a question (maybe I already asked in the past). Is the audit/history safeguarded by the same ACLs? I do not understand this part well but it either should be or the audit would need to be only for the privileged users not to expose anything above the ACLs.

osoukup: I see below in the migration that the audit table has `acl_read`/`acl_write` so I think my question is probably unnecessary and it seem everything is as it should be.

osoukup (approved):
Looks great!

docs/developer/DEVELOP.md
```diff
@@ -877,3 +877,82 @@ def forwards_func(...):
...
+
+which we might employ in django migrations.
+
+Alternately, we can temporarily disable event triggers at the pg table level with 'brute force'
+by directly disabling in postgres with something like:
+
+```
+ALTER TABLE osidb_flaw DISABLE TRIGGER ALL;
+ALTER TABLE osidb_affect DISABLE TRIGGER ALL;
+ALTER TABLE osidb_flawmeta DISABLE TRIGGER ALL;
```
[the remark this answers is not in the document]
JimFuller-RedHat: ah I missed that, thx

docs/developer/DEVELOP.md
```diff
@@ -877,3 +877,82 @@ def forwards_func(...):
...
+```
+This approach helpfully returns diff of changes from previous events.
+
+We can retrieve all events specific to any entity including any foreign key
+relationships using **references** method. 
+```python
+flaw1 = Flaw.objects.all()[1]
+print(pghistory.models.Events.objects.references(flaw1).count())
+```
+This would include events on the specific Flaw and any FlawMeta and Affect events
```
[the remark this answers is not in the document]
JimFuller-RedHat: will remove

docs/developer/DEVELOP.md
```diff
@@ -877,3 +877,82 @@ def forwards_func(...):
 
 The `set_user_acls` is the important part here, you should apply it to both
 the forwards and backwards function before performing any database operations.
+
+### History audit
+
+We generate an audit of activity on Flaw, FlawMeta and Affect. This information is
```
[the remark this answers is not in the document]
JimFuller-RedHat: will remove

PR #484, opened 2024-03-26
Add fields to sorting flaw api

costaconrado:
This PR includes additional fields into flaw API for sorting.

Closes OSIDB-2367.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -3029,6 +3029,7 @@ paths:
             - -affects__type
             - -affects__updated_dt
             - -affects__uuid
+            - -bz_id
             - -components
             - -created_dt
             - -cve_id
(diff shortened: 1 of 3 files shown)


openapi.yml
```diff
@@ -3095,6 +3102,7 @@ paths:
             - affects__type
             - affects__updated_dt
             - affects__uuid
+            - bz_id
```
osoukup, line 3105: I am either hallucinating or I see some items twice in the list :eyes: can you please double check it?

osoukup (approved):
LGTM but it seems to me that there are some duplicates in the list

openapi.yml
JakubFrejlach: There is `-bz-id` and `bz-id` for descending and ascending order.

costaconrado: it adds all items twice one for asc and one for desc like: `bz_id` and `-bz_id` in this case.

osoukup: I see! The difference is not too visible for a tired human :smile:

PR #486, opened 2024-04-02
Create API for Alerts

dmonzonis:
Alerts have been refactored into their own model instead of being stored in a JSON field, and an API endpoint to retrieve alerts for all models has been created.

Closes OSIDB-325.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -2192,6 +2192,160 @@ paths:
                   version:
                     type: string
           description: ''
+  /osidb/api/v1/alerts:
+    get:
+      operationId: osidb_api_v1_alerts_list
+      description: List existing alerts for all models.
+      parameters:
+      - in: query
+        name: alert_type
+        schema:
+          type: string
+          enum:
+          - ERROR
+          - WARNING
+      - in: query
+        name: exclude_fields
+        schema:
+          type: array
+          items:
+            type: string
+        description: 'Exclude specified fields from the response. Multiple values
+          may be separated by commas. Dot notation can be used to filter on related
+          model fields. Example: `exclude_fields=field,related_model_field.field`'
+      - in: query
+        name: include_fields
+        schema:
+          type: array
+          items:
+            type: string
+        description: 'Include only specified fields in the response. Multiple values
+          may be separated by commas. Dot notation can be used to filter on related
+          model fields. Example: `include_fields=field,related_model_field.field`'
+      - name: limit
+        required: false
+        in: query
+        description: Number of results to return per page.
+        schema:
+          type: integer
+      - in: query
+        name: name
+        schema:
+          type: string
+        description: Retrieve only Alerts with the specified name, which is given
+          by the model's validation process.
+      - name: offset
+        required: false
+        in: query
+        description: The initial index from which to return the results.
+        schema:
+          type: integer
+      - in: query
+        name: parent_model
+        schema:
+          type: string
+          enum:
+          - affect
+          - affectcvss
+          - flaw
+          - flawacknowledgment
+          - flawcomment
+          - flawcvss
+          - flawmeta
+          - flawreference
+          - package
+          - snippet
+          - tracker
+        description: Retrieve only Alerts related to the specified model, e.g. flaw
+          or affect.
+      - in: query
+        name: parent_uuid
+        schema:
+          type: string
+          format: uuid
+        description: Retrieve only Alerts related to a model with the given UUID.
+      - in: query
+        name: uuid
+        schema:
+          type: string
+          format: uuid
+      tags:
+      - osidb
+      security:
+      - OsidbTokenAuthentication: []
+      - {}
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                allOf:
+                - $ref: '#/components/schemas/PaginatedAlertList'
+                - type: object
+                  properties:
+                    dt:
+                      type: string
+                      format: date-time
+                    env:
+                      type: string
+                    revision:
+                      type: string
+                    version:
+                      type: string
+          description: ''
+  /osidb/api/v1/alerts/{uuid}:
+    get:
+      operationId: osidb_api_v1_alerts_retrieve
+      parameters:
+      - in: query
+        name: exclude_fields
+        schema:
+          type: array
+          items:
+            type: string
+        description: 'Exclude specified fields from the response. Multiple values
+          may be separated by commas. Dot notation can be used to filter on related
+          model fields. Example: `exclude_fields=field,related_model_field.field`'
+      - in: query
+        name: include_fields
+        schema:
+          type: array
+          items:
+            type: string
+        description: 'Include only specified fields in the response. Multiple values
+          may be separated by commas. Dot notation can be used to filter on related
+          model fields. Example: `include_fields=field,related_model_field.field`'
+      - in: path
+        name: uuid
+        schema:
+          type: string
+          format: uuid
+        description: A UUID string identifying this alert.
+        required: true
+      tags:
+      - osidb
+      security:
+      - OsidbTokenAuthentication: []
+      - {}
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                allOf:
+                - $ref: '#/components/schemas/Alert'
+                - type: object
+                  properties:
+                    dt:
+                      type: string
+                      format: date-time
+                    env:
+                      type: string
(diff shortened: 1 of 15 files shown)


osidb/api_views.py
```diff
@@ -828,3 +831,47 @@ class TrackerView(ModelViewSet):
     filterset_class = TrackerFilter
     http_method_names = get_valid_http_methods(ModelViewSet, excluded=["delete"])
     permission_classes = [IsAuthenticatedOrReadOnly]
+
+
+@include_meta_attr_extend_schema_view
```
osoukup, line 836: `include_meta_attr_extend_schema_view` should probably go away

osidb/filters.py
```diff
@@ -779,3 +780,16 @@ class Meta:
             # versions fields
             "versions__version": ["exact"],
         }
+
+
+class AlertFilter(IncludeFieldsFilterSet, ExcludeFieldsFilterSet):
+    owner_uuid = CharFilter(field_name="object_id")
+    owner_model = CharFilter(field_name="content_type__model")
```
osoukup, line 787: I am thinking whether the `owner_` naming is good. At least for me it sounds like something about a person which is responsible for (resolving) the alert. I am thinking how to denote the relation by other name. What would you say about eg. `parent_` ?

JakubFrejlach, line 787: Suggestion: This might be turned into a ChoiceFilter, I think we can retrieve all models which are subclassing some mixin like this approx:
```python
In [32]: def get_mixin_subclases(mixin):
    ...:     result = []
    ...:     for subclass in mixin.__subclasses__():
    ...:         if subclass._meta.abstract:
    ...:             result.extend(get_mixin_subclases(subclass))
    ...:         else:
    ...:             result.append(subclass.__name__)
    ...:     return result
    ...:

In [33]: get_mixin_subclases(AlertMixin)
Out[33]:
['Flaw',
 'Affect',
 'FlawCVSS',
 'AffectCVSS',
 'FlawComment',
 'FlawAcknowledgment',
 'FlawReference',
 'Package',
 'Snippet',
 'Tracker',
 'FlawMeta']
```

osidb/api_views.py
```diff
@@ -828,3 +831,47 @@ class TrackerView(ModelViewSet):
...
+            ),
+            OpenApiParameter(
+                "owner_uuid",
+                type=OpenApiTypes.STR,
+                location=OpenApiParameter.QUERY,
+                description=(
+                    "Retrieve only Alerts related to a model with the given UUID."
+                ),
+            ),
+            OpenApiParameter(
```
JakubFrejlach, line 859: If my suggestion with the owner model choices will be added I think this might be removed and drf-spectacular will pick up the choices from filter itself.

osidb/migrations/0123_alert_refactor.py
```diff
@@ -0,0 +1,153 @@
...
+        model.objects.exclude(_alerts={}).exclude(_alerts__isnull=True).iterator()
+    ):
+        for key in instance._alerts.keys():
+            values = instance._alerts.get(key)
+            yield Alert(
+                name=key,
+                description=values.get("description", ""),
+                alert_type=values.get("type", "warning"),
+                resolution_steps=values.get("resolution_steps", ""),
+                content_object=instance,
```
JakubFrejlach, line 51: Cannot apply the migration, I am getting prompted with a traceback on this particular line:
```
...
TypeError: Alert() got an unexpected keyword argument 'content_object'
```

JakubFrejlach (changes requested):
Added some suggestions which are not really urgent, however locally I was not able to run the migration so not approving yet as not sure if migration is working or not.

openapi.yml
```diff
@@ -7343,6 +7348,33 @@ components:
       - AFFECTED
       - NOTAFFECTED
       type: string
+    Alert:
+      type: object
+      description: Alert serializer to expose alerts stored in models that inherit
+        from AlertMixin.
```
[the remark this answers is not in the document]
osoukup: It is not very intuitive and the docs creation is not even very documented as far as I can remember.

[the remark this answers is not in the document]
osoukup: The meaningful and understandable description for the user is the preference here. The developer can live with it :smile:

osidb/filters.py
dmonzonis: To avoid computational overhead (if the subclasses have to be recomputed each time we get the choices) I have just added the possible options in the OpenApiParameter definition; this way we also get a list of options in the docs, although they will have to be revised in the event that the architecture changes.

osidb/mixins.py
```diff
@@ -148,68 +149,75 @@ class Meta:
         abstract = True
 
 
+class Alert(ValidateMixin):
```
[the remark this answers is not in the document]
dmonzonis: It was added mostly to not allow incorrect alert types, as there was a test case for that (`TestModelDefinitions::test_alert_incorrect_type`). But it's true that this case would be a programmer error since alerts are not created by users. So I will remove this test case as well.

osidb/filters.py
JakubFrejlach: Sure, that is a good idea. That little function I included might go into `osidb/helpers.py` and it can be used to detect any changes from dev point of view. We already have some similar helper functions there as well (eg. get_unique_meta_attr_keys).

dmonzonis: I will add it there. I did use a pretty similar function to get the list of models for the migration, but it was done ad-hoc,  didn't think about keeping it somewhere for future reference.

osoukup (approved):
LGTM

osidb/mixins.py
```diff
@@ -148,68 +149,79 @@ class Meta:
         abstract = True
 
 
+class Alert(models.Model):
```
jsvob, line 152: Nitpick: I'd prefer to store the model itself in `osidb/models.py` for consistency with most other models.

dmonzonis: The only reason I added it here is to avoid a circular dependency: the `AlertMixin` uses `Alert`, and models in the `osidb/models.py` file use `AlertMixin`

osoukup, line 152: @dmonzonis I was thinking about using the alerts in the collector framework to store some data collection issues so we can query them easily. While thinking about it I realized one issue here.

The alerts do not have any ACL controls while they can still be freely queried on API by anyone authenticated. That means that even though a user would have only a public access-level, (s)he could eg. see alerts of the embargoed flaws and OSIDB would be leaking an information about their existence and potentially much more. The embargo is always taken quite seriously in ProdSec so I would see it as quite an issue.

I see two way how to overcome it. Probably easier way would be to restrict the API endpoint to those with the embargo access only. Anyone could see instance-specific alerts among its data so this general endpoint would have more narrowed usage anyway. A cleaner solution would be to provide the `Alert` model with ACLs which would be inherited from the parent instance.

I know this is quite late in game so feel free to create a separate Jira for this improvement and plan it for the next sprint so it does not interfere with your already planned work.

osoukup (changes requested):

osidb/mixins.py
dmonzonis: I changed it so that alerts now properly use the parent's ACLs.

dmonzonis:
@JakubFrejlach may I get a re-review from you as well? The original issues were fixed a while ago, and the only new change is that now `Alert` uses ACLs.

osoukup (approved):
LGTM. Thank you for addressing the ACL issue!

osidb/filters.py
```diff
@@ -795,3 +796,16 @@ class Meta:
             # versions fields
             "versions__version": ["exact"],
         }
+
+
+class AlertFilter(IncludeFieldsFilterSet, ExcludeFieldsFilterSet):
+    parent_uuid = CharFilter(field_name="object_id")
+    parent_model = CharFilter(field_name="content_type__model")
```
JakubFrejlach, line 803: I think this could be turned into `ChoiceFilter` and thus the whole query parameter annotation in the `osidb/api_views.py` might not be needed then but I have no strong preference and will leave it up to you.

JakubFrejlach (approved):
LGTM now! Great job.

osidb/migrations/0123_alert_refactor.py
JakubFrejlach: @dmonzonis So this was actually some problem on my side, I issued `make compose-down` and build everything from scratch and it works.

PR #487, opened 2024-04-03
Show cvss_version as enum in API schema for CVSS

JakubFrejlach:
This PR:
* changes `cvss_version` for FlawCVSS and AffectCVSS on API schema from string to enum


JakubFrejlach:
Schema changes should be OK as again this is just a description change, cvss_version is in fact an enum and we just shown it as a string until now.

dmonzonis (approved):
LGTM

osoukup (approved):
LGTM

PR #488, opened 2024-04-05
Generate CC lists for new Jira trackers

jsvob:
Closes OSIDB-2191


apps/trackers/tests/test_bugzilla.py
```diff
@@ -478,3 +478,106 @@ def test_generate_keywords(self):
...
+            expected_cc = sorted(
+                set(
+                    f"{n}@redhat.com" if "@" not in n else n
+                    for n in (
+                        expected_component_cc + expected_private_tracker_cc + default_cc
+                    )
+                )
+            )
+
+            if flaw.embargoed and not ps_module.private_trackers_allowed:
```
jsvob, line 577: This is explicitly coded (but without a comment) in AffectCCBuilder, but I can't find a similar thing in SFM2 for _trackers_. Is this a flaw-only affect-only thing not applicable to trackers?

apps/trackers/jira/query.py
```diff
@@ -267,3 +271,89 @@ def generate_additional_fields(self):
...
+        #      TODO: Is this a bug?
+        # Parse BZ component
+        if self.ps_component and "/" in self.ps_component:
+            bz_component = self.ps_component.split("/")[-1]
+        else:
+            bz_component = self.ps_component
+
+        # embargoed value unused here
+        affect_cc_builder = AffectCCBuilder(
+            self.tracker.affects.first(), embargoed=None
```
osoukup, line 301: `AffectCCBuilder` was mean to be used for each affect and the results were mean to be merged. Why is it only used for the first (which is probably close to random) affect here?

apps/trackers/jira/query.py
```diff
@@ -267,3 +271,89 @@ def generate_additional_fields(self):
...
+        # Replaces contact aliases with appropriate emails/usernames if alias
+        # has the contact set for current BTS. Other records are made intact.
+        # NOTE: Similar functionality in AffectCCBuilder.expand_alias(),
+        #       but only for a single contact at a time.
+        contacts = dict(
+            PsContact.objects.all()
+            .values_list("username", "jboss_username")
+            .filter(username__in=cc_list)
+        )
+        cc_list = {contacts.get(cc, cc) for cc in cc_list}
```
osoukup, line 322: I would be for extending the `expand_alias` method to support the bulk expansion rather than implementing the expansion on two different places.

apps/trackers/jira/query.py
```diff
@@ -267,3 +271,89 @@ def generate_additional_fields(self):
...
+        )
+        # TODO: Why does AffectCCBuilder set bz_component to None?
+        affect_cc_builder.bz_component = bz_component
+        cc_list = set(affect_cc_builder.component_cc())
+
+        if self.ps_module.default_cc:
+            # Default CC List
+            cc_list.update(self.ps_module.default_cc)
+        if self.tracker.is_embargoed and self.ps_module.private_tracker_cc:
+            cc_list.update(self.ps_module.private_tracker_cc)
```
osoukup, line 311: I see that there is no `private_tracker_cc` in `AffectCCBuilder.module_cc`. That may be a bug. Could you rather improve that method and use it here instead of implementing it again?

apps/trackers/bugzilla/query.py
```diff
@@ -91,7 +91,24 @@ def generate_cc(self):
...
+
+        # NOTE: AffectCCBuilder doesn't implement private_tracker_cc (as of April 2024).
+        #       This is not a problem because we will file BZ trackers only for public
+        #       projects (probably only Fedora).
+        if self.tracker.external_system_id:
+            # Add CCs only on creation.
+            return
+
+        affect_cc_builder = AffectCCBuilder(
+            self.tracker.affects.first(),
```
osoukup, line 103: I am not sure we agree on supporting multi-flaw trackers (another comment) but if so we should adjust this piece of code accordingly.

apps/trackers/tests/test_bugzilla.py
osoukup: When private trackers are not allowed but the flaw is embargoed, we should not create any trackers at all so then the tracker CC list seems irrelevant to me. I am not very sure what is the reasoning behind this functionality but it is probably some exceptional product workflow.

osoukup (changes requested):
I raised a few comments but they does not necessarily mean that your solution is incorrect :smile: as it may really be just my bad understanding. I do not feel as an expert on CC lists. If you think your code is OK, just provide an answer to the question. Generally, this seems as a good approach and result of a complicated research. Good job!

apps/trackers/jira/query.py
```diff
@@ -267,3 +271,89 @@ def generate_additional_fields(self):
...
+        # NOTE That SFM2 for Jira tracker creation uses ps_module.component_overrides only for
+        #      generating Jira "components" field, which OSIDB doesn't do, but not for CC lists;
+        #      CC list creation is based solely on ps_component, not on bz_component.
+        #      Therefore AffectCCBuilder.ps2bz_component is not reused here.
+        #      TODO: Is this a bug?
+        # Parse BZ component
+        if self.ps_component and "/" in self.ps_component:
+            bz_component = self.ps_component.split("/")[-1]
+        else:
+            bz_component = self.ps_component
```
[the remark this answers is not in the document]
jsvob: See https://gitlab.cee.redhat.com/prodsec-dev/sfm2/-/blob/master/sfm2/backends/jira/rest/request.py#L143 and the FIXME and the `parse_bz_component` call that does the splitting for Jira trackers. Whether that is correct or something is appreciably different in SFM2 vs OSIDB, I don't know.

How to divine out what is the correct way? For example the following investigation didn't help me a bit:

```
>>> from osidb.models import PsModule
>>> for m in PsModule.objects.all():
...  if "/" in m.default_component:
...   print(m.default_component)
...   print(m.name)
...   print(m.bts_name)
...   print()
... 
source/security
rhsso-7
jboss

source/security
rhbk
jboss

>>> 

```

apps/trackers/jira/query.py
jsvob: It seemed like a great idea for a reason absolutely eluding me now, and I swear I wasn't under influence of anything. Thanks for pointing this out, will fix, this is wrong as it is now.

apps/trackers/bugzilla/query.py
jsvob: Yes, will fix by creating per-affect CC lists and then merging them.

apps/trackers/jira/query.py
osoukup: It is too complicated indeed. But I think that the significant part of the complicated matter is that SFM2 code is a complete mess. There are patterns and building blocks I tried to identify, eg.

- flaw CC is composition affect CCs
- affect CC is composition of module, component, and Bugzilla CCs
- RHSCL affect CC is affect CC with a different component handling
- etc.

so I think that decomposition and deduplication is valuable here. I think this code will live with us for some time and I wanted to avoid repeating the SFM2 unmaintainable mess - we might be soon able to cut the sync from Bugzilla but the other direction does not seem easy to be thrown away.

However, I do not want to dictate it here. If you think that deduplication would be too complicated or too risky in the sense of introducing new errors then do it somehow so it done :smiley:

apps/trackers/jira/query.py
jsvob: Yes, will fix accordingly.

Confirming that SFM2 does generate affect CC lists from private_tracker_cc, so adding this in OSIDB as you suggest will fix an OSIDB bug and bring it in line with SFM2 (the class is called AffectCCBuilder after all).

- cc_list_for_affects method calls cc_list_from_prod_defs method which generates ccs from private_tracker_cc
- cc_list_for_affects method is called by cc_list_for_flaw method which is called from FlawBase class so its's used in SFM2

Also confirming that AffectCCBuilder is used to build affect CC lists in OSIDB already (given the [confusion](https://redhat-internal.slack.com/archives/C061QHHTM2M/p1712342079811019) I wasn't sure).

apps/trackers/tests/test_bugzilla.py
jsvob: Makes sense. Leaving the check in as it seems to not hurt anything if trackers are created in this case that shouldn't happen.

apps/bbsync/cc.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, affect: Affect, embargoed: bool) -> None:
         self.ps_module_obj = PsModule.objects.get(name=affect.ps_module)
         self.embargoed = embargoed
 
+        # TODO why is it None if BTS is not bugzilla?
```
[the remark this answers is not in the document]
jsvob: Indeed. Bugzilla-based affect/tracker CC lists use BugzillaComponent objects while Jira-based don't. In the revised version Jira path has separate logic.

apps/trackers/jira/query.py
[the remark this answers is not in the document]
jsvob: Functionality replicated but explained and (hopefully) clearly separated into topics in the revised version.

apps/trackers/jira/query.py
jsvob: Affect-based lists are generated for each affect in the revised version.

apps/trackers/jira/query.py
jsvob: Fixed in the revised version.

osoukup (approved):
LGTM. I think this is more then enough! Great job!

apps/bbsync/tests/test_cc.py
```diff
@@ -435,26 +435,33 @@ def test_embargoed_blacklist(self):
...
+                "cat@redhat.com",
+                "dog@redhat.com",
+                "duck@fedora.org",
+            ]
+        else:
+            assert add_cc == [
+                "cat",
+                "dog",
+                "duck@fedora.org",
+            ]
```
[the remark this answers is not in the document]
jsvob: Fixed, thanks.

osidb/cc.py
```diff
@@ -0,0 +1,486 @@
...
+        #      generating Jira "components" field, which OSIDB doesn't do, but not for CC lists;
+        #      CC list creation is based solely on ps_component, not on bz_component.
+        #      Therefore ps2bz_component is not reused here.
+        # Parse BZ component
+        if self.ps_component and "/" in self.ps_component:
+            self.bz_component = self.ps_component.split("/")[-1]
+        else:
+            self.bz_component = self.ps_component
+
+    pass
```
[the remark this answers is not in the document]
jsvob: Fixed, thanks.

PR #489, opened 2024-04-09
OSIDB-2354: create restore script for use with sql backup from uat/stage

JimFuller-RedHat:
To use

```
> scripts/restore_pg.sh osidb-dump.sql
```

**Note** - we will add uat groups when that environment is created.

### DIFF
--- a/config/settings_local.py
+++ b/config/settings_local.py
@@ -7,25 +7,29 @@ from .settings import *
 SECRET_KEY = get_random_secret_key()  # pragma: allowlist secret
 
 # Minimal group for read access of public flaws in OSIDB
-PUBLIC_READ_GROUPS = ["data-prodsec"]
+PUBLIC_READ_GROUPS = [
+    "data-prodsec",
+    "osidb-stage-public-read",
+]
 # Minimal group for write access of public flaws in OSIDB
-PUBLIC_WRITE_GROUP = "data-prodsec-write"
+PUBLIC_WRITE_GROUP = ["data-prodsec-write", "osidb-stage-public-write"]
 # Minimal group for read access of embargoed flaws in OSIDB
-EMBARGO_READ_GROUP = "data-topsecret"
+EMBARGO_READ_GROUP = ["data-topsecret", "osidb-stage-embargo-read"]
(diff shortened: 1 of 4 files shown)


config/settings_local.py
```diff
@@ -7,25 +7,29 @@
...
+EMBARGO_READ_GROUP = ["data-topsecret", "osidb-stage-embargo-read"]
 # Minimal group for write access of embargoed flaws in OSIDB
-EMBARGO_WRITE_GROUP = "data-topsecret-write"
+EMBARGO_WRITE_GROUP = ["data-topsecret-write", "osidb-stage-embargo-write"]
 # Minimal group for read access of internal flaws in OSIDB
-INTERNAL_READ_GROUP = "data-internal-read"
+INTERNAL_READ_GROUP = ["data-internal-read", "osidb-stage-internal-read"]
 # Minimal group for write access of internal flaws in OSIDB
-INTERNAL_WRITE_GROUP = "data-internal-write"
+INTERNAL_WRITE_GROUP = ["data-internal-write", "osidb-stage-internal-write"]
```
osoukup, line 23: The groups from the settings are used throughout the code and wherever there is a singular really just one group is expected so I think this will break the local environment on number of places.

osoukup (changes requested):
I think this change breaks number of places in the code where the group constants in singular are being used. I think it should be done different way or the code must be adjusted.

config/settings_local.py
JimFuller-RedHat: ok if that is the design then I can change this the other way around eg. just rewrite everything to use local group (and not add any groups here.

osoukup: I think it is a bad design somehow organically achieved. But it is as it is now. We can fix the design which may be more costly or you can implement your change some other way

JimFuller-RedHat: as discussed in chat - for now we will just rewrite acls and not add any new groups here and raise an issue about future redesign in this area.

JimFuller-RedHat:
I have changed restore_pg.sh script to explicitly set acls to local dev read/write acls which is probably the best/safest approach for now.

osoukup (approved):
LGTM

PR #491, opened 2024-04-11
enable custom configuration of product-definitions repo branch

osoukup:
I have sanitized/stripped the product-definitions repo branch
https://gitlab.cee.redhat.com/prodsec/product-definitions/-/tree/osidb-stage?ref_type=heads
so it has no more issues when used on the stage but currently it is not possible to switch from **master**. The PR is to enable it so I can switch the stage


JakubFrejlach (changes requested):
This should be added to the example `.env` file in `docs/DEVELOP.md`, there are actually a lot of such config vars which we haven't documented.

collectors/product_definitions/core.py
```diff
@@ -115,8 +125,8 @@ def sync_ps_products_modules(ps_products_data: dict, ps_modules_data: dict):
         }
         related_ps_modules = filtered_product_data.pop("ps_modules")
 
-        ps_product, _ = PsProduct.objects.update_or_create(
-            short_name=product_short_name, defaults=filtered_product_data
+        ps_product = PsProduct.objects.create(
+            short_name=product_short_name, **filtered_product_data
         )
```
JakubFrejlach, line 130: This has been here for quite some time but I never actually realize that having it like this won't erase anything no longer relevant :smile: Good job.

JakubFrejlach (approved):
LGTM now.

PR #493, opened 2024-04-16
Allow editing flaws in NEW workflow state

dmonzonis:
There is a validation in flaws which enforces them to have at least one affect, which is only disabled when creating a flaw. This commit also disables this validation when editing 'draft' flaws, i.e. flaws which are in a NEW state, and instead issues an alert. The validation is still active in other states and when trying to move to the next state.

Closes OSIDB-2452.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Make workflows API RESTful (OSIDB-1716)
 - Minor change to enable perf tests to run in CI (OSIDB-2447)
+- Allow editing flaws without affects in NEW state (OSIDB-2452)
 
 ## [3.7.0] - 2024-04-17
 ### Added
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1277,7 +1277,17 @@ class Flaw(
             return
 
         if not Affect.objects.filter(flaw=self).exists():
-            raise ValidationError("Flaw does not contain any affects.")
+            err = ValidationError("Flaw does not contain any affects.")
+            # When a flaw in a "new" workflow state is modified, allow saving
+            # with no affects but issue an alert
+            if self.workflow_state == WorkflowModel.WorkflowState.NEW:
+                self.alert(
+                    "_validate_flaw_without_affect",
+                    err.message,
+                    _type="error",
+                )
+            else:
+                raise err
 
     def _validate_nonempty_impact(self):
         """
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -1876,15 +1876,24 @@ class TestFlawValidators:
         flaw.save(raise_validation_error=False)
 
     def test_validate_flaw_without_affect(self):
-        """test that flaws without affect raises an error on editing"""
+        """
+        test that flaws without affect raises an error on editing,
+        unless it is in a new workflow state
+        """
         flaw1 = FlawFactory()
         AffectFactory(flaw=flaw1)
         assert flaw1.save() is None
 
-        flaw2 = FlawFactory()
+        flaw2 = FlawFactory(workflow_state=WorkflowModel.WorkflowState.TRIAGE)
         with pytest.raises(ValidationError) as e:
             flaw2.save()
         assert "Flaw does not contain any affects." in str(e)
+
+        # Flaws in the 'new' state can be edited without any affects, but
+        # an alert is raised
+        flaw3 = FlawFactory(workflow_state=WorkflowModel.WorkflowState.NEW)
+        assert flaw3.save() is None
+        assert bool("_validate_flaw_without_affect" in flaw3._alerts)
 
     def test_no_impact(self):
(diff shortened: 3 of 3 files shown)


osidb/models.py
```diff
@@ -1259,7 +1259,17 @@ def _validate_flaw_without_affect(self):
...
         if not Affect.objects.filter(flaw=self).exists():
-            raise ValidationError("Flaw does not contain any affects.")
+            err = ValidationError("Flaw does not contain any affects.")
+            # When a flaw in a "new" workflow state is modified, allow saving
+            # with no affects but issue an alert
+            if self.workflow_state == WorkflowModel.WorkflowState.NEW:
+                self.alert(
+                    "_validate_flaw_without_affect",
+                    err.message,
+                    _type="error",
```
dmonzonis, line 1269: this will have to be updated if #486 gets merged first

osidb/tests/test_flaw.py
```diff
@@ -1886,6 +1889,14 @@ def test_validate_flaw_without_affect(self):
             flaw2.save()
         assert "Flaw does not contain any affects." in str(e)
 
+        # Flaws in the 'new' state can be edited without any affects, but
+        # an alert is raised
+        flaw3 = FlawFactory(
+            workflow_state=WorkflowModel.WorkflowState.NEW,
+        )
+        assert flaw3.save() is None
+        assert bool("_validate_flaw_without_affect" in flaw3._alerts)
```
dmonzonis, line 1898: same as above

docs/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SLA for compliance priority brought to parity with SFM2 (OSIDB-2257)
 - Migrate data with outdated workflow_state values to the current ones (OSIDB-1718)
 - Flaw CVSS score and Affect CVSS score are now readonly (OSIDB-2347)
+- Allow editing flaws in NEW state (OSIDB-2452)
```
osoukup, line 30: I would expand it to `Allow editing flaws without affects in NEW state`

osidb/tests/test_flaw.py
```diff
@@ -1886,6 +1889,14 @@ def test_validate_flaw_without_affect(self):
             flaw2.save()
         assert "Flaw does not contain any affects." in str(e)
 
+        # Flaws in the 'new' state can be edited without any affects, but
```
osoukup, line 1892: In the old code above there is twice created a flaw by the factory without providing any `workflow_state` value which seems to result in a weird situation when there is an empty value as it has no default but the empty value is not among the choices. I am surprised that it is even possible and therefore I suspect that the `ValidationError` might actually be raised because of this and not because of the missing affect (or there are two errors at the same time). I would provide a value different from `NEW` to prevent the weirdness.

osoukup (changes requested):
This change is an improvement and therefore I would be OK with it (if the test comment is addressed or disprove) but the final solution should probably be rather done in `workflows` module in way similar to
https://github.com/RedHatProductSecurity/osidb/blob/master/apps/workflows/workflows/default.yml#L32

Originally it worked in a cumulative way so it would be enough to just move the condition from `PRE_SECONDARY_ASSESSMENT` to `TRIAGE` to do the trick but I think it does not work this way any more (correct me if I am wrong @costaconrado) so the condition should actually be put everywhere except `NEW` - meaning that the file is probably wrong now so it will require quire a rework anyway.

osidb/tests/test_flaw.py
dmonzonis: It is indeed created with an empty state, and any state other than `NEW` raises the error. I will make it have another state to make it more clear.

dmonzonis:
@osoukup The workflow requirements are checked when we promote a flaw to the next state, but not when editing a flaw in its same state. To allow editing a flaw in the `NEW` state, I believe dealing with the validation is the only option.

Moreover, I'm not sure if the `has affects` requirement is even necessary, as this is already checked at every state (except `NEW` after this commit), since promoting a flaw ultimately calls `save` which runs the validations, namely `_validate_flaw_without_affect`, so even without the requirement the flaw will not be able to switch to the next state if it doesn't have affects.

osoukup:
> @osoukup The workflow requirements are checked when we promote a flaw to the next state, but not when editing a flaw in its same state. To allow editing a flaw in the `NEW` state, I believe dealing with the validation is the only option.
> 
> Moreover, I'm not sure if the `has affects` requirement is even necessary, as this is already checked at every state (except `NEW` after this commit), since promoting a flaw ultimately calls `save` which runs the validations, namely `_validate_flaw_without_affect`, so even without the requirement the flaw will not be able to switch to the next state if it doesn't have affects.

@dmonzonis you are correct. I just mean that we should eventually shift from state-less alerts to state-full workflow. Most of the validations only make sense after the flaw reached a certain point in the workflow and they should even serve as a guidance on which steps to take in a certain state. However, for now probably let us forgot the workflow and just go with your improvement. It will be all we need now and we can utilize the workflow later.

osoukup (approved):
LGTM

PR #494, opened 2024-04-16
fix endpoint tests

osoukup:
Closes OSIDB-2373


jobselko (approved):
LGTM (only `test_promote_endpoint` in `workflows` is failing, but this is a different issue)

osoukup:
> The tests work for me locally because I have set `JIRA_TASKMAN_AUTO_SYNC_FLAW=1` in my `.env` file.

@jobselko are you sure that this is the issue? I have changed it to 0 and checked
```
podman exec -it testrunner bash
env | grep TASK
...
JIRA_TASKMAN_AUTO_SYNC_FLAW=0
```
and all my tests in `workflows` app are passing for me locally.

jobselko:
> > The tests work for me locally because I have set `JIRA_TASKMAN_AUTO_SYNC_FLAW=1` in my `.env` file.
> 
> @jobselko are you sure that this is the issue? I have changed it to 0 and checked
> 
> ```
> podman exec -it testrunner bash
> env | grep TASK
> ...
> JIRA_TASKMAN_AUTO_SYNC_FLAW=0
> ```
> 
> and all my tests in `workflows` app are passing for me locally.

I meant failing CI tests in `apps/taskman/tests/test_flaw_model_integration.py` (these work for me when `JIRA_TASKMAN_AUTO_SYNC_FLAW=1` is set, and my `.env` also uses `export`).

As I mentioned earlier, I think that `test_promote_endpoint` in `workflows` fails on my localhost due to a different issue (but it is strange that it works for you).

osoukup:
@jobselko my question was rather meant whether you tried `JIRA_TASKMAN_AUTO_SYNC_FLAW=0` too because for me it behaves the same way regardless there is `0` or `1`. It just succeeds.

jobselko:
> @jobselko my question was rather meant whether you tried `JIRA_TASKMAN_AUTO_SYNC_FLAW=0` too because for me it behaves the same way regardless there is `0` or `1`. It just succeeds.

Yes. With `0`, it fails for me like the CI.

PR #495, opened 2024-04-16
Fix component backward compatibility

costaconrado:
This PR fix the retro compatibility for sorting API.

This PR also adds non-model fields in the sorting mechanism.

Context: during the creation of release 3.7.0 I noticed I've introduced a break on the backwards compatibility of sorting by component [here](https://github.com/RedHatProductSecurity/osidb/commit/358be8a41399a12bcb500477171a1f3007925032#diff-5016bdc926b6c0f649f363739d09159a1db664aae19a08fa6b38b03406f040dcL6269)
I also noticed other models had the same issue of non-model fields being missing in the sorting API.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1323,11 +1323,13 @@ paths:
             - -cvss3_score
             - -cvss_scores__comment
             - -cvss_scores__created_dt
+            - -cvss_scores__cvss_version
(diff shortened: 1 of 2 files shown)


osoukup (approved):
LGTM

PR #496, opened 2024-04-16
Prep release 3.7.0

costaconrado:


costaconrado:
I have divided this in 2 commits so the tag that goes to master do not list features that are not present there.

osoukup:
This PR is not raised between the right branches. I am closing it and will re-raise differently.

PR #497, opened 2024-04-17
Collect errata not linked to any flaws

dmonzonis:
Before this commit, errata which are not linked to any flaws were skipped, i.e. not stored in OSIDB.

With this commit, these errata will also be saved in the database, since it is somewhat common to have errata with trackers but no flaws, and right we need to have these stored.

Closes OSIDB-1527.


jsvob (approved):

osoukup (approved):
LGTM

PR #501, opened 2024-04-22
Allow searching flaws by CVE

dmonzonis:
Before this commit, the search query parameter allowed to search text-based fields (like title or description) by keywords. With this commit, we can now also search by similar CVEs.

This is useful if the user forgot part of a CVE or is not completely sure about one or two digits. For example, the user may be looking for the flaw associated with CVE-2001-0414 but doesn't remember the whole CVE, so they search for CVE-2001-04 which should return the flaw they're looking for (among others that look like that). Or they may search for CVE-2001-0464 because they are not sure about the second to last digit, which will also return the one they are looking for.

Note that this is different from passing `cve_id` as a query param, which matches the CVE exactly.

Closes OSIDB-2482.


docs/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Make workflows API RESTful (OSIDB-1716)
 - Minor change to enable perf tests to run in CI (OSIDB-2447)
+- Allow searching by CVE similarity (OSIDB-2482)
```
jobselko, line 11: This should be in the `### Added` part of the changelog (also, rebase is probably necessary).

jobselko (approved):

PR #503, opened 2024-04-25
Enable flaw draft creation in BZ

jobselko:
This PR enables collectors to create flaw drafts in BZ. A collector can only create a flaw draft, updating is not possible. The `alias` field prevents duplicate creation.

Flaw drafts created by NVD:
- A source flaw always contains only one cve_id, which is also used in `alias` and `external_ids`.

Flaw drafts created by OSV:
- If a source flaw does not contain cve_id, osv_id is used in both `alias` and `external_ids`.
- If a source flaw contains cve_id, cve_id is used in `alias` and osv_id/cve_id in `external_ids`.
- If a source flaw contains more cve_ids, a new flaw draft is created for each.

Closes OSIDB-2261

### DIFF
--- a/apps/bbsync/tests/test_integration.py
+++ b/apps/bbsync/tests/test_integration.py
@@ -6,13 +6,22 @@ from freezegun import freeze_time
 
 from apps.bbsync.exceptions import UnsaveableFlawError
 from collectors.bzimport.collectors import BugzillaTrackerCollector, FlawCollector
-from osidb.models import Affect, Flaw, FlawAcknowledgment, FlawCVSS, Tracker
+from osidb.models import (
+    Affect,
+    Flaw,
+    FlawAcknowledgment,
+    FlawCVSS,
+    FlawReference,
+    Snippet,
+    Tracker,
+)
 from osidb.tests.factories import (
     AffectFactory,
     FlawAcknowledgmentFactory,
     FlawFactory,
     PsModuleFactory,
     PsUpdateStreamFactory,
+    SnippetFactory,
     TrackerFactory,
 )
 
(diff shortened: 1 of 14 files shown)


osidb/tests/factories.py
```diff
@@ -513,33 +513,35 @@ class Meta:
...
     @factory.lazy_attribute
     def content(self):
+        # contains all NVD fields and only the currently used OSV fields
         data = {
             "cve_id": self.cve_id,
             "cvss_scores": [
                 {
                     "score": 8.1,
-                    "issuer": FlawCVSS.CVSSIssuer.NIST,
+                    "issuer": FlawCVSS.CVSSIssuer.REDHAT,
```
osoukup, line 544: nitpick: That sounds a bit weird that the source is NVD and the CVSS is issued by Red Hat.

osidb/tests/test_snippet.py
```diff
@@ -49,38 +49,22 @@ def test_create_flaw_from_snippet(
...
+        for i in ["type", "url"]:
+            assert getattr(flaw_ref, i) == snippet.content["references"][0][i]
 
         flaw_snippet = flaw.snippets.all().first()
         assert flaw_snippet == snippet
-        assert flaw_snippet.content == content
-        assert flaw_snippet.external_id == snippet.external_id
-        assert flaw_snippet.source == snippet.source
+        for i in ["content", "external_id", "source"]:
+            assert getattr(flaw_snippet, i) == getattr(snippet, i)
```
osoukup, line 62: I do not think that these shortcuts are a good idea for the tests. It makes it harder to debug when something goes wrong. A simple `assert X == Y` without any cycle, getter, indexing, etc. is maybe a verbose but easy to investigate structure. I do not have too strong preference however so leaving it up to you.

apps/bbsync/tests/test_integration.py
```diff
@@ -832,3 +841,74 @@ def test_flaw_update_multi_cve_restricted(self, auth_client, test_api_uri):
...
+    )
+    def test_flaw_draft_create(self, source, cve_id, ext_id):
+        """
+        test creating a flaw draft with Bugzilla two-way sync
+        """
+        content = {
+            "cve_id": cve_id,
+            "cvss_scores": [
+                {
+                    "issuer": FlawCVSS.CVSSIssuer.REDHAT,
```
osoukup, line 865: Will it work when the issue is not Red Hat? That will most probably be the majority of the cases.

apps/bbsync/tests/test_query.py
```diff
@@ -208,6 +208,26 @@ def test_generate_component(self, is_draft, result):
...
+        else:
+            snippet = SnippetFactory(
+                source=Snippet.Source.OSV, ext_id=alias_id, cve_id=None
+            )
+            flaw = FlawFactory(cve_id=None, meta_attr={"external_ids": [alias_id]})
+            snippet.flaw = flaw
+            snippet.save()
+
+        bbq = FlawBugzillaQueryBuilder(flaw)
+        assert bbq.query["alias"] == [alias_id]
```
osoukup, line 229: This test has only two lines of the join code between the parametrized cases and the parametrization adds two extra lines, branching, and conditions. I do not see it beneficial. Having two separate but much simpler test cases would be IMO much better.

osoukup (commented):
I have one question to answer and few suggestions. If the answer to the question is yes then I will approve.

apps/bbsync/tests/test_integration.py
osoukup: Discussed in person.

osoukup (approved):
Raised question discussed and concluded in person. The status qua is actually acceptable as Red Hat score is being stored in Bugzilla, NIST score is being collected by a different collector, and any other CVSS score issuers are no important at this point so if we do not sync their scores through Bugzilla and miss them in some OSIDB instance we can live with it without any issues. LGTM

osidb/tests/factories.py
jobselko: Also discussed in person.

PR #504, opened 2024-04-26
Improve collectors

osoukup:
This PR reworks the way flaw-affect-tracker linking is done. It completely removes the tracker stuff from the flaws sync and introduces the linking as part of the tracker sync (without a need to sync the flaws there). It significantly lower the FlawCollector burden and also makes the linking more resilient so even the cloned Jira trackers without any linking information on the flaw side gets linked. There is also a number of smaller improvements and tests.

Closes OSIDB-1012
Closes OSIDB-2587

### DIFF
--- a/collectors/jiraffe/collectors.py
+++ b/collectors/jiraffe/collectors.py
@@ -30,7 +30,7 @@ class JiraTrackerCollector(Collector):
     BEGINNING = timezone.datetime(2014, 1, 1, tzinfo=timezone.get_current_timezone())
     # Jira API does not seem to have either issues returning large number of results
     # or any restrictions on the maximum query period so we can be quite greedy
-    BATCH_PERIOD_DAYS = 365
+    BATCH_PERIOD_DAYS = 10
 
     def __init__(self):
         super().__init__()
--- a/collectors/jiraffe/tests/test_collectors.py
+++ b/collectors/jiraffe/tests/test_collectors.py
@@ -30,6 +30,7 @@ class TestJiraTrackerCollector:
         test that getting the next batch of Jira issues works
         """
         collector = JiraTrackerCollector()
+        collector.BATCH_PERIOD_DAYS = 365
         assert collector.BEGINNING == timezone.datetime(2014, 1, 1, tzinfo=timezone.utc)
         assert collector.metadata.updated_until_dt is None
 
@@ -56,6 +57,7 @@ class TestJiraTrackerCollector:
         correctly taking into account the FlawCollector freshness
         """
         collector = JiraTrackerCollector()
+        collector.BATCH_PERIOD_DAYS = 365
         assert collector.BEGINNING == timezone.datetime(2014, 1, 1, tzinfo=timezone.utc)
         assert collector.metadata.updated_until_dt is None
 
@@ -91,6 +93,7 @@ class TestJiraTrackerCollector:
         test the Jira collector run
         """
         collector = JiraTrackerCollector()
+        collector.BATCH_PERIOD_DAYS = 365
         assert collector.BEGINNING == timezone.datetime(2014, 1, 1, tzinfo=timezone.utc)
         assert collector.metadata.updated_until_dt is None
 
@@ -135,6 +138,7 @@ class TestJiraTrackerCollector:
         test that Jira collector data status is changed to complete when the data are current
         """
         collector = JiraTrackerCollector()
+        collector.BATCH_PERIOD_DAYS = 365
         collector.metadata.updated_until_dt = timezone.now()
         CollectorMetadata(
             name="collectors.bzimport.tasks.flaw_collector",


osoukup:
@skontar these are the linking changes I was talking about

jobselko (approved):
LGTM, but reviews from others would be useful too

skontar (approved):
Looks good to me. Some QoL improvements suggested.

collectors/jiraffe/convertors.py
```diff
@@ -16,6 +17,50 @@
...
+        save the tracker with its context to DB
+        """
+        # wrap this in an atomic transaction so that
+        # we don't query this tracker during the process
+        with transaction.atomic():
+            # re-create all affect links
+            # in case some where removed
+            self.tracker.affects.clear()
+            self.tracker.affects.add(*self.affects)  # bulk add
+            self.tracker.save(
```
jsvob, line 54: Suggesting to add a comment:
```
# NOTE: When enabling independent read & write DB replicas, this might break, because the just-written values will be read before they can be propagated. This will again cause something like OSIDB-2487.
```

jsvob (approved):
LGTM

Tested only locally so far due to complications detailed in OSIDB-2487 and OSIDB-2576.

Moreover, the testing is limited due to the fact that I wasn't able to download enough flaws to make jira_tracker_collector reach the current time and collect the recent trackers, nor to at least reach an older flaw mentioned in a [comment in OSIDB-2487](https://issues.redhat.com/browse/OSIDB-2487?focusedId=24618565&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24618565)

```
from collectors.bzimport.task import flaw_collector
flaw_collector()
from collectors.jiraffe.tasks import jira_tracker_collector
for i in range(12): jira_tracker_collector()
# The newest tracker is sometime at the beginning of 2024
# This is the state in which I performed the testing described below

flaw_collector()
flaw_collector()
flaw_collector()
# repeat a few times
flaw_collector() # now it takes really long and downloads more and more flaws... the work day is over
```

When a flaw, affects and trackers were created through the local instance, it behaved correctly and didn't lose data over about 12 hours.

When the local instance was wiped, `from collectors.ps_constants.tasks import ps_constants_collector ; ps_constants_collector() ; from collectors.product_definitions.tasks import product_definitions_collector; product_definitions_collector() ; from collectors.jiraffe.tasks import metadata_collector ; metadata_collector()` run in console and only `python3 manage.py syncflaws --verbosity=3 2029189` run in bash, it didn't sync the trackers. That seems understandable given the comment `the tracker collector should never outrun the flaw one since it creates the linkage which might then be missed`.

After syncing a few trackers manually...

```
from collectors.jiraffe.collectors import JiraTrackerCollector; jc=JiraTrackerCollector(); jc.collect("OCPBUGS-29606") ; jc.collect("OCPBUGS-29606") ; jc.collect("OCPBUGS-29607") ; jc.collect("OCPBUGS-29608") ; jc.collect("OCPBUGS-29609") ; jc.collect("OCPBUGS-29610")
```

...OSIDB started listing these (and only these) trackers in `curl -H "Authorization: Bearer $token" ${SERVICE_URL}/osidb/api/v1/flaws/CVE-2384-0216 | python3 -mjson.tool | less`.

Subsequent edit of the flaw caused the yet-unsynced trackers to be deleted from srtnotes. This is the same behavior as before PR504.

After manually syncing yet another tracker...

```
from collectors.jiraffe.collectors import JiraTrackerCollector; jc=JiraTrackerCollector(); jc.collect("OCPBUGS-29611") 
```

...OSIDB started listing that tracker as well in `api/v1/flaws/CVE-2384-0216` and a subsequent edit of the flaw **added** the tracker back to srtnotes. **This is a massive improvement in PR504**. The new tracker linking mechanism seems to work very well here. Not only fetching a disconnected tracker reconnects it immediately from API perspective, it also repairs srtnotes on next flaw edit.

I hope that this improvement, along with the elimination of parallel sync (which may or may not hide potentially misunderstood breakage), and the elimination of [this part of code that unlinked trackers whenever jira hit a rate limit](https://github.com/RedHatProductSecurity/osidb/blob/release-3.7.0/collectors/bzimport/collectors.py#L471) (which seems fairly often since jira lowered the rate limits), it will hopefully fix OSIDB-2487 in a technical debt-reducing and maintainable way.

Very nice, thank you, Ondřej!

collectors/bzimport/convertors.py
```diff
@@ -163,21 +163,63 @@ def bug(self):
...
+            affects.append(affect)
+        return affects
+
     def _normalize(self) -> dict:
         """
         raw data normalization
         """
         ps_module, ps_component = tracker_summary2module_component(self._raw["summary"])
+        ps_update_stream = tracker_parse_update_stream_component(self._raw["summary"])[
+            0
```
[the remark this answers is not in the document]
osoukup: It all seems to have the correct names and the formatting is just a result of **black**.

collectors/jiraffe/convertors.py
osoukup: I am not sure that it is the issue of this code. I would consider it as an issue of the replica logic/setup. The atomic transaction is declared and I do not think that Python/Django developer should care here for how it is done in the database/Django framework. Without the atomicity someone else might get inconsistent data.

jsvob: @JimFuller-RedHat FYI
@osoukup :+1:

PR #499, opened 2024-04-17
merge 3.7.0 to master

osoukup:
I am going the **cherry-pick** way as merging **release-3.7.0** to **master** does not seem easy.

PR #506, opened 2024-04-29
OSIDB-2585: backing out db routing until we deprecate BZ sync

JimFuller-RedHat:
Backed out using db routing which seems unstable with usage of BZsync ... also added the ability for gunicorn to reload on changes when using local dev environment.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -30,7 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mainly Flaw, Affect, Tracker (OSIDB-2065)
 - Add support for additional_fields in Jira BTS (OSIDB-696)
 - Add scripts/restore_pg.sh script for restoring sql dump
-- Add db routing taking advantage of write and read pg replicas
 
 ### Changed
 - Ignore hosts on VCR recording (OSIDB-1678)
--- a/config/__init__.py
+++ b/config/__init__.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import os
+
 import django
 
 # This will make sure the app is always imported when
@@ -10,3 +12,8 @@ from .celery import app as celery_app  # noqa: F401
 
 django.setup()
 __all__ = ["osidb"]
+
+
+def get_env() -> str:
+    # return "config.settings.env" -> ["config", "settings", "env"] -> "env"
+    return os.getenv("DJANGO_SETTINGS_MODULE", "").split("_")[-1]
--- a/config/settings_prod.py
+++ b/config/settings_prod.py
@@ -95,26 +95,7 @@ DATABASES = {
         },
         "CONN_MAX_AGE": 120,
     },
-    "read-replica-1": {
-        "NAME": get_env("OSIDB_DB_NAME", default="osidb"),
-        "USER": get_env("OSIDB_DB_USER"),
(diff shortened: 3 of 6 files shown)


docs/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minor change to enable perf tests to run in CI (OSIDB-2447)
 - Allow editing flaws without affects in NEW state (OSIDB-2452)
 - Fixed read replica to perform HTTP requests as atomic transactions (OSIDB-2585)
+- Removed db routing
```
osoukup, line 20: I am checking the changelog. The DB routing is on master said to be added in 3.7.0 but when I check out to 3.7.0 tag the item is not present in the CHANGELOG. So it is incorrectly added here under 3.7.0 and should under Unreleased instead. Then we would announce adding and removing it in a single release which does not make much sense. I suggest removing it from here and from 3.7.0 below together with the replica fix (hope this is all related) as part of this change basically reverting the CHANGELOG so the users get clearer information.

osoukup (approved):
LGTM but I would like to include the CHANGELOG modification as described.

jsvob (approved):
I'm not persuaded that the problem lies in bzsync, since the issues indicated at a different problem even before any bzsync/bzimport occurred: A write randomly appeared and disappeared when read back in the timespan of 0 to 40 milliseconds after the write. And possibly data written inside an open transaction could be read outside in a different thread.

I think that inside a request's transaction, any changes must not be visible outside of that transaction (i.e. tasks and other requests) until that request&transaction finish successfully (how django's ATOMIC_REQUESTS worked with a single database replica until now). If using replicas with ATOMIC_REQUESTS doesn't keep that behavior, we'd have a problem because the whole osidb architecture relies on that. If we can keep the old behavior of ATOMIC_REQUESTS with DB replicas, then I think deprecating or not deprecating BZ sync is fairly irrelevant.

JimFuller-RedHat:
> I'm not persuaded that the problem lies in bzsync, since the issues indicated at a different problem even before any 

I hear you - the problem is that we do not align db read/write with HTTP methods (either in collectors or sync command or rest endpoints) ... using larger scale techniques like replicas highlights misalignments. 

We will back out of db routing and wait for BZsync to deprecate ... then (as you mentioned) it should be much easier to align.

docs/CHANGELOG.md
jobselko: I think that Jim's change is correct. The release was made from the `merge-3.7.0-to-master` branch, see CHANGELOG in #499.

jobselko: @JimFuller-RedHat @osoukup Sorry for the merge interruption, but could you please check the above?

JimFuller-RedHat: as discussed with @osoukup - we are not adding and removing db routing in a single release - and we never released db routing in last release (was a mistake).

jsvob: Jitka is right. Yes, the line `Add db routing taking advantage of write and read pg replicas` was included in 3.7.0 already. And `git blame` and searching the commit f529baef in `git log -v --source  --log-size  --notes --graph --name-status --show-signature` shows that it's been added before 3.7.0.

jobselko: Ondrej is right in the end :). That change was not released, CHANGELOG in #499 is incorrect and Ondrej already fixed the release notes.

PR #507, opened 2024-04-29
Preserve existing JiraProjectField models when fetching from Jira breaks

jsvob:
Preserve existing JiraProjectField models when fetching from Jira breaks
Related to OSIDB-2576
Closes OSIDB-2583


osoukup (approved):
LGTM

PR #508, opened 2024-04-30
Fix incorrect ACLs for flaw drafts

jobselko:
This PR fixes group settings when reading data from BZ, so flaw drafts always have internal ACLs.

Fixes OSIDB-2263

### DIFF
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -76,30 +76,44 @@ class BugzillaGroupsConvertorMixin:
         """
         appropriate read LDAP groups
         """
-        if "security" not in self.bug.get("groups", []):
-            return settings.PUBLIC_READ_GROUPS
-
-        if not BZ_ENABLE_IMPORT_EMBARGOED:
-            raise self.FlawConvertorException(
-                f"Bug {self.bz_id} is embargoed but BZ_ENABLE_IMPORT_EMBARGOED is set to False"
-            )
-
-        return [settings.EMBARGO_READ_GROUP]
+        return self.get_group("read")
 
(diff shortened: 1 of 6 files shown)


collectors/bzimport/convertors.py
```diff
@@ -76,6 +76,9 @@ def groups_read(self):
         """
         appropriate read LDAP groups
         """
+        if self.bug.get("groups", []) == ["redhat"]:
```
osoukup, line 79: The whole branching logic should be re-thinked a bit more. Assume that for some reason the groups are `["redhat", "devel"]`. Then, we would endup with a public flaw in OSIDB. Despite encountering unexpected group we should not extend the visibility about Bugzilla.

I suggest to make the logic in a way that `security` group present will always make it embargoed, otherwise any other group present (including `redhat` but also any other unexpected group) will make it internal, no groups will make it public.

collectors/bzimport/convertors.py
```diff
@@ -91,6 +94,9 @@ def groups_write(self):
         """
         appropriate write LDAP groups
         """
+        if self.bug.get("groups", []) == ["redhat"]:
+            return [settings.INTERNAL_WRITE_GROUP]
```
osoukup, line 98: Same here. As the mapping from Bugzilla to `["public", "internal", "embargoed"]` is used on two places, I would place it into a separate method and just call it from both `groups_read` and `groups_write` with an appropriate substitution.

osoukup (changes requested):
The logic needs to be a bit improved.

osoukup (approved):
LGTM

PR #510, opened 2024-05-02
Fix workflow integration

costaconrado:
This PR fix integration logic problems in Workflows/Taskman.

This PR sets a default value for workflow state since we removed automatic updates.
This PR also change update logic to rely on the states Jira returns from update query.
This PR also removes tests that rely only on automatic change of state and add manual update in tests that need it.

Partially closes OSIDB-2216.
Closes OSIDB-2456.


osidb/tests/test_mixins.py
```diff
@@ -479,3 +481,196 @@ def test_tracking_mixin_manager_auto_timestamps(self):
...
+        assert flaw.bz_id
+        assert flaw.task_key
+
+        AffectFactory(flaw=flaw, ps_module="ps-module-0")
+        flaw = Flaw.objects.get(pk=flaw.uuid)
+
+        flaw.promote(jira_token=jira_token)
+        assert flaw.workflow_state == WorkflowModel.WorkflowState.TRIAGE
+
+        from apps.taskman.service import JiraTaskmanQuerier
```
osoukup, line 596: This import is a duplicate - there is already one at the top.

osoukup (approved):
LGTM. The integration tests are great! To the thoughts on adding that extra validation count test, yeah. The more tests the better. I do not think there should ever be a relevant case when the validations should run multiple times during a single save.

PR #511, opened 2024-05-02
Bulk Affect update endpoint

jsvob:
Closes OSIDB-2407

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -14,7 +14,7 @@ from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
 from packageurl import PackageURL
-from rest_framework.decorators import api_view, permission_classes
+from rest_framework.decorators import action, api_view, permission_classes
 from rest_framework.permissions import AllowAny, IsAuthenticatedOrReadOnly
 from rest_framework.request import Request
 from rest_framework.response import Response
@@ -38,6 +38,7 @@ from .filters import (
 from .mixins import Alert
 from .models import Affect, AffectCVSS, Flaw, Tracker
 from .serializer import (
+    AffectBulkPutResponseSerializer,
     AffectCVSSPostSerializer,
     AffectCVSSPutSerializer,
     AffectCVSSSerializer,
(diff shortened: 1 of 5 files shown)


osidb/api_views.py
```diff
@@ -680,6 +681,49 @@ class FlawPackageVersionView(
     permission_classes = [IsAuthenticatedOrReadOnly]
 
 
+def action_at_empty_prefix():
```
jsvob, line 684: Hi @JakubFrejlach any idea, please, whether there's a nicer way to generate valid openapi.yml for a bulk PUT handler at `^affects/$` while the usual PUT at `^affects/(?P<pk>[^/.]+)$` is still there? Without this magic and `@action`, drf-spectacular ignores the custom `put` method and only sees the usual single-instance PUT at `^affects/(?P<pk>[^/.]+)$` .

JakubFrejlach: @JakubFrejlach I've went through your research and indeed this is a very delicate corner case :smile: Unfortunately I don't think there will be a better solution. The only thing which comes to my mind and would probably worth to consider is to just go with some name for the bulk endpoint and avoid using `action_at_empty_prefix`.

Eg. having:
`PUT /affects/bulk`

and later probably also `DELETE /affects/bulk` and `POST /affects/bulk` to really emphasize that these are the bulk endpoints. I don't see a problem with referencing bulk endpoints with this extra keyword in URL. Wdyt?

osidb/api_views.py
```diff
@@ -738,6 +739,80 @@ class AffectView(SubFlawViewDestroyMixin, ModelViewSet):
...
+    def put(self, request, *args, **kwargs):
+        """
+        Bulk update endpoint. Expects a list of dict Affect objects.
+        """
+
+        bz_api_key = request.META.get("HTTP_BUGZILLA_API_KEY")
+        if not bz_api_key:
+            raise ValidationError({"Bugzilla-Api-Key": "This HTTP header is required."})
+
+        # TODO sometime: Some of these actions probably belong to another layer, perhaps serializer.
```
jobselko, line 756: I agree with moving this logic to a serializer. See https://www.django-rest-framework.org/api-guide/serializers/#customizing-listserializer-behavior

jobselko: I will check the rest later, as I do not have enough time today. My apologies.

jobselko: I will not review/block this PR since it was decided to stick with the current version at today's meeting.

osidb/api_views.py
osoukup: I agree.

PR #512, opened 2024-05-06
Ensure that snippet does not contain duplicate references

jobselko:
This PR ensures that snippet's external references never contain a URL that is also in a source reference. If this happens, `ValidationError` is raised when saving `FlawReference` with the same URL for the second time (for a given `Flaw`).

For example, see https://nvd.nist.gov/vuln/detail/CVE-2023-48733, where a source reference is already included in a list that we treat as a list of external references.


osoukup (approved):
LGTM

PR #514, opened 2024-05-07
Update ubi8 base image version

dmonzonis:
There have been reports of vulnerabilities in UBI version 8.7. This commit updates it to the latest major version, 8.9, where the vulnerabilities should have been patched, in the Dockerfile.


JimFuller-RedHat (approved):
LGTM

PR #515, opened 2024-05-09
add uat as option for api responses

costaconrado:
This PR adds uat as option for API responses instead of default to local.

This also PR creates a separated file for uat access logs.

### DIFF
--- a/scripts/run-service.sh
+++ b/scripts/run-service.sh
@@ -14,6 +14,8 @@ ACCESS_LOG_FILE="-"
 
 if [ "$DJANGO_SETTINGS_MODULE" = "config.settings_stage" ]; then
     ACCESS_LOG_FILE="/var/log/stage-access.log"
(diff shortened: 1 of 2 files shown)


scripts/run-service.sh
```diff
@@ -14,6 +14,8 @@ ACCESS_LOG_FILE="-"
 
 if [ "$DJANGO_SETTINGS_MODULE" = "config.settings_stage" ]; then
     ACCESS_LOG_FILE="/var/log/stage-access.log"
+elif [ "$DJANGO_SETTINGS_MODULE" = "config.settings_uat" ]; then
+    ACCESS_LOG_FILE="/var/log/uat-access.log"
```
jsvob, line 18: Is this file something we get access to automatically?

jsvob (approved):
LGTM but I'm not familiar with this matter

scripts/run-service.sh
costaconrado: no, we need to open a ticket with splunk team (I've already opened it)

PR #516, opened 2024-05-13
Fix flaw instance not being up to date after sync

dmonzonis:
After syncing a flaw with Bugzilla upon creating it from the API, the
flaw instance was not updated and if the Jira sync was enabled, a save
within its task sync would override the data and the Bugzilla data would
be lost. In particular, the `meta_attr` field would not be updated, even
if the Bugzilla bug is actually created. This caused errors when trying
to update the data from the API.

With this change, after the Bugzilla sync, the instance is updated so
that we are working with the latest data, and other mixins don't clash
with it.

Closes OSIDB-2628.

### DIFF
--- a/apps/bbsync/tests/test_integration.py
+++ b/apps/bbsync/tests/test_integration.py
@@ -444,7 +444,12 @@ class TestBBSyncIntegration:
         )
         ps_module1 = PsModuleFactory(
             bts_name="bugzilla",
-            bts_groups={"public": ["devel"]},
+            bts_groups={
+                "public": ["devel"],
+                "embargoed": [
+                    "private",
+                ],
+            },
             bts_key="Red Hat Certification Program",
             name="rhcertification-8",
             default_component="redhat-certification",
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1518,6 +1518,8 @@ class Flaw(
         # fetch from Bugzilla
         fc = FlawCollector()
         fc.sync_flaw(self.bz_id)
+        # Make sure the flaw instance has the latest data
+        self.refresh_from_db()
 
     def tasksync(
         self,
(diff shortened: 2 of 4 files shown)


osidb/models.py
```diff
@@ -1518,6 +1518,8 @@ def bzsync(self, *args, bz_api_key, **kwargs):
         # fetch from Bugzilla
         fc = FlawCollector()
         fc.sync_flaw(self.bz_id)
+        # Make sure the flaw instance has the latest data
+        self.refresh_from_db()
```
osoukup, line 1522: I think that recently with @jobselko we dealt with something in Assembler collectors and this would have probably fixed it in a better way. I am not sure what exactly it was but something with difference between runtime and DB after the Bugzilla sync. So just FYI @jobselko

osoukup (approved):
LGTM. I should have added the sync from DB when I introduced this to prevent issues like this. Thank you for fixing my mess :smile:

osidb/models.py
jobselko: Yeah, a "runtime" `flaw` with uuid `1` had different data than the same flaw retrieved by `Flaw.objects.get(uuid=1)` :).

apps/bbsync/tests/test_integration.py
```diff
@@ -444,7 +444,12 @@ def test_flaw_unembargo_complex(
         )
         ps_module1 = PsModuleFactory(
             bts_name="bugzilla",
-            bts_groups={"public": ["devel"]},
+            bts_groups={
+                "public": ["devel"],
+                "embargoed": [
+                    "private",
```
jobselko, line 450: Why `"embargoed": ["private"]` is necessary now?

jobselko (approved):
LGTM, great catch!

apps/bbsync/tests/test_integration.py
dmonzonis: An "embargoed" group seems to be necessary when a flaw is embargoed and synced with bugzilla, as per this line which otherwise causes a `KeyError` https://github.com/RedHatProductSecurity/osidb/blob/90f7256b55a60606a134a157cae50835bf8e1760/apps/trackers/bugzilla/query.py#L221

This test started failing with this commit's modification, probably because this inconsistency wasn't caught before due to the same error solved by it.

PR #517, opened 2024-05-14
Ignore SLA if update stream specifies it's not applicable

dmonzonis:
The rhsa_sla_applicable field in update streams specifies whether SLA is applicable or not, and in case it's not, SLA policies should not be applied to the tracker related to the update stream. Before this commit, SLA policies were applied regardless of the value of this field. With this commit, this field is properly used to determine whether SLA policies apply or not.


osidb/tests/factories.py
```diff
@@ -495,7 +495,6 @@ class Meta:
     name = factory.sequence(lambda n: f"ps_update_stream_{n}")
     version = factory.Faker("word")
     target_release = factory.Faker("word")
-    rhsa_sla_applicable = factory.Faker("boolean")
```
osoukup, line 498: Why is this being removed?

dmonzonis: Faker will produce random boolean values, but for most tests where `rhsa_sla_applicable` is not specified, we want to use the default which is `True`. Otherwise it will get `rhsa_sla_applicable = False` for some tests and with the change introduced in this commit, the SLA policy will not be generated and the test will fail.

osoukup: I did not realize that there is the default. Makes sense!

osoukup (approved):
LGTM

PR #518, opened 2024-05-14
Do not try to generate custom text for triage trackers

jsvob:
Do not try to generate custom text for triage trackers

Functionally, there's no charge because `Tracker.is_triage` is always `False`. This is just cleanup and future-proofing to meet the goal of OSIDB-2494 both now and in the future where `Tracker.is_triage` might be non-False. Hence `technical` label, no changelog.

Closes OSIDB-2494

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -130,9 +130,6 @@ class TrackerQueryBuilder:
         ):
             prefixes.append("[CISA Major Incident]")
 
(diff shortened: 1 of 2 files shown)


osoukup (approved):
LGTM

jsvob:
Left the tests in, just adjusted, so that if we need to reintroduce changes in triage texts, we have easier work with tests.

PR #519, opened 2024-05-15
Update actions in GH workflows

jobselko:
This PR updates GH workflows to use the latest versions of actions. The current actions use EOL Node.js and will be deprecated.


osoukup (approved):
LGTM

PR #521, opened 2024-05-15
release 3.7.1 prep

osoukup:
Patch release to fix the collector issues.


jsvob (approved):
LGTM, but don't take me seriously.

.secrets.baseline
```diff
@@ -162,7 +162,11 @@
         "filename": "apps/bbsync/tests/test_integration.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
+<<<<<<< HEAD
         "line_number": 70,
+=======
+        "line_number": 80,
+>>>>>>> 7943700 (keep generating BZ ID Jira tracker label)
```
costaconrado: this merge conflict will break the secrets.

.secrets.baseline
```diff
@@ -162,11 +162,7 @@
         "filename": "apps/bbsync/tests/test_integration.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
-<<<<<<< HEAD
-        "line_number": 70,
-=======
-        "line_number": 80,
->>>>>>> 7943700 (keep generating BZ ID Jira tracker label)
```
costaconrado: nitpick: I believe this change belongs to the last commit.

apps/trackers/tests/test_integration.py
```diff
@@ -26,153 +25,146 @@
...
-#     # 7) check that the update actually happened
-#     assert "updated_dt" in loaded_tracker.meta_attr
-#     assert updated_dt != loaded_tracker.meta_attr["updated_dt"]
+class TestTrackerSaver:
+    @pytest.mark.vcr
+    def test_tracker_create_bugzilla(self):
+        """
+        test basic Bugzilla tracker creation
+        """
+        # 1) define all the context
```
costaconrado: Now I see why you commented this code in "remove tracker sync from flaw sync".

I would rather remove there and add here for cleaner/easier "reverse cherry-picking" if needed in future but I believe this is fine, you can freely ignore both this and the past comment about this test.

costaconrado (approved):
LGTM.
I left few nitpicks but since most of this is already merged on master and will be used for deployment purpose, feel free to ignore them.

.secrets.baseline
osoukup: Oh :smile: I had to screw the rebasing

.secrets.baseline
osoukup: The final version is OK. I can fix this in the history.

collectors/bzimport/collectors.py
```diff
@@ -553,8 +553,9 @@ def collect_flaw(self, flaw_id, total_retries=0):
 
 class BugzillaTrackerCollector(Collector):
 
-    # version 0.0.1 of OSIDB was released on January 21st 2022
-    BEGINNING = timezone.datetime(2022, 1, 21, tzinfo=timezone.get_current_timezone())
+    # according to the Bugzilla advanced search the longest non-updated
+    # security trackers going chronologically were last updated in 2010
```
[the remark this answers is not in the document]
osoukup: I have to agree :smile: the actual rephrasing makes sense on **master** however so I will just leave it as it is here

apps/trackers/tests/test_integration.py
```diff
@@ -7,6 +7,7 @@
 from rest_framework import status
 
 from apps.trackers.jira.query import JiraPriority
+
```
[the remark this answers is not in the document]
osoukup: I probably mix them during the rebase. Never mind, on **master** it should be OK

apps/trackers/tests/test_integration.py
```diff
@@ -7,7 +7,7 @@
 from rest_framework import status
 
 from apps.trackers.jira.query import JiraPriority
-from apps.trackers.save import TrackerSaver
+# from apps.trackers.save import TrackerSaver
```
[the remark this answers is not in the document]
osoukup: I left it there so the reviewer can follow the implementation process or at least I though it will be better to understand :smile: no I see it is maybe rather confusing. It is already on **master** so I will just keep it the same way here

PR #522, opened 2024-05-16
Filter by null or empty CVE IDs in the API

dmonzonis:
With this commit, there is now the possibility to filter, at the API level, for null or empty CVE IDs.

To do this, one may use the query parameter cve_id__isempty and set it to 1 (show only flaws with null or empty CVE IDs) or 0 (show flaws with CVE IDs that are neither null nor empty).

The change has been done in a way that it could be reused by other nullable string fields in the future, if needed.

Closes OSIDB-2625.


osoukup (approved):
LGTM. Good job!

JakubFrejlach (approved):
LGTM!

PR #530, opened 2024-05-21
Fix FlawReference article count validation

costaconrado:
This PR changes article validation to consider creation scenario.

This PR also remove `.save()` from the validation test, this would make the test pass since it saves the model twice.

Closes OSIDB-2651.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -3290,14 +3290,23 @@ class FlawReference(ACLMixin, BugzillaSyncMixin, TrackingMixin):
         """
         Checks that a flaw has maximally one article link.
         """
+        old_reference = FlawReference.objects.filter(uuid=self.uuid).first()
+        article_count = 0
+        if self.type == FlawReference.FlawReferenceType.ARTICLE:
+            if (
+                not old_reference
+                or old_reference.type != FlawReference.FlawReferenceType.ARTICLE
+            ):
+                article_count = 1
+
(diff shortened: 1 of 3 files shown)


costaconrado:
The reason it works in STAGE and not in UAT is the `BBSYNC_SYNC_TO_BZ`, due the current state of mixins this variable makes this entity save more than one time so the creation scenario could be ignored since validations will run twice.

osidb/models.py
```diff
@@ -3293,11 +3293,11 @@ def _validate_article_links_count_via_flawreferences(self):
...
         )
-
-        if article_links.count() > 1:
+        article_count = 1 if self._state.adding else 0
+        article_count += article_links.count()
+        if article_count > 1:
             raise ValidationError(
-                f"A flaw has {article_links.count()} article links, "
-                f"but only 1 is allowed."
+                f"A flaw has {article_count} article links, " f"but only 1 is allowed."
```
jobselko, line 3300: f"A flaw has {article_count} article links, but only 1 is allowed."

jobselko (approved):

osidb/models.py
costaconrado: nice catch, thanks!

osoukup (approved):
LGTM

jobselko (approved):

PR #532, opened 2024-05-22
OSIDB-2734 add JIRA stage http forwarder

JimFuller-RedHat:
Provide http forwarder for accessing JIRA stage.
```
http://localhost:8000/osidb/api/v1/jira_stage_forwarder?path=/rest/api/latest/version
```
where the **path** url param defines JIRA rest api path.

**Note**: this is an ugly hack and we should remove at some point.


osoukup (approved):
Why should this be a part of OSIDB? If this is really needed which I am not convinced about, I would be for making it an app module disabled in any environment except the local

osoukup (changes requested):
I accidentally approved while I wanted to continue discussing

JimFuller-RedHat:
completely agree - I rather we document away the problem (eg. stage/uat users would have to set proxy in their browsers) then use this abomination - I offered up this PR just in case (for some reason) that is not an option.

JimFuller-RedHat:
@osoukup mentioned in chat we should apply logic so this only enables in stage/uat environments ... we might also hive off to a standalone app

osoukup (approved):
Still quite a hack but I am OK with it on non-production environments

PR #535, opened 2024-05-27
Remove FlawHistory

jobselko:
This PR removes the `FlawHistory` model and related code. There is no data to keep as we have never used that model.

Closes OSIDB-2742


jobselko:
I am unsure about the `technical` label. Any preferences?

osoukup:
> I am unsure about the `technical` label. Any preferences?

This should be no difference for the user and not much difference at all (mostly a cleanup) so makes sense to me it marked as a change of purely technical nature.

PR #536, opened 2024-05-27
OSIDB-2743: Remove  Affect type

MrMarble:
Removed the `AffectType` enum and the `type` field from the `Affect` model as it only has one possible value and is not needed at all.

Closes OSIDB-2743


osoukup (approved):
LGTM

JakubFrejlach (changes requested):
I would update the `CHANGELOG.md` this is field removal and it directly influences API so it is something which potential user would like to see in the changelog. Otherwise LGTM!

JakubFrejlach (approved):

dmonzonis (approved):

osidb/migrations/0129_remove_affect_type.py
jobselko, line 1: There already is migration `0129_delete_flawhistory.py`, so you have to create a new one.

MrMarble: Oh yeah, I did a rebase because of that and totally forgot to rename the migration file, changed

PR #542, opened 2024-05-28
Add reported_dt to flaw drafts

jobselko:
This PR adds the `reported_dt` field to flaws created by collectors. `reported_dt` is set according to `created_dt`.

Fixes OSIDB-2771


dmonzonis (approved):
LGTM, but why were we allowing these flaws to be created to begin with? Is there a strong reason why we skip validations [here](https://github.com/RedHatProductSecurity/osidb/pull/542/files#diff-8d7ea87df8ca210c5df403a76b8228d0d1e1d2ae7fc99f54b5c6f7e02d396deeR1595)?

jobselko:
@dmonzonis These flaws serve as drafts for the VM team and are intended for triage. They often contain incomplete data, but we cannot require all validations to pass because we want to preserve every flaw.

PR #543, opened 2024-05-29
OSIDB-2736: Remove flaw state field

MrMarble:
The attribute `Flaw.state` holds the value of the Bugzilla state which is no more relevant in the future workflow and is at the same time confusing often being considered as THE state which it is not.

The field was marked as deprecated in #171


JakubFrejlach (approved):
LGTM!

dmonzonis (commented):
This introduces a breaking change in the API. I think we should deprecate the field instead of outright removing it. We should be extra careful with stuff like this especially now that we have several environments live and are close to going to production, as it is hard to know with 100% certainty that a breaking change will not affect any users of the API.

MrMarble:
> This introduces a breaking change in the API. I think we should deprecate the field instead of outright removing it. We should be extra careful with stuff like this especially now that we have several environments live and are close to going to production, as it is hard to know with 100% certainty that a breaking change will not affect any users of the API.

Hi Daniel! I don't fully understand your comment, this field was already deprecated (see #171 and https://github.com/RedHatProductSecurity/osidb/pull/543/files#diff-8d7ea87df8ca210c5df403a76b8228d0d1e1d2ae7fc99f54b5c6f7e02d396deeL635-L642) but maybe I don't have the full context so @JakubFrejlach can jump in and clarify things. 
This change is aimed to the 4.0 release of OSIDB as well

dmonzonis:
> > This introduces a breaking change in the API. I think we should deprecate the field instead of outright removing it. We should be extra careful with stuff like this especially now that we have several environments live and are close to going to production, as it is hard to know with 100% certainty that a breaking change will not affect any users of the API.
> 
> Hi Daniel! I don't fully understand your comment, this field was already deprecated (see #171 and https://github.com/RedHatProductSecurity/osidb/pull/543/files#diff-8d7ea87df8ca210c5df403a76b8228d0d1e1d2ae7fc99f54b5c6f7e02d396deeL635-L642) but maybe I don't have the full context so @JakubFrejlach can jump in and clarify things. This change is aimed to the 4.0 release of OSIDB as well

I didn't have the full context either. If this field was already deprecated and the change will be released with a major version, then my concerns have been addressed.

dmonzonis (approved):

osoukup:
@dmonzonis your point on deprecation is totally relevant and we should make sure that this change does not go to eg. UAT before OSIM is ready. On the other hand releasing this to the stage is I think desired so the data consumers can test their adjustions to the change. This is part of the major release preparation and from the very beginning I am in close communication with SDEngine and OSIM and they have plan to account for the change plus I check Konflux and CPaaS usage of our data to confirm that they are no impacted - these are all the data consumers we know about.

PR #544, opened 2024-05-29
OSIDB-2734 refine /osidb/api/v1/jira_stage_forwarder?path= 

JimFuller-RedHat:
So we remove json boilerplate just for this endpoint - and pass thru headers/url params.


osoukup (approved):
LGTM

PR #545, opened 2024-05-29
Make flaw comments independent of Bugzilla

dmonzonis:
This commit redesigns the FlawComment model as well as the API to make them independent of Bugzilla sync. Before this commit, the comment info was being stored in meta_attr, which is filled during the sync with Bugzilla. Some info was also shared in the base model (text) but wasn't being sent through the API, while other info was completely lost (creator, is_private, time).

With this commit, the info is stored in the model, and during sync it is updated. The info in the model is also serialized and sent via the API.

Closes OSIDB-2760.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2943,6 +2943,14 @@ class FlawComment(
     # text of the comment
     text = models.TextField()
 
+    # creator of the comment, which can be passed as an argument when creating it,
+    # similar to the flaw's owner field, or if BZ sync is enabled, then it will be
+    # implied from the BZ API key owner during sync
+    creator = models.CharField(max_length=100, blank=True)
+
+    # whether the comment is internal or not
+    is_private = models.BooleanField(default=False)
+
     # comment meta data
     meta_attr = HStoreField(default=dict)
 
--- a/collectors/bzimport/convertors.py
+++ b/collectors/bzimport/convertors.py
@@ -843,6 +843,8 @@ class FlawConvertor(BugzillaGroupsConvertorMixin):
                 ),
                 order=comment["count"],
                 text=comment["text"],
+                creator=comment["creator"],
+                is_private=comment.get("is_private", False),
                 type=FlawComment.FlawCommentType.BUGZILLA,
                 acl_read=self.acl_read,
                 acl_write=self.acl_write,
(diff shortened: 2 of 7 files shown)


osidb/models.py
```diff
@@ -2988,6 +2996,14 @@ class Meta:
             GinIndex(fields=["acl_read"]),
         ]
 
+    @property
+    def time(self):
+        """
+        Return the time of the BZ comment if sync is enabled, or the
+        record's creation datetime otherwise.
+        """
+        return self.meta_attr.get("time", self.created_dt)
```
osoukup, line 3005: I am thinking why is this necessary. Wouldn't it be better to just set `created_dt` to the Bugzilla `time` value in the `FlawConvertor` if fetched from Bugzilla and just leave it to be automatically set by the `TrackingMixin` otherwise? When/why should they be different?

osoukup (changes requested):
Mostly LGTM but I raised a few points to answer/address.

osidb/models.py
dmonzonis: I did that because `meta_attr.time` is what OSIM checks to make the transition more clear, but thinking it twice, I think you're right since we shouldn't bend OSIDB to adhere to OSIM standards, but the other way around. So I will definitely change that.

osidb/migrations/0132_flawcomment_redesign.py
```diff
@@ -0,0 +1,48 @@
...
+    set_user_acls(settings.ALL_GROUPS)
+    FlawComment = apps.get_model("osidb", "FlawComment")
+
+    # Move existing comment data from meta_attr to the model
+    comments = FlawComment.objects.all().iterator(chunk_size=BATCH_SIZE)
+
+    batch = []
+    for i, comment in enumerate(comments, 1):
+        comment.creator = comment.meta_attr.get("creator", "")
+        # is_private seems to be stored as a string in old data
```
JakubFrejlach, line 20: Note: This is not entirely true. `meta_attr` is hstore field type and such field can only hold key:values records where each value is a string, meta_attr simply cannot hold any other type. I believe in old data, there is a boolean in place.

JakubFrejlach (approved):
LGTM

osidb/migrations/0132_flawcomment_redesign.py
dmonzonis: I see. I will check for both types then just in case. Inconsistencies like this are yet another good reason to make the switch.

osoukup (approved):
LGTM

PR #547, opened 2024-05-30
Fix not being able to set CVE to empty string through API

dmonzonis:
The `cve_id` field allows both empty and null values, but through the API it was rejecting an empty string with an error, because the serializer did not accept empty strings.

To make it consistent with the model, the serializer has been modified to allow empty strings as well.

Closes OSIDB-2702.


jobselko (approved):

osoukup (approved):
LGTM

MrMarble (approved):
LGTM

PR #548, opened 2024-05-30
Remove several cvss fields from Flaw

jobselko:
This PR updates the NVD collector to use only the `FlawCVSS` model for cvss. It also removes the following cvss fields from the `Flaw` model: `cvss2`, `cvss2_score`, `cvss3`, `cvss3_score`, `nvd_cvss2` and `nvd_cvss3` (deprecated in #335).

Related to OSIDB-2749


osoukup (approved):
LGTM

PR #557, opened 2024-06-04
Remove is_major_incident field from Flaw

jobselko:
This PR removes the `is_major_incident` field from the `Flaw` model (deprecated in #267). It also removes its replacement `is_major_incident_temp` and reworks SLA.

Closes OSIDB-2741


osidb/models.py
```diff
@@ -642,8 +642,8 @@ class FlawRequiresSummary(models.TextChoices):
 
     def is_major_incident_temp(self):
         """
-        This function mirrors behaviour of the is_major_incident field.
-        Once the original field is removed, this should be turned into property.
+        This function mirrors behaviour of the old "is_major_incident" field.
+        All of its usage should be checked and replaced by "major_incident_state" field.
```
jobselko, line 646: I will create Jira for this.

osoukup (commented):
Could the `is_major_incident_temp` be addressed here too? It does not seem to be necessary any more so why do we need to split it out to a new task?

jobselko:
@osoukup I think this is not so straightforward and it could require a discussion, so I did not want to mix it with a field removal.

jobselko:
Ok, I will try :)

osoukup:
@jobselko let us discuss whatever you see not so straightforward :wink:

apps/workflows/tests/test_models.py
```diff
@@ -660,7 +660,7 @@ def test_adjust(self):
                 "description": "random description",
                 "priority": 1,  # is more prior than default one
                 "conditions": [
-                    "is_major_incident_temp"
+                    "major_incident_is_APPROVED"
```
jobselko, line 663: @osoukup @costaconrado I am not sure what value is expected here and why there are so many failing tests. I would be glad if we could discuss it via call/slack.

osoukup: You need to use `major_incident_state` instead. I think that the correct value might be something like `major_incident_state_is_approved` (the enum value might even work in uppercase).

jobselko: I also tried `major_incident_state_is_approved`, but it did not help me. I had to update SLA policy...

apps/sla/sla.yml
```diff
@@ -24,13 +24,12 @@
...
 description: >
-  SLA policy applied to requested or approved
-  Major Incident or CISA Major Incident
+  SLA policy applied only to approved Major Incident
 conditions:
   affect:
     - is not community
   flaw:
-    - is major incident
+    - major incident state is approved
```
osoukup, line 32: So this change is unfortunately 1:1 to the original. The similar policy should apply for the CISA MI and also for requested but not yet approved MI - at least without we have an approval that it should be otherwise as in the old tooling it was this way for and intentionally. The right solution could be to create separate policy for each of the three cases with all the same except this condition.

osoukup (changes requested):
The SLA policies has to be fixed. Also this
https://github.com/RedHatProductSecurity/osidb/blob/master/apps/workflows/checks.py#L28
should be removed

apps/workflows/tests/test_models.py
```diff
@@ -54,9 +54,9 @@ def test_function_value(self):
         """
         test if a function without parameters can be used as condition can be checked
         """
-        check = Check("is_major_incident_temp")
+        check = Check("major_incident_state")
         # alias
-        check_alias = Check("is major incident")
+        check_alias = Check("major incident state")
```
jobselko, line 59: @costaconrado Can I completely remove `test_function_value()` as there is no `is_major_incident_temp`?

osoukup (approved):
LGTM

osidb/models.py
```diff
@@ -640,22 +640,6 @@ class FlawRequiresSummary(models.TextChoices):
     # mitigation to apply if the final fix is not available, from srtnotes "mitigation"
     mitigation = models.TextField(blank=True)
 
-    def is_major_incident_temp(self):
```
dmonzonis, line 643: Would it make sense to keep this, make it a property, rename it to `is_major_incident` (replacing the old field) so that we can use it in other parts of code instead of having to type
```py
if flaw.major_incident_state in [
    Flaw.FlawMajorIncident,
]
```
each time?

apps/workflows/tests/test_models.py
costaconrado: its fine to delete this test. 
Now that is_major_incident_temp don't exists we really don't have a good function in Flaw model to test this, so its fine to remove this test.

costaconrado (approved):
LGTM.

osidb/models.py
jobselko: I discussed it with @osoukup. The property was too ambiguous (it also contained the incorrect `REQUESTED` state) so we decided to get rid of it.

osoukup: Yeah. When someone sees `is_major_incident` the usual understanding is to assume it means `APPROVED` which was not true and therefore quite confusing. The property was brought from the old tooling where it was some kind of shortcut but going forward we should start distinguishing which type of MI it is and whether only requested or already approved.

PR #558, opened 2024-06-04
Set target release in Jira trackers

dmonzonis:
Using the data from the PsUpdateStream, set the Target Release field (or if not available, use the Target Version field as fallback) when creating/updating Jira trackers. This info is available in the `target_release` field in PsUpdateStream.

Closes OSIDB-2727.


osoukup (approved):
LGTM

PR #559, opened 2024-06-04
OSIDB-2734 refine stage jira http forwarder

JimFuller-RedHat:
Require stage **jira-api-key** (via HTTP header) - which removes the need for using osidb-stage own jira token ... and will now throw HTTP 400 error if missing Jira-Api-Key header.


osoukup (approved):
This is much better! I did not realize before that we are actually using the service API key. Using the user one is much better from the audit point of view

PR #560, opened 2024-06-04
Implement is_triage in trackers and use it in SLA policies

dmonzonis:
This commit implements the `is_triage` property in trackers. A tracker is considered in triage when all the linked affects have a NEW affectedness state. This property is now used in all SLA policies, which will not be accepted if the tracker is in triage.

Closes OSIDB-2729.


dmonzonis:
Closing this, since after the discussion in Jira, this feature is not wanted anymore. Instead, we will remove the "triage" process completely.

PR #561, opened 2024-06-04
OSIDB-2747: Remove meta_attr from FlawComment

MrMarble:
~I've seen that this field seems to be used by OSIM in https://github.com/RedHatProductSecurity/osim/blob/main/src/components/FlawComments.vue#L43, so instead of removing it it should be set to read_only~

Finally decided to remove the field entirely 

Closes OSIDB-2747


osoukup (commented):
What do they use it for? I have a feeling that it may be a result of some previous bad comment model design and we should rather change the OSIM to use the new model fields (@dmonzonis recently reworked the model to have real fields instead of meta attributes). There should not be any value any more which is important and is not a real model field.

MrMarble:
> What do they use it for? I have a feeling that it may be a result of some previous bad comment model design and we should rather change the OSIM to use the new model fields (@dmonzonis recently reworked the model to have real fields instead of meta attributes). There should not be any value any more which is important and is not a real model field.

As far as I can see, the `is_private`, `creator`, `time` and `text` attributes of `meta_attr` are being used, but they all seem to exist outside of `meta_attr`.

osoukup:
> > What do they use it for? I have a feeling that it may be a result of some previous bad comment model design and we should rather change the OSIM to use the new model fields (@dmonzonis recently reworked the model to have real fields instead of meta attributes). There should not be any value any more which is important and is not a real model field.
> 
> As far as I can see, the `is_private`, `creator`, `time` and `text` attributes of `meta_attr` are being used, but they all seem to exist outside of `meta_attr`.
> 
> Should I remove it and update OSIM usage then?

So this is all used for the historical reasons and should be switched to the new attributes. OSIDB `meta_attr` removal still make sense to me. @JakubFrejlach it seems to me that this is not being tracked among the OSIM tasks in the major release epic however.

dmonzonis:
@MrMarble the usage of the new fields in OSIM is already being worked on in [OSIDB-2700](https://issues.redhat.com/browse/OSIDB-2700).

As to whether we can completely remove `meta_attr` from FlawComment or not, it still contains some data that is not currently being used in OSIDB and is just copied from BZ (e.g. `attachment_id` or `tags`). I'm all for removing the field if all of this is just boilerplate from BZ and is never going to be needed in the future.

MrMarble:
@dmonzonis I can't decide if those fields are going to be used in OSIDB or not, @osoukup or @JakubFrejlach are the ones that may know the answer, but whatever is decided I'm open to do the changes

osoukup:
@MrMarble @dmonzonis let us remove the `meta_attr` :wink: those last pieces of data there are not to be used and the useful once are already full-blown attributes

JakubFrejlach:
@osoukup @MrMarble @dmonzonis This confusion probably comes from some miscommunication. I've review all the OSIDB 4.0 prereqs and created counterpart tasks for OSIM, but regarding `meta_attr` there was not many communication so I simply didn't created any Jiras regarding this. We can remove it here and for OSIM I will file a new issue which will be about removing `meta_attr` usage

jobselko:
@osoukup @MrMarble Maybe we could keep this PR to only set `meta_attr` as `read_only` and leave the removal for another Jira/PR?

JimFuller-RedHat:
alternately I am going to ask why we are considering removing meta_attr at all ? Initially they served the purpose of stabilising the official logical model -  we could just simply not allow them to be accessible via REST API.

osoukup:
@JimFuller-RedHat this is a relevant point. For some model they definitely serve the purpose and will not be removed. For models like these they only hold the copy of the other attributes so no additional value. Removing them from the API is I think necessary as we keep encountering people using the meta attributes as regular attributes and then we have unnecessarily break someones' integrations. Leaving them in the model would be do no harm but also not bring any value at least at this moment. I would prefer cleaning up unused attributes at this point already. If we see any potential usage, we can discuss it further.

JimFuller-RedHat:
@osoukup it is debatable if we should have ever exposed **meta_attr** to REST API at all (and this solves the 'people using them problem') ... the idea of using meta_attr was to avoid directly loading entities with business logic which was a big problem in the past - granted I am mostly concerned about Flaw entity and these secondary entities are probably stable enough now to remove **meta_attr** but lets be mindful that the original purpose of **meta_attr** was to avoid overloading business logic and random meta data directly onto data entities ... we are always free to conflate such things at the django model level.

osoukup (approved):
LGTM and thank you for your patience here :smile:

JimFuller-RedHat:
why are we removing meta_attr from flaw ?

osidb/serializer.py
jobselko, line 1: Please remove `IncludeMetaAttrMixin` from `FlawCommentSerializer`

jobselko, line 1: Also remove `@include_meta_attr_extend_schema_view` from `FlawCommentView` in `api_views.py`

jobselko (changes requested):

jobselko (approved):

PR #563, opened 2024-06-04
support bz trackers for non bz flaws

osoukup:
This PR enhances the tracker creation and linking logic to work with non-Bugzilla flaws. To keep the links in the tracker data the `whiteboard` field originally used by ProdSec (but no more) was chosen. It is then used as another/fallback linking option during the sync from Bugzilla.

Closes OSIDB-2845


osoukup:
@costaconrado this is the Bugzilla counterpart, please review

costaconrado (approved):
Nice job! LOGTM.

PR #564, opened 2024-06-05
Remove is_triage property from Tracker

dmonzonis:
This property will not be needed anymore, and it was never properly implemented, so it can be safely removed.

Closes OSIDB-2853.


osoukup (approved):
Thank you for cleaning this up! LGTM

PR #569, opened 2024-06-06
BZ trackers to UAT

osoukup:
Fix to non-Bugzilla flaw tracker creation and linking plus NEW affect tracker suggestion.

Closes OSIDB-2808


osoukup:
@costaconrado so this should deliver the remaining tracker fixes to the UAT

PR #570, opened 2024-06-06
OSIDB-2734 more tweaks to /osidb/api/v1/jira_stage_forwarder?path=

JimFuller-RedHat:
apparantly stage jira absolutely needs a User-Agent for HTTP POST to be set appropriately and also fix my stupid mistakes so HTTP POST forwarding against stage jira actually works.


osoukup (approved):
LGTM

PR #566, opened 2024-06-05
OSIDB-2740: Rename Flaw summary and description

MrMarble:
It was identified that the naming OSIDB introduced in its ancient times is not appropriate and people are constantly confused by it.

We should rename `Flaw.description` which corresponds to the initial Bugzilla flaw bug comment to `comment_zero`. Additionally, we should rename `Flaw.summary` to `cve_description` - we **CANNOT** use just description as it would make the transition very complicated by conflicting with the current name of the initial Bugzilla flaw comment.

To keep the value aligned this will also require to rename `Flaw.requires_summary` to `Flaw.requires_cve_description`.

Closes OSIDB-2740

> This task was bigger than I initially though, should have done it in two separate pull requests... also my local environment went crazy and had to rely on CI for tests, sorry for the rebase mess

PR #574, opened 2024-06-07
Add HTTPS_PROXY to docker-compose.test.yml

jobselko:
This PR adds the `HTTPS_PROXY` variable to the `docker-compose.test.yml` file to ensure that the custom proxy works in containers. Without setting this, it is impossible to update test cassettes which use Jira.


jobselko:
@MrMarble This should fix the issue in #566

MrMarble (approved):

MrMarble:
Just a question, why don't we use the [env_file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute)  attribute of compose instead of passing all the vairables one by one?

JakubFrejlach:
> Just a question, why don't we use the [env_file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) attribute of compose instead of passing all the vairables one by one?

Is this attribute also available for podman-compose? We are not using docker-compose and I know it is far from being feature complete compared to docker-compose.

MrMarble:
> > Just a question, why don't we use the [env_file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) attribute of compose instead of passing all the vairables one by one?
> 
> Is this attribute also available for podman-compose? We are not using docker-compose and I know it is far from being feature complete compared to docker-compose.

It seems to be available yes, the podman-compose uses the compose spec as documentation  https://github.com/compose-spec/compose-spec/blob/master/05-services.md#env_file

Also there's one example using it https://github.com/containers/podman-compose/blob/main/examples%2Fnodeproj%2Fdocker-compose.yml#L22

osoukup (approved):
LGTM. Leaving the question of whether this is the correct way of passing the variables non-addressed as I have a little idea what are the actual options and their advantages :smile_cat:

jsvob:
> > > Just a question, why don't we use the [env_file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute) attribute of compose instead of passing all the vairables one by one?
> > 
> > 
> > Is this attribute also available for podman-compose? We are not using docker-compose and I know it is far from being feature complete compared to docker-compose.
> 
> It seems to be available yes, the podman-compose uses the compose spec as documentation https://github.com/compose-spec/compose-spec/blob/master/05-services.md#env_file
> 
> Also there's one example using it https://github.com/containers/podman-compose/blob/main/examples%2Fnodeproj%2Fdocker-compose.yml#L22

Once we have the cycles to clean up and improve our container-related tooling, we'll find many such things to improve. Historically, we needed to support multiple versions of docker-compose and podman-compose, each with its own quirks, and some of them old RHEL CSB versions at that. Also some old podman versions had bugs we had to work around. There was a lot of functionality that didn't work universally and `env_file` may or may not have been one of them. Some quirks were difficult to figure out and isolate and we ended up with a number of undocumented Chesterton’s Fences despite documentation efforts back in the day (e.g. we did document healthchecks but I have no idea about `env_file`, although I dimly remember there was something about it in a specific configuration that didn't work right). The historical reasons are also why each container has a script with its own dependency checks implemented in bash instead of using compose's healthchecks and long-form depends_on. Nowadays the situation is different and we can probably say we only care about what's in Fedora 40.

When revamping this one day, we'll just have to check that all modes of use and features we depend on work on all kinds of dev environments we care about: the biggest pain in the past was mysterious breakages with unknown origin and then repeatedly discovering a few days late it was a change in the container tooling that broke a specific usecase or in a specific situation _sometimes_ (for example `"Failed to start transient timer unit"` mentioned in the repo). That'll mean also reviewers should test it extensively.

PR #575, opened 2024-06-07
Fix openapi and swagger to document BZ & JIRA API key requirements

jsvob:
Fix openapi and swagger to document BZ & JIRA API key requirements
Closes OSIDB-2748


jsvob:
I will squash commits before merge.

PR #576, opened 2024-06-10
Implement major incident start date field

dmonzonis:
This field is set to the major incident BZ flag (hightouch and hightouch-lite) creation time, or automatically set when the flaw is set to be an approved major incident in OSIDB.

Closes OSIDB-2728.


apps/sla/models.py
```diff
@@ -54,7 +54,11 @@ def start(self, instance):
         """
         compute SLA start moment for the given instance
         """
-        return self.get_start(getattr(instance, date) for date in self.dates)
+        return self.get_start(
+            getattr(instance, date)
+            for date in self.dates
+            if getattr(instance, date) is not None
```
osoukup, line 60: This fix makes sense and I am at the same time hesitant about it :smile: I agree that the situation when the date is not really set would result in a pretty non-obvious error here. On the other hand the error means that the policy is written wrong - eg. we check Major Incident start date on something not necessarily being a Major Incident. Just skipping the date comparison by `is not None` might hide the bogus policy and result in some wrong/empty SLA deadline being set.

I am thinking about a proper fix. What would you say about the following?

Replace the `getattr` with something like `self.get_date` which would do the `getattr`, then check `is (not) None`, and if empty it would raise
https://github.com/RedHatProductSecurity/osidb/blob/master/apps/sla/exceptions.py#L18
with some descriptive error message that according to the policy a set date was expected but it is in fact empty.

collectors/bzimport/convertors.py
```diff
@@ -585,6 +585,20 @@ def major_incident_state(self):
...
+        If the flaw is an approved major incident, the start date is assumed to be
+        the date when the hightouch flag was created.
+        """
+        if self.major_incident_state in {
+            Flaw.FlawMajorIncident.APPROVED,
+            Flaw.FlawMajorIncident.CISA_APPROVED,
+        }:
+            for flag in self.flags:
+                if flag["name"] == "hightouch":
+                    return flag["creation_date"]
```
osoukup, line 600: There is one tricky thing. In Bugzilla the regular Major Incident which when confirmed stands for `APPROVED` here is represented by `hightouch` flag there and the CISA Major Incident which when confirmed stands for `CISA_APPROVED` here is represented by `hightouch_lite` flag there. That means that for each of the two states you have to get the date from a different Bugzilla flag (the other is usually not existing). Also there may be rare but existing situation when both flags are raised - which is not really valid but we have to handle it somehow. In such case we prioritize the regular Major Incident (`hightouch`) and ignore the CISA one.

osidb/migrations/0143_rename_flaw_fields.py
```diff
@@ -6,7 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('osidb', '0141_remove_flawacknowledgment_meta_attr'),
+        ('osidb', '0142_flaw_major_incident_start_dt'),
```
osoukup, line 9: I do not think that it is very safe to mess with the existing migrations. They are most probably already applied somewhere which is recorded in the DB etc. I would expect this to result in some nasty Django errors. You should rather rename the new validation.

collectors/bzimport/convertors.py
osoukup: Actually I am wrong on the invalid case. We do not prioritize anything in this scenario but simply use an `INVALID` Major Incident state. You can do whatever you want I guess :smile:

osoukup (approved):
I have a few comments to address. Generally this goes in the right direction

osoukup (changes requested):
I accidentally approved :smile: probably it is a habit when I see your code but this time I would like (at least to discuss) some changes

osidb/migrations/0143_rename_flaw_fields.py
dmonzonis: Oops, this is actually a mistake from when I rebased

apps/sla/models.py
dmonzonis: I think that it makes sense to have a policy with something like
```yaml
start:
  latest:
    - reported date
    - unembargo date
    - major_incident_start_dt
```
for flaws that can be a MI or not, and that `major_incident_start_dt` should be ignored if it's not a MI, which is what would happen with the current code. If we raise an error then we lose this flexibility by enforcing that this policy only works for MI flaws, but for that we already have the `is major incident` condition.

osidb/tests/test_flaw.py
```diff
@@ -603,6 +603,23 @@ def test_ps_product_affect(self, ps_module_name, ps_product_name):
             affect = AffectFactory()
         assert affect.ps_product == ps_product_name
 
+    @pytest.mark.enable_signals
+    def test_major_incident_start_dt(self):
+        """
+        Test that the signal to set the MI start date is correctly called.
+
```
jobselko, line 610: redundant space

osidb/signals.py
```diff
@@ -76,8 +88,9 @@ def populate_cvss_score(sender, instance, **kwargs):
 
 
 @receiver(pre_save, sender=Flaw)
-def update_local_updated_dt_flaw(sender, instance, **kwargs):
+def update_flaw_fields(sender, instance, **kwargs):
     instance.local_updated_dt = timezone.now()
+    update_major_incident_start_dt(instance)
```
jobselko, line 93: I would prefer to keep this separate as more functions update flaw fields.

docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support Bugzilla tracker creation/linking for non-Bugzilla flaws (OSIDB-2845)
 - Add bulk-enabling parameter "sync_to_bz" to POST for Trackers (OSIDB-2609)
 - Add bulk POST, DELETE for Affects (OSIDB-2722)
+- Add major incident start date field (OSIDB-2728)
```
jobselko, line 23: I would mention the whole name of a field for transparency, i.e. Add "major_incident_start_dt" field to Flaw

osidb/serializer.py
```diff
@@ -1526,6 +1526,7 @@ class Meta:
                 "reported_dt",
                 "mitigation",
                 "major_incident_state",
+                "major_incident_start_dt",
```
jobselko, line 1529: It is just an idea, but maybe it would be worth adding it to filters as well?

collectors/bzimport/tests/test_convertors.py
```diff
@@ -1451,6 +1460,53 @@ def test_groups(self, request, bz_groups, ldap_read_group, ldap_write_group):
...
+    )
+    def test_major_incident_start_dt(self, hightouch, hightouch_lite, creation_date):
+        """
+        Test that if the appropriate flags are set to be a MI, then the MI start
+        date is properly set from the flag creation date.
+        """
+        flaw_bug = self.get_flaw_bug()
+        flaw_bug["flags"] = []
+
+        if hightouch is not None and hightouch_lite is not None:
```
jobselko, line 1479: Maybe a stupid question, but why both `hightouch` and `hightouch_lite` must be `not None`?

apps/sla/models.py
osoukup: One thing is that MIs have a very specific handling and more strict policy so they should not be consider by a standard policies - currently even the CISA MI has its own policy defined

https://github.com/RedHatProductSecurity/osidb/blob/master/apps/sla/sla.yml#L42

The other thing is that the policies should not be very general but rather as specific as possible so people could very easily connect the SLA context with the resulting deadline only by looking at the SLA file. There should ideally be no logic in the code like what to do when something is empty.

osidb/signals.py
osoukup: You mean having multiple different receivers for the `pre_save` and `sender=Flaw`?

jobselko: Oh, I overlooked that `@receiver(pre_save, sender=Flaw)` is unique - originally, I thought that `populate_cvss_score` has the same one, but actually, it has `FlawCVSS`. So my note can probably be ignored :).

collectors/bzimport/tests/test_convertors.py
dmonzonis: These flags are optional in Bugzilla. I'm testing the case in which these flags aren't present, in which case the date shouldn't be set either (as it would have a major incident state of `NOVALUE`)

osoukup (commented):
LGTM. Thank you for addressing my opinionated feedback :smile_cat:

dmonzonis:
@osoukup I think I have to take a different approach to the SLA containing a null `major_incident_start_dt`. When we check for policies that apply, the SLA is loaded and if it finds a `major_incident_start_dt=None` it will fail, but maybe it doesn't apply because it did not meet the conditions to begin with (the instance is not a major incident so it shouldn't even load that policy).

So the problem comes from the error being raised too soon, before checking if conditions apply. I think I should add this validation somewhere else, maybe after the SLA is completely loaded, add a method that checks for inconsistencies and enforces some rules: if it uses `major_incident_start_dt`, then it must enforce it being a major incident, if it uses `unembargo_dt` then it must enforce that it is not embargoed, etc.

osoukup:
@dmonzonis oh I see! Then it means that I was insisting on something unreasonable... Sorry for that. Then probably your original solution (which was if I remember correctly just skipping the `None` values) is an acceptable way to go. If the policy is written carefully it should not be an issue anyway and if it eventually is we will deal with it later.

jobselko:
@dmonzonis You have to recreate migrations as they will alter the `FlawAudit` model as well.

PR #578, opened 2024-06-11
Do not allow empty strings in CVE ID

dmonzonis:
The field `cve_id` is set as null=True, blank=True and unique=True. One issue with this is that it can have both `None` and empty string values, but only one value with empty string is allowed due to uniqueness. What we actually want, is for the values to either have a valid CVE, or be null.

Django doesn't execute the validations if the field is blank=True and we have an empty string. However we do not want to set blank=False since this would not allow sending a null value either. In https://code.djangoproject.com/ticket/4136 this issue was addressed, but only for forms, not on the model itself, and we want to make sure that the data is consistent no matter where we modify it from (API, shell, etc).

To overcome this issue, this commit modifies the save method so that an empty string value is properly converted to `None` before saving.

Closes OSIDB-2702.


osoukup (approved):
LGTM

osidb/models.py
```diff
@@ -705,6 +705,13 @@ def __str__(self):
         """convert to string"""
         return str(self.uuid)
 
+    def save(self, *args, **kwargs):
+        # Automatically convert empty string values to null, since validation is skipped for
+        # empty strings
+        if self.cve_id == "":
+            self.cve_id = None
+        super().save(*args, **kwargs)
```
jobselko, line 713: We usually do these things via `signals.py`. Would it be possible to move it there? I am not saying you have to do it, just asking.

JakubFrejlach: Actually for converting None (NULL) values for Char based fields, for data which come from Bugzilla, I've used mixin and it was done in the clean function, so I have no strong preference on this one.

PR #579, opened 2024-06-11
Implement search on emptiness for several fields

dmonzonis:
Like it was done in #522, filter on emptiness by more fields has been implemented: cve_description, cwe_id, statement, mitigation, owner, and cvss_scores.

Like with cve_id, to use this filter through the API one may set {field_name}__isempty to either 1 or 0 (empty or non-empty).

Closes OSIDB-2815.

### DIFF
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -71,6 +71,25 @@ class EmptyOrNullStringFilter(BooleanFilter):
         return method(query)
 
 
+class EmptyCvssFilter(BooleanFilter):
+    """Filter by whether the flaw has any CVSS with the given issuer and version."""
+
+    def __init__(self, issuer, version, *args, **kwargs):
+        self.issuer = issuer
+        self.version = version
+        super().__init__(*args, **kwargs)
+
+    def filter(self, queryset, value):
+        if value in EMPTY_VALUES:
+            return queryset
+
+        method = queryset.exclude if value else queryset.filter
+        query = Q(cvss_scores__issuer=self.issuer) & Q(
+            cvss_scores__version=self.version
+        )
+        return method(query).distinct()
+
+
 class DistinctFilterSet(FilterSet):
     """
     Custom FilterSet which enforces the distinct for field names that starts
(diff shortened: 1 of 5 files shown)


osidb/filters.py
```diff
@@ -276,6 +274,15 @@ class FlawFilter(DistinctFilterSet, IncludeFieldsFilterSet, ExcludeFieldsFilterS
...
     cvss_scores__cvss_version = CharFilter(field_name="cvss_scores__version")
 
+    # Emptiness filters
+    cve_id__isempty = EmptyOrNullStringFilter(field_name="cve_id")
+    cve_description__isempty = EmptyOrNullStringFilter(field_name="cve_description")
+    statement__isempty = EmptyOrNullStringFilter(field_name="statement")
+    mitigation__isempty = EmptyOrNullStringFilter(field_name="mitigation")
+    owner__isempty = EmptyOrNullStringFilter(field_name="owner")
+    cwe_id__isempty = EmptyOrNullStringFilter(field_name="cwe_id")
+    cvss_scores__isempty = BooleanFilter(field_name="cvss_scores", lookup_expr="isnull")
```
JakubFrejlach, line 284: This only checks whether some CVSS is attached to a Flaw without any context about version or issuer, right?

I am not 100% sure about the usecase but I would imagine users would like to explicitly list only Flaws without RH CVSS v3 judging by the wording in the Jira issue. :thinking:

JakubFrejlach (commented):
LGTM, but I have one consideration regarding CVSS emptiness.

osidb/filters.py
dmonzonis: Okay, but then I would rename the filter to something like `cvss3__isempty` to avoid confusion, as if I use the name of a field + `__isempty` I would expect emptiness on the whole field, not only an arbitrary subset. How does that sound to you?

JakubFrejlach: Yeah, definitely, we need to be explicit here. In the future, there will be also cvss4, etc. Also not sure, whether they are interested only in RH cvss3 or NVD cvss3 as well :thinking:

dmonzonis: I will go ahead and just add all of the possibilities

JakubFrejlach (approved):
LGTM now!

jobselko (approved):

osoukup (approved):
LGTM

MrMarble (approved):

PR #582, opened 2024-06-13
Improve affect validation error messages

dmonzonis:
Some validation error messages in the Affect model have been improved to include more info on which affect is raising the error. This is useful when performing bulk operations on affects, since a validation error will pop up with no info on which affect caused it, and the flaw may have lots of affects.

Closes OSIDB-2893.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1775,7 +1775,9 @@ class Affect(
                 or self.resolution != old_affect.resolution
             ):
                 raise ValidationError(
-                    f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                    f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an invalid "
+                    f"affectedness/resolution combination: {self.resolution} is not a valid resolution "
+                    f"for {self.affectedness}."
                 )
 
             # If modifying something else from a record with an invalid combination, e.g. the
@@ -1783,7 +1785,8 @@ class Affect(
             self.alert(
                 "flaw_historical_affect_status",
                 f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has a "
-                "historical affectedness/resolution combination which is not valid anymore.",
+                "historical affectedness/resolution combination which is not valid anymore: "
+                f"{self.resolution} is not a valid resolution for {self.affectedness}.",
             )
 
     def _validate_affect_status_resolution(self):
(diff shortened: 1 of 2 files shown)


osidb/models.py
```diff
@@ -1769,15 +1769,18 @@ def _validate_historical_affectedness_resolution(self):
                 or self.resolution != old_affect.resolution
             ):
                 raise ValidationError(
-                    f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                    f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an invalid "
+                    f"affectedness/resolution combination: {self.resolution} is not a valid resolution "
```
jsvob, line 1773: Nitpick: says "affectedness/resolution" first, prints resolution/affectedness second.

```suggestion
                    f"resolution/affectedness combination: {self.resolution} is not a valid resolution "
```

osidb/models.py
```diff
@@ -1775,15 +1775,18 @@ def _validate_historical_affectedness_resolution(self):
                 or self.resolution != old_affect.resolution
             ):
                 raise ValidationError(
-                    f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                    f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an invalid "
+                    f"resolution/affectedness combination: {self.resolution} is not a valid resolution "
+                    f"for {self.affectedness}."
```
osoukup, line 1780: The usual order to refer to the affect status is in order of first `affectedness` and then `resolution`. This way people are used to see it in the old tooling and I think the opposite order may be confusing for them. Eventually we should merge it into a single `status` value but at least for the time being it would be better to stick with the current order.

osidb/models.py
```diff
@@ -1797,7 +1800,9 @@ def _validate_affect_status_resolution(self):
             not in AFFECTEDNESS_HISTORICAL_VALID_RESOLUTIONS[self.affectedness]
         ):
             raise ValidationError(
-                f"{self.resolution} is not a valid resolution for {self.affectedness}."
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has an invalid "
+                f"resolution/affectedness combination: {self.resolution} is not a valid resolution "
```
osoukup, line 1804: The same order issue here.

osoukup (approved):
I would be for using the order `affectedness` first and `resolution` second. Otherwise LGTM

PR #583, opened 2024-06-13
FIx install-certs.sh script

JakubFrejlach:
This PR:
* fixes install-certs.sh to account for RH_CERT_URL both with and without trailing slash


osoukup (approved):
I see this syntax for the first time :smile: but I am no bash expert and if it works then LGTM

jsvob (approved):

PR #584, opened 2024-06-13
Allow null in cvss comment.

jsvob:
Closes OSIDB-2907

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2109,7 +2109,7 @@ class AffectCVSSManager(ACLMixinManager, TrackingMixinManager):
             )
 
 
-class CVSS(ACLMixin, BugzillaSyncMixin, TrackingMixin):
+class CVSS(ACLMixin, BugzillaSyncMixin, NullStrFieldsMixin, TrackingMixin):
     class CVSSVersion(models.TextChoices):
         VERSION2 = "V2", "version 2"
         VERSION3 = "V3", "version 3"
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -27,6 +27,7 @@ from .exceptions import DataInconsistencyException
 from .helpers import differ, ensure_list
 from .mixins import ACLMixin, Alert, AlertMixin, TrackingMixin
 from .models import (
+    CVSS,
     Affect,
     AffectCVSS,
     Erratum,
@@ -919,30 +920,44 @@ class JiraTaskSyncMixinSerializer(JiraAPIKeyMixin, serializers.ModelSerializer):
         abstract = True
 
 
-class AffectCVSSSerializer(
+class AbstractCVSSSerializer(
     ACLMixinSerializer,
     AlertMixinSerializer,
     BugzillaSyncMixinSerializer,
     IncludeExcludeFieldsMixin,
     TrackingMixinSerializer,
 ):
+    """Abstract serializer for FlawCVSS and AffectCVSS serializer"""
+
+    cvss_version = serializers.ChoiceField(choices=CVSS.CVSSVersion, source="version")
+    score = serializers.FloatField(read_only=True)
+    comment = serializers.CharField(required=False, allow_null=True, allow_blank=True)
+
+    class Meta:
+        abstract = True
+        fields = (
+            # Also add "affect" or "flaw" for AffectCVSS or FlawCVSS
+            ["comment", "cvss_version", "issuer", "score", "uuid", "vector"]
+            + ACLMixinSerializer.Meta.fields
+            + AlertMixinSerializer.Meta.fields
+            + TrackingMixinSerializer.Meta.fields
+        )
+
+
+class AffectCVSSSerializer(
+    AbstractCVSSSerializer,
+):
     """AffectCVSS serializer"""
 
     cvss_version = serializers.ChoiceField(
         choices=AffectCVSS.CVSSVersion, source="version"
     )
-    score = serializers.FloatField(read_only=True)
 
     class Meta:
         """filter fields"""
 
         model = AffectCVSS
-        fields = (
-            ["affect", "comment", "cvss_version", "issuer", "score", "uuid", "vector"]
-            + ACLMixinSerializer.Meta.fields
-            + AlertMixinSerializer.Meta.fields
-            + TrackingMixinSerializer.Meta.fields
-        )
+        fields = ["affect"] + AbstractCVSSSerializer.Meta.fields
 
 
 @extend_schema_serializer(exclude_fields=["affect", "updated_dt"])
@@ -1442,29 +1457,19 @@ class FlawReferencePutSerializer(FlawReferenceSerializer):
 
 
 class FlawCVSSSerializer(
-    ACLMixinSerializer,
-    BugzillaSyncMixinSerializer,
-    IncludeExcludeFieldsMixin,
-    TrackingMixinSerializer,
-    AlertMixinSerializer,
+    AbstractCVSSSerializer,
 ):
     """FlawCVSS serializer"""
 
     cvss_version = serializers.ChoiceField(
         choices=FlawCVSS.CVSSVersion, source="version"
     )
-    score = serializers.FloatField(read_only=True)
 
     class Meta:
         """filter fields"""
 
         model = FlawCVSS
-        fields = (
-            ["comment", "cvss_version", "flaw", "issuer", "score", "uuid", "vector"]
-            + ACLMixinSerializer.Meta.fields
-            + TrackingMixinSerializer.Meta.fields
-            + AlertMixinSerializer.Meta.fields
-        )
+        fields = ["flaw"] + AbstractCVSSSerializer.Meta.fields
 
 
 @extend_schema_serializer(exclude_fields=["updated_dt", "flaw"])
--- a/osidb/tests/endpoints/test_affects.py
+++ b/osidb/tests/endpoints/test_affects.py
@@ -169,6 +169,55 @@ class TestEndpointsAffects:
         assert response.status_code == status.HTTP_200_OK
         assert response.data["uuid"] == cvss_uuid
 
+    @pytest.mark.parametrize(
+        "skip_value,request_provided_value,resulting_value",
+        [
+            (False, "something", "something"),
+            (False, "", ""),
+            (False, None, ""),  # null converted to ""
+            (True, None, ""),  # omitted value converted to ""
+        ],
+    )
+    @pytest.mark.enable_signals
+    def test_affectcvss_create_comment(
+        self,
+        auth_client,
+        test_api_uri,
+        skip_value,
+        request_provided_value,
+        resulting_value,
+    ):
+        """
+        Test the behavior of comment in creation of AffectCVSS records via a REST API POST request.
+        """
+        flaw = FlawFactory()
+        affect = AffectFactory(flaw=flaw)
+        cvss_data = {
+            "comment": request_provided_value,
+            "issuer": AffectCVSS.CVSSIssuer.REDHAT,
+            "cvss_version": AffectCVSS.CVSSVersion.VERSION3,
+            "vector": "CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H",
+            "embargoed": flaw.embargoed,
+        }
+        if skip_value:
+            del cvss_data["comment"]
+
+        # Tests "POST" on affects/{uuid}/cvss_scores
+        response = auth_client().post(
+            f"{test_api_uri}/affects/{str(affect.uuid)}/cvss_scores",
+            data=cvss_data,
+            format="json",
+            HTTP_BUGZILLA_API_KEY="SECRET",
+        )
+        cvss_uuid = response.data["uuid"]
+        assert response.data["comment"] == resulting_value
+
+        # Tests "GET" on affects/{uuid}/cvss_scores/{uuid}
+        response = auth_client().get(
+            f"{test_api_uri}/affects/{str(affect.uuid)}/cvss_scores/{cvss_uuid}"
+        )
+        assert response.data["comment"] == resulting_value
+
     @pytest.mark.enable_signals
     def test_affectcvss_update(self, auth_client, test_api_uri):
         """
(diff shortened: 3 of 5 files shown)


osidb/models.py
```diff
@@ -2128,7 +2128,8 @@ class CVSSIssuer(models.TextChoices):
 
     issuer = models.CharField(choices=CVSSIssuer.choices, max_length=16)
 
-    comment = models.TextField(blank=True)
+    # Nullable to allow clients to send a null value.
+    comment = models.TextField(blank=True, null=True, default="")  # noqa: DJ01
```
JakubFrejlach, line 2132: passing `null=True` for any Char or Text based field in Django should be strictly avoided (https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.null) or we will end up with the two possible empty representation of a string based field (NULL vs empty string). Only exception for null=True should be, when such field needs to be also unique. I think we should rather allow null values on serializer and then convert them to empty strings.

JakubFrejlach (changes requested):
I disagree with this change, we already dealt with this in the past and I would like to avoid bringing this back.

osidb/models.py
jsvob: Hmm. That doesn't work. 

```
curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -H "Bugzilla-Api-Key: ${BZIMPORT_BZ_API_KEY}"  -H "Jira-Api-Key: ${JIRA_AUTH_TOKEN}" -X POST -d '{
  "comment": null,
  "cvss_version": "V3",
  "issuer": "RH",
  "vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:H/A:H",
  "embargoed": false
       }' ${SERVICE_URL}/osidb/api/v1/flaws/c96cb300-5f0f-4771-8687-0044c2087fd8/cvss_scores

{"comment":["This field may not be null."],"dt":"2024-06-13T18:04:45.701300Z","revision":"unknown","version":"3.7.3","env":"local"}[root@17e893e3c809 src]# 

```

I tried all of these and none of them work:

```
comment = serializers.CharField(required=False, allow_null=True, allow_blank=True)
comment = serializers.CharField(required=False, allow_blank=True)
comment = serializers.CharField(required=False, allow_null=True)
```

I guess if we have to fix it on OSIDB side instead of OSIM side, the only viable option is to override `create` of AffectCVSSSerializer, but I originally wanted to avoid it.

What we agree on here is that this PR as it is (null in database) must not be merged. Marking as WIP to make it clear.

I'll wait for the Slack discussion and then rework it with custom `create` if necessary.

Thanks!

JakubFrejlach: Hmm, that is weird, I just tried via my local OSIM after I made those changes and it succeeded with the following payload:
image

superbuggy:
I am pretty sure [this is better](https://github.com/RedHatProductSecurity/osim/pull/279)... @JakubFrejlach thoughts?

osidb/models.py
jsvob: Thank you for trying it out. I can't figure out why it doesn't work for me.

```
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ make migrate
make migrations-create
make[1]: Entering directory '/home/jsvoboda/work-prodsec/osidb'
>creating migrations in osidb-service
`which podman` exec -it osidb-service python3 manage.py makemigrations --settings config.settings_local
No changes detected
make[1]: Leaving directory '/home/jsvoboda/work-prodsec/osidb'
make migrations-apply
make[1]: Entering directory '/home/jsvoboda/work-prodsec/osidb'
>appyling migrations in osidb-service
`which podman` exec -it osidb-service python3 manage.py migrate --settings config.settings_local
Operations to perform:
  Apply all migrations: auth, bbsync, contenttypes, exploits, framework, osidb, pghistory, sessions, trackers
Running migrations:
  No migrations to apply.
make[1]: Leaving directory '/home/jsvoboda/work-prodsec/osidb'
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ make kill_django 
>Note that the django server restarts automatically, as long as osidb-service is running.
`which podman` exec -it osidb-service pkill -f "python3 manage.py runserver"
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ git diff
diff --git a/osidb/models.py b/osidb/models.py
index 5eacadb..fc05e6c 100644
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -2103,7 +2103,7 @@ class AffectCVSSManager(ACLMixinManager, TrackingMixinManager):
             )
 
 
-class CVSS(ACLMixin, BugzillaSyncMixin, TrackingMixin):
+class CVSS(ACLMixin, BugzillaSyncMixin, TrackingMixin, NullStrFieldsMixin):
     class CVSSVersion(models.TextChoices):
         VERSION2 = "V2", "version 2"
         VERSION3 = "V3", "version 3"
diff --git a/osidb/serializer.py b/osidb/serializer.py
index 86268e3..98442ef 100644
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -932,6 +932,7 @@ class AffectCVSSSerializer(
         choices=AffectCVSS.CVSSVersion, source="version"
     )
     score = serializers.FloatField(read_only=True)
+    comment = serializers.CharField(required=False, allow_null=True, allow_blank=True)
 
     class Meta:
         """filter fields"""
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ git log | head -n1
commit c194eb2155c420ceaefc1835002a987f0b55ad52
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   osidb/models.py
	modified:   osidb/serializer.py

no changes added to commit (use "git add" and/or "git commit -a")
(venv) [jsvoboda@jsvoboda-thinkpadx1carbongen9 osidb]$ 
```

Tried adding `NullStrFieldsMixin` both to `CVSS` (what it should be probably) and `AffectCVSS` (just to make sure) with the same result.

Tried all 3 of the variants of `serializers.CharField` listed above. Always did `make migrate` and `make kill_django` or `make restart_django_foreground` before testing each change.

Also did compose-down a few times in between (with appropriate setup before continuing - `from collectors.ps_constants.tasks import ps_constants_collector ; ps_constants_collector() ; from collectors.product_definitions.tasks import product_definitions_collector; product_definitions_collector() ; from collectors.jiraffe.tasks import metadata_collector ; metadata_collector()` and `python3 manage.py syncflaws --verbosity=3 2291272` and modifying the flaw to pass validations `. .env ; osdsetp ; curl -H "Authorization: Bearer $token" -H "Content-Type: application/json" -H "Bugzilla-Api-Key: ${BZIMPORT_BZ_API_KEY}"  -H "Jira-Api-Key: ${JIRA_AUTH_TOKEN}" -X PUT -d '{ "title": "From NVD collector", "comment_zero": "A medium severity vulnerability in BIPS has been identified where an authenticated attacker with high privileges can access the SSH private keys via an information leak in the server response.", "updated_dt": "2024-06-14T06:47:27Z", "unembargo_dt": "2024-06-11T00:00:00Z", "components":["testcomp", "ssh"], "impact": "MODERATE", "embargoed": false }' ${SERVICE_URL}/osidb/api/v1/flaws/c96cb300-5f0f-4771-8687-0044c2087fd8`).

jobselko: What about this solution?

```
    comment = serializers.SerializerMethodField()

    def get_comment(self, obj):
        return (obj.comment if obj.comment else "")
```

JakubFrejlach: I think you are editing `AffectCVSSSerializer` instead of `FlawCVSSSerializer` there are two serializers for CVSS as there are two models which are taking advantage of CVSS models. The curl command you've send was definitely aiming for Flaw CVSS creation, so unless you will edit also FlawCVSSSerializer, it won't work.

JakubFrejlach: > What about this solution?
> 
> ```
>     comment = serializers.SerializerMethodField()
> 
>     def get_comment(self, obj):
>         return obj.comment if obj.comment else ""
> ```

This won't work as using SerializerMethodField is for readonly purposes.

jsvob:
For posterity, I didn't succeed in quickly making a mixin that would allow nullable fields in a generic way.

```
class NullStrFieldsSerializerMixin(serializers.Serializer):
    def get_fields(self):
        """
        Modifies CharField fields that are required=False, allow_blank=True to
        be also allow_null=True. That means it only modifies fields that
        are already allowed to be omitted by the client. This makes it possible
        for the client to also omit the field by sending a null value.

        The corresponding model must inherit from NullStrFieldsMixin.
        """
        fields = super().get_fields()
        for field_name, field in fields.items():
            if isinstance(field, serializers.CharField):
                if field.allow_blank and not field.required and not field.allow_null:
                    field.allow_null = True
                # ^ doesn't work

                # doesn't work either
                field.allow_null = True
                field.allow_blank = True
                field.required = False
                
                # Debugging needed. Maybe validation has to be tweaked too?
        return fields
```

Based on 1:1 with @JakubFrejlach we might want to explore that path in the future when/if we have many such fields, but not worth it now.

osidb/models.py
jsvob: Thank you, I was indeed modifying only half of the serializers!

osidb/models.py
```diff
@@ -2109,7 +2109,7 @@ def create_cvss(affect, issuer, version, **extra_fields):
             )
 
 
-class CVSS(ACLMixin, BugzillaSyncMixin, TrackingMixin):
+class CVSS(ACLMixin, BugzillaSyncMixin, TrackingMixin, NullStrFieldsMixin):
```
osoukup, line 2112: nitpick: When there is no specific reason for a certain order, we should stick with the alphabet

osoukup (approved):
LGTM

osidb/models.py
jsvob: Thanks, fixed.

osidb/serializer.py
```diff
@@ -919,19 +919,40 @@ class Meta:
...
+
+    class Meta:
+        abstract = True
+        fields = (
+            # Also add "affect" or "flaw" for AffectCVSS or FlawCVSS
+            ["comment", "cvss_version", "issuer", "score", "uuid", "vector"]
+            + ACLMixinSerializer.Meta.fields
+            + AlertMixinSerializer.Meta.fields
+            + TrackingMixinSerializer.Meta.fields
+        )
```
jobselko, line 945: I am not sure if the current abstract class makes much sense. Can the changes be made directly in the respective classes?

jobselko (changes requested):
Could you please provide tests? Thank you

osidb/serializer.py
jsvob: The changes certainly can be made directly in the respective classes, but the dichotomy that the base CVSS abstract model is used in two different models that are used in two different serializers was lost on me and resulted in a day of lost time.

Given that and after a chat with @JakubFrejlach I deemed it worthwhile to have the common abstract serializer to make it non-confusing when the next change on CVSS occurs.

osidb/tests/endpoints/test_affects.py
```diff
@@ -168,6 +169,7 @@ def test_affectcvss_create(self, auth_client, test_api_uri):
         )
         assert response.status_code == status.HTTP_200_OK
         assert response.data["uuid"] == cvss_uuid
+        assert response.data["comment"] == ""  # null converted to ""
```
jobselko, line 172: Could you please test this functionality in a separate test with the following scenarios?
1. string -> string
2. empty string -> empty string
3. None -> empty string

osidb/serializer.py
```diff
@@ -919,19 +919,40 @@ class Meta:
...
+
+    cvss_version = serializers.ChoiceField(
+        choices=FlawCVSS.CVSSVersion, source="version"
+    )
+    score = serializers.FloatField(read_only=True)
+    comment = serializers.CharField(required=False, allow_null=True, allow_blank=True)
+
+    class Meta:
+        abstract = True
+        fields = (
```
JakubFrejlach, line 939: In both `FlawCVSSSerilizer` and `AffectCVSSSerializer` you can now do:

```python
class Meta:
    ...
    fields = AbstractCVSSSeralizer.Meta.fields + ["flaw"]

class Meta:
    ...
    fields = AbstractCVSSSeralizer.Meta.fields + ["affect"]
```

JakubFrejlach (approved):
One suggestion, otherwise LGTM!

osidb/serializer.py
```diff
@@ -919,19 +919,40 @@ class Meta:
...
     ACLMixinSerializer,
     AlertMixinSerializer,
     BugzillaSyncMixinSerializer,
     IncludeExcludeFieldsMixin,
     TrackingMixinSerializer,
+):
+    """Abstract serializer for FlawCVSS and AffectCVSS serializer"""
+
+    cvss_version = serializers.ChoiceField(
+        choices=FlawCVSS.CVSSVersion, source="version"
```
jobselko, line 932: Please use `CVSS.CVSSVersion` instead of `FlawCVSS.CVSSVersion`

osidb/serializer.py
jobselko: I briefly checked the DRF documentation (https://www.django-rest-framework.org/api-guide/serializers/), and there seem to be several options on how to solve the "None to empty string" issue, e.g.:
1. Overriding `.to_representation()` and `.to_internal_value()` methods in a serializer for custom serialization/deserialization (i.e. converting None to an empty string in serializer)
2. Creating a custom char field and automatically convert None to an empty string

If I understand it correctly, currently, the serializer enables both None and empty string, and the conversion is done in models via `NullStrFieldsMixin`. But in future, we should handle the conversion directly in a serializer and then save either string or empty string in models.

osoukup: This may be a good suggestion for the future but for today we cannot afford any bigger changes. Let us only focus on making this working properly and let us leave the improvements for the next time.

osidb/serializer.py
jsvob: Thanks, fixed. Moved the `[]` to the beginning to make the generated schema a bit clearer because these are IMO one of the most important fields. Maybe it's just a cosmetic preference.

osidb/serializer.py
jsvob: Thanks, fixed.

osidb/tests/endpoints/test_affects.py
jsvob: Thanks, fixed.

jobselko (approved):
LGTM. Maybe the `CHANGELOG` could be changed because there are changes in API.

PR #589, opened 2024-06-18
SLA calculation logic simplification

p-rog:
I simplified the sla calculation logic based on the new Engineering requirements.


osoukup (approved):
LGTM

PR #592, opened 2024-06-20
Rudimentary API request logging

jsvob:
Closes OSIDB-2514

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -3,6 +3,7 @@ implement osidb rest api views
 """
 
 import logging
+from datetime import datetime
 from typing import Type
 from urllib.parse import urljoin
 
@@ -73,7 +74,8 @@ from .serializer import (
 )
 from .validators import CVE_RE_STR
 
-logger = logging.getLogger(__name__)
+# Use only for RudimentaryUserPathLoggingMixin
+api_logger = logging.getLogger("api_req")
 
 
 @api_view(["GET"])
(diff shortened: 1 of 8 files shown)


jsvob:
No idea how to test logging without spending time on researching that, which I (we all) don't have. I think it's enough that all tests run fine and that it uses logging that's already well-used and tested by actual use.

osidb/api_views.py
```diff
@@ -83,7 +84,92 @@ def healthy(request: Request) -> Response:
...
+
+        request._rudimentary_user_path_logging = {
+            "method": method,
+            "path": path,
+            "start": datetime.now(),
+            "user": user,
+        }
+
+        log_message = f"API REQUEST {user} at {method} {path}"
+        logger.debug(log_message)
```
osoukup, line 143: Is `debug` level enough? Shouldn't it be maybe `info`? The production (where it matters) does not run in the debug mode so I expect that this logging will have no effect there. Or do miss something?

osidb/api_views.py
```diff
@@ -83,7 +84,92 @@ def healthy(request: Request) -> Response:
...
+        timediff = end - request._rudimentary_user_path_logging["start"]
+        ms_number = timediff.total_seconds() * 1000
+        ms_formatted = "{:6.0f}".format(ms_number)
+
+        user = request._rudimentary_user_path_logging["user"]
+        method = request._rudimentary_user_path_logging["method"]
+        path = request._rudimentary_user_path_logging["path"]
+
+        log_message = f"REQ END in {ms_formatted} ms {user} at {method} {path}"
+        logger.debug(log_message)
```
osoukup, line 167: The `debug` level again.

osidb/api_views.py
jsvob: Good question. Honestly, I have very little idea how that works.  I see there's some configuration in `settings_prod.py` and based on that and Splunk it seems prod-django.log (which would most probably be the log where these API request logging messages would go to) is set to WARNING level, whereas everything else seems to have (implicit?) INFO. I'll study this and ask around the team to understand it better. Thank you!

osoukup: If you find out that the log level is not set appropriately, it is a viable approach to just adjust it. WARNING level seems to me too strict, I would expect INFO level. However, DEBUG level on the other hand seems to be to unrestricted. I would consider debug logs as something necessary for debugging so not something one would expect to be used in production.

jsvob: Thanks, that's partly what I did in the end.

After learning a bit about logging, it seems it would be too complicated to have e.g. 'osidb' module at level WARNING and nested 'osidb.api_views' module at level INFO (seems Python likes it in the opposite order of levels, which doesn't fit our usecase).

So I added a separate logging module 'api_req' and modified the 'console' level for prod.

Here's the change for easier review: https://github.com/RedHatProductSecurity/osidb/compare/f66d43619d22034b36d57befd367961e83c0c017..ecab2150f3b929d6b244a6547ed24d0af4dc8223

osoukup (approved):
LGTM. Great job!

PR #596, opened 2024-06-21
Set Jira tracker "components" field.

jsvob:
Closes OSIDB-2988


jsvob:
If the tests fail, it might be independent breakage on master. I'll investigate on Monday.

osoukup (approved):
LGTM

PR #598, opened 2024-06-24
OSIDB-3011: Add PUT to jira forwarder

MrMarble:
# OSIDB-3011: Add PUT to jira forwarder

Currently Jira Forwarder only implements `GET`  and `POST` methods, we need to support `PUT` as well for editing issues and other jira entities

Closes OSIDB-3011

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -1058,6 +1058,35 @@ class JiraStageForwarderView(APIView):
         )
         return Response(response.json(), status=response.status_code)
 
+    def put(self, request, *args, **kwargs):
+        """perform JIRA stage HTTP PUT"""
+
+        path_value = request.GET.get("path")
+        target_url = f"{JIRA_SERVER}{path_value}"
+        headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+            "Accept-Encoding": "gzip, deflate, br, zstd",
+            "User-Agent": "OSIM",
+        }
+        jira_api_key = request.headers.get("Jira-Api-Key")
+        if jira_api_key:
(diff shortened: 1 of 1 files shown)


osidb/api_views.py
```diff
@@ -1058,6 +1058,35 @@ def post(self, request, *args, **kwargs):
...
+        else:
+            raise ValidationError({"Jira-Api-Key": "This HTTP header is required."})
+
+        response = requests.put(
+            target_url,
+            data=request.body,
+            proxies=self.proxies,
+            headers=headers,
+            timeout=60,
+        )
```
jsvob, line 1084: Why is params not in PUT but is in POST?

Is that not used by the complementary code that calls the forwarder?

MrMarble: I did some testing and the https://github.com/RedHatProductSecurity/osidb/blob/f5f0f304155da618d8201c40b064d1752f7996c6/osidb/api_views.py#L1064 also contains the query params.

The https://github.com/RedHatProductSecurity/osidb/blob/f5f0f304155da618d8201c40b064d1752f7996c6/osidb/api_views.py#L1044 is a dict with only one key `path`, so the query was being doubled.

```
# Generated url
 https://example.com/path/to/view?query1=true&path=path%2Fto%2Fview%3Fquery1%3Dtrue

target_url =>  https://example.com
path_value => /path/to/view?query1=true
params => &path=path%2Fto%2Fview%3Fquery1%3Dtrue
```

jsvob (approved):

JakubFrejlach (approved):

osoukup (approved):
LGTM

PR #600, opened 2024-06-24
Flaw comments respect is_private flag when created

jsvob:
Closes OSIDB-3003

### DIFF
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -409,9 +409,10 @@ class FlawBugzillaQueryBuilder(BugzillaQueryBuilder):
         pending_comments = FlawComment.objects.pending().filter(flaw=self.flaw)
 
         if pending_comments.exists():
+            comment = pending_comments.first()
             self._query["comment"] = {
-                "body": pending_comments.first().text,
-                "is_private": False,
+                "body": comment.text,
+                "is_private": comment.is_private,
             }
 
     def generate_fixed_in(self):
--- a/apps/bbsync/tests/test_query.py
+++ b/apps/bbsync/tests/test_query.py
@@ -1184,7 +1184,14 @@ class TestGenerateGroups:
 
 
 class TestGenerateComment:
-    def test_generate_comment(self):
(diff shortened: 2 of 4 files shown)


apps/bbsync/query.py
```diff
@@ -411,7 +411,7 @@ def generate_comment(self):
         if pending_comments.exists():
             self._query["comment"] = {
                 "body": pending_comments.first().text,
-                "is_private": False,
+                "is_private": pending_comments.first().is_private,
```
JakubFrejlach, line 414: I think this is not atomic and could potentially lead to a mismatch (text from one comment and is_private flag from another one) when new pending comment is created between the two `.first()` calls as Django queries are lazy.

Example with affects:
```python
In [31]: skynet_affects = Affect.objects.filter(ps_component__contains="skynet")

In [32]: skynet_affects.first()
Out[32]: <Affect: c6c31ad6-d57d-4419-9ccd-48474594ac05>

In [33]: # new affect with "skynet" ps_component created

In [34]: skynet_affects.first()
Out[34]: <Affect: 2e2dd85f-181a-4742-bd63-00d5b3142b84>
```

Querying for the comment just once should solve the problem:
```python
comment = pending_comments.first()
self._query["comments"] = {
...
```

osoukup: BBSync is currently triggered by an API call and Django should ensure atomic behavior but if we can make it not relying on that it would be better as it should soon become an async action triggered by backend worker.

osoukup (approved):
LGTM

apps/bbsync/query.py
jsvob: Fixed, thanks.

Just a note, if this is turned into an async action without the requisite cleanup of pending comments (now done synchronously during import from bugzilla), it will send seemingly random comments to bugzilla, because it just selects one of the pending ones.

osoukup: Oh :smile: that is probably not what would people expect but at least it would be never boring.

PR #601, opened 2024-06-24
Make Taskman service validate Jira token

costaconrado:
This PT add Jira token validation for Taskman service.

Closes OSIDB-2203.

### DIFF
--- a/apps/taskman/service.py
+++ b/apps/taskman/service.py
@@ -16,7 +16,11 @@ from collectors.jiraffe.core import JiraQuerier
 from osidb.models import Flaw, Impact, PsProduct
 
 from .constants import JIRA_TASKMAN_PROJECT_KEY, JIRA_TASKMAN_URL
-from .exceptions import JiraTaskErrorException, MissingJiraTokenException
+from .exceptions import (
+    JiraTaskErrorException,
+    MissingJiraTokenException,
+    TaskWritePermissionsException,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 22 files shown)


jsvob (approved):

apps/taskman/service.py
```diff
@@ -79,6 +83,19 @@ def __init__(self, token) -> None:
...
+        # This request raises exception for unauthenticated users
+        permission_response = self.jira_conn._session.get(token_validation_url)
+
+        permissions = permission_response.json().get(
+            "permissions", {"CREATE_ISSUES": {"havePermission": False}}
+        )
+        if not permissions["CREATE_ISSUES"]["havePermission"]:
+            raise TaskWritePermissionsException(
+                f"Token is valid for {JIRA_TASKMAN_URL} but user doesn't have write permission in {JIRA_TASKMAN_PROJECT_KEY} project."
+            )
```
osoukup, line 97: These safeguards are great! I still a bit wonder how it could happen before that the `taskman` could fail eg. on invalid token but then we proceeded to the flaw creation. I assume that such failure would happen a few lines below during some consecutive Jira call and these safeguards does not change anything there. It should however solve the current issues so definitely good to proceed with it.

osoukup (approved):
LGTM

costaconrado:
Note for the future:
 I've submitted a new commit with the cassettes rewritten as I mention in some dailies that were only possible after OSIDB-3029.
There is no changes in code (only some IDs in tests because of cassettes) since the last review you guys did so I'm not requesting a new review.

PR #602, opened 2024-06-25
Fix Flaw CC list builder to produce CCs in BZ format

JakubFrejlach:
This PR adds new ` bts_name_override` parameter to `BaseAffectCCBuilder` so we can easily override the CC format to Bugzilla even for Jira tracker PS modules. This is important for `BugzillaFlawCCBuilder` as currently it is generating a mixture of Bugzilla and Jira formated CCs which Bugzilla refuses to work with. This hot fix kinda violates the initial separation of concerns philosophy in the `osidb/cc.py` module, however this is probably the fastest way how to fix it and not spend more time refactoring it. In the future we would like to refactor this anyway (and also when leaving Bugzilla). 

Closes OSIDB-2985

### DIFF
--- a/apps/bbsync/tests/test_cc.py
+++ b/apps/bbsync/tests/test_cc.py
@@ -2,7 +2,7 @@ import json
 
 import pytest
 
-from apps.bbsync.cc import CCBuilder, RHSCLAffectCCBuilder
+from apps.bbsync.cc import AffectCCBuilder, CCBuilder, RHSCLAffectCCBuilder
 from apps.bbsync.constants import RHSCL_BTS_KEY, USER_BLACKLIST
 from apps.bbsync.tests.factories import BugzillaComponentFactory, BugzillaProductFactory
 from osidb.models import Affect, Flaw, PsModule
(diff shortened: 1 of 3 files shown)


jsvob (approved):
LGTM!

Sorry for the bug!

> This hot fix kinda violates the initial separation of concerns philosophy in the osidb/cc.py module,

Actually, I think it solves the issue cleanly. I thought the problem was two-dimensional (affect/flaw, jira/bugzilla) and that was the way cc.py was laid out. Turns out the problem is three-dimensional (affect/flaw, tracker goes to jira / tracker goes to bugzilla, CC list goes to jira / CC list goes to bugzilla) and the way you solved is is probably as clear as it would get.

> however this is probably the fastest way how to fix it and not spend more time refactoring it.

Agreed.

> In the future we would like to refactor this anyway (and also when leaving Bugzilla).

Not having bugzilla will simpliy things a lot. Before that, not much sense in refactoring it IMHO.

osoukup (approved):
LGTM

dmonzonis (approved):

PR #603, opened 2024-06-25
Allow creating Jira task for a flaw on demand

dmonzonis:
OSIDB-authored flaws should create a Jira task when created. But for old flaws inherited from SFM2, although they should be considered OSIDB-authored (they don't come from a collector), they don't have an associated Jira task and right now there is no way to create and associate a task for them. These flaws can be recognised as they should have an empty workflow state.

This commit adds the option to force the creation of a Jira task if the flaw does not have one. When updating a flaw through the API, setting the query parameter `create_jira_task` will trigger the creation of the Jira task if it does not exist, and set the `task_key` and workflow state fields.

Closes OSIDB-2882.

### DIFF
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -1659,6 +1659,13 @@ class FlawSerializer(
         if old_flaw.is_embargoed and self._is_public(new_flaw, validated_data):
             new_flaw.unembargo()
 
+        # Force Jira task creation if requested
+        request = self.context.get("request")
+        if request:
+            create_jira_task = request.query_params.get("create_jira_task")
+            if create_jira_task:
+                new_flaw.tasksync(jira_token=self.get_jira_token(), force_creation=True)
+
         # perform regular flaw update
         new_flaw = super().update(new_flaw, validated_data)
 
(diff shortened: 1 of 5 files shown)


MrMarble (approved):

osidb/serializer.py
```diff
@@ -1659,6 +1659,13 @@ def update(self, new_flaw, validated_data):
         if old_flaw.is_embargoed and self._is_public(new_flaw, validated_data):
             new_flaw.unembargo()
 
+        # Force Jira task creation if requested
+        request = self.context.get("request")
+        if request:
+            create_jira_task = request.query_params.get("create_jira_task")
```
JakubFrejlach, line 1665: This parameter should be added to OpenAPI schema via `@extend_schema_view` decorator in [Flaw API view](https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/api_views.py#L419)

JakubFrejlach (changes requested):
API schema should adjusted, otherwise LGTM.

osidb/serializer.py
jobselko, line 1: Could you please add some test?

JakubFrejlach (approved):
LGTM now!

jobselko (approved):

PR #607, opened 2024-06-26
run task collector every minute

osoukup:
Jira issues underlying the task management may potentially change frequently in there (unlike eg. project metadata) and each change like this might lead into a mid-air collision in OSIDB. Therefore I think that we should check for the changes frequently. Most of the time there will probably be nothing so it should be a low volume data sync anyway.


osoukup:
@costaconrado what do you think about this change? It seems to me that we should treat the tasks rather as trackers and not as metadata? I do not think that it would be an issue to check frequently but if people start touching them for some reason it might be an issue otherwise

PR #614, opened 2024-06-27
Fix test_create_jira_task_param randomly failing

dmonzonis:
This test used a mock for `create_or_update_task` which included a redundant `save` call to the flaw. This seemingly harmless extra save could create a discrepancy between the `updated_dt` passed as a parameter in the original PUT call and the `updated_dt` that the flaw has by the time it hits its actual save call.

Since the save is usually fast and `updated_dt` has a precision of seconds, the test usually passed, but sometimes there would be a discrepancy of one second and the test would fail.

This commit removes the redundant save, making sure the `updated_dt` stays consistent throughout the flaw update process and the test does not fail.


jsvob (approved):

osoukup (approved):
LGTM

PR #621, opened 2024-07-02
task collector fixes

osoukup:
Closes OSIDB-3083

### DIFF
--- a/collectors/jiraffe/tests/test_collectors.py
+++ b/collectors/jiraffe/tests/test_collectors.py
@@ -1,4 +1,5 @@
 import json
+import uuid
 
 import pytest
 from django.utils import timezone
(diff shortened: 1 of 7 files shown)


collectors/jiraffe/tests/test_collectors.py
```diff
@@ -82,6 +82,18 @@ def test_collect(self, monkeypatch):
         flaw = Flaw.objects.get(uuid=flaw.uuid)
         assert flaw.workflow_state == "TRIAGE"
 
+    @pytest.mark.vcr
+    def test_link_on_cve(self, monkeypatch):
+        monkeypatch.setattr(collectors, "JIRA_TOKEN", "SECRET")
+
+        # some random UUID
+        flaw = FlawFactory(cve_id="CVE-2024-34703")
```
costaconrado, line 90: I believe this don't test the cve linking since it will rely on UUID linking when CVE_ID fails, I believe you will have to change the UUID of the flaw from the database or from the labels in Jira to test this behavior

costaconrado (changes requested):

collectors/jiraffe/tests/test_collectors.py
osoukup: How probable it is that the generated UUID will be ` 9d9132a4-0484-48a5-b484-185abf39b771`? I can put there an assert to be absolutely sure but it seemed so rare to me that I did not bother with it

osoukup:
@costaconrado one more review please

costaconrado (approved):
LGTM!

PR #622, opened 2024-07-02
Implement message throttling in the API

dmonzonis:
This commit introduces message throttling for the prod environment, allowing only a certain amount of requests per hour for authenticated and unauthenticated users. The rates can be individually modified through env variables.

Closes OSIDB-894.

### DIFF
--- a/config/settings_prod.py
+++ b/config/settings_prod.py
@@ -132,6 +132,24 @@ CISA_COLLECTOR_CRONTAB = crontab(minute=0)
 LOG_FILE_SIZE = 1024 * 1024 * 10  # 10mb
 LOG_FILE_COUNT = 3
 
+# Global API throttle rates. Note that these can be customized per view if needed.
+ANON_API_THROTTLE_RATE = get_env("ANON_API_THROTTLE_RATE", default="10/minute")
+USER_API_THROTTLE_RATE = get_env("USER_API_THROTTLE_RATE", default="100/minute")
+
+# Set up message throttling in the API
+REST_FRAMEWORK.update(
+    {
+        "DEFAULT_THROTTLE_CLASSES": [
+            "rest_framework.throttling.AnonRateThrottle",
+            "rest_framework.throttling.UserRateThrottle",
+        ],
+        "DEFAULT_THROTTLE_RATES": {
+            "anon": ANON_API_THROTTLE_RATE,
+            "user": USER_API_THROTTLE_RATE,
+        },
+    }
+)
+
 # To not disrupt the logging of OSIDB instance running in PSI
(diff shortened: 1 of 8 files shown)


config/settings_prod.py
```diff
@@ -132,6 +132,24 @@
 LOG_FILE_SIZE = 1024 * 1024 * 10  # 10mb
 LOG_FILE_COUNT = 3
 
+# Global API throttle rates. Note that these can be customized per view if needed.
+ANON_API_THROTTLE_RATE = get_env("ANON_API_THROTTLE_RATE", default="10/minute")
+USER_API_THROTTLE_RATE = get_env("USER_API_THROTTLE_RATE", default="100/minute")
```
dmonzonis, line 137: These rates are up to discussion.
I chose what I believe are sane values as I don't see why a user would need to make more than 100 requests per minute. If there is some automated process that requires more, let me know, although if that is the case maybe the process needs to be redesigned so that we don't hit the API that much (e.g. bulk operations).

JakubFrejlach: @dmonzonis I am thinking about VM team scripts. I haven't seen all of them but minimally one of them is creating kernel Flaws and adding like 10 affects for each of those Flaws, that might potentially exceed that limit. This can be mitigated by using a bulk actions however they are not yet exposed in the osidb-bindings, I would need to prioritize that work.

jsvob (approved):

config/settings_prod.py
jsvob: I agree with @JakubFrejlach. I think one user could generate up to 400 requests per minute using this approach and lately maybe even more because it has been optimized I think. Given that there are flaws with 600 affects, I propose to set this to 600/minute and lower it later when the dust settles.

dmonzonis: Is the issue only with affects? I could set a higher limit for the affect API endpoint for the time being, instead of setting it globally.

osoukup: The bindings could ideally handle that and lower the load but it is some additional work of course

osoukup (approved):
LGTM. I think that the rates look reasonable. However I would maybe check with OSIM to see how many API requests eg. loading a flaw page means.

config/settings_prod.py
jsvob: Probably yes, I'm aware only of affects needing this. Mass tracker filing is probably slow enough not to hit the limit.

JimFuller-RedHat:
I missed this - if I had not missed this I would have asked 'why are we throttling at the app layer ?' eg. spending python cycles for every request where this could be handled by nginx ... there are various rh api gateways I guess we could contemplate but I have always found nginx sufficient (and also can dramatically speed things up). sorry for belated comment.

PR #626, opened 2024-07-03
fix duplicate comment issue leading in internal server error

osoukup:
This is probably a rare situation but today it lead to blocking on flaw sync and consequently to breaking the task collector. The main issue was the missing `return` statement.

Closes OSIDB-3086

### DIFF
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -611,6 +611,26 @@ class TestFlaw:
             assert f.major_incident_start_dt == timezone.now()
 
 
+class TestFlawComment:
+    def test_multiple_objects_returned(self):
+        """
+        test that the MultipleObjectsReturned case of create_flawcomment is handled well
+        this is the reproducer for https://issues.redhat.com/browse/OSIDB-3086
+        """
+        flaw = FlawFactory()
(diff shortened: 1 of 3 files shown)


JakubFrejlach (approved):
LGTM!

osidb/tests/test_flaw.py
```diff
@@ -611,6 +611,26 @@ def test_major_incident_start_dt(self):
             assert f.major_incident_start_dt == timezone.now()
 
 
+class TestFlawComment:
+    def test_multiple_objects_returned(self):
+        """
+        test that the MultipleObjectsReturned case of create_flawcomment is handled well
+        this is the reproducer for https://issues.redhat.com/browse/OSIDB-3086
```
jsvob, line 618: Internal URL.

```suggestion
        this is the reproducer for OSIDB-3086
```

osoukup: I already hit auto-merge and went home :smile: I agree that it is internal but I am a bit hesitant to consider it an issue. Red Hat Jira itself is public and issue ID does not hold any information above its existence so putting there only issue ID seems to me as providing completely the same information. Or do you think otherwise? Anyway we have a plenty of such cases throughout the code base.

PR #629, opened 2024-07-04
Correctly handle flaw comments with&without bzimport or bifurcated history

jsvob:
Closes OSIDB-3030


osoukup (approved):
This was more complicated then I though! LGTM

PR #631, opened 2024-07-08
temporarily move has trackers workflow requirement

osoukup:
so the community trackers absence does not block the embargoed flaws promotion before we can properly rework the workflows

Closes OSIDB-3098


costaconrado (approved):
LGTM.

PR #632, opened 2024-07-09
update TUTORIAL.md after the major release changes

osoukup:
as the desription is now comment_zero


jobselko (approved):
LGTM. A similar issue is also in `DEVELOP.md`, but it would take some time to address everything now.

PR #634, opened 2024-07-10
Allow setting empty impact value

jobselko:
This PR removes impact validation and thus enables saving a flaw with an empty impact. This logic should be revisited as part of the workflow rework.

Closes OSIDB-3128

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1095,16 +1095,6 @@ class Flaw(
             else:
                 raise err
 
-    def _validate_nonempty_impact(self):
-        """
-        check that the impact is not empty
-
-        we cannot enforce this by model definition
-        as the old flaws may have no impact
-        """
-        if not self.impact:
-            raise ValidationError("Impact value is required.")
-
     def _validate_nonempty_components(self):
         """
         check that the component list is not empty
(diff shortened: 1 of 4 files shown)


osidb/models.py
```diff
@@ -1103,7 +1103,7 @@ def _validate_nonempty_impact(self):
         as the old flaws may have no impact
         """
         if not self.impact:
-            raise ValidationError("Impact value is required.")
+            self.alert("empty_impact", "Impact value is required.")
```
osoukup, line 1106: I am not sure this is the correct approach. According to the conversation which triggered this task there are valid cases when the flaw has no impact. Then the alert is false positive and it will make people ignore it. As we do not have better solution now (should come as part of workflow rework) I would be for leaving the responsibility of providing the value on users and just not alert at all.

jobselko: Updated, please re-review.

osoukup (changes requested):
Definitely better but one more detail :smile: I am checking the workflow definitions and this https://github.com/RedHatProductSecurity/osidb/blob/master/apps/workflows/workflows/default.yml#L33 would not accept the empty impact value

osoukup (approved):
LGTM

PR #635, opened 2024-07-10
Workaround for possible race condition

skontar:
In sync manager, it should never happen that sync task is started without it being scheduled (based on saved timestamps).

However, it sometimes happens for some reason. To workaround the issue, ignore those sync managers in re-scheduling logic.

Workaround for OSIDB-3131

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -180,6 +180,17 @@ class SyncManager(models.Model):
         """
         for sync_manager in cls.objects.all():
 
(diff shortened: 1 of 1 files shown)


jsvob (approved):
LGTM for a stop-gap solution. I don't see a logical issue that this would create in the resulting state. We discussed how to fix it for long-term resolution.

osoukup (approved):
LGTM

JakubFrejlach (approved):

PR #636, opened 2024-07-11
prepare conditional BBSync flaw sync-back

osoukup:
Implementing the pre-async cleaup of imports from BZ based on introduced env var to enable switch into an asynchronous way of synchronizing flaws.

Closes OSIDB-3104
Closes OSIDB-3105


costaconrado (approved):
LGTM!

PR #638, opened 2024-07-11
Duplicate Alerts created concurrently in multiple threads handled correctly

jsvob:
Related to OSIDB-3048

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -7,7 +7,7 @@ from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelatio
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres import fields
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.db import models
+from django.db import IntegrityError, models
 from django.utils import timezone
 
 from osidb.exceptions import DataInconsistencyException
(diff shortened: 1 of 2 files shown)


osidb/mixins.py
```diff
@@ -657,16 +657,47 @@ def alert(
...
+                object_id=self.uuid,
+                content_type=ContentType.objects.get_for_model(self),
+                description=description,
+                alert_type=alert_type,
+                resolution_steps=resolution_steps,
+                acl_read=acl_read,
+                acl_write=acl_write,
+            ).save()
+        except IntegrityError:
+            pass
```
jobselko, line 700: I am not sure if writing general information into code brings any value. The code becomes longer, and the important part is buried under less important parts. If I understood correctly, the only necessary information is why there is `pass` in `except`. And that could be simplified as follows:
```
        except IntegrityError:
            # skip as alerts of the same name have the same meaning
            pass
```

jsvob: OK, changed the comment as you suggested, but kept the change in a separate commit that I'd like _not_ being squashed. The reason is that this was a footgun we spent days on and we're not exactly experts on distributed computing. If there's an issue in it, there's a chance the explanation in git history will help. At least I personally use `git log -S` and other ways of traversing history for pieces of code difficult to understand, and it would help at least me, if I knew this.

JakubFrejlach: I think the simplified comment completely throws away the reason why it was done. If I would be someone without context I would have no idea what is going on and why it matters that same name alerts have same meaning.

JakubFrejlach (approved):
LGTM! Made a small comment about the explanation comment but I will leave it up to you.

osidb/mixins.py
jobselko: I am okay with having it more detailed in a reasonable way, but a paragraph with 20 lines seems to me to be too much.

osidb/mixins.py
```diff
@@ -657,16 +657,47 @@ def alert(
...
+        # Since all Alerts of the same name have practically the same meaning,
+        # it is fine to give up in this thread without retrying to rewrite the
+        # just-created Alert by practically the same information.
+        #
+        # Not using select_for_update() because we don't care if an alert (row) is
+        # overwritten, because all such versions have the same meaning.
+        #
+        # References:
+        # - https://www.postgresql.org/docs/16/index-unique-checks.html#:~:text=At%20present%2C%20only-,b%2Dtree,-supports%20it.)%20Columns
+        # - https://www.postgresql.org/docs/16/locking-indexes.html#:~:text=B%2Dtree%2C,without%20deadlock%20conditions.
```
jobselko: We use postgresql version 13, not 16.

JakubFrejlach: Just locally, stage/prod/uat are on aurora anyway :smile:

jsvob: It explains core concepts that are practically timeless anyway. This hasn't changed for 22 years, which is the furthest into the past the relevant documentation can be found quickly. ([1](https://www.postgresql.org/docs/7.2/indexes-unique.html#:~:text=only%20B%2Dtree%20indexes%20can%20be%20declared%20unique), [2](https://www.postgresql.org/docs/7.2/locking-indexes.html#:~:text=B%2Dtree%20indexes,for%20concurrent%20applications.))

osidb/mixins.py
jsvob: The original @jobselko 's feedback indicated a slight misunderstanding caused by a low information density of the original block of comments (it would be `alerts of the same name and age have the same meaning` without the slight misunderstanding).

Since both @JakubFrejlach and I prefer to have this footgun documented, I rewrote the block of comments with higher information density.

Now I'm merging it so that I can continue working on OSIDB-3048 (Blocker priority) and test it on stage. Feel free to submit a follow-up PR.

PR #639, opened 2024-07-11
Make sync managers less prone to race conditions

skontar:
Change the sync managers code so it is making  changes directly in the query and does not hold and refresh its objects in memory. This should avoid race conditions when updating database objects.

OSIDB-3131

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -53,84 +53,92 @@ class SyncManager(models.Model):
         """
         raise NotImplementedError("Update synced links not implemented.")
 
-    def schedule(self):
+    @classmethod
+    def schedule(cls, sync_id):
         """
         Schedule sync_task to Celery queue.
+
+        :param sync_id: Unique ID for synchronized data object.
         """
-        with transaction.atomic():
-            self.refresh_from_db()
-            self.last_scheduled_dt = timezone.now()
-            self.save()
+        cls.objects.get_or_create(sync_id=sync_id)
+        cls.objects.filter(sync_id=sync_id).update(last_scheduled_dt=timezone.now())
         try:
-            self.sync_task.apply_async(args=[self.sync_id])
+            cls.sync_task.apply_async(args=[sync_id])
         except AttributeError:
             raise NotImplementedError(
                 "Sync task not implemented or not implemented as Celery task."
             )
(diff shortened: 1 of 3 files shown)


osidb/sync_manager.py
```diff
@@ -53,84 +53,92 @@ def update_synced_links(self):
...
         except AttributeError:
             raise NotImplementedError(
                 "Sync task not implemented or not implemented as Celery task."
             )
-        logger.info(f"{self.__class__.__name__} {self.sync_id}: Sync scheduled")
+        logger.info(f"{cls.__name__} {sync_id}: Sync scheduled")
 
-    def started(self, celery_task):
+    @classmethod
+    def started(cls, sync_id, celery_task):
```
jsvob, line 74: Observation: This is still theoretically prone to a race condition where multiple workers in parallel pick up the task, from the local perspective of each the task is not started, and both let it start.

Shouldn't be a problem, hopefully:
- I hope that celery is smart enough not to let two workers pick the exact same task at the same time.
- Therefore it could only happen if it was scheduled multiple times. That hopefully doesn't happen normally based on the structure of the callers.

jsvob (approved):
I like the pattern of CRUD usage, neatly using the database's internal locking for update atomicity where it matters, and leaving the rest very neat, _New Jersey style_.

There's a miniscule chance of race conditions in most of the methods if two tasks with the same id are executed almost exactly in sync in two threads, but for most of them the impact is having the resulting fields being overwritten by the "wrong" thread (e.g. having `last_consecutive_*` off by one) and for `started()` the risk is a very miniscule chance of being started in parallel and then perhaps failing during the processing, with probably negligible impact thanks to the reschedule functionality. It might be possible to implement it absolutely correct, but it would get unwieldy. Also this level of scrutiny is inspired only by dealing with concurrency issues lately, but most of OSIDB has slight race conditions that don't matter and implementing the [ostrich algorithm](https://en.wikipedia.org/wiki/Ostrich_algorithm) is right.

It is very clean how it is and IMO it will resolve all concurrency issues we've actually observed. LGTM.

skontar:
I do not think that in Celery a task is supposed to ever by picked and executed twice, that would make it horrible asynchronous "queue" processor.

PR #641, opened 2024-07-12
Sensible Bugzilla errors from API

JakubFrejlach:
This PR:
* handles any BugzillaError exception when thrown during any API requests and returns it as 422 response with the exception text (full traceback is not included) instead of basic 500 internal server error

Closes OSIDB-3126


jsvob (approved):

osoukup (approved):
LGTM

PR #642, opened 2024-07-12
Do not try to work in a failed transaction

jsvob:
Related to OSIDB-3048

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -7,7 +7,7 @@ from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelatio
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres import fields
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
-from django.db import IntegrityError, models
+from django.db import IntegrityError, models, transaction
 from django.utils import timezone
 
 from osidb.exceptions import DataInconsistencyException
@@ -685,6 +685,10 @@ class AlertMixin(ValidateMixin):
         # constraint gracefully (if the row doesn't exist yet, there's nothing
         # to lock).
         #
+        # transaction.atomic() is there because after IntegrityError, the
+        # transaction can't continue and OSIDB uses ATOMIC_REQUESTS, see
+        # https://stackoverflow.com/a/48836554 for use of nested tx.
+        #
         # References:
         # - https://www.postgresql.org/docs/16/index-unique-checks.html#:~:text=At%20present%2C%20only-,b%2Dtree,-supports%20it.)%20Columns
         # - https://www.postgresql.org/docs/16/locking-indexes.html#:~:text=B%2Dtree%2C,without%20deadlock%20conditions.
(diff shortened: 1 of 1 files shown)


jvaught-rh:
I was reading the changes for my own fun. This is making a save point atomic? Which means it rolls back if an exception is thrown instead of committing something partially?

jsvob:
> I was reading the changes for my own fun. This is making a save point atomic? Which means it rolls back if an exception is thrown instead of committing something partially?

If the save happens without an exception, it is guaranteed to be atomic. If theoretically two updates to the same Alert were happening concurrently and theoretically the DB didn't serialize the updates (which it most probably does), now it would be guaranteed that all the fields of the updated Alert are set by a single thread, thus being mutually consistent. Because DBs serialize updates to the same row, having or not having the `with transaction.atomic()` block here doesn't matter.

However, if the creation of a new Alert fails with `IntegrityError` because a concurrent thread created it a miniscule amount of time earlier (which sadly happens about 50% of the time on stage, we should probably spend a lot of money on lottery tickets), the transaction within which the `IntegrityError` was raised is not allowed to receive further database operations. Before this PR, it was attempted (through the `except` and `pass`) and so the database protested and the whole transaction (the whole request handler) errored out with `TransactionManagementError`. Because OSIDB wraps the whole request handlers in transactions that take the whole time the request is handled.

(The creation of two new identical Alerts in two concurrent thread fails with `IntegrityError` because there's a `UniqueConstraint` on the `Alert` model. That's in there from previous attempts at handling the issue that we have multiple threads that try to create the same Alert at the same time, and each from their own local perspective think they are creating the first instance ever, and only later it assembles into multiple instances. I didn't want to work around it by deleting the extra instances in the same place they are created/updated because when we weeks or months later start using multiple DB replicas, reads will be delayed about 20ms after writes, and so we should minimize code that does both reading and writing at the same time.)

So with this PR, there's a small transaction within the large (request) transaction, so that when the creation of the Alert fails because a concurrent Celery thread _just_ created the same Alert, only the small transaction is let to abort, and the overall request handler is allowed to proceed. This piece of code is called both from Celery during bzimport and from POST and PUT request handlers, e.g. when creating new Affects.

costaconrado:
@jvaught-rh if I understood correctly, this would do the opposite you said, we would treat the alert creation as a transaction, but because it is inside a try-catch, if it fail we don't update/create the alert but still save the flaw. 

@jsvob if it only happens in the scenario described in the comments (same name, description, parent object), this is a fair solution but I believe we should in the future figure why multiple threads are being called for the same user request, this behavior is not only confusing but also very hard to do maintenance.

costaconrado (approved):
LGTM.

jsvob:
@costaconrado Thanks, I agree that it is confusing and we should try to make it so that it doesn't happen. AFAIK it happens with the following coincidence:

- A client adds/edits an Affect or something that generates an Alert (in web pod).
- BZImport (running in Celery) notices an update of the flaw and imports it, during which it generates the same alert.
- A client adds/edits another Affect, which manipulates the same Flaw, and generates the same alert.

If the client doesn't issue requests rapidly one after another, it doesn't happen.

If BZImport doesn't run, it doesn't happen. On UAT, it's fine. Flower logs show concurrent work of Celery on the very same flaw that the client just edited.

AFAIK alerts are created by the models' structure in such a deeply baked-in way, there's little chance for us to rework it.

Most importantly, we'll disable BZImport soon. We can then remove the transaction and even the try/except, because it would be then theoretically only created by multiple clients manipulating the same object at the same time, so it would be fine if one of them fails with any error (be it IntegrityError, TransactionManagementError, or anything else).

PR #643, opened 2024-07-16
Make sync managers unique based on sync_id

skontar:


jsvob (approved):
I find the migration fine as long as we either deploy to prod everything before it **and** this migration, or nothing from sync managers at all. (Otherwise, it would need the part that deletes all duplicates.)

LGTM

PR #647, opened 2024-07-22
ignore issue if query is outdated

costaconrado:
This PR changes task collector in order to ignore outdated issues.

Currently task collector get all tasks updated in the last couple minutes and sequentially update the flaws, but because it runs every minute, sometimes users updates will be reverted since the query were performed before the user updated. This PR fix this behavior by adding the Jira timestamp in WorkflowModel and comparing them before changing flaw.

Closes OSIDB-3085.


dmonzonis (approved):
LGTM

PR #648, opened 2024-07-24
Alert lockfree duplicate handling

jsvob:
This is more of a question than something useful.

When two clients edit the same Flaw (or Affects of the same Flaw), they create a deadlock with a console message that looks like this

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.DeadlockDetected: deadlock detected
DETAIL:  Process 1180 waits for ShareLock on transaction 9013684; blocked by process 1187.
Process 1187 waits for ShareLock on transaction 9012362; blocked by process 1180.
HINT:  See server log for query details.
CONTEXT:  while updating tuple (15939,175) in relation "osidb_affect"
```

I thought that's because each thread deletes and replaces the Alert objects, so I created this PR. However, the PR, which minimizes changes done to Alerts, behaves the same.

(Alerts are created by `_validation_*` methods and each of them has only one mode of operation: Create an Alert if there's something wrong. So Alerts were updated by deleting all Alerts related to the given model (Flaw, Affect, etc.) and calling the `_validation_*` methods again.)

Any idea how to allow two clients work with the same Flaw without deadlocks?

(I haven't yet tried to use postgresql to see what it can tell me about the deadlocks, because I don't know anything about how to work with postgres yet.)

Related to OSIDB-3048


jsvob:
This is irrelevant. Noticed `affect` is not `alert` but my brain read it wrong until now. Tests show that if there are two transactions that edit two rows in the DB in the opposite order, they generate a deadlock. (Shown in internal Slack.) It's currently unavoidable. And has nothing to do with Alerts (it can happen on any model, it did happen a few times on Alerts, but also on Affects as seen in the error message above).

PR #649, opened 2024-07-24
Handle OperationalError when thrown during concurrent requests

jsvob:
An example of a response a client sees in case of an aborted transaction due to a deadlock (generated by OSIDB when `DEBUG = False`):

```
{"detail":"[OperationalError] Concurrent access to the same model by you and someone else. For example, two clients editing the same Flaw. Exception details: deadlock detected\nDETAIL:  Process 423 waits for ShareLock on transaction 59013; blocked by process 415.\nProcess 415 waits for ShareLock on transaction 61660; blocked by process 423.\nHINT:  See server log for query details.\nCONTEXT:  while updating tuple (362,9) in relation \"osidb_alert\"\n","dt":"2024-07-24T18:10:45.589941Z","revision":"unknown","version":"4.1.2","env":"local"}
```

Closes OSIDB-3048

### DIFF
--- a/osidb/exception_handlers.py
+++ b/osidb/exception_handlers.py
@@ -1,5 +1,6 @@
 from bugzilla.exceptions import BugzillaError
 from django.core.exceptions import ValidationError as DjangoValidationError
+from django.db.utils import OperationalError
 from rest_framework import status
 from rest_framework.exceptions import ValidationError as DRFValidationError
 from rest_framework.response import Response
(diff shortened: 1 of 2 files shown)


osidb/exception_handlers.py
```diff
@@ -23,4 +24,12 @@ def exception_handler(exc, context):
 
     if isinstance(exc, DjangoValidationError):
         exc = DRFValidationError(as_serializer_error(exc))
+
+    if isinstance(exc, OperationalError):
+        data = {
+            "detail": "[OperationalError] Concurrent access to the same model by you and someone else. For example, two clients editing the same Flaw. Exception details: "
+            + str(exc)
```
jobselko, line 31: I am not sure if every `OperationalError` is caused by concurrent access. Can we not use only output from `exc` instead of creating our own?

osoukup: We could check the error message for some keyword (deadlock?) to be certain that it is the concurrent access issue. Otherwise it would probably be an implementation error so re-raising might be OK.

osoukup (approved):
LGTM

osidb/exception_handlers.py
jsvob: Good point. Turns out it can happen with other DB errors too. Thanks! Fixed.

PR #650, opened 2024-07-25
Make OSV flaw creation independent of BZ data

jobselko:
This PR updates the way how OSV creates new flaws. `meta_attr` is now created directly during the flaw creation and not by BZ sync, making the code independent of BZ.
(If `cve_id` is absent, `external_id` from `meta_attr` is used to determine if a flaw already exists)

Closes OSIDB-3173


osoukup (approved):
LGTM

PR #651, opened 2024-07-25
UnackedHandler only recommends active unacked streams (OSIDB-3160)

costaconrado:
This PR modifies UnackedHandler so it only recommends unacked streams when listed in active streams.

Closes OSIDB-3160.


costaconrado:
I decided for removing the UnackedHandler instead of disabling it, in case we need it in future I can restore it from the history in git.

osoukup (approved):
LGTM

PR #653, opened 2024-07-25
Make flaw draft creation independent of BZ data

jobselko:
This PR simplifies checking whether a flaw is a draft. It removes the dependency on full BZ sync and only checks the state of the associated Jira task and BZ ID, which should be sufficient.

This PR also included the ACLs change when promoting/rejecting a flaw. However, I struggled with tests, so @costaconrado helped me finish this change in #654. Thank you one more time for all your help, I really appreciate it!

Closes OSIDB-3175


osoukup (approved):
LGTM

PR #656, opened 2024-07-29
add contract priority description in trackers

costaconrado:
This PR adds a description and a FAQ url into trackers tagged as `contract-priority`.

Closes OSIDB-3165.

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -101,7 +101,7 @@ class TrackerQueryBuilder:
             if len(summary) <= MAX_SUMMARY_LENGTH:
                 break
 
-            cves, description = self._summary_shorten(cves, description)
+            cves, description = TrackerQueryBuilder._summary_shorten(cves, description)
 
         return summary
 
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM

apps/trackers/common.py
```diff
@@ -329,6 +333,19 @@ def _description_embargoed(self):
 
         return description_parts
 
+    def _description_contract_priority(self):
```
dmonzonis, line 336: Can this be a static method?

apps/trackers/tests/test_common.py
```diff
@@ -448,6 +448,48 @@ def test_bugzilla_virtualizaiton(self, ps_component):
...
+
+        tqb = TrackerQueryBuilder()
+        tqb.instance = tracker
+
+        description = """\
+Contract priorities are scoped by at the \
+product/version level and include all severities \
+and all components within that product/version. \
+All severities (including low) must be resolved \
+within the allotted SLA time."""
```
jobselko, line 483: You can use sth like this to keep the indentation (this will not include any extra blanks or newlines):

```
description = (
    "word1"
    "word2"
)
```

apps/trackers/common.py
costaconrado: sure, I've changed all methods that can be static in this class, in a new commit, please take a look again.

apps/trackers/tests/test_common.py
costaconrado: I've fixed, I've also changed other tests in this file to follow the same pattern in a separated commit, please take  a look again.

PR #657, opened 2024-07-29
Keep the save to Jira and BZ separate for collector flaws

jobselko:
This PR fixes the partial save for collector flaws caused by the asynchronous BZ. Separating the save for Jira and BZ enables Jira changes to be done immediately.

Fixes OSIDB-3188


jsvob (approved):
LGTM but I don't understand it enough.

PR #658, opened 2024-07-29
remove sync from Bugzilla from the async sync to Bugzilla

osoukup:
Closes OSIDB-3199

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Duplicate Alerts created concurrently in multiple threads handled correctly (OSIDB-3048)
 - Make task collector ignore outdated issues (OSIDB-3085)
 - Allow Flaw API to proper unassign owner in Jira (OSIDB-3145)
+- Remove sync from Bugzilla from the async sync to Bugzilla (OSIDB-3199)
 
 ## [4.1.3] - 2024-07-25
 ### Changed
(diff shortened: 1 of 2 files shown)


osoukup:
@jsvob so this is the solution I am proposing. I went the other way than you proposed to make `_perform_bzsync` basically what the async sync needs only and leaving the conditional branching in `bzsync`. The idea is that the **else** branch will eventually go away but `_perform_bzsync` will be kept using by the `sync_manager`.

I still need to properly test it as started with the empty DB and a queue full of scheduled tasks so the on demand schedules are somewhere far in the queue. I have to learn how to test the new collectors.

osoukup:
I turned off all the collectors and was finally able to test the change! The empty task queue makes it significantly easier to get scheduled immediately and see it clearly :smile:

jsvob (approved):
LGTM.

Tested locally with FLAW_COLLECTOR_ENABLED=0; BBSYNC_SYNC_FLAWS_TO_BZ_ASYNCHRONOUSLY=1 and other collectors disabled to have an otherwise empty celery queue.

Now the duplicated tasks (while the duplication will be dealt with separately under OSIDB-3190) all fail only with

> BugzillaError("Red Hat Bugzilla Stage's database reported a query serialization error. Most likely this occurred because another user or process attempted to change the same data that you were attempting to change.

And I don't see any more of deadlocks or the client being blocked on suddenly-slow requests.

I was a little worried about the unconditional save below `# save in case a new Bugzilla ID was obtained` but it apparently creates no issues whatsoever.

Thank you!

PR #659, opened 2024-07-29
Avoid parallelism-induced duplicately-executed tasks.

jsvob:
Related to OSIDB-3198

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -542,6 +542,78 @@ class BZSyncManager(SyncManager):
     Sync manager class for OSIDB => Bugzilla synchronization.
     """
 
+    @classmethod
+    def schedule(cls, sync_id, *args, **kwargs):
+        """
+        Schedule BZSyncManager's sync_task to Celery queue.
+
+        TODO:
+          - This is a workaround. SyncManager.schedule should not
+            decide to skip or postpone tasks, but we need it to:
+            - Reduce noise in flower logs that would be generated
+              by duplicate tasks failing.
+            - Reduce amount of requests sent needlessly to Bugzilla.
+            - Avoid sending potentially incomplete data immediately.
+          - Move the deduplication logic to Flaw.bzsync:
+            - No BZSyncManager-related action upon client request.
+            - A periodic task to query recently changed tasks once per
+              X minutes and call BZSyncManager.schedule for them.
+            - BZSyncManager.schedule to be left not implemented,
+              inherited SyncManager.schedule used.
+          - OSIDB-3205
+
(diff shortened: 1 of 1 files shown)


jsvob:
This shouldn't be scheduled just yet. I'll fix a slight regression regarding updating `last_scheduled_dt` introduced in 870dc3d56aba0b8159735858414e840ba7f654a0 (didn't notice it on review) and made worse by this PR. And @skontar should probably review / we'll have to discuss it on a 1:1.

jsvob:
@skontar The regression regarding updating last_scheduled_dt and an extraneous save fixed in the revised commit. Seems to work fine.

osoukup (commented):
I am not sure how much draft this is :smile_cat: but so far LGTM

jsvob:
@skontar @osoukup please re-review.

osidb/sync_manager.py
```diff
@@ -18,6 +18,16 @@ class SyncManager(models.Model):
...
     tasks were processed, and re-scheduling tasks when necessary.
+
+    The philosophy:
+    - Provide visibility (when and how many times a task ran or failed, why it failed).
+    - Straightforward (short implementation, easy to understand, errors not obscured
+                       or masked by complicated logic).
+    - Queue in order (SyncManager tries to send tasks to Celery without reordering,
+                      which happens only on retries).
+    - The queued tasks should be idempotent or close to idempotent
+        (it shouldn't matter if a task fails or if it runs multiple twice, the end
```
osoukup, line 29: nitpick: I like the phrase `multiple twice` a lot :smile: but I think it is not what was supposed to be there

osoukup (approved):
LGTM. Nicely and clearly structured!

costaconrado (approved):
LGTM.

osidb/sync_manager.py
```diff
@@ -542,6 +542,77 @@ class BZSyncManager(SyncManager):
     Sync manager class for OSIDB => Bugzilla synchronization.
     """
 
+    @classmethod
+    def schedule(cls, sync_id, *args, **kwargs):
+        """
+        Schedule BZSyncManager's sync_task to Celery queue.
+
+        TODO:
+          - This is a workaround. SyncManager.schedule should not
```
jobselko, line 551: If this is only a workaround, we should create a new Jira to track the correct change so we do not forget about it.

jsvob: Filed as OSIDB-3205 and it will probably become a part of a new epic based on our future team talks about parallelism-related design changes.

skontar:
For the record, I looked at it and LGTM, just forgot to set the review state.

PR #662, opened 2024-07-30
Keep the save to Jira and BZ separate everywhere

jobselko:
This PR is a follow-up to #657. The save to Jira and BZ should be always separate.

Fixes OSIDB-3203


osoukup (approved):
LGTM

PR #664, opened 2024-07-31
Log exceptions in exception handler

JakubFrejlach:
This PR:
* adds exception logging in exception handler so we can preserve the full exception with traceback in splunk

### DIFF
--- a/osidb/exception_handlers.py
+++ b/osidb/exception_handlers.py
@@ -1,3 +1,5 @@
+import logging
+
 from bugzilla.exceptions import BugzillaError
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db.utils import OperationalError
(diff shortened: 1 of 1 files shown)


jobselko (approved):
LGTM, but looks more like a technical change

osidb/exception_handlers.py
```diff
@@ -39,4 +44,7 @@ def exception_handler(exc, context):
             }
             return Response(data, status=status.HTTP_409_CONFLICT)
 
+    if isinstance(exc, DjangoValidationError):
+        exc = DRFValidationError(as_serializer_error(exc))
```
jobselko, line 48: Just a question, but why is there no logger?

JakubFrejlach: as this just pre-process exception and passes it down the `drf_exception_handler` at the end of the exception_handler function, this handler then handles the exception accordingly, normal API errors are handled with the response to the user and are usually not worth logging I guess and anything unknown and special will be reraised as internal server error 500 which raises exception anyway.

JakubFrejlach: We can theoretically log the exception everytime, but this might result in some duplicities, not sure whether to do this or not

jobselko: Up to you :)

jobselko: We can add it later to check the results on stage (and keep production clean for now).

costaconrado (approved):
LGTM.

osoukup (approved):
LGTMk

PR #667, opened 2024-07-31
Prevent bbsync (OSIDB->BZ) from saving data back to OSIDB

jsvob:
Prevent bbsync (OSIDB->BZ) from saving data back to OSIDB unless absolutely necessary.


osoukup (approved):
LGTM

PR #668, opened 2024-08-01
Revert message throttling in the API

JakubFrejlach:
This PR reverts changes introduced in #622 and thus it reverts OSIDB-894.

The revert is being done because on OSIDB production, users are being limited and logged out because the IP is used rather than user id for cache.

From @MrMarble investigation:
* _The throttle middleware uses the user id as cache and the IP as fallback_
* _Its called for all `ApiView` classes_
* _the `TokenRefreshView` does not have access to the session, so even if the user is logged in, it will use the IP for cache_


dmonzonis (approved):

MrMarble (approved):

osoukup (approved):
LGTM

PR #666, opened 2024-07-31
rollback transaction when handling exception

osoukup:
This was probably causing number of issues but will no more.

PR #669, opened 2024-08-01
Proper rollback on atomic requets

JakubFrejlach:
This PR improves changes introduced in #666 so the rollback is performed more safely and the behaviour matches DRF rollback.


jobselko (approved):

osoukup (approved):
LGTM

PR #670, opened 2024-08-01
remove forgotten flaw collector dependencies

osoukup:
... so that the collectors does not wait forever for something which does not happen any more.


JakubFrejlach (approved):
LGTM!

collectors/jiraffe/collectors.py
```diff
@@ -147,15 +147,6 @@ def get_batch(self) -> (List[Issue], timezone.datetime):
...
         period_end = period_start + timezone.timedelta(days=self.BATCH_PERIOD_DAYS)
-        # the tracker collector should never outrun the flaw one
-        # since it creates the linkage which might then be missed
-        period_end = min(
-            period_end,
-            CollectorMetadata.objects.get(
-                name="collectors.bzimport.tasks.flaw_collector"
-            ).updated_until_dt
-            or period_end,
-        )
```
MrMarble, line 158: Do we have to do the same here https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/bzimport/collectors.py#L607-L613 or is not needed?

jobselko: I think this is needed when running the collector manually.

osoukup: Very good catch! This should be removed too

MrMarble (approved):

PR #673, opened 2024-08-01
Cannot modify CVE of existing flaw

dmonzonis:
Since commit 654f09ddd69769b1b479773a51678002fc3e7ef1, the serializer order for flaws changed, now running the Jira serializer before the Bugzilla sync one. This causes the flaw instance to be updated to the database before the bzsync is run, but the bzsync relied on the old instance not being updated yet, and therefore the bzsync changes that required the old instance for comparison are not being applied correctly anymore.

With this commit, instead of relying on the instance not being saved by the time bzsync is run, we copy the old instance without saving it to the db, and pass it to the bzsync for it to use as needed.

Closes OSIDB-3102.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Fixed
 - Cannot fill trackers concurrently (OSIDB-3230)
+- Cannot modify CVE of existing flaws (OSIDB-3102)
 
 ## [4.1.5] - 2024-08-01
 ### Removed
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1326,19 +1326,19 @@ class Flaw(
 
             # sync to Bugzilla
             bs = FlawBugzillaSaver(self, bz_api_key)
-            self = bs.save(old_instance)
+            flaw_instance = bs.save(old_instance)
             kwargs[
                 "auto_timestamps"
             ] = False  # the timestamps will be get from Bugzilla
             kwargs["raise_validation_error"] = False  # the validations were already run
             # save in case a new Bugzilla ID was obtained
-            self.save(*args, **kwargs)
+            flaw_instance.save(*args, **kwargs)
 
             # fetch from Bugzilla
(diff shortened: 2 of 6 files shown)


osoukup (approved):
LGTM

docs/CHANGELOG.md
jobselko, line 1: Please rebase changelog

osidb/models.py
```diff
@@ -1350,7 +1352,8 @@ def _perform_bzsync(self, *args, bz_api_key, **kwargs):
 
         # sync to Bugzilla
         bs = FlawBugzillaSaver(self, bz_api_key)  # prepare data for save to BZ
-        self = bs.save()  # actually send to BZ (but not save to DB)
+        old_instance = kwargs.pop("old_instance", None)
+        self = bs.save(old_instance)  # actually send to BZ (but not save to DB)
```
jsvob, line 1356: What is `self` now?

Technically, `self` is one of the parameters. It's a variable that exists in this scope only. Assigning anything to it just changes the value. The original object referenced by the previous reference saved in the variable named `self` is not changed.

Before this PR, this weird construction conveyed (or so I hoped, since it was there for quite some time) that `bs.save()` actually returns the identical reference to the object referenced by `self` that was passed to `FlawBugzillaSaver(self, ...`.

Assigning something else than the original reference to the variable named `self` is technically valid in Python, but semantically it's confusing. If we do that, we can also create variables that shadow built-in functions. 

While I didn't find a PEP that would directly discourage this, there's https://peps.python.org/pep-0008/#function-and-method-arguments that says _"Always use `self` for the first argument to instance methods."_.

This is a slippery slope :) We don't want ending up with writing programs that have something like this in it :)

```
int = 1
id = 2
str = 3
bool = 4
len = 5
all = 6
list = 7
dict = 8
print = 9
type = 10
open = 11
```

So...

```suggestion
        flaw_instance = bs.save(old_instance)  # actually send to BZ (but not save to DB)
```
... and then later use that for the rest of the calls that use `self`

dmonzonis: Agreed, assigning to self is very confusing

jsvob: Thank you!

PR #677, opened 2024-08-05
Gin index for Alert acl_read

JakubFrejlach:
This PR:
* adds Gin index for Alert RBS field `acl_read` for better performance

Note: VACUUM ANALYZE should be issued after Gin index is applied to respective instances

Closes OSIDB-3218


JimFuller-RedHat (approved):
LGTM

PR #679, opened 2024-08-06
Move SLA policies to ps_constants

dmonzonis:
Instead of having the SLA policies in a file in OSIDB's codebase, the file has been moved to the ps_constants repo, and the ps_constants collector in OSIDB now picks the SLA policies from there.

To facilitate the collection of the SLA policies into OSIDB, some SLA related objects have been converted to Django models, so that the data from the collector can be stored and updated as with other ps_constants data.

Closes OSIDB-3220.


dmonzonis:
This PR requires that https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/merge_requests/93 is merged first, or the collector will not be able to find the SLA definitions in the ps_constants repo.

apps/sla/tests/test_framework.py
```diff
@@ -179,13 +184,13 @@ def test_single(self):
...
-                sla_framework = SLAFramework()
-                sla_context = sla_framework.classify(tracker)
+            load_sla_policies(sla_file)
+            sla_context = sla_classify(tracker)
 
-                assert sla_context.sla
-                assert sla_context.start == flaw.created_dt
-                assert sla_context.end == flaw.created_dt + timedelta(days=5)
+            assert sla_context.sla
+            print(sla_context.sla)
```
jsvob, line 191: Leftover `print`

collectors/ps_constants/core.py
```diff
@@ -88,3 +95,15 @@ def sync_special_consideration_packages(sc_packages):
...
         package = SpecialConsiderationPackage(name=name)
         package.save()
+
+
+@transaction.atomic
+def sync_sla_policies(sla_policies):
+    """
+    Sync SLA policy data
+    """
+    SLAPolicy.objects.all().delete()
```
jsvob, line 105: What happens if ps_constants become unavailable? Will all SLA policies be deleted? I realize this is in principle a part you didn't change, but I think this now became `ps_constants_collector`'s biggest responsibility that would pose a bigger issue if the state was deleted than the constants before it.

jsvob (commented):
Some nitpicks, otherwise looks good to me, but someone else should review too. This is complex and IDK enough about SLAs.

apps/sla/tests/test_framework.py
dmonzonis: Oops

collectors/ps_constants/core.py
dmonzonis: The data is fetched by `fetch_ps_constants`, and if there is an error either in the request or loading the YAML, this function will raise an error and the `ps_constants_collector` will not continue.

jsvob (approved):

apps/sla/models.py
```diff
@@ -2,18 +2,33 @@
...
 
+    DURATION_TYPES = [
+        ("business days", "Business Days"),
+        ("calendar days", "Calendar Days"),
+    ]
+
+    START_CHOICES = [
+        ("earliest", "Earliest"),
+        ("latest", "Latest"),
+    ]
```
JakubFrejlach, line 30: Just a question, why not to utilize `models.TextChoices` in this case?

apps/sla/models.py
```diff
@@ -24,39 +39,50 @@ class SLA:
...
-        """
-        # allow the parsering of the shorter
-        # style definitions the same way
-        if isinstance(start_desc, str):
-            start_desc = {"latest": [start_desc]}
+    duration = models.IntegerField()
+    duration_type = models.CharField(max_length=20, choices=DURATION_TYPES)
+    start_criteria = models.CharField(max_length=20, choices=START_CHOICES)
+    start_dates = fields.ArrayField(
+        models.CharField(max_length=100), default=list, null=True
```
JakubFrejlach, line 46: Is it expected that `start_dates` can be NULL (None)? It might be sufficient to go with just `blank=True` and always count on empty list.

JakubFrejlach (approved):
Just couple of questions, otherwise LGTM!

apps/sla/models.py
osoukup: A valid policy must have some start date defined (not saying that it must be necessarily required here).

apps/sla/migrations/0001_initial.py
```diff
@@ -0,0 +1,85 @@
...
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("name", models.CharField(max_length=100)),
```
osoukup, line 68: The name should be unique so maybe we could set it as the DB restriction?

apps/sla/migrations/0001_initial.py
```diff
@@ -0,0 +1,85 @@
...
+                ("description", models.TextField()),
+                ("condition_descriptions", models.JSONField(default=dict)),
+                ("order", models.IntegerField()),
```
osoukup, line 71: The order must also be unique.

apps/sla/models.py
```diff
@@ -156,22 +190,47 @@ class SLAPolicy:
...
+    def create_from_description(self, policy_desc, order=None):
+        """Creates an SLA policy from a YAML description."""
+        name = policy_desc["name"]
+        description = policy_desc["description"]
+        sla = SLA.create_from_description(policy_desc["sla"])
+        sla.save()
+
+        if order is None:
+            # Order is implied by the number of already existing SLA policies
+            order = SLAPolicy.objects.count()
```
osoukup, line 213: This sounds to me a bit random (maybe it is not). The order must correspond to the vertical ordering of the SLA policies in the file as it is used for prioritization.

osoukup (approved):
LGTM

apps/sla/models.py
dmonzonis: You're right, `TextChoices` is more appropriate here

apps/sla/models.py
dmonzonis: Actually if the SLA is well defined it should neither be blank or null, so I'll just remove the `null=True`

apps/sla/models.py
dmonzonis: They are loaded sequentially from the YAML in the same order they are defined, so the order is maintained this way.

PR #682, opened 2024-08-07
OSIDB-3248: Object of type Flaw is not JSON serializable

MrMarble:
>[!IMPORTANT]
> This needs to be merged before #673 is deployed to production
> unless we want to break bugzilla synchronization 

### Summary

`BZSyncManager` schedules a task to synchronize the flaw changes to bugzilla asynchronously:
https://github.com/RedHatProductSecurity/osidb/blob/d96e3dac5bd102901ad53734eb6adb64cc4ec622/osidb/sync_manager.py#L600-L606
on #673 the serializer was modified to add an attribute `old_instance` that is used to detect if we are creating a new flaw or updating an existing one:
https://github.com/RedHatProductSecurity/osidb/blob/c6287f30d685f78fc8c5bc1e16acecf34ac743a7/osidb/serializer.py#L873-L875
This parameter is an instance of `Flaw` model that is sent as part of the message to the celery queue for future processing 
image
> Example of the args 

By default, celery uses `json` serialization, but a Django Model is not json serializable, hence the error, this means that the synchronization to bugzilla fails, normaly silently since this error is thrown on the worker, but sometimes, if the workers are idle, when a user updates a flaw, the  `BZSyncManager.schedule_task` is called right away and the **error is propagated to the user**

#### Pickle (Chosen)
Celery has support for the `pickle` serialization, that allows to serialize almost everything, including Django models.
Enabling the `pickle` serializer in a **root container** is discouraged for security reasons as the pickled could be executable, this can be overriden by the environment variable `C_FORCE_ROOT` but instead I opted for creating an user inside the docker image for running the application. 

#### JSON
If we want to keep using `json` serialization we need to rewrite part (if not all) of the `FlawBugzillaQueryBuilder` to use a `dict` instead of a `FlawModel`.
We can't utilize the `FlawSerializer` on the `BZSyncManager` as it creates a circular dependency, also not all fields of the flaw are needed, so the best approach would be to create a simplified json dict with the required parameters and use them on `FlawBugzillaQueryBuilder`

Closes OSIDB-3248


JimFuller-RedHat:
initial impressions - this whole area needs a bit more thinking ... I would include @skontar on these conversations and at least wait for @osoukup to comment.

dmonzonis (commented):
Initially the old flaw was passed as an argument to avoid having to touch code in the query builder which used it. But since this is a problem due to having to serialize the whole flaw for the celery task, I believe we should try a different approach, like extracting only the info we need for the query builder (CVE, some meta attributes, whether the flaw existed...) and passing that instead of the flaw, and modify the query builder to correctly collect those parameters instead of relying on the old flaw.

osoukup:
This seems to me to introduce quite dramatic changes only because we need JSON serialization. The idea of the query builder was that it compares the old and the new model instance to see what has changed. Probably this does not even work correctly any more - eg. compare new CC list with the old one when we do not fetch it from Bugzilla at all... I think that the query builder changes would be more appropriate and maybe having a UUID as a Celery task argument would then be enough.

osoukup:
@MrMarble this one is being solved differently by https://github.com/RedHatProductSecurity/osidb/pull/695 so the PR can be closed

PR #684, opened 2024-08-09
Allow start dates to come from multiple sources in SLA

dmonzonis:
Before this commit, the start dates in the SLA were always taken from
the flaw. With this commit, now the SLA file can specify where the date
comes from: flaw, affect or tracker.

For example, this is now valid in the SLA file:

```
latest:
  tracker:
    - created date
  flaw:
    - unembargo date
```

and it will pick the latest date from either the tracker's creation date
or the flaw's unembargo date.

Closes OSIDB-3221.


osoukup (approved):
LGTM

PR #654, opened 2024-07-25
make internal flaws public on promote

costaconrado:
This PR change Workflows in order to only allow internal flaws in new/draft state and rejected flaws.

PR #685, opened 2024-08-09
Update ACLs of linked objects to match collector flaw

jobselko:
This PR is a follow-up to #654 and ensures that all related objects of a collector flaw will have adjusted ACLs if necessary after the flaw is promoted.

This will fix a situation like when an affect had internal ACLs because it was created for an internal (not promoted) flaw.

Closes OSIDB-3253

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -210,6 +210,10 @@ class ACLMixin(models.Model):
     def is_internal(self):
         return set(self.acl_read + self.acl_write) == self.acls_internal
 
+    @property
+    def is_public(self):
+        return set(self.acl_read + self.acl_write) == self.acls_public
+
     def acl2group(self, acl):
         """
         transform back to human readable group name or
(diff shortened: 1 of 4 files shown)


osidb/mixins.py
```diff
@@ -555,6 +559,46 @@ def unembargo(self):
...
+                for related in self._meta.related_objects
+                # only the models with ACLs other than "snippets" are subject of this
+                if (
+                    issubclass(related.related_model, ACLMixin)
+                    and related.related_name != "snippets"
+                )
+            ]
+        ):
+            # continue deeper into the related context
+            related_instance.set_public_nested()
```
osoukup, line 600: Correct me if I am wrong but this seems to potentially change the tracker visibility too. There are however PS modules (and it is quite frequent - eg. https://gitlab.cee.redhat.com/prodsec/product-definitions/-/blob/master/data/platforms/ps_modules.json?ref_type=heads#L92) where even when the flaw is public the trackers should stay internal. Just blindly setting all the related stuff public would break this assumption.

osoukup (changes requested):

osidb/mixins.py
osoukup: So after discussing this I found out that my comment was actually not relevant. As there is no internal visibility level defined for the trackers in Bugzilla/Jira we just use the public one from the product definitions. Then, the change from internal to public in OSIDB is actually a no-op for the tracker visibility in there.

osoukup (approved):
LGTM

PR #686, opened 2024-08-09
Validations improvements

costaconrado:
This PR changes ValidateMixin to allow skip validations in between mixins.

Most endpoints were recreating alerts multiple times, for instance `/osidb/api/v1/flaw`  recreates 17x the Alerts for the Flaw being created. Similar behavior can be noticed an other models using ValidateMixin like Affect and Trackers. This PR ensures that saves needed for OneToMany relationships or external sync can be performed calling the validate method as `dry_run`.

Closes OSIDB-3108.


osoukup (approved):
LGTM

jsvob (approved):

PR #688, opened 2024-08-12
Ignore up-to-date collected trackers

costaconrado:
This PR make Trackers not run validations for the parent Flaw.

This PR also adds a manual call to duplication validation in Tracker::save method to avoid having two equal entries in BTS in case multiple workers use the same entity at the same time.

This PR also changes how collectors and converters work for Trackers, only calling the `save` method in case a Tracker doesn't already have the latest BTS timestamp.

This PR also do changes in serializer so Trackers are created with default fake date (1/1/1970) that will be replaced the first time the collectors or the sync manager get the BTS timestamp.

Closes OSIDB-3244.
Closes OSIDB-3234.


osidb/serializer.py
```diff
@@ -687,8 +688,12 @@ def create(self, validated_data):
 
         # create new tracker instance in the local DB first
         # so we can create the links before the backend sync
-        # alerts will be created after links
-        tracker.save(no_alerts=True)
+        # alerts will be created after links and timestamp
+        # should be generated by BTS and will be fixed during sync
```
jsvob, line 692: Nitpick: Is there an extra word in the comment?

jsvob (approved):
Very nice! LGTM! Thank you!

Didn't test it and someone else should review as well.

osidb/models.py
```diff
@@ -2424,6 +2424,10 @@ def save(self, *args, bz_api_key=None, jira_token=None, **kwargs):
             and bz_api_key is not None
             and self.type == self.TrackerType.BUGZILLA
         ):
+            # avoid creating creating tracker in duplicity
```
osoukup, line 2427: nitpick: double `creating`

osoukup (approved):
LGTM. Good job!

osidb/serializer.py
costaconrado: Message improved, thanks.

osidb/models.py
costaconrado: Duplicated word removed, thanks.

PR #690, opened 2024-08-13
Fix migration in SLA app

dmonzonis:
Fix error in migration which causes it to fail if there is existing SLA data.

Related to OSIDB-3221.

### DIFF
--- a/apps/sla/migrations/0002_alter_sla_start_dates.py
+++ b/apps/sla/migrations/0002_alter_sla_start_dates.py
@@ -2,14 +2,13 @@
 
 import json
 
-import django.contrib.postgres.fields.jsonb
 from django.db import migrations, models
 
 
 def convert_array_to_json(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


costaconrado (approved):
LGTM.

apps/sla/migrations/0002_alter_sla_start_dates.py
```diff
@@ -2,14 +2,13 @@
...
 
-import django.contrib.postgres.fields.jsonb
 from django.db import migrations, models
 
 
 def convert_array_to_json(apps, schema_editor):
     SLA = apps.get_model("sla", "SLA")
     for instance in SLA.objects.all():
-        instance.temp_json_field = json.dumps(instance.array_field)
+        instance.temp_json_field = json.dumps(instance.start_dates)
```
JakubFrejlach, line 11: Can you please elaborate? I don't understand at all. `.array_field` would work in some cases?

PR #694, opened 2024-08-15
OSIDB-3281: Move DEFER from historical to current possible affect resolution

superbuggy:
Move DEFER from historical to current possible affect resolution, disallows filing trackers for such affects.

Closes OSIDB-3290. Closes OSIDB-3330.

### DIFF
--- a/apps/trackers/tests/test_file_offer.py
+++ b/apps/trackers/tests/test_file_offer.py
@@ -40,7 +40,21 @@ class TestTrackerSuggestions:
                 Affect.AffectResolution.WONTFIX,
                 False,
             ),
-            (Affect.AffectAffectedness.NEW, Affect.AffectResolution.WONTFIX, False),
+            (
+                Affect.AffectAffectedness.NEW,
+                Affect.AffectResolution.WONTFIX,
+                False,
+            ),
+            (
+                Affect.AffectAffectedness.AFFECTED,
+                Affect.AffectResolution.DEFER,
+                False,
+            ),
+            (
+                Affect.AffectAffectedness.NEW,
+                Affect.AffectResolution.DEFER,
+                False,
+            ),
         ],
     )
     def test_trackers_file_offer_invalid(
--- a/apps/workflows/tests/test_models.py
+++ b/apps/workflows/tests/test_models.py
@@ -253,6 +253,11 @@ class TestCheck:
             ),
             (
                 Affect.AffectAffectedness.NEW,
+                Affect.AffectResolution.DEFER,
+                False,
+            ),
+            (
+                Affect.AffectAffectedness.NEW,
                 Affect.AffectResolution.OOSS,
                 False,
             ),
@@ -264,6 +269,11 @@ class TestCheck:
             (
                 Affect.AffectAffectedness.AFFECTED,
                 Affect.AffectResolution.WONTFIX,
+                False,
+            ),
+            (
+                Affect.AffectAffectedness.AFFECTED,
+                Affect.AffectResolution.DEFER,
                 False,
             ),
             (
@@ -298,6 +308,7 @@ class TestCheck:
 
         affect = AffectFactory(
             flaw=flaw,
+            impact=Impact.LOW,
             affectedness=affectedness,
             resolution=resolution,
         )
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use OSIDB Bugzilla service account API key for majority of bzsync
   instead of user ones (OSIDB-3261)
 - Adjust synchronous bzsync to only work one-way
+- Move DEFER from historical to current possible affect resolution (OSIDB-3281)
 
 ### Fixed
 - Cannot modify CVE of existing flaws (OSIDB-3102)
--- a/osidb/constants.py
+++ b/osidb/constants.py
@@ -61,11 +61,13 @@ SERVICES_PRODUCTS = [
 AFFECTEDNESS_VALID_RESOLUTIONS = {
     Affect.AffectAffectedness.NEW: [
         Affect.AffectResolution.NOVALUE,
+        Affect.AffectResolution.DEFER,
         Affect.AffectResolution.WONTFIX,
         Affect.AffectResolution.OOSS,
     ],
     Affect.AffectAffectedness.AFFECTED: [
         Affect.AffectResolution.DELEGATED,
+        Affect.AffectResolution.DEFER,
         Affect.AffectResolution.WONTFIX,
         Affect.AffectResolution.OOSS,
     ],
@@ -77,12 +79,9 @@ AFFECTEDNESS_VALID_RESOLUTIONS = {
 
 # Historical affectedness/resolution combinations that were valid in the past
 AFFECTEDNESS_HISTORICAL_VALID_RESOLUTIONS = {
-    Affect.AffectAffectedness.NEW: [
-        Affect.AffectResolution.DEFER,
-    ],
+    Affect.AffectAffectedness.NEW: [],
     Affect.AffectAffectedness.AFFECTED: [
         Affect.AffectResolution.FIX,
-        Affect.AffectResolution.DEFER,
         Affect.AffectResolution.WONTREPORT,
     ],
     Affect.AffectAffectedness.NOTAFFECTED: [],
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1893,6 +1893,19 @@ class Affect(
                 "WONTFIX but has open tracker(s).",
             )
 
+    def _validate_defer_open_trackers(self, **kwargs):
+        """
+        Prevent an affect with open trackers from having a resolution of DEFER
+        """
+        if (
+            self.resolution == Affect.AffectResolution.DEFER
+            and self.trackers.exclude(status__iexact="CLOSED").exists()
+        ):
+            raise ValidationError(
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} cannot have "
+                "resolution DEFER with open tracker(s).",
+            )
+
     def _validate_unknown_component(self, **kwargs):
         """
         Alerts that a flaw affects a component not tracked in Bugzilla.
@@ -2528,6 +2541,16 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
         if not self.is_closed and affect:
             raise ValidationError(
                 f"The tracker is associated with a WONTFIX affect: {affect.uuid}",
+            )
+
+    def _validate_defer_open_tracker(self, **kwargs):
+        """
+        Check whether DEFER affects have open trackers.
+        """
+        affect = self.affects.filter(resolution=Affect.AffectResolution.DEFER).first()
+        if not self.is_closed and affect:
+            raise ValidationError(
+                f"The tracker is associated with a DEFER affect: {affect.uuid}",
             )
 
     def _validate_multi_flaw_tracker(self, **kwargs):
--- a/osidb/tests/test_flaw.py
+++ b/osidb/tests/test_flaw.py
@@ -141,14 +141,14 @@ class TestFlaw:
         assert affect1 in all_affects
         assert affect2 in all_affects
 
-        comment1 = FlawCommentFactory(flaw=vuln_1)
+        comment1 = FlawCommentFactory(flaw=vuln_1, order=0)
         comment2 = FlawComment(
             flaw=vuln_1,
             external_system_id="9999991",
             acl_read=self.acl_read,
             acl_write=self.acl_write,
             text="some comment text",
-            order=0,
+            order=1,
         )
         comment2.save()
         all_comments = vuln_1.comments.all()
@@ -367,6 +367,12 @@ class TestFlaw:
 
         new_affect = AffectFactory(affectedness=Affect.AffectAffectedness.NEW)
         assert new_affect.delegated_resolution is None
+        undelegated_affect = AffectFactory(
+            affectedness=Affect.AffectAffectedness.AFFECTED,
+            resolution=Affect.AffectResolution.DEFER,
+        )
+        assert undelegated_affect.delegated_resolution is None
+
         undelegated_affect = AffectFactory(
             affectedness=Affect.AffectAffectedness.AFFECTED,
             resolution=Affect.AffectResolution.WONTFIX,
@@ -1841,6 +1847,14 @@ class TestFlawValidators:
                 Affect.AffectAffectedness.AFFECTED,
                 Affect.AffectAffectedness.AFFECTED,
                 Affect.AffectResolution.DELEGATED,
+                Affect.AffectResolution.DEFER,
+                False,
+                False,
+            ),
+            (
+                Affect.AffectAffectedness.AFFECTED,
+                Affect.AffectAffectedness.AFFECTED,
+                Affect.AffectResolution.DELEGATED,
                 Affect.AffectResolution.FIX,
                 True,
                 False,
@@ -1890,7 +1904,7 @@ class TestFlawValidators:
 
         affect.affectedness = affectedness_new
         affect.resolution = resolution_new
-        affect.impact = Impact.CRITICAL
+
         if should_raise:
             with pytest.raises(
                 ValidationError,
@@ -1911,11 +1925,10 @@ class TestFlawValidators:
         "affectedness,resolution,should_raise",
         [
             (Affect.AffectAffectedness.NEW, Affect.AffectResolution.NOVALUE, False),
-            (Affect.AffectAffectedness.NEW, Affect.AffectResolution.DEFER, True),
+            (Affect.AffectAffectedness.NEW, Affect.AffectResolution.DEFER, False),
             (Affect.AffectAffectedness.NEW, Affect.AffectResolution.WONTFIX, False),
             (Affect.AffectAffectedness.NEW, Affect.AffectResolution.OOSS, False),
             (Affect.AffectAffectedness.AFFECTED, Affect.AffectResolution.FIX, True),
-            (Affect.AffectAffectedness.AFFECTED, Affect.AffectResolution.DEFER, True),
             (
                 Affect.AffectAffectedness.AFFECTED,
                 Affect.AffectResolution.DELEGATED,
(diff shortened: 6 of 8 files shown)


osidb/models.py
```diff
@@ -1872,6 +1887,16 @@ def _validate_affect_status_resolution(self, **kwargs):
...
+        
+        if (
+            self.resolution == Affect.AffectResolution.DEFER
+            and self.impact != Impact.LOW
+        ):
+            raise ValidationError(
+                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has resolution "
+                f"{self.resolution} {Affect.AffectResolution.DEFER}, but has {self.impact} impact/severity."
+                f"{Affect.AffectResolution.DEFER} is only valid on affects with {Impact.LOW} impact/severity."
+            )
```
osoukup: This second check seems duplicite to me. Also the it extends the validation scope as it was originally supposed to only check the status and resolution combinations - not impact.

osidb/constants.py
```diff
@@ -61,12 +61,12 @@
 AFFECTEDNESS_VALID_RESOLUTIONS = {
     Affect.AffectAffectedness.NEW: [
         Affect.AffectResolution.NOVALUE,
-        Affect.AffectResolution.WONTFIX,
+        Affect.AffectResolution.DEFER,
         Affect.AffectResolution.OOSS,
     ],
     Affect.AffectAffectedness.AFFECTED: [
         Affect.AffectResolution.DELEGATED,
-        Affect.AffectResolution.WONTFIX,
```
osoukup, line 69: I am not aware of that we should stop using WONTFIX. Is there any reference about it?

osoukup (changes requested):
I have a few comments I would like to clarify or address but generally this goes the right direction

osidb/models.py
superbuggy: Indeed, I had written a separate dedicated validation where I separated this logic but hadn't deleted my original code here when I first created the PR. I have since added a commit to remove it to this PR which I can squash later in the review process.

osidb/constants.py
superbuggy: It was requested and documented OSIDB-3829, but this might be better to tackle in a later PR

superbuggy: typo: OSIDB-3289

osoukup: Then it is OK

osoukup (approved):
I got explanation for what I missed the context for and now LGTM

superbuggy:
I am noticing that `apps/bbsync/tests/test_cc.py::TestCCBuilder::test_notaffected` doesn't have deterministic behavior on my local, is this a known issue for anyone else?

osidb/models.py
```diff
@@ -1918,6 +1933,19 @@ def _validate_wontfix_open_tracker(self, **kwargs):
                 "WONTFIX but has open tracker(s).",
             )
 
+    def _validate_defer_open_trackers(self, **kwargs):
+        """
+        Prevent an affect with trackers from having a resolution of DEFER
```
jsvob, line 1938: nitpick
```suggestion
        Prevent an affect with open trackers from having a resolution of DEFER
```

osoukup:
> I am noticing that `apps/bbsync/tests/test_cc.py::TestCCBuilder::test_notaffected` doesn't have deterministic behavior on my local, is this a known issue for anyone else?
> 
> https://github.com/RedHatProductSecurity/osidb/blob/64bb15d67157cc5eb8300ec0ea5d193d9322fd32/apps/bbsync/tests/test_cc.py#L175-L203
> 
> This error will occur sometimes but not others
> 
> ```
> E       AssertionError: assert ['you@redhat.com'] == []
> E         Left contains one more item: 'you@redhat.com'
> E         Use -v to get the full diff
> 
> apps/bbsync/tests/test_cc.py:203: AssertionError
> ```
> 
> for
> 
> https://github.com/RedHatProductSecurity/osidb/blob/64bb15d67157cc5eb8300ec0ea5d193d9322fd32/apps/bbsync/tests/test_cc.py#L203

I have not noticed this but the module is getting quite a rework in https://github.com/RedHatProductSecurity/osidb/pull/695

jsvob (approved):

apps/bbsync/tests/test_integration.py
```diff
@@ -773,14 +775,16 @@ def test_affect_update(self, auth_client, test_api_uri):
             HTTP_BUGZILLA_API_KEY="SECRET",
             HTTP_JIRA_API_KEY="SECRET",
         )
+        print(response.content)
```
osoukup: There seems to be forgotten debugging `print`

apps/trackers/tests/test_file_offer.py
```diff
@@ -63,6 +71,7 @@ def test_trackers_file_offer_invalid(
         ps_module_regular = PsModuleFactory(name="regular-module")
         affect = AffectFactory(
             flaw=flaw,
+            impact=impact,
```
osoukup, line 74: The impact does not seem to have any effect on the test. Could there be just LOW hard-coded here? Or is there any reason for the parametrizing?

apps/workflows/tests/test_models.py
```diff
@@ -298,6 +305,7 @@ def test_has_trackers(self, affectedness, resolution, trackers_required):
 
         affect = AffectFactory(
             flaw=flaw,
+            impact=impact,
```
osoukup, line 308: Hard-code the parameter here instead of providing a parametrization which is actually the same all the time

osidb/models.py
```diff
@@ -1887,16 +1887,6 @@ def _validate_affect_status_resolution(self, **kwargs):
...
-        
-        if (
-            self.resolution == Affect.AffectResolution.DEFER
-            and self.impact != Impact.LOW
-        ):
-            raise ValidationError(
-                f"Affect ({self.uuid}) for {self.ps_module}/{self.ps_component} has resolution "
-                f"{self.resolution} {Affect.AffectResolution.DEFER}, but has {self.impact} impact/severity."
-                f"{Affect.AffectResolution.DEFER} is only valid on affects with {Impact.LOW} impact/severity."
-            )
```
osoukup: This should be rebased

osidb/models.py
```diff
@@ -2704,11 +2692,19 @@ def fix_state(self):
             self.resolution = self.resolution.lower()
 
         # Eg. GITOPS-1472, AAH-682
-        if self.status in ("won't fix", "obsolete"):
-            return Affect.AffectFix.WONTFIX
+        if self.status in ("won't fix", "obsolete", "deferred"):
+            # won't fix should display/should handle as as DEFER
+            return Affect.AffectFix.DEFER
         if self.status in ("done", "resolved", "closed"):
-            if self.resolution in ("won't do", "won't fix", "wontfix", "obsolete"):
```
osoukup, line 2660: This mapping is wrong. We should just take the values from BTS as they were. There is no change in the meaning of the old or BTS values. Still **won't fix** etc. means WONTFIX even though it is historical value now. Additionally, we are changing the value which then go to the CVE pages - the new and the historical. This should be all reverted back

osidb/tests/test_flaw.py
```diff
@@ -2065,7 +2079,13 @@ def test_validate_ooss_open_tracker(
             ps_module=ps_module.name,
             affectedness=Affect.AffectAffectedness.AFFECTED,
             resolution=resolution,
+            impact=Impact.LOW if resolution == Affect.AffectResolution.DEFER else "",
         )
+
+        # DEFER resoltuion affects must have LOW impact to successfully validate
+        if resolution == Affect.AffectResolution.DEFER:
+            affect.impact = Impact.LOW
```
osoukup, line 2087: There are two checks for the DEFER resolution here. One should be enough

osidb/tests/test_flaw.py
```diff
@@ -2129,9 +2150,9 @@ def test_validate_wontfix_open_tracker(
         tracker.affects.add(affect)
 
         match = (
-            f"The tracker is associated with a WONTFIX affect: {affect.uuid}"
+            f"The tracker is associated with a {Affect.AffectResolution.DEFER} affect: {affect.uuid}"
             if entity == "tracker"
-            else f"{affect.uuid}.*is marked as.*but has open tracker"
+            else f"{affect.uuid}.*cannot have resolution {Affect.AffectResolution.DEFER} with open tracker"
```
osoukup, line 2155: I wonder what is the meaning of `.*` there. There used to be two like it was some highlight but we do not have any such mechanism. It should be probably replaced by a space

apps/bbsync/tests/cassettes/test_integration/TestBBSyncIntegration.test_affect_update.yaml
```diff
@@ -774,7 +774,7 @@ interactions:
         "If docs needed, set a value", "cf_internal_whiteboard": "", "target_release":
         ["---"], "cf_conditional_nak": [], "target_milestone": "---", "cf_srtnotes":
         "{\"affects\": [{\"ps_module\": \"rhel-8\", \"ps_component\": \"kernel\",
-        \"affectedness\": \"affected\", \"resolution\": \"wontfix\", \"impact\": \"important\",
```
osoukup, line 777: These changes should be squashed when we later remove the cassette

apps/bbsync/tests/test_cc.py
```diff
@@ -179,12 +179,14 @@ def test_notaffected(self, cc, embargoed):
...
         flaw = FlawFactory(embargoed=embargoed)
         affect1 = AffectFactory(
-            flaw=flaw, affectedness=Affect.AffectAffectedness.NOTAFFECTED
+            flaw=flaw,
+            affectedness=Affect.AffectAffectedness.NOTAFFECTED,
         )
         affect2 = AffectFactory(
             flaw=flaw,
             affectedness=Affect.AffectAffectedness.AFFECTED,
-            resolution=Affect.AffectResolution.WONTFIX,
```
osoukup, line 187: This test is being modified in two commits for no apparent reason

osoukup (changes requested):
I re-review the changes and found additional issues. Also the changes should be squashed so we do not introduce updates which are then just deleted. The commits must be signed before the PR can be merged.

apps/trackers/tests/test_file_offer.py
superbuggy: There is a validation to permit only LOW impact flaws with DEFER resolution, and thought parameterizing was better than hardcoding for one special case, and better for diversity of test data.

apps/bbsync/tests/test_integration.py
superbuggy: This was removed in a prior commit.

osidb/models.py
superbuggy: There was a request to change the data value to 'DEFER' and that there was not a value to having the historical data. See OSIDB-3289 for more details.

osidb/tests/test_flaw.py
superbuggy: As I understand it, the meaning is any number of characters in between the affect uuid and the character sequence `cannot have resolution` and so on

apps/bbsync/tests/test_cc.py
superbuggy: These have been squashed.

osidb/models.py
superbuggy: This requirement has reverted, WONTFIX is to be supported. Will revert this.

apps/exploits/tests/test_api.py
```diff
@@ -61,14 +61,16 @@ def load_data():
         ps_component="ps-component-2",
     )
     AffectFactory(
-        resolution=Affect.AffectResolution.WONTFIX,
```
osoukup, line 64: Is there any reason to change WONTFIX to DEFER here?

apps/exploits/tests/test_api.py
```diff
@@ -61,14 +61,16 @@ def load_data():
...
+        resolution=Affect.AffectResolution.DEFER,
+        impact=Impact.LOW,
         affectedness=Affect.AffectAffectedness.AFFECTED,
         flaw=f3,
         ps_module="rhel-6",
         ps_component="firefox",
     )
     AffectFactory(
-        resolution=Affect.AffectResolution.WONTFIX,
```
osoukup, line 71: Same here.

apps/exploits/tests/test_api.py
```diff
@@ -353,7 +355,7 @@ def test_exploits_flaw_data(self, auth_client):
                         "ps_module": "rhel-6",
                         "ps_component": "firefox",
                         "affectedness": "AFFECTED",
-                        "resolution": "WONTFIX",
```
osoukup, line 356: Same here.

apps/exploits/tests/test_api.py
```diff
@@ -365,7 +367,7 @@ def test_exploits_flaw_data(self, auth_client):
                         "ps_module": "rhel-5",
                         "ps_component": "kernel",
                         "affectedness": "AFFECTED",
-                        "resolution": "WONTFIX",
```
osoukup, line 368: Same here.

apps/trackers/tests/test_file_offer.py
```diff
@@ -40,7 +40,11 @@ class TestTrackerSuggestions:
                 Affect.AffectResolution.WONTFIX,
                 False,
             ),
-            (Affect.AffectAffectedness.NEW, Affect.AffectResolution.WONTFIX, False),
```
osoukup, line 43: Same here.

apps/trackers/tests/test_file_offer.py
```diff
@@ -40,7 +40,11 @@ class TestTrackerSuggestions:
...
+            (
+                Affect.AffectAffectedness.NEW,
+                Affect.AffectResolution.DEFER,
+                False,
```
osoukup, line 46: It makes sense to test the DEFER resolution but we should just add it and not replace the WONTFIX case.

apps/trackers/tests/test_file_offer.py
osoukup: It is already hard-coded, just with LOW for one case and MODERATE for the rest. It is the same number of options as if it was all just LOW and one extra condition above. Is MODERATE important for the other cases?

docs/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow start dates to come from multiple sources in SLA (OSIDB-3221)
 - Update public date for collector flaws (OSIDB-3212)
 - Tracker collector ignores up-to-date entries (OSIDB-3244)
+- Retire WONTFIX handle its associated values with DEFER (OSIDB-3281)
```
osoukup, line 17: At the end we do not retire WONTFIX

osidb/models.py
```diff
@@ -2653,7 +2691,7 @@ def fix_state(self):
         if self.resolution:
             self.resolution = self.resolution.lower()
 
-        # Eg. GITOPS-1472, AAH-682
+            # Eg. GITOPS-1472, AAH-682
```
osoukup, line 2694: There does not seem to be a reason of this change and it shifts the comment on and incorrect indention

osidb/tests/test_flaw.py
```diff
@@ -1890,7 +1906,11 @@ def test_validate_historical_affectedness_resolution(
 
         affect.affectedness = affectedness_new
         affect.resolution = resolution_new
-        affect.impact = Impact.CRITICAL
+        affect.impact = (
+            Impact.LOW
+            if resolution_new == Affect.AffectResolution.DEFER
+            else Impact.CRITICAL
+        )
```
osoukup, line 1913: There does not seem to be a reason for any specific impact so LOW all the time should be OK

osidb/tests/test_flaw.py
```diff
@@ -2065,7 +2098,9 @@ def test_validate_ooss_open_tracker(
             ps_module=ps_module.name,
             affectedness=Affect.AffectAffectedness.AFFECTED,
             resolution=resolution,
+            impact=Impact.LOW if resolution == Affect.AffectResolution.DEFER else "",
```
osoukup, line 2101: Wouldn't just LOW do the job? An empty impact is rather invalid value

osidb/tests/test_flaw.py
osoukup: On I see! You are correct

osoukup (changes requested):
The commit description (and ideally also the PR one) should be updated. I have still a few comments mostly to the tests. The core functionality looks OK

apps/trackers/tests/test_file_offer.py
superbuggy: I don't think it is. After this last round of changes, I removed impact from the parameters after all.

apps/exploits/tests/test_api.py
superbuggy: Do you think there is no value to doing so? If not, I will remove. I figured it made sense to test a similar affect resolution status here.

osoukup: I mean I am not against testing DEFER too, it just no makes sense to me to stop testing WONTFIX when we keep using it. If you prefer testing DEFER, just introduce another test(s) or parametrize the existing

apps/trackers/tests/test_file_offer.py
osoukup: I mean if it is just directly set to LOW there does not have to be any **if else** conditioning

docs/CHANGELOG.md
superbuggy: Updated

apps/exploits/tests/test_api.py
superbuggy: Ah here I'm testing both

```py
    AffectFactory(
        resolution=Affect.AffectResolution.WONTFIX,
        affectedness=Affect.AffectAffectedness.AFFECTED,
        flaw=f2,
        ps_module="ps-module-2",
        ps_component="ps-component-2",
    )
    AffectFactory(
        resolution=Affect.AffectResolution.DEFER,
        affectedness=Affect.AffectAffectedness.AFFECTED,
        flaw=f3,
        ps_module="rhel-6",
        ps_component="firefox",
    )
    AffectFactory(
        resolution=Affect.AffectResolution.DEFER,
        affectedness=Affect.AffectAffectedness.AFFECTED,
        flaw=f4,
        ps_module="rhel-5",
        ps_component="kernel",
    )
```

osidb/tests/test_flaw.py
superbuggy: Removed in light of not needing validation.

apps/exploits/tests/test_api.py
superbuggy: I had been testing a combination, but I don't think adding coverage for DEFER here adds any value, so I reverted it.

apps/trackers/tests/test_file_offer.py
superbuggy: Removed in light of not needing validation.

osoukup (approved):
LGTM

dmonzonis (approved):

PR #695, opened 2024-08-15
adjust BBSync to one-way

osoukup:
This PR first makes OSIDB to stop syncing SRT notes to the Bugzilla. This is quite a significant change as it means eg. no more sync of the affects etc. which simplifies the logic a lot and will allow to make more significant changes without worrying about the backwards compatibility.

Then, it introduces number of modifications to make BBSync work in one-way sync mode.

Partially it is because the sync manager does not allow passing any extra arguments so we have to deal with the old flaw reference a different way. Partially it is because we do not sync the flaw from the Bugzilla so a number of values of the old flaw were actually no more available in the metadata.

Wherever we still have to use them we now build our own metadata on the OSIDB which should hold the expected Bugzilla status and allow us to perform the changes based on the old and current state difference.

I also decided to drastically simplify the CC list builder only accumulating the unique additions never removing anyone and never re-adding anyone already added before - they can remove themselves if needed.

Closes OSIDB-3251


osoukup:
@RedHatProductSecurity/osidb-devs please review. I suggest to go commit-wise so that way it should be clear what was the thought process and how the changes connect together

JakubFrejlach (changes requested):
Found couple of things, but rest looking good.

osidb/models.py
```diff
@@ -1333,10 +1333,11 @@ def bzsync(self, *args, bz_api_key, **kwargs):
 
         # switch of sync/async processing
         if SYNC_FLAWS_TO_BZ_ASYNCHRONOUSLY:
-            # Process the bzsync asynchronously
+            # save to OSIDB now
+            kwargs.pop("raise_validation_error", None)
+            super(ValidateMixin, self).save(*args, **kwargs)
```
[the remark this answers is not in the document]
osoukup: You are right! I removed the `save` from there

apps/bbsync/query.py
```diff
@@ -86,11 +101,38 @@ def flaw(self):
...
+    def cve_id(self):
+        cves = filter_cves(json.loads(self.meta_attr.get("alias", "[]")))
+        # BBSync does not accept multi-CVE flaws
+        # therefore we may safely pick the first
+        return cves[0] if cves else None
+
+    @cve_id.setter
+    def cve_id(self, cve_id):
+        aliases = filter_cves(
+            json.loads(self.meta_attr.get("alias", "[]")), revert=True
```
jobselko, line 121: `revert` instead of `inverse`

jobselko (approved):

JakubFrejlach (approved):

osidb/cc.py
```diff
@@ -430,30 +377,11 @@ def generate(self) -> Tuple[List[str], List[str]]:
...
-
-        # do not remove CCs which are being added from another source
-        remove_cc = [cc for cc in remove_cc if cc not in add_cc]
-
-        return add_cc, remove_cc
+        add_cc = [cc for cc in all_cc if cc not in self.old_cc]
+        # simplify the functionality by only adding people
+        # who where never added before (accumulating)
+        # and never removing anyone - they can do it
+        return add_cc, []
```
dmonzonis, line 384: If we aren't using the `remove_cc` array anymore, we can just return the `add_cc` instead of also returning an empty list that will be ignored on every call anyway

dmonzonis (approved):

osidb/cc.py
osoukup: You are completely right. I wanted to save a bit of work on adjusting tests etc. as I am in a bit of hurry. This functionality should hopefully go completely away soon enough so if you do not insist on cleaning it I would just leave it as it is for the rest of its days

PR #697, opened 2024-08-21
Alert when Bugzilla sync failed

dmonzonis:
Now that bzsync can be asynchronous, it can be the case that the flaw is correctly saved, but the sync to Bugzilla failed for some reason. In this scenario, there may be data discrepancies between OSIDB and Bugzilla, and as of now, there is no way for the user to know this happened.

With this commit, when there is an error during bzsync, an alert is raised on the flaw, so that the user knows that there may be data not synced to Bugzilla.

Closes OSIDB-3361.

### DIFF
--- a/osidb/models.py
+++ b/osidb/models.py
@@ -1355,20 +1355,37 @@ class Flaw(
 
         creating = self.bz_id is None
 
-        # sync to Bugzilla
-        bs = FlawBugzillaSaver(self, bz_api_key)  # prepare data for save to BZ
-        flaw_instance = bs.save()  # actually send to BZ (but not save to DB)
+        try:
+            # sync to Bugzilla
+            bs = FlawBugzillaSaver(self, bz_api_key)  # prepare data for save to BZ
+            flaw_instance = bs.save()  # actually send to BZ (but not save to DB)
 
-        if creating:
-            # Save bz_id to DB
-            kwargs["auto_timestamps"] = False  # no timestamps changes on save to BZ
-            kwargs["raise_validation_error"] = False  # the validations were already run
-            # save in case a new Bugzilla ID was obtained
-            # Instead of self.save(*args, **kwargs), just update the single field to avoid
-            # race conditions.
-            flaw_instance.save(
-                *args, update_fields=["meta_attr"], no_alerts=no_alerts, **kwargs
+            if creating:
+                # Save bz_id to DB
+                kwargs["auto_timestamps"] = False  # no timestamps changes on save to BZ
+                kwargs[
+                    "raise_validation_error"
+                ] = False  # the validations were already run
+                # save in case a new Bugzilla ID was obtained
+                # Instead of self.save(*args, **kwargs), just update the single field to avoid
+                # race conditions.
+                flaw_instance.save(
+                    *args, update_fields=["meta_attr"], no_alerts=no_alerts, **kwargs
+                )
+        except Exception as e:
+            # Sync failed but if it was done async the original flaw may be saved, resulting in
+            # incosnsitent data between OSIDB and BZ.
+            logger.error(f"Error when syncing flaw {self.uuid} to Bugzilla: {e}.")
+            self.alert(
+                "bzsync_failed",
+                "The Bugzilla sync for this flaw failed in the last save, so there may be data discrepancies. "
+                "The Vulnerability Tooling team has been notified about this and is looking into it.",
+                Alert.AlertType.ERROR,
             )
+            raise e
+
+        # If the bzsync was performed correctly, remove any possible alert on previously failed bzsync
+        self.alerts.filter(name="bzsync_failed").delete()
 
     def tasksync(
         self,
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Update Vulnerability trackers on components change (OSIDB-3323)
 
+### Changed
+- Alert users when Bugzilla sync failed (OSIDB-3252)
+
 ## [4.3.2] - 2024-09-19
 ### Changed
 - Update the release documentation (OSIDB-3384)


osidb/models.py
```diff
@@ -1344,22 +1344,36 @@ def bzsync(self, *args, bz_api_key, **kwargs):
...
+            except Exception as e:
+                # Sync failed but if it was done async the original flaw may be saved, resulting in
+                # incosnsitent data between OSIDB and BZ.
+                logger.error(f"Error when syncing flaw {self.uuid} to Bugzilla: {e}")
+                self.alert(
+                    "bzsync_failed",
+                    "The Bugzilla sync for this flaw failed in the last save, so there may be data discrepancies.",
+                    Alert.AlertType.ERROR,
+                )
+                raise e
```
JakubFrejlach, line 1376: This section of the code is run only when the synchronous sync is being executed and that is basically no longer used, in my PR which I will be raising soon, it will be used only for syncing BZ comments as we need to preserve authors of the commits.

Problem here is with asynchronous, which happens in the `if` branch of this condition block. BZsyncManager is scheduled there which basically then runs [`_perform_bzsync` method](https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/sync_manager.py#L630), so the actual logging should happen there. If any problem happens in the synchronous branch, it is not a big deal since it fails directly and user is greeted with error message.

JakubFrejlach (changes requested):
This needs to be reworked a bit as it should target asynchronous bzsync and not synchronous one. Also I am not sure whether just alerting on the Flaw is enough as this mechanism still relies on fact that someone will open that particular Flaw, notice that Alert and actually report it or take any actions.

dmonzonis:
@JakubFrejlach from OSIDB's perspective I'm not sure we can do much more to warn the user that the BZSync failed other than making use of the alerting system and logging, since it runs on a parallel task and the user may be doing other operations on other flaws by the time it fails.

On OSIM's side we could maybe make use of the notification system, and periodically check when there are flaws which failed their BZSync to notify the user to check them out.

Do you have any suggestions?

JakubFrejlach:
> @JakubFrejlach from OSIDB's perspective I'm not sure we can do much more to warn the user that the BZSync failed other than making use of the alerting system and logging, since it runs on a parallel task and the user may be doing other operations on other flaws by the time it fails.
> 
> On OSIM's side we could maybe make use of the notification system, and periodically check when there are flaws which failed their BZSync to notify the user to check them out.
> 
> Do you have any suggestions?

@dmonzonis I don't think we need to warn user about such thing as once user correctly make changes in OSIDB it becomes OSIDB responsibility to push such changes to Bugzilla. If we would notify users, only thing which will happen will be that they will notify us and we need to make actions and only when user notices it. So imho this should be more about making us aware of such failed syncs. 

There is the notifications service for example, we could maybe expose some of the sync managers data to API and possibly make same daily/weekly report of such failed sync jobs? Just an idea. Wdyt?

dmonzonis:
@JakubFrejlach Using the notifications system to warn us about this is definitely a good idea, will look into it.

However I still think that it's not such a bad idea to somehow also warn the user that the data is not synced with Bugzilla. If they somehow find out the data is out of sync they may freak out. If there's an error in the flaw (maybe we can change the message to let the user know we have been notified about it and are looking into it) it could give the end user some sense of security that we are aware of it and working to resolve it. Or they may freak out as well if they don't open the error message. But still better than nothing IMO, and doesn't really hurt to have the alert there.

JakubFrejlach:
> @JakubFrejlach Using the notifications system to warn us about this is definitely a good idea, will look into it.
> 
> However I still think that it's not such a bad idea to somehow also warn the user that the data is not synced with Bugzilla. If they somehow find out the data is out of sync they may freak out. If there's an error in the flaw (maybe we can change the message to let the user know we have been notified about it and are looking into it) it could give the end user some sense of security that we are aware of it and working to resolve it. Or they may freak out as well if they don't open the error message. But still better than nothing IMO, and doesn't really hurt to have the alert there.

@dmonzonis Indeed that would be a good approach, I haven't realized that initially. So lets alert on such flaws which fails to sync (I believe there is like a 5 attempts before the Flaw is proclaimed to have failed sync permanently) and then figure out some notifications for us (possibly via the notifications microservice)

JakubFrejlach:
@dmonzonis sorry I missed the message, I will review ASAP

osidb/models.py
```diff
@@ -1349,20 +1349,34 @@ def _perform_bzsync(self, *args, bz_api_key=None, no_alerts=False, **kwargs):
...
+                    *args, update_fields=["meta_attr"], no_alerts=no_alerts, **kwargs
+                )
+        except Exception as e:
+            # Sync failed but if it was done async the original flaw may be saved, resulting in
+            # incosnsitent data between OSIDB and BZ.
+            logger.error(f"Error when syncing flaw {self.uuid} to Bugzilla: {e}.")
+            self.alert(
+                "bzsync_failed",
+                "The Bugzilla sync for this flaw failed in the last save, so there may be data discrepancies. "
+                "Our team has been notified about this and is looking into it.",
```
osoukup, line 1376: `Our team` does not sound very specific. I would either replace it with `Vulnerability Tooling team` or somehow mention the OSIDB developers instead

osoukup (changes requested):
The error does not seem to be deleted on a successful sync to Bugzilla. If we fix the issue preventing the sync and resync or eg. if a consecutive automatic re-sync succeeds, the error will just keep hanging there until the flaw is saved.

dmonzonis:
@osoukup I was under the impression that when a bzsync happened it would trigger the flaw's save, clearing the alerts, but I was wrong. I have fixed this now.

osoukup:
> @osoukup I was under the impression that when a bzsync happened it would trigger the flaw's save, clearing the alerts, but I was wrong. I have fixed this now.

It was this way it the past but after deprecating the sync from Bugzilla it only happens when the flaw is created and there is new Bugzilla ID to be stored. On update it is done no more.

osoukup (approved):
LGTM

JakubFrejlach (approved):
LGTM now! Apologies for the late review!

PR #698, opened 2024-08-21
Use OSIDB service account BZ API key

JakubFrejlach:
This PR:
* removes usage of users BZ API key from majority of bzsync, only comments are still needed to be synced via user API key since we need to preserve the author
* changes bzsync so any change (except comments) are done via OSIDB BZ api key, this prevents leaking the API keys in flower and removes the problems with invalid API keys and asynchronous sync
* adjusts synchronous bzsync to be truly one-way

Note: disregard @osoukup commits, they are part of #695 and this PR builds on top of it

Closes OSIDB-3261


osidb/models.py
```diff
@@ -1576,7 +1555,9 @@ def _create_flaw(self, *args, **kwargs) -> Flaw:
 
         shared_acl = {"acl_read": self.acl_read, "acl_write": self.acl_write}
         # ensure that a flaw always contains external id (even if BZ sync is disabled)
-        shared_flaw = shared_acl | {"meta_attr": {"external_ids": self.external_id}}
+        shared_flaw = shared_acl | {
+            "meta_attr": {"external_ids": json.dumps([self.external_id])}
+        }
```
jobselko, line 1560: We should create a migration for this change in the near future to keep the data consistent. This is my mistake, sorry for it.

osoukup (approved):
LGTM

PR #700, opened 2024-08-22
implement validation of PS module and PS update stream correspondance

osoukup:
and quite a number of test updates as up until now we were allowed to be imprecise and leave the PS module and PS updates stream incorrespondence which is not possible any more

Closes OSIDB-3584

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -354,7 +354,7 @@
         "filename": "osidb/tests/endpoints/flaws/test_flaws.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
-        "line_number": 1323,
+        "line_number": 1344,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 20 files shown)


osidb/models.py
```diff
@@ -2485,6 +2486,29 @@ def _validate_tracker_ps_update_stream(self, **kwargs):
...
             )
 
+    def _validate_tracker_ps_module_ps_update_stream(self, **kwargs):
+        """
+        check that the tracker is associated with a PS update stream corresponding to its PS module
+        """
+        if not self.affects.exists():
+            return
+
+        ps_module = PsModule.objects.filter(name=self.affects.first().ps_module)
```
costaconrado, line 2496: I believe you need to add `first()` or this will fail at 2506:
```suggestion
        ps_module = PsModule.objects.filter(name=self.affects.first().ps_module).first()
```

osidb/models.py
```diff
@@ -2485,6 +2486,29 @@ def _validate_tracker_ps_update_stream(self, **kwargs):
...
+        # PS module is checked by a different validation
+        if not ps_module:
+            return
+
+        ps_update_stream = PsUpdateStream.objects.filter(name=self.ps_update_stream)
```
costaconrado, line 2501: I believe you need to add first() or this will fail at 2508:
```suggestion
        ps_update_stream = PsUpdateStream.objects.filter(name=self.ps_update_stream).first()
```

osidb/models.py
osoukup: You are correct, thank you!

dmonzonis (approved):
LGTM

PR #710, opened 2024-08-29
New trackers not using the tracker creation date in SLA computation

dmonzonis:
The new SLAs use the tracker creation date, instead of the flaw creation date, as one of the parameters. However, when creating a new tracker, the SLA is computed with the instance before it is fully saved, therefore not having a `created_dt`, and resulting in an incorrect Due Date in the Jira tracker.

This commit introduces a simple workaround to not have to change the logic of when the Jira query is done, by setting a temporary created_dt for the tracker as the current date when the tracker is being newly created.

Closes OSIDB-3376.

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -3,7 +3,10 @@ Jira tracker query generation module
 """
 import json
 import logging
+from datetime import datetime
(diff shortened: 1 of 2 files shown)


costaconrado (approved):
LGTM.

PR #711, opened 2024-08-29
fix date format error

osoukup:
Closes OSIDB-3364

### DIFF
--- a/apps/bbsync/save.py
+++ b/apps/bbsync/save.py
@@ -166,7 +166,8 @@ class BugzillaSaver(BugzillaQuerier):
             if "last_change_time" in self.instance.meta_attr
             else self.instance.meta_attr["updated_dt"]
         )
-        return make_aware(datetime.strptime(last_change_time, DATETIME_FMT))
+        # let us support multiple date formats
+        return datetime.fromisoformat(last_change_time.replace("Z", "+00:00"))
 
 
 class FlawBugzillaSaver(BugzillaSaver):
--- a/apps/bbsync/tests/test_saver.py
+++ b/apps/bbsync/tests/test_saver.py
@@ -37,3 +37,14 @@ class TestBugzillaSaver:
                 DataInconsistencyException, match="Failed to write back to Bugzilla"
             ):
                 bs.save()
+
+    @pytest.mark.parametrize(
+        "last_change_time",
+        ["2024-06-20T23:08:18Z", "2024-08-29 13:27:08+00:00"],
+    )
+    def test_stored_last_change(self, last_change_time):
+        """
+        https://issues.redhat.com/browse/OSIDB-3364 reproducer
+        """
+        bs = BugzillaSaver(Flaw(meta_attr={"last_change_time": last_change_time}))
+        bs.stored_last_change  # no exception here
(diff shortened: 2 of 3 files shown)


costaconrado (approved):
LGTM.

apps/bbsync/save.py
```diff
@@ -166,7 +166,8 @@ def stored_last_change(self):
             if "last_change_time" in self.instance.meta_attr
             else self.instance.meta_attr["updated_dt"]
         )
-        return make_aware(datetime.strptime(last_change_time, DATETIME_FMT))
+        # let us support multiple date formats
+        return datetime.fromisoformat(last_change_time.replace("Z", ""))
```
jsvob, line 170: I don't understand. Was the problem the implied timezone?

```
>>> datetime.datetime.fromisoformat("2024-06-20T23:08:18Z")
datetime.datetime(2024, 6, 20, 23, 8, 18, tzinfo=datetime.timezone.utc)
>>> datetime.datetime.fromisoformat("2024-06-20T23:08:18")
datetime.datetime(2024, 6, 20, 23, 8, 18)
>>> 
```

jsvob: Also, with make_aware, the method returned datetime with tzinfo, whereas now it will return datetime without tzinfo. Trying to understand the implications and looking at tests...

jsvob: Funny thing, when I wrap the whole expression being returned with `make_aware()`, it makes tests pass. But that was just quick shotgun programming. I actually have no idea whether it will do something wrong w.r.t. timezone conversions/corruptions (https://www.youtube.com/watch?v=-5wpm-gesOY comes to mind :D ).

jsvob: Looking at the referenced OSIDB-3364 and at the new test, another observation comes to mind:

```
>>> datetime.datetime.fromisoformat("2024-06-20T23:08:18")
datetime.datetime(2024, 6, 20, 23, 8, 18)
>>> datetime.datetime.fromisoformat("2024-08-29 13:27:08+00:00")
datetime.datetime(2024, 8, 29, 13, 27, 8, tzinfo=datetime.timezone.utc)
```

That is, the removal of `Z` makes the result be without the timezone in the first example and with timezone in the second example.

Looking at the code of `make_aware`, it doesn't complain as long as the offset from UTC is zero, which should be fine with the data produced by `datetime.datetime.fromisoformat` (unless I'm missing something).

Because I don't know the actual reason for the `.replace("Z", "")`, I wonder whether we can spend just the least possible amount of effort and keep the status quo of _doing something with the timestamps and hoping it's fine_ and do the multi-format equivalent of the old code: `return make_aware(datetime.fromisoformat(last_change_time))`

osoukup: The issue was that the change done in https://github.com/RedHatProductSecurity/osidb/commit/8ee28e9b0afc09b277cde7e8407d0060c74b0e97#diff-ebe1c224bdd0ffc9f247b35caecd1b03746276b968e7cbf57200af9c5958cd76R189 made tracker collector to store dates in a different format - **+00:00** instead of **Z** at the end. That lead to the traceback mentioned in the task as the format was suddenly not compliant with the date format string.

I first though that I will change the format back to make it unified but when I realized that Jira trackers have the new format from the beginning I was not sure it is worth to do it. I decided to only enhance this single string-date transformation to accept both formats. `fromisoformat` looks promising accepting various formats but it does not accept **Z** in the end so may instead of removing it I should rather replace it with **+00:00**. It may be the reason of the failing tests.

osoukup: Replace instead of the removal seems to be all the trick.

jsvob: Interesting!

jsvob (approved):
I don't understand it much, but if it works, it works :)

PR #718, opened 2024-08-30
Add Severity field mapping for additional schemes

jsvob:
Add Severity field mapping for additional schemes since not all projects are able to convert based on the original spec.
Closes OSIDB-3386


osoukup (approved):
LGTM

PR #722, opened 2024-09-03
Allow trackers to have manually set SLAs

costaconrado:
This PR allows trackers to have manually set SLAs by setting a custom label.

Closes OSIDB-3374.


osoukup (commented):
Just to understand the change well, this means that a tracker will first be normally created with the automatically set deadline but a person can come and set it manually providing the label which will prevent any further automatic deadline set?

rehakm:
Yes Ondro, that is the desired usecase.

PR #723, opened 2024-09-04
Update version to 4.2.1

jsvob:


docs/CHANGELOG.md
jobselko, line 1: This should be a minor release instead of a patch. `git diff` mentioned in our [docs](https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/OPERATIONS.md) clearly shows that there are changes in `openapi.yml`.

Also, I discussed with Zdenek and Ondrej that we cannot release the current master until we are sure all osidb-ops repo changes are correctly applied to all environments. But I do not know if somebody mentioned it to you.

jsvob: Oh right, the new CVEORG value. Sorry and thanks!

PR #728, opened 2024-09-04
Fix error on unembargo when trackers exist

dmonzonis:
The SLA computation is expecting a value for the `unembargo_dt` when we unembargo, but we are using a version of the tracker that is not saved yet and therefore does not have this updated information.

This commit fixes this by getting the latest version of the affect, which is already saved (and therefore the latest version of the flaw), instead of the version of the affect/flaw that the unsaved tracker has, to perform the SLA computation.

Closes OSIDB-3398.


jsvob (approved):

osoukup (approved):
LGTM

PR #729, opened 2024-09-06
add list of skipped labels instead of single label

costaconrado:
This PR adds a list of labels that should not compute SLA in Jira.

This is a request to allow multiple labels to have SLA skipped since we have different reasons to add it manually. This currently does not change the behavior but allows a constant list of labels in the future.

This is a improvement of OSIDB-3374.


JakubFrejlach (approved):
LGTM!

osoukup (commented):
I do not want to block this as there is no real change yet but I would like to first understand why a single label is not enough. From the tooling point of view there is only a single use case - ignore this issue for setting the SLA.

costaconrado:
Closing this PR, check OSIDB-3374 comments to raise additional questions.

PR #740, opened 2024-09-19
Update version to 4.3.2

jsvob:
see the changelog


docs/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+## [4.3.2] - 2024-09-19
 ### Changed
 - Update the release documentation (OSIDB-3384)
```
jobselko, line 11: What exactly we want to release? It seems that the main change is missing in the changelog.

jsvob: It's a bunch of technical changes.

jobselko: OK, thanks!

jobselko: I thought those changes would not be technical if they were worthy of release.

jsvob: They wouldn't be if it were not for an unfortunate timing of two minor changes being synchronized in such a manner that it breaks production.

One of the technical changes was the cease of setting values relevant to compliance priority and contract priority. Nobody pretty much cared, it was just something in the backlog to eventually do. And it was merged days ago.

However, Jira team was much quicker than I anticipated and changed Vulnerability issuetype values relevant to compliance and contract priority. I didn't think it would be an issue because we already removed use of these values... right? Or so I thought.

It resulted in production refusing to create trackers. So we have to release the change that's been in master for quite some time.

jobselko:
Our new documentation says to `submit release-X.Y.Z-prep as a PR against the release-X.Y.Z branch`. Why is the target branch different?

docs/CHANGELOG.md
jobselko: I understand, but when I want to release X, I expect that the X will be in changelog.

jsvob:
Good catch, I missed that! Fixed. Thanks!

jobselko (approved):

PR #743, opened 2024-09-20
remove mixin loop

costaconrado:
This PR removes a infinite recursion in save method when sync to bugzilla is disabled.

This commit https://github.com/RedHatProductSecurity/osidb/pull/686/commits/db2cc6f588e453ee7e04b519f0e40fd0559fc74d introduced a infinite recursion in environments that don't sync to bugzilla, when validations were separated from BugzillaSyncMixin, was no need to bzsync method to skip mixins / call the next save method, making this infinite loop.

The way the 409 used to happen was by calling save method multiple times until the current save call have a different timestamp from the 1st save call, throwing the exception presented in  https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/mixins.py#L52

Closes OSIDB-3430.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Alert users when Bugzilla sync failed (OSIDB-3252)
 
+### Fixed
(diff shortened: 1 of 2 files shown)


osoukup (approved):
Aha! That was the UAT issue! LGTM

MrMarble (approved):

JakubFrejlach (approved):
LGTM! Nice detective work!

PR #744, opened 2024-09-20
Update Vulnerability issuetype Jira trackers on Flaw components change.

jsvob:
Closes OSIDB-3323


osidb/serializer.py
```diff
@@ -1748,7 +1748,11 @@ def mi_differ(flaw1, flaw2):
         # but that drastically increases the code complexity and brings only a little
         # value - would prevent a rare extra update attempt without any real effect
         if (
-            not differ(old_flaw, new_flaw, ["cve_id", "is_embargoed", "unembargo_dt"])
+            not differ(
+                old_flaw,
+                new_flaw,
+                ["cve_id", "is_embargoed", "unembargo_dt", "components"],
```
osoukup, line 1754: If there is no specific reason to sort the list this way I would prefer the alphabet :smile_cat:

osoukup (approved):
LGTM

osidb/serializer.py
jsvob: Sorry, missed the neat fact that it was sorted nicely before I touched it! Thanks for noticing! Fixed :-)

jsvob:
I'm adding a single extra set of pytest parameters to test the conditions more thoroughly.

PR #747, opened 2024-09-24
Disable flaw drafts creation for NVD

jobselko:
This PR removes the flaw creation from the NVD collector (CVSS scores assigned by NIST are still collected). 

Closes OSIDB-3256


collectors/nvd/collectors.py
```diff
@@ -248,7 +178,7 @@ def collect(self, cve: Union[str, None] = None) -> str:
             try:
                 flaw = Flaw.objects.get(cve_id=cve_id)
 
-                # update NVD CVSS2 or CVSS3 data if necessary
+                # update NVD CVSS2, CVSS3 and CVSS4 data if necessary
                 if self.update_cvss_via_flawcvss(flaw, item):
                     updated_flaws.append(cve_id)
                     # no automatic timestamps as those go from Bugzilla
```
jobselko, line 184: I am not sure if this is still true because we made several Bugzilla changes.

osoukup: Totally not true any more :smile: no flaw timestamps go from Bugzilla now

jobselko: Okay, I will change it accordingly :)

osoukup (approved):
LGTM. Regarding your comment, I think we should add flaw **updated_dt** update when the CVSS score is fetched so eg. SDEngine knows that there is something new to be fetched. Could you please either add it here or file a Jira otherwise?

osoukup (approved):
LGTM

PR #749, opened 2024-09-26
Update CVEorg collector for UAT/Prod

jobselko:
This PR brings the final changes for UAT and Production environments. The date from which flaw drafts will be created has been agreed with the VM team.

Related to OSIDB-3473 and OSIDB-3255


osoukup (approved):
LGTM

PR #750, opened 2024-09-26
OSIDB-3479: Allow searching for flaw components

MrMarble:
DjangoQL lacks suport for `ArrayField` type, so searching for flaw components was not working, neither the suggestions.

I implemented a custom field that allows searching for the components, I added it below the `FlawQLSchema` class but it can be moved somewhere else, also the naming can be improved, right now this is the only custom field but I can imagine creating a couple more, so maybe is worth having them in a separate file?

Some example queries and explanations:

* `components = "openssl"`:  Any flaw that has at least one component named *openssl*
* `components in ("openssl", "kernel")`: Any flaw that has at least one component named *openssl* **or** *kernel*
* `components = "openssl,kernel"`: Any flaw that has both *openssl* **and** *kernel* components
* `components = ""`: Any flaw that does not have any components
* `components != ""`: Any flaw that hast at least one component

Created following the documentation from https://docs.djangoproject.com/en/5.1/ref/models/querysets/#field-lookups and https://github.com/ivelum/djangoql?tab=readme-ov-file#custom-search-fields

Closes OSIDB-3479

### DIFF
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -16,7 +16,7 @@ from django_filters.rest_framework import (
     OrderingFilter,
 )
 from djangoql.queryset import apply_search
-from djangoql.schema import DjangoQLSchema
+from djangoql.schema import DjangoQLSchema, StrField
 
 from apps.workflows.workflow import WorkflowModel
 from osidb.dmodels.package_versions import Package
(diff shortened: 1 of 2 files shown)


osoukup (approved):
LGTM

jobselko (changes requested):
Wrong section in changelog

osidb/filters.py
```diff
@@ -292,10 +292,45 @@ def get_fields(self, model):
...
+    model = Flaw
+    name = "components"
+    suggest_options = True
+
+    def get_options(self, search):
+        options = super(FlawComponentField, self).get_options(search)
+        flat_list = []
+        for option in options:
+            flat_list += option
+        return flat_list
```
jobselko, line 310: Could this be simplified?

jobselko (approved):

PR #752, opened 2024-09-26
Upgrade Django and other dependencies

dmonzonis:
This PR aims to upgrade the versions of several dependencies to versions that are currently supported, and to get security fixes for vulnerabilities which affect the versions we were using before.

Closes OSIDB-3392.


openapi.yml
```diff
@@ -7127,28 +7114,6 @@ components:
...
-              type: string
-            ps_component:
-              type: string
-            ps_module:
-              type: string
-            ps_product:
-              type: string
-            resolution:
-              type: string
-          readOnly: true
```
dmonzonis, line 7151: With the upgrade of `drf-spectacular` (which is required due to a breaking change in the upgrade of drf) these disappeared for only some of the serializers. In trying to figure out why and how the schema is generated, it appears the issue is that to generate the schema it performs some requests to the various endpoints but with no query params, so it does not include the `include_meta_attr` params and the field gets removed in `IncludeMetaAttrMixin`.

Why this happened only for some serializes (for instance, it happened for `AffectSerializer` but not for `AffectPostSerializer` which just inherits from the former, even though in this case it shouldn't even have them since it's read-only) is beyond me. But this was also happening before for the `history` field on all serializers, since the requests do not include the `include_history` param either, which makes me wonder if these should even be here to begin with. If we do, we might want to show the `history` field as well, but this will require some extensive work to understand how the schema is generating these fields that can be dynamically removed in the serializer.

devel-requirements.txt
```diff
@@ -119,9 +119,9 @@ distlib==0.3.3 \
     --hash=sha256:c8b54e8454e5bf6237cc84c20e8264c3e991e824ef27e8f1e81049867d861e31 \
     --hash=sha256:d982d0751ff6eaaab5e2ec8e691d949ee80eddf01a62eaa96ddb11531fe16b05
     # via virtualenv
-django==3.2.25 \
-    --hash=sha256:7ca38a78654aee72378594d63e51636c04b8e28574f5505dff630895b5472777 \
-    --hash=sha256:a52ea7fcf280b16f7b739cec38fa6d3f8953a5456986944c3ca97e79882b4e38
+django==4.2.16 \
```
osoukup, line 122: The commit message says Django 4.12 which does not seem to be the case.

openapi.yml
osoukup: I am thinking how to understand this change from the versioning point of view. The API does not get changed, only the documentation, correct?

osoukup (approved):
LGTM. Good job!

devel-requirements.txt
dmonzonis: Oops, that's definitely a typo

openapi.yml
dmonzonis: Correct, the API does not change and the `meta_attr` are still returned only if you pass the `include_meta_attr` param. But the documentation stopped showing this param, which is probably due to a change in how the documentation is generated in the new version.

osoukup: Maybe that is actually for good as those meta attributes have no insurance of even existence but they were already being used as production data multiple times.

JakubFrejlach (approved):
LGTM! Good job. Made some small, rather informative comments, nothing needs changes I think.

PR #753, opened 2024-09-27
Add VCR test for creating and updating Vulnerability issuetype trackers.

jsvob:
Closes OSIDB-3441

### DIFF
--- a/apps/trackers/tests/test_integration.py
+++ b/apps/trackers/tests/test_integration.py
@@ -9,14 +9,15 @@ from django.utils import timezone
 from rest_framework import status
 
 from apps.trackers.jira.query import JiraPriority
-from apps.trackers.models import JiraBugIssuetype
+from apps.trackers.models import JiraBugIssuetype, JiraProjectFields
 from apps.trackers.save import TrackerSaver
 from apps.trackers.tests.factories import JiraProjectFieldsFactory
 from collectors.bzimport.collectors import BugzillaTrackerCollector, FlawCollector
 from collectors.bzimport.constants import BZ_DT_FMT
+from collectors.jiraffe.collectors import JiraTrackerCollector
 from osidb.dmodels.tracker import Tracker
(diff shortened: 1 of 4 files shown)


apps/trackers/tests/test_integration.py
```diff
@@ -614,3 +615,359 @@ def test_tracker_update_jira(
         assert flaw3.affects.count() == 1
         assert flaw3.affects.first().trackers.count() == 1
         assert flaw3.affects.first().trackers.first().uuid == tracker.uuid
+
+    @pytest.mark.vcr
+    def test_tracker_create_update_jira_vulnerability_issuetype(
```
osoukup, line 620: The tests is quite massive and I have just quite fresh experience with the file offer module and some long tests (but about a third in length) which were quite complicated to understand what actually got broken. It also seems to me that it tests multiple different things. I understand that we want to test the whole stack here but if there is a way to split it, I would be for it - eg. when after the create you check everything (so we know the state from which we continue onward) would it be different to just make another test with only update setting all the asserted context as it is supposed to be?

osoukup (approved):
LGTM

apps/trackers/tests/test_integration.py
jsvob: Maybe I'm missing something, but the major difference would be the state of Bugzilla and Jira. Since the operations hit both and they both have to have the correct data on their sides, splitting the test would just massively duplicate the setup work. I did try to make the create and update tests separate, but met this complication.

jsvob: It will definitely be fun :-( No better idea though. Thanks for thinking about a better way.

jsvob: Although I did try to make it so that the test creates the flaw and tracker itself. Therefore, it will require only changing as far as the logic and asserts are concerned. Not manual BTS setup work.

costaconrado (approved):
LGTM!

PR #754, opened 2024-09-30
Record last impact increase in trackers

dmonzonis:
This commit adds a new field `last_impact_increase_dt` which records the last time a tracker's aggregated impact increased. This will be used for SLA computation, using this date instead of the tracker's creation date for computing the due date.

Closes OSIDB-3448.


osidb/signals.py
```diff
@@ -117,3 +117,14 @@ def update_local_updated_dt_tracker(sender, instance, **kwargs):
             no_alerts=True,  # recreating alerts from nested entities can cause deadlocks
             raise_validation_error=False,
         )
+
+
+@receiver(pre_save, sender=Affect)
```
osoukup, line 122: We will also need to account for the flaw change - the tracker aggregated impact is composition of its flaw impacts which may or may not be overridden by the affect impacts.

osidb/signals.py
```diff
@@ -117,3 +117,14 @@ def update_local_updated_dt_tracker(sender, instance, **kwargs):
...
+def update_last_impact_increase_dt(sender, instance, **kwargs):
+    if not instance._state.adding and Impact(instance.impact) > Impact(
+        Affect.objects.get(pk=instance.pk).impact
+    ):
+        for tracker in instance.trackers.all():
+            if Impact(instance.impact) > tracker.aggregated_impact:
+                tracker.last_impact_increase_dt = timezone.now()
+                tracker.save(auto_timestamps=False, raise_validation_error=False)
```
osoukup, line 130: As we do not need to auto-set timestamps or run validations I am thinking whether this would not be a case when **update** can be easily used instead of **save**. The reasons are described in https://docs.google.com/document/d/1-3bCyQky6c0OboSHRVEFm4BSl3nM79LHIVXfQ8HyRqc/edit?usp=sharing and I do not remember then in detail :smile: but **save** can have some unintended side effects together with the parallelism.

osoukup (changes requested):
I think that this needs an extension.

JakubFrejlach (approved):
Apart from a already mentioned things, LGTM!

osidb/signals.py
dmonzonis: Makes sense, I changed it

docs/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Alert users when Bugzilla sync failed (OSIDB-3252)
+- Record last impact increase in trackers (OSIDB-3448)
```
jobselko, line 18: Please rebase changelog. You are updating release 4.3.3

osoukup (approved):
LGTM. Thank you for incorporating my feedback!

docs/CHANGELOG.md
jobselko: The `Unreleased` section is still outdated, compare with current [changelog](https://github.com/RedHatProductSecurity/osidb/blob/master/docs/CHANGELOG.md).

PR #761, opened 2024-10-02
Enable one SLA test for Jira trackers

dmonzonis:
Two of these tests were recently disabled, and when enabled simultaneously they fail. However we can enable one of them to not lose coverage of this functionality, and both of the tests are similar anyway so this is much better than having both disabled.

Closes OSIDB-3505.


osoukup (approved):
Makes sense. LGTM

jsvob (approved):
Thank you and sorry for disabling the tests.

PR #762, opened 2024-10-03
Fix IntegrityError during Tracker filing

JakubFrejlach:
This PR:
* fixes occasional `IntegrityError` duplicate external system id  happening during the tracker filing (see commit description reasoning)

Note: as the actual situation happens concurrently and in transactions, it is very nontrivial to actually write some meaningful tests in this case.

Closes OSIDB-3433


osoukup (approved):
LGTM

dmonzonis (approved):

PR #765, opened 2024-10-04
Select most relevant CVE, CVSS, CWE, Source for multi-flaw trackers with Vulnerability issuetype.

jsvob:
Closes OSIDB-3348

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -26,8 +26,7 @@ from apps.trackers.exceptions import (
 from apps.trackers.models import JiraProjectFields
 from collectors.jiraffe.constants import JIRA_BZ_ID_LABEL_RE
 from osidb.cc import JiraAffectCCBuilder
-from osidb.helpers import cve_id_comparator
-from osidb.models import Affect, AffectCVSS, Flaw, FlawSource, Impact
+from osidb.models import Affect, AffectCVSS, Flaw, FlawCVSS, FlawSource, Impact
 from osidb.validators import CVE_RE_STR
 
 from .constants import (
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM

docs/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Record last impact increase in trackers (OSIDB-3448)
+- Select most relevant CVE, CVSS, CWE, Source for Vulnerability trackers (OSIDB-3348)
```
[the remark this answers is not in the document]
jsvob: :facepalm: Thank you for noticing, sorry.

apps/trackers/jira/query.py
```diff
@@ -585,6 +584,117 @@ def generate(self):
...
+
+        issuers = list(affect.flaw.cvss_scores.values_list("issuer", flat=True))
+        if not issuers:
+            return
+        selected_issuer = issuers[0]
+        prefered_order = [
+            FlawCVSS.CVSSIssuer.REDHAT,
+            FlawCVSS.CVSSIssuer.NIST,
+            FlawCVSS.CVSSIssuer.CVEORG,
+            FlawCVSS.CVSSIssuer.OSV,
```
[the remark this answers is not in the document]
jsvob: You are right. I'll fix it to ignore non-RH CVSS. There's actually nothing in the spec that would say otherwise.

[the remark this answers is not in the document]
jsvob: Fixed. Thank you!

apps/trackers/jira/query.py
```diff
@@ -585,6 +584,105 @@ def generate(self):
...
+            # Affect override present.
+            return (
+                affect.cvss_scores.filter(issuer=AffectCVSS.CVSSIssuer.REDHAT)
+                .order_by("-version")
+                .first()
+            )
+
+        return (
+            affect.flaw.cvss_scores.filter(issuer=FlawCVSS.CVSSIssuer.REDHAT)
+            .order_by("-version", "-score")
```
JakubFrejlach, line 649: Why do we sort by `score` here on Flaw model but not above in the Affect model?

JakubFrejlach (approved):
One question, otherwise LGTM!

apps/trackers/jira/query.py
```diff
@@ -585,6 +584,105 @@ def generate(self):
...
+            if impact == affects_impacts[-1][1]
+        ]
+        flaw_created_dt_affects = sorted(
+            [
+                (affect.flaw.created_dt, affect)
+                for affect, impact in greatest_impact_affects_impacts
+            ],
+            key=lambda pair: pair[0],
+        )
+        most_important_affect = flaw_created_dt_affects[0][1]
```
dmonzonis, line 625: Maybe directly using the queryset's `order_by` would be shorter (and probably more performant?) than using a list comprehension and sorting in Python?

jobselko (approved):

apps/trackers/jira/query.py
jsvob: Because I didn't realize there's already validation that ensures that there can't be multiple CVSS scores with a given issuer and version for a given Flaw/Affect, and that the CVSS test factories already ensure that too. And because in a previous quite different WIP version of the tests and algorithm I somehow came up with this sorting for some reason and forgot to treat both AffectCVSS and FlawCVSS in the same way. Removed the sorting by score. Thank you for noticing.

apps/trackers/jira/query.py
jsvob: `aggregated_impact` is not a DB field.

```
django.core.exceptions.FieldError: Cannot resolve keyword 'aggregated_impact' into field. Choices are: acl_read, acl_write, affectedness, alerts, created_dt, cvss_scores, embargoed, events, exploitreview, flaw, flaw_id, impact, meta_attr, ps_component, ps_module, resolution, trackers, updated_dt, uuid
```

So I do all sorting in python to at least keep the principles similar inside the single function.

The amount of objects iterated this way in `greatest_impact_affects_impacts` and `flaw_created_dt_affects` is IMO small enough not to be a performance issue.

dmonzonis (approved):

PR #767, opened 2024-10-07
extend incident types

osoukup:
Two new incident types were requested to be introduced. **Minor Incident** has only a little bit of special handling and is mostly informative. **0-day** is on the other hand a totally big deal and has **Major Incident** special handling in most cases and will have more strict SLA on top of it - which is going to be handled via SLA policies.

Closes OSIDB-3390


dmonzonis (approved):
LGTM

osidb/models.py
```diff
@@ -1130,13 +1135,17 @@ def _validate_unsupported_impact_change(self, **kwargs):
         ] or old_flaw.major_incident_state in [
             Flaw.FlawMajorIncident.APPROVED,
             Flaw.FlawMajorIncident.CISA_APPROVED,
+            # Minor Incident is only informative
+            Flaw.FlawMajorIncident.ZERO_DAY,
```
JakubFrejlach, line 1139: Maybe it is only me, but the way the Minor incident is commented kinda feels like the whole comment regards to `ZERO_DAY` instead and it creates confusion that Zero day is Minor. I got actually confused during the review. Maybe I would include the whole name in the comment `Flaw.FlawMajorIncident.MINOR is only informative`. This repeats several times.

osidb/models.py
```diff
@@ -1130,13 +1135,17 @@ def _validate_unsupported_impact_change(self, **kwargs):
...
         ]
         is_high_impact = self.impact in [
             Impact.CRITICAL,
             Impact.IMPORTANT,
         ] or self.major_incident_state in [
             Flaw.FlawMajorIncident.APPROVED,
             Flaw.FlawMajorIncident.CISA_APPROVED,
+            # Minor Incident is only informative
+            Flaw.FlawMajorIncident.ZERO_DAY,
```
JakubFrejlach, line 1148: Same as above.

osidb/signals.py
```diff
@@ -66,6 +66,9 @@ def update_major_incident_start_dt(flaw: Flaw) -> None:
     is_major_incident = flaw.major_incident_state in {
         Flaw.FlawMajorIncident.APPROVED,
         Flaw.FlawMajorIncident.CISA_APPROVED,
+        # Minor Incident has no impact on the SLA
+        # so do not record its start in any way
+        Flaw.FlawMajorIncident.ZERO_DAY,
```
JakubFrejlach, line 71: Same as above.

JakubFrejlach (approved):
One suggestion which applies to multiple places in code, but I will leave it up to you, maybe it is OK this way.

apps/taskman/service.py
```diff
@@ -174,6 +174,8 @@ def _generate_task_data(self, flaw: Flaw):
         if flaw.major_incident_state in [
             Flaw.FlawMajorIncident.APPROVED,
             Flaw.FlawMajorIncident.CISA_APPROVED,
+            Flaw.FlawMajorIncident.MINOR,
```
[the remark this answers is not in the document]
osoukup: Good point. I was also a bit hesitant here. So far we used only a single label and it is probably only our initiative to set it in this case as do not think people care for it in the OSIM tasks. As it is internal for PSIRT (not engineers) however and they own those tasks I at the end decided to put it there.

I am not sure how to approach this label. We can have different labels for different incident types. We can omit minor there. We can just leave it as it is until someone cares. Do you have a preference? I think I do not :smile: 

The whole confusing thing is now that we have a variable **major_incident_state** and it can actually hold an information saying that we have a **minor incident**. The rename would be ideal but laborious.

osidb/models.py
osoukup: I will think about how to make it less confusing. I just wanted to explicitly mention that the minor one is not included on purpose and not just forgotten etc.

apps/trackers/common.py
```diff
@@ -116,22 +116,37 @@ def _summary_prefixes(self) -> str:
...
+        ):
+            prefixes.append("[Minor Incident]")
+
+        elif any(
+            flaw
+            for flaw in self.flaws
+            if flaw.major_incident_state == Flaw.FlawMajorIncident.ZERO_DAY
+        ):
+            prefixes.append("[0-day]")
+
```
MrMarble, line 149: Could this be rewritten to reduce the O(n) complexity? if the flaw is `ZERO_DAY` we have to loop the entire flaw list at least 3 times.

This is my suggestion (could be completely wrong 😆 ): 
```python
prefix_map = {
    Flaw.FlawMajorIncident.APPROVED: "[Major Incident]",
    Flaw.FlawMajorIncident.CISA_APPROVED: "[CISA Major Incident]",
    Flaw.FlawMajorIncident.MINOR: "[Minor Incident]",
    Flaw.FlawMajorIncident.ZERO_DAY: "[0-day]"
}

for flaw in sorted(self.flaws, key=lambda flaw: flaw.major_incident_state):
    if flaw.major_incident_state in prefix_map:
        prefixes.append(prefix_map[flaw.major_incident_state])
        break
```

osidb/models.py
osoukup: I reworded it and used your advice

apps/trackers/common.py
osoukup: Oh sure! This way is really silly :smile: silly copy-pasting

osoukup: And good suggestion! I will use it. Thank you!

MrMarble (approved):

PR #768, opened 2024-10-09
Set Impact for collector flaws based on CVSS severity

jobselko:
This PR reworks and extends the CVSS section for collector flaws to be able to calculate the resulting impact. The impact is created based on the highest CVSS score. The PR also updates the OSV collector to collect CVSSv4.

Closes OSIDB-3487

### DIFF
--- a/collectors/utils.py
+++ b/collectors/utils.py
@@ -1,10 +1,11 @@
 import re
-from typing import Optional
+from decimal import Decimal
+from typing import Optional, Union
 
 from django.core.exceptions import ValidationError
 from django.core.validators import URLValidator
 
-from osidb.models import FlawReference, PsUpdateStream
+from osidb.models import FlawReference, Impact, PsUpdateStream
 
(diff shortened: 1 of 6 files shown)


collectors/utils.py
```diff
@@ -98,3 +99,27 @@ def handle_urls(references: list, source_ref: str) -> list:
             pass
 
     return urls
+
+
+CVSS_SCORE_TO_IMPACT = {
```
jsvob, line 104: Is this modeled after https://www.first.org/cvss/specification-document#Qualitative-Severity-Rating-Scale ?

jsvob (approved):
LGTM but I'm not knowledgeable in this area

collectors/utils.py
jobselko: Partially yes. It matches both Impact in OSIDB and CVSS severity defined by FIRST. However, FIRST does not define severity for CVSSv2, so I hope this simplification is good.

osoukup (approved):
LGTM

PR #769, opened 2024-10-10
Update UBI image and vulnerable deps

dmonzonis:
Update the version of the UBI image that the osidb-service pod is based on, as well as some dependencies which contain vulnerabilities.

Related to OSIDB-3326.


osoukup (approved):
LGTM

PR #770, opened 2024-10-10
hotfix flaw comments

MrMarble:
#545 Added the field `order` to the `FlawComments` model but the field was excluded from the comment creation endpoint, so an error was thrown when trying to create a comment.
image
 
---
### Edit

After further (and long) testing, the error seems to come from this #752 pull request, that adds the field as required, also, there must be something different with the version of Django, as I could not reproduce the error locally until I rebuilt the containers.

Before the update of Django, it was okay to create a required field on the serializer, but after the update, the field needs to be part of the body, as Django throws an error before entering the serializer

I've marked the field as deprecated on the API and removed the constraint on the comments as bzimport is not used anymore with the idea of updating it to use a proper field once the `order` field is removed in a major release

Fixes OSIDB-3547


MrMarble:
> This does not seem to me as a step to the right direction. User should not care for the any order here. (S)he is always creating the very last comment and the backend should make sure that the order will be applied in this sense. Or do I miss something?
> 
> If it is mean as some temporary hotfix which is going to be removed later, it is not a good idea. API removals are breaking changes and require major release.

After further discussion and testing with @dmonzonis we think that removing the field is a better solution, internally OSIDB creates the field sequentially so the user does not need to do it

openapi.yml
```diff
@@ -7497,11 +7497,6 @@ components:
         external_system_id:
           type: string
           maxLength: 100
-        order:
-          type: integer
-          maximum: 2147483647
-          minimum: -2147483648
-          nullable: true
```
jobselko, line 7504: This change will make another release major, right?

dmonzonis (approved):

osoukup (changes requested):
I agree that the field should not be there. However, when it already is, we should just mark it as deprecated and start ignoring it in the query. The removal can happen as part of the next major release but I do not see a reason why it should trigger one right now.

jsvob (changes requested):
* I don't see https://github.com/RedHatProductSecurity/osidb/pull/545 adding `order` as a required parameter for POST in openapi.yml. `FlawCommentSerializer.create()` injects `order` into `validated_data`. I do see the error on stage, it is strange, seems like some sort of validation error. `git blame openapi.yml` doesn't reveal any recent changes related to `order`.
* `FlawCommentPostSerializer` has been unchanged for a long time and it specifically removes `order` from the required parameters.
* Isn't the issue a different version of some library that interprets something differently?
* If we start creating comments without `order`, it will break listing comments in order AFAIK, so any change will have to test that it will continue behaving correctly. I'm not sure whether removing the field from the serializer will make django ignore it when injected into `validated_data` by `FlawCommentSerializer.create()`.
* The `order` serializer field that is being removed here has been there ever since commit f980e888cfe2b434745723244a488f1c97d2701a (the `order` field defined in CommentSerializer is also used by FlawCommentSerializer, which is then inherited by FlawCommentPostSerializer with the `order` field removed for the purposes of schema generation). So the reasoning in this PR seems not in line with the history, I'd guess the root cause is different.
* Has this change been tested?

MrMarble:
@jsvob The field was added as required on this pull request https://github.com/RedHatProductSecurity/osidb/pull/752
seems like the behavior of django changed after the update because I was unable to reproduce it locally until I rebuilt the containers with the new version

osoukup (approved):
LGTM

osidb/migrations/0172_remove_flawcomment_unique_per_flaw_comment_nums.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('osidb', '0171_alter_flawcomment_order'),
+    ]
+
+    operations = [
+        migrations.RemoveConstraint(
```
dmonzonis, line 13: Can we put this in a single migration file?

dmonzonis (approved):

osidb/models.py
```diff
@@ -2219,11 +2219,12 @@ class Meta:
 
         # Ensure that it's not possible to have two bzimports running concurrently
         # and succeed while creating numbering conditions impossible to handle later.
-        constraints = [
-            models.UniqueConstraint(
-                fields=["flaw", "order"], name="unique_per_flaw_comment_nums"
-            ),
-        ]
+        # TODO: Re-enable this constraint once the "order" field is removed from the model.
```
jsvob, line 2222: We can't re-enable the constraint if we introduce comments that lack the order, or have duplicate ordering, or it will be difficult to do so.

This effectively changes the API because the client can't use the `order` field for sorting comments anymore, and I don't see any sorting being used when serializing on OSIDB side other than the model's `Meta` which will necessarily cause very out-of-chronological-ordering comment order, which is bound to be very confusing:

```
    class Meta:
        """define meta"""

        ordering = (
            "order",
            "external_system_id",
            "uuid",
            "created_dt",
        )
```

The commit with message `Make order field optional` doesn't explain why.

jsvob (changes requested):
Sorry for blocking this but I don't see how this won't cause further breakage. Maybe change the ordering in the model's Meta and sync with OSIM that this API change will be fine for them?

osidb/models.py
jsvob: What I wrote ^ is incorrect.

We had a call. Thank you for explaining the changes and changes in Django. I was confused and not knowledgeable enough (and didn't have the time to do my own research without blocking it even longer). For posterity, the `order` field is still being set by `FlawCommentSerializer.create()`, so it will continue behaving exactly as it used to behave.

jsvob (approved):
LGTM, thank you for updating the TODO and explaining the changes. Sorry for the holdup!

jobselko:
I think that all commits should be merged into one as they relate to the same thing.

PR #775, opened 2024-10-15
do not restrict tracker filing by future support scope

osoukup:
as the ProdSec support can predate the GA

Closes OSIDB-3559

### DIFF
--- a/apps/trackers/api.py
+++ b/apps/trackers/api.py
@@ -43,7 +43,7 @@ class TrackerFileSuggestionView(RudimentaryUserPathLoggingMixin, APIView):
         ).distinct()
         # remove unsupported
         active_module_names = [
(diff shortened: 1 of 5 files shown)


jsvob (approved):

apps/trackers/api.py
```diff
@@ -43,7 +43,7 @@ def post(self, request, *args, **kwargs):
         ).distinct()
         # remove unsupported
         active_module_names = [
-            module.name for module in active_modules if module.is_supported
```
dmonzonis, line 46: Can we remove the `is_supported` property then? I don't see it being used anywhere else

osoukup: I can remove it

dmonzonis (approved):

JakubFrejlach (approved):

PR #777, opened 2024-10-16
Check title for keywords in CVEorg collector

jobselko:
This PR extends keywords checking in the CVEorg collector. The `title` is newly checked together with already used `comment_zero`. If at least one of them contains only blacklisted keywords, a flaw is skipped.

Closes OSIDB-3545


collectors/cveorg/collectors.py
```diff
@@ -230,16 +230,21 @@ def save_snippet_and_flaw(self, content: dict) -> tuple[bool, bool]:
...
             return False, False
 
-        if should_create_snippet(content["comment_zero"]):
-            snippet, snippet_created = Snippet.objects.get_or_create(
-                source=Snippet.Source.CVEORG,
-                external_id=content["cve_id"],
-                defaults={"content": content},
+        if not should_create_snippet(content["comment_zero"]) or (
+            not should_create_snippet(content["title"])
+            and content["title"] != "From CVEorg collector"
```
osoukup, line 235: Just a question out curiosity. Why do we need to check that the title is different from `From CVEorg collector` to refuse the snippet creation?

osoukup (approved):
LGTM

collectors/cveorg/collectors.py
jobselko: I wanted to ensure that we are examining only real titles and not the custom ones. Lets us say that some part of the custom title will get to a keywords list, we forget to update this check and we can end up with wrong results. But this situation is very unlikely, so removing the title check should be save (if you prefer to do that).

osoukup: Makes sense. I just wanted to understand it. Thank you!

JakubFrejlach (approved):

PR #779, opened 2024-10-16
fix metadata update on flaw un-privating

osoukup:
TODO ...not ready for review


osoukup:
The reproducer does not work. Locally all behaves fine.

PR #784, opened 2024-10-18
fix Bugzilla flaw summary exceeding

osoukup:
Closes OSIDB-3551

### DIFF
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -4,12 +4,41 @@ from itertools import chain
 
 from django.utils import timezone
 
+from apps.bbsync.constants import MAX_SUMMARY_LENGTH, MULTIPLE_DESCRIPTIONS_SUBSTITUTION
+from apps.bbsync.exceptions import UnsavableModelError
 from collectors.bzimport.constants import ANALYSIS_TASK_PRODUCT
 from osidb.helpers import cve_id_comparator, filter_cves
 from osidb.models import Flaw, FlawComment, Impact, PsModule
 
 from .cc import CCBuilder
 from .constants import DATE_FMT
+
+
+def summary_shorten(cves, description):
+    """
+    shorten the aligable parts of the tracker summary
+    """
+    # first shorten CVE list
+    if len(cves) > 1:
+
+        # remove the last CVE
+        cves = cves[0:-1]
+        # add the dots to the new last
+        cves[-1] = cves[-1] + " ..."
+
+    # finally shorten the description
+    else:
+        # when we cannot preserve at least a minimal meaningful description
+        # something is fairly wrong and we cannot create such tracker
+        if len(description) <= len(MULTIPLE_DESCRIPTIONS_SUBSTITUTION):
+            raise UnsavableModelError(
+                f"Summary generated for the tracker is longer than {MAX_SUMMARY_LENGTH}"
+            )
+
+        # simply shorten the desciption by one
+        description = description[0:-5] + " ..."
+
+    return cves, description
 
 
 class BugzillaQueryBuilder:
--- a/apps/bbsync/query.py
+++ b/apps/bbsync/query.py
@@ -248,17 +248,29 @@ class FlawBugzillaQueryBuilder(BugzillaQueryBuilder):
         # 3) filter out the empty CVE IDs
         #    there can be one in case we are removing the CVE ID
         cve_ids = [cve_id for cve_id in cve_ids if cve_id]
-        # 4) filter out eventual duplicates | sort by CVE ID | stringify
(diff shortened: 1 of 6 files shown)


apps/bbsync/query.py
```diff
@@ -12,6 +14,33 @@
...
+        cves[-1] = cves[-1] + " ..."
+
+    # finally shorten the description
+    else:
+        # when we cannot preserve at least a minimal meaningful description
+        # something is fairly wrong and we cannot create such tracker
+        if len(description) <= len(MULTIPLE_DESCRIPTIONS_SUBSTITUTION):
+            raise UnsavableModelError(
+                f"Summary generated for the tracker is longer than {MAX_SUMMARY_LENGTH}"
+            )
```
dmonzonis, line 36: I know this is copied from old code, but we can move this to the top of the function. We avoid doing the operations on the `cves` array and remove this `else` clause making the code cleaner too.

apps/bbsync/query.py
```diff
@@ -219,17 +248,29 @@ def generate_summary(self):
...
+
+            cve_string = " ".join(cve_ids) + " " if cve_ids else ""
+            summary = f"{embargoed}{cve_string}{components}{description}"
+
+            if len(summary) <= MAX_SUMMARY_LENGTH:
+                break
+
+            cve_ids, description = summary_shorten(cve_ids, description)
+
+        self._query["summary"] = summary
```
dmonzonis, line 273: If we have a long list of `components`, we may reach a point when `summary_shorten` can't shorten `cve_ids` and `description` further but we are still over the limit, ending in an infinite loop.

It might be a good idea to use a variable to keep track of whether we are still reducing the length at each step, and if we reach the limit (length is not reducing anymore but we are still over the max length) just exit the loop and shorten it the barbarian way
```py
summary = summary[:MAX_SUMMARY_LENGTH-3] + "..."
```

dmonzonis (changes requested):

apps/bbsync/query.py
osoukup: I can do it

apps/bbsync/query.py
osoukup: Makes sense

apps/bbsync/query.py
osoukup: @dmonzonis actually I do not fully understand. How do you mean that we can avoid doing operations on the `cves`? The first mechanism to shorten the tracker summary is removing CVE IDs one at a time. If we avoid it eg. the issue I am trying to solve here (flaw with 35 CVE IDs) will be blocking the sync.

dmonzonis: What I mean is raising the error first if it needs to be raised. If we do the operations on the `cves` and then raise an error, it won't matter anyway, or will it?

osoukup: That is possible. I was asking about removing the else clause as it does not seem possible to me. I assume that historically there were rarely an issue with the description length so probably therefore first the algorithm deals with CVEs which might be many.

osoukup: Nowadays it can happen that the description is too long as we do not set any restriction there on the title but it is probably still less probably than having too many CVE in case of some multi-CVE tracker

dmonzonis: I thought the shortening on the `description` was outside of the else clause, I read the indentation wrong. It can stay as is then.

apps/bbsync/query.py
osoukup: @dmonzonis so added a condition to error out instead of shortening in case we already have the very last CVE ID and the description is minimal - so the reason of the length are the components. However, I see that this is actually no change from the original behavior which would try to shorten once more and fail with the completely same error. Do you really think that there is any change needed?

dmonzonis: It should be okay now, but I still think that relying on a method that is called inside the `while True` loop to raise an error in order to exit the loop can be dangerous. If someone else in the future modifies this method without knowing this, and removes the `raise`, we could end up with an infinite loop, which would be a rare case but still possible. I still think a safeguard in the loop doesn't add a lot of complexity and can prevent this.

But I'll leave it up to you, the code as it is now works well.

dmonzonis (approved):

docs/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Check title for keywords in CVEorg collector (OSIDB-3545)
 
+### Fixed
+- Fix Bugzilla flaw summary exceeding (OSIDB-3551)
```
jobselko, line 16: Wrong changelog

PR #785, opened 2024-10-18
update delegated resolution mapping

osoukup:
so low impact won't fix changes to fix deferred

Closes OSIDB-3575


dmonzonis (approved):
I don't really understand the reasoning behind this change, and these special cases can be tricky to remember later. But I guess this was requested to work like this so LGTM

PR #786, opened 2024-10-21
Refactor OSV collector

jobselko:
This PR refactors the OSV collector to ease the upcoming changes regarding the partial disabling of flaw creation. The only thing that this PR changes is data sent to a logger. IDs of created snippets and flaws replaced a count of created and updated snippets.

Related to OSIDB-3351


dmonzonis (approved):
LGTM

PR #787, opened 2024-10-22
OSIDB-3589: Handle ValidationError on tracker filing

MrMarble:
With the update of Django 4.2 constraints now run as part of the  model validation, so we need to handle `ValidationError` 

Closes OSIDB-3589


JakubFrejlach (approved):
LGTM!

PR #788, opened 2024-10-22
Fix handling of generator failure in OSV collector

jobselko:
This PR fixes a bug introduced in #786, where the OSV vulns generator got outside the try block, so its failure was not handled correctly. This caused the whole code to stop instead of logging an error and continuing.

Related to OSIDB-3351


MrMarble (approved):

costaconrado (approved):
LGTM.

PR #792, opened 2024-10-23
Update version to 4.5.0

JakubFrejlach:
Same approach as in #783, rebasing the whole models splitting would be so much pain. We need to do a full release soon so we clean this thing up so it doesn't pile up like this.


osoukup (approved):
LGTM

PR #793, opened 2024-10-23
Fix incorrect changes in CHANGELOG

jobselko:
This PR fixes the incorrect change in CHANGELOG introduced in #784.


osoukup (approved):
Thank you! I though I would fix it with the release changelog PRs but sooner is definitely better :+1:

PR #794, opened 2024-10-23
avoid tracker creation conflicts by async tracker sync

osoukup:
Closes OSIDB-3593


osidb/models/tracker.py
```diff
@@ -170,19 +168,11 @@ def save(self, *args, bz_api_key=None, jira_token=None, **kwargs):
...
-            btc.no_alerts = True
-            btc.sync_tracker(tracker_instance.external_system_id)
-            BZTrackerLinkManager.link_tracker_with_affects(
-                tracker_instance.external_system_id
-            )
+            # no save or fetch to prevent collisions
+            # only schedule an asynchronous sync
+            BZTrackerDownloadManager.schedule(tracker_instance.external_system_id)
+            # at the end delete the temporary empty tacker
+            Tracker.objects.filter(external_system_id="").delete()
```
JakubFrejlach, line 175: I am just thinking, do we need to actually create the Tracker instance with empty `external_system_id` and store it in the database? TrackerSaver operates with the django model instance and it shouldn't matter whether it is saved or not, right? If that is the case, we could just not save the tracker [here in serializer](https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/serializer.py#L702) and work with the instance instead and never save it. Or I am missing something?

osidb/models/tracker.py
```diff
@@ -209,19 +199,11 @@ def save(self, *args, bz_api_key=None, jira_token=None, **kwargs):
...
-            jtc.no_alerts = True
-            jtc.collect(tracker_instance.external_system_id)
-            JiraTrackerLinkManager.link_tracker_with_affects(
-                tracker_instance.external_system_id
-            )
+            # no save or fetch to prevent collisions
+            # only schedule an asynchronous sync
+            JiraTrackerDownloadManager.schedule(tracker_instance.external_system_id)
+            # at the end delete the temporary empty tacker
+            Tracker.objects.filter(external_system_id="").delete()
```
JakubFrejlach, line 206: Same question as above.

JakubFrejlach (approved):
LGTM! I have one question regarding the whole dummy Tracker with empty id, but maybe I am missing something or there is something being initialized needed for the query builder when we get only by saving the Tracker on our side and not doing that would just create a need for reworking the whole saver, which is not desired at the moment. Leaving that up to you.

osidb/models/tracker.py
osoukup: You are right but making it without the actual temporary saving to DB would require a deeper rework of the whole `save` machinery and query builder as it dives somehow to created that temporary tracker to later use it in the query builder (not sure if elsewhere) and I did not want to touch it more than necessary :smile: 

As it creates and then removes it in a transaction for the rest of the world it will look like it never happened.

JakubFrejlach: Agreed!

PR #798, opened 2024-10-24
Adjust delays in CVEorg and OSV collectors

jobselko:
This PR adjusts delays in CVEorg and OSV collectors to wait only in a case when a flaw is created.


osoukup (approved):
LGTM

MrMarble (approved):

PR #799, opened 2024-10-24
decrease jira queries in Taskman

osoukup:
Let us check the token validity only when it is a user one and do not query for static metadata.


costaconrado (approved):
LGTM.

PR #801, opened 2024-10-24
Add upstream references to Jira trackers on creation

dmonzonis:
When filing a tracker, the related upstream links (flaw references that are of type UPSTREAM for any related flaws) will be added as remote links in the Jira issue.

Closes OSIDB-3148.

### DIFF
--- a/apps/taskman/service.py
+++ b/apps/taskman/service.py
@@ -265,3 +265,19 @@ class JiraTaskmanQuerier(JiraQuerier):
         query_list = self.query_tasks(query_list)
         query_list = self.query_updated(query_list, updated_after, updated_before)
         return self.run_query(query_list)
+
+    def add_link(self, issue_key, url, title):
+        """Add a remote link to a task."""
+        try:
+            data = {"url": url, "title": title}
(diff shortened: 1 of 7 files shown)


apps/taskman/service.py
```diff
@@ -265,3 +265,19 @@ def get_task_period(
         query_list = self.query_tasks(query_list)
         query_list = self.query_updated(query_list, updated_after, updated_before)
         return self.run_query(query_list)
+
+    def add_link(self, issue_key, url, title):
```
osoukup, line 269: Why is this defined in the **taskman** when it is only used for the trackers? I assume it probably follows the pattern of the **create_comment** method which seems to me to have set a bad precedence - the commenting seems to be used for both tasks and trackers so is also not **taskman** specific. What about moving it (ideally both) to the general **JiraQuerier**? That could relieve some unnecessary dependencies between the task and tracker management - which would probably have to happen in the future anyway if we successfully migrate the tracker functionality elsewhere.

osoukup (changes requested):
The code LGTM but I would be for moving it a bit

osoukup (approved):
Thank you for the refactoring! This looks much better

PR #804, opened 2024-10-25
Concurrent logging

JakubFrejlach:
This PR:
* implements JSON TCP socket log handler with TLS/SSL support which will be handling the logs instead of the rotating file handler which does not support concurrent logging by sending them to new logstash pod (will be introduced in the dependant ops repository PR)
* changes configuration of stage, uat and production to use socket log handler
* TLS certs for the logstash will be added to the hashicorp vault (stage is already done)

This PR is ready for review but should not be merged before the ops repo changes. (concurrent-logging branch)

Partially addresses OSIDB-3357

### DIFF
--- a/osidb/helpers.py
+++ b/osidb/helpers.py
@@ -4,7 +4,9 @@ Helpers for direct or development shell usage
 
 import json
 import logging
+import logging.handlers
 import re
+import ssl
 import sys
 import warnings
 from distutils.util import strtobool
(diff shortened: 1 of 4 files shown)


osidb/helpers.py
```diff
@@ -238,3 +240,98 @@ def safe_get_response_content(response: Response):
...
+        super().__init__(*args, **kwargs)
+        self.logfile = logfile
+        # self.retries = retries
+        # self.delay = delay
+
+        # self.fallback_handler = None
+        # if fallback_logfile is not None:
+        #     self.fallback_handler = logging.FileHandler(filename=fallback_logfile)
+
+    def makePickle(self, record: logging.LogRecord) -> bytes:
```
osoukup, line 259: For the method names we use the snake-case instead of the camel-case so I would be for preserving the consistency and got with the camel-case here too.

osoukup: I see below that the parent class uses the camel-case too so you probably preserved that. Then, I have no strong preference. The mixture seems to be unavoidable.

osidb/helpers.py
```diff
@@ -238,3 +240,98 @@ def safe_get_response_content(response: Response):
...
+    #                     # Log to the fallback file after all retries fail
+    #                     if self.fallback_handler is not None:
+    #                         print("fallback")
+    #                         self.fallback_handler.emit(record)  # Fallback to file
+    #                     # self.handleError(record)  # Handle error logging
+    #     except Exception as e:
+    #         # Handle unexpected exceptions and log to fallback file
+    #         if self.fallback_handler is not None:
+    #             print("fallback")
+    #             self.fallback_handler.emit(record)
```
osoukup: Is there any purpose of leaving these commented lines here?

osoukup: I see they are being removed in the following commit. Then it makes even less sense to me to introduce them here but at least they are not part of the final result. I have no strong preference then.

osoukup (approved):
Leaving a few not minor comments but generally LGTM. Great that this gets solved!

osidb/helpers.py
JakubFrejlach: Yes, this done in order to override the method which serializes the data in the handler and unfortunately they are using camel-case :/

osidb/helpers.py
JakubFrejlach: Seems like I messed up rebasing :smile:  I will fix that, thanks!

PR #805, opened 2024-10-28
fix internal acls during serializer update

costaconrado:
This PR changes `ACLMixinSerializer` in order to support other ACLs other than embargoed and public.

Closes OSIDB-3578.


costaconrado:
@reviewer I've added a test that may not be very useful to the codebase but was needed to check the validity of this bug fix, if you think I should remove it, let me know.

osoukup (approved):
Oh I see! So for the internal it resolved to not embargoed and got set to public here in the serializer... Good catch! LGTM

jobselko:
@costaconrado `embargoed2acls` is used several times in `serializer.py`. Is it sufficient to change only that one method?

PR #806, opened 2024-10-29
Collect all product definitions in a single transaction

dmonzonis:
Before this commit, the product definitions are collected in 3 different transactions. The creation of update streams and modules are done in different transactions, but the linkage between both is important and if the user performs some action that requires it, like updating a tracker, while this linkage is not done, it can lead to issues like trackers unlinking from the flaw.

To avoid this, with this commit all product definitions are collected in a single transaction. This ensures that we have all modules, update streams and the links between them at all times.

Closes OSIDB-3590


JakubFrejlach (approved):
LGTM now!

PR #807, opened 2024-10-30
OSIDB-3581: Validate zero cvss score for flaws without impact

MrMarble:
Added validation to the `Flaw` Model and the `FlawCVSSSerializer` to check that if the flaw has impact `''`, the CVSS score **must** be 0.

I had to add the validation to the serializer as adding it to the Model was using the value already on the database instead of the one that comes from the request.

Closes OSIDB-3581

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -162,7 +162,7 @@
         "filename": "apps/bbsync/tests/test_integration.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
-        "line_number": 94,
+        "line_number": 95,
         "is_secret": false
       }
     ],
@@ -467,5 +467,5 @@
       }
     ]
   },
-  "generated_at": "2024-10-31T12:11:24Z"
+  "generated_at": "2024-11-04T10:36:13Z"
 }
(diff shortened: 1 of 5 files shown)


osidb/serializer.py
```diff
@@ -1488,6 +1489,34 @@ class Meta:
...
+
+        # Check that the CVSS score is 0.0 for flaws with no impact
+        if flaw.impact == Impact.NOVALUE and cvss.cvss_object.base_score != Decimal(
+            "0.0"
+        ):
+            raise serializers.ValidationError(
+                {"cvss_score": ["CVSS score must be 0.0 for flaws with no impact"]}
+            )
+
+        return super().update(new_cvss, validated_data)
```
JakubFrejlach, line 1518: Why is this done on the serializer level? Don't we have a validation mixin for such things? :thinking:

JakubFrejlach: Ah, I see, that is because the CVSS score is populated based on the vector by signal after the save.
https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/signals.py#L89

When we implemented that, there was a kinda chicken/egg problem as to populate the score value, you need to first validate the string but the validations are run after the values were changed and altering the score in the validation would feel wrong.

Actually I am just thinking about it and it might mean that some other validations are wrong because of this functionality.

MrMarble: I was able to access the `cvss_object.base_score` field on the Model (should have tried after I found the field while tinkering with the serializer), so I've moved the validation from the serializer to its own `_validate_` method

JakubFrejlach (approved):
LGTM now!

dmonzonis (approved):

PR #808, opened 2024-10-30
validate currrent ACL before making tracker public

costaconrado:
Closes OSIDB-3578.


osidb/serializer.py
jobselko, line 1: `_is_public()` from `FlawSerializer` works incorrectly because it uses `embargoed2acls()` which ignores internal ACLs

jobselko (approved):
LGTM

PR #809, opened 2024-10-31
Fix some of warnings and prevent new ones

jobselko:
This PR fixes warnings regarding `django.utils.timezone.utc` being deprecated and ensures that no new warnings will be introduced in the code (only the already present warnings are allowed).


osoukup (approved):
LGTM

PR #810, opened 2024-11-04
OSIDB-3588: Implement resolution steps for duplicate tracker validation

MrMarble:
- Added the external system id of the duplicated trackers to the error message so it is easier to identify in cases with multiple affects/trackers.
- Modified the alert creation to accept the `params` field of the `ValidationError` as `kwargs`, to pass additional parameters.
- Added some resolution steps to `_validate_tracker_duplicate`

Closes OSIDB-3588

### DIFF
--- a/osidb/models/tracker.py
+++ b/osidb/models/tracker.py
@@ -387,14 +387,20 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
         validate that there is only one tracker with this update stream associated with each affect
         """
         for affect in self.affects.all():
-            if (
-                affect.trackers.filter(ps_update_stream=self.ps_update_stream).count()
-                > 1
-            ):
+            trackers = affect.trackers.filter(ps_update_stream=self.ps_update_stream)
+            if trackers.count() > 1:
+
                 raise ValidationError(
-                    f"Tracker with the update stream {self.ps_update_stream} "
+                    f"Tracker with the update stream {self.ps_update_stream} ({self.external_system_id}) "
                     "is already associated with the affect "
-                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid})"
+                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid}) "
+                    f"by the tracker(s) {', '.join([str(tracker.external_system_id) for tracker in trackers if tracker.uuid != self.uuid])}",
+                    params={
+                        "resolution_steps": "When manually cloning a tracker, please ensure that each affect has a unique update stream, "
+                        "which can be identified in the tracker's title. "
+                        "If the tracker is a duplicate, remove the 'SecurityTracking' label from the tracker in the external system. "
+                        "If the tracker is not expected to be a duplicate, please contact the Vulnerability Tooling team for further assistance."
+                    },
                 )
 
     def can_unembargo(self):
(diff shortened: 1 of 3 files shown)


osidb/models/tracker.py
```diff
@@ -387,14 +387,20 @@ def _validate_tracker_duplicate(self, **kwargs):
...
+                    f"Tracker with the update stream {self.ps_update_stream} ({self.external_system_id}) "
                     "is already associated with the affect "
-                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid})"
+                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid}) "
+                    f"by the tracker(s) {', '.join([str(tracker.external_system_id) for tracker in trackers if tracker.uuid != self.uuid])}",
+                    params={
+                        "resolution_steps": "When manually cloning a tracker, please ensure that each affect has a unique update stream, "
+                        "which can be identified in the tracker's title. "
+                        "If the tracker is a duplicate, remove the duplicate tracker. "
+                        "If the tracker is not expected to be a duplicate, please contact the OSIDB team for further assistance."
```
JakubFrejlach, line 402: I would rather use `Vulnerability Tooling team` as we already used it in bzsync alert, so we should keep that consistent:
https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/models/flaw/flaw.py#L993

osidb/models/tracker.py
```diff
@@ -387,14 +387,20 @@ def _validate_tracker_duplicate(self, **kwargs):
...
-                    f"Tracker with the update stream {self.ps_update_stream} "
+                    f"Tracker with the update stream {self.ps_update_stream} ({self.external_system_id}) "
                     "is already associated with the affect "
-                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid})"
+                    f"{affect.ps_module}/{affect.ps_component} ({affect.uuid}) "
+                    f"by the tracker(s) {', '.join([str(tracker.external_system_id) for tracker in trackers if tracker.uuid != self.uuid])}",
+                    params={
+                        "resolution_steps": "When manually cloning a tracker, please ensure that each affect has a unique update stream, "
+                        "which can be identified in the tracker's title. "
+                        "If the tracker is a duplicate, remove the duplicate tracker. "
```
JakubFrejlach, line 401: By removing the tracker you mean removing the SecurityTracking label so it is not considered a tracker anymore?

JakubFrejlach (commented):
Couple of notes.

osidb/models/tracker.py
osoukup: Good point. This is generally quite complicated and people need a bit of inside to understand what to do. Ideally the tooling would just ignore the duplicates but before it is smart enough manual intervention is needed. I think that a resolution of making sure that every affect's tracker has a unique update stream and contacting us when not sure what to do might be enough :smile: I can imagine various creative ways how people could remove a tracker. I am not sure they would pick removing `SecurityTracking` keyword/label :sweat_smile:

osoukup (approved):
LGTM

osidb/models/tracker.py
JakubFrejlach: Also will removing the label on its own work? I don't think we have a mechanism which will remove the tracker from the database when the label is removed, right?

JakubFrejlach (approved):
LGTM!

dmonzonis (approved):

osidb/models/tracker.py
osoukup: Good point. I know I implemented one for the flaws in the past but I am not about the trackers

PR #811, opened 2024-11-04
Add history to several other models

dmonzonis:
History was only enabled for flaws and affects. This commit introduces history to several smaller models: Snippet, Tracker, FlawAcknowledgment, FlawComment, FlawCVSS and FlawReference.

Closes OSIDB-3466.


JimFuller-RedHat:
As I was involved with this feature at the beginning - a few thoughts popped into my head - have these changes been tested at scale eg. do we understand the impact/increase in terms of storage and processing this PR might incur ? ACLS of the corresponding audit table should always inherit from the model being audited ... was the work ever done for when we unembargo (to make sure we change related audit acls ) ?

osidb/migrations/0174_add_history_to_more_models.py
```diff
@@ -0,0 +1,1068 @@
...
+                    table="osidb_flawacknowledgmentaudit",
+                    when="BEFORE",
+                ),
+            ),
+        ),
+        # Enable row level security for the new history models
+        migrations.RunSQL(
+            reverse_sql=migrations.RunSQL.noop,
+            sql="""
+    ALTER TABLE osidb_snippetaudit FORCE ROW LEVEL SECURITY;
```
JimFuller-RedHat, line 861: I think we need to add this to etc/pg/setup-rls-audit-tables.sql as well ?

JimFuller-RedHat (approved):
This PR seems to just add the audit tables for the models - I do not see anywhere where we are surfacing audit info in REST API (dedicated audit endpoint or in entity endpoints). Added a few comments in review though I think this PR should be heavily tested at scale before going into production so we understand fully impact on data size and processing.

osidb/migrations/0174_add_history_to_more_models.py
dmonzonis: I guess? I'm not sure what that file is used for. Isn't it sufficient with running the migration?

JimFuller-RedHat: maybe this is no longer used (I think this was originally just used by docker for dev purposes) ... might be worth checking previous migrations take care of it (and removing)

osidb/models/flaw/cvss.py
```diff
@@ -24,6 +25,12 @@ def create_cvss(flaw, issuer, version, **extra_fields):
             return FlawCVSS(flaw=flaw, issuer=issuer, version=version, **extra_fields)
 
 
+@pghistory.track(
+    pghistory.InsertEvent(),
+    pghistory.UpdateEvent(),
+    pghistory.DeleteEvent(),
+    model_name="FlawCVSSAudit",
+)
```
JakubFrejlach, line 33: There are two CVSS models, `FlawCVSS` and `AffectCVSS` since the Jira issue talks about enabling history for CVSS scores I get the feeling that `AffectCVSS` should be included too.

JakubFrejlach (changes requested):

osidb/tests/test_audit.py
```diff
@@ -129,25 +140,79 @@ def test_access_flawevent(
...
+
+        flaw_ref = FlawReferenceFactory(flaw=flaw)
+        assert flaw_ref.acl_read == embargoed_read_groups
+        assert flaw_ref.acl_write == embargoed_write_groups
+
+        flaw_com = FlawCommentFactory(flaw=flaw)
+        assert flaw_com.acl_read == embargoed_read_groups
+        assert flaw_com.acl_write == embargoed_write_groups
+
+        flaw_cvss = FlawCVSSFactory(flaw=flaw)
```
JakubFrejlach, line 184: Can we test AffectCVSS audit as well?

JakubFrejlach (approved):
Just a minor comment regarding test, other than that LGTM. I also agree with @JimFuller-RedHat that we should fully test this PR so we understand the scale of the audit data generated.

PR #815, opened 2024-11-06
Fix conversion of CVSS severity to impact

jobselko:
This PR ensures that when converting CVSS severity to impact, numbers in the `Decimal` type are always compared so the impact is correctly set.

Fixes OSIDB-3661

### DIFF
--- a/collectors/utils.py
+++ b/collectors/utils.py
@@ -116,6 +116,9 @@ def convert_cvss_score_to_impact(score: Union[Decimal, float]) -> Impact:
     This function converts CVSS score to Flaw Impact.
     Flaw Impact matches CVSS severity, which is defined by CVSS score ranges.
     """
+    # Ensure the score is always of Decimal type rounded to 1 decimal place
+    score = round(Decimal(score), 1)
+
     impact = Impact.NOVALUE
     for key, value in CVSS_SCORE_TO_IMPACT.items():
         lower, upper = value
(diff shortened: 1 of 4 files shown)


collectors/utils.py
```diff
@@ -111,11 +111,15 @@ def handle_urls(references: list, source_ref: str) -> list:
 }
 
 
-def convert_cvss_score_to_impact(score: Union[Decimal, float]) -> Impact:
+def convert_cvss_score_to_impact(score: Decimal) -> Impact:
```
dmonzonis, line 114: This implies that we alwasy get a `Decimal` yet there have logic in case it's not?
If the score can be passed as a float I think it should be reflected here, even if later on it's all converted to a `Decimal`

jobselko: You are right, I overlooked this when moving the condition. Thanks!

costaconrado (changes requested):

collectors/utils.py
```diff
@@ -116,6 +116,10 @@ def convert_cvss_score_to_impact(score: Union[Decimal, float]) -> Impact:
     This function converts CVSS score to Flaw Impact.
     Flaw Impact matches CVSS severity, which is defined by CVSS score ranges.
     """
+    # Ensure the score is always of Decimal type rounded to 1 decimal place
+    if not isinstance(score, Decimal):
+        score = round(Decimal(score), 1)
```
JakubFrejlach, line 121: Is it possible that we get `Decimal` instance but not rounded to 1 decimal place? Such case would pass unnoticed here. We can simply do:
```python
score = round(Decimal(score), 1)
```
as creating a new Decimal from Decimal and rounding it to the 1 place won't change it if it is already in the right format.

jobselko: `cvss` lib should always return numbers like `1.2`, but your suggestion is probably better. Thanks!

costaconrado (approved):
LGTM. Thanks for addressing.

JakubFrejlach (approved):

PR #816, opened 2024-11-06
Ignore invalid CVSS from OSV collector

jobselko:
This PR temporarily updates exception handling when encountering invalid CVSS data from OSV like `{'type': 'CVSS_V4', 'score':'CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:H/A:N'}`. This case is now only logged but does not block the OSV collector from running. Once the data gets fixed in OSV, the change will be removed.

Fixes OSIDB-3663

### DIFF
--- a/collectors/osv/collectors.py
+++ b/collectors/osv/collectors.py
@@ -289,7 +289,12 @@ class OSVCollector(Collector):
                     # Skip unsupported score types
                     continue
                 vector = cvss["score"]
-                score = self.CVSS_TO_CVSSLIB[cvss["type"]](vector).base_score
+                try:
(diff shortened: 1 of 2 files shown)


collectors/osv/collectors.py
jobselko, line 1: There are no tests because there is only one OSV item with invalid CVSS data, and this item will not be converted into a flaw because it contains CVE ID. (OSV collector creates flaws only for items without CVE ID)

costaconrado (approved):
LGTM.

collectors/osv/collectors.py
costaconrado: noted :+1:

PR #819, opened 2024-11-07
publish internal flaws only when the triage is completed

osoukup:
to prevent sending notification and creating CVE page too early

Closes OSIDB-3669


costaconrado (approved):
LGTM.

PR #820, opened 2024-11-07
preserve nist flag on score adoption

osoukup:
... instead of removing it. Let us adjust the flag accordingly and relieve the related validation.

Closes OSIDB-3672


JakubFrejlach (approved):
LGTM!

PR #821, opened 2024-11-07
Adjust flaw impact if NIST CVSS is changed

jobselko:
This PR updates the NVD collector to correctly set a flaw `impact` if NIST CVSS was changed, and the `impact` was originally empty.

Fixes OSIDB-3661

### DIFF
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -1,3 +1,4 @@
+from decimal import Decimal
 from typing import Union
 
 import nvdlib
@@ -7,8 +8,9 @@ from django.utils import timezone
 from nvdlib.classes import CVE
 
 from collectors.framework.models import Collector
+from collectors.utils import convert_cvss_score_to_impact
 from osidb.core import set_user_acls
-from osidb.models import Flaw, FlawCVSS
+from osidb.models import Flaw, FlawCVSS, Impact
 
 logger = get_task_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


collectors/nvd/collectors.py
```diff
@@ -295,6 +324,11 @@ def update_cvss_via_flawcvss(self, flaw: Flaw, item: dict) -> bool:
                     continue
 
                 if original_cvss != new_cvss:
+                    # set impact if it was empty because a validation requires it
+                    if not flaw.impact:
+                        flaw.impact = self.calculate_highest_impact(
+                            [cvss2_score, cvss3_score, cvss4_score]
+                        )
```
osoukup, line 331: I think that the procedure should probably be a little bit different favoring the newer score versions before the old ones as they might have better expressing capabilities and therefore more precise results but I might be wrong and the caution is OK as a first estimate. Definitely leave it as it is. We can reconsider if someone complains.

osoukup (approved):
LGTM

collectors/nvd/collectors.py
```diff
@@ -246,27 +248,54 @@ def get_original_nvd_cvss(flaw: Flaw, cvss_version: str) -> Union[str, None]:
...
+        """
+        Return the highest impact calculated from `cvss_data`.
+        """
+        # get a score based on CVSS of the highest severity
+        highest_score = Decimal("0.0")
+        for cvss in cvss_data:
+            if cvss:
+                score = round(Decimal(cvss), 1)
+                if score > highest_score:
+                    highest_score = score
```
jobselko, line 273: this one is unnecessarily complicated, I will update it

collectors/nvd/collectors.py
jobselko: `[cvss2_score, cvss3_score, cvss4_score]` is created from new scores, e.g. `new_cvss4, cvss4_score = self.get_new_nvd_cvss(item, FlawCVSS.CVSSVersion.VERSION4)`

PR #825, opened 2024-11-08
Properly save updated flaw in NVD collector

jobselko:
This PR fixes the bug introduced in #821. The flaw impact was updated before saving a CVSS score, but the flaw itself was not saved.

Fixes OSIDB-3661

### DIFF
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -324,9 +324,11 @@ class NVDCollector(Collector, NVDQuerier):
                 if original_cvss != new_cvss:
                     # set impact if it was empty because a validation requires it
                     if not flaw.impact:
-                        flaw.impact = self.calculate_highest_impact(
+                        impact = self.calculate_highest_impact(
                             [new_cvss2_score, new_cvss3_score, new_cvss4_score]
                         )
+                        Flaw.objects.filter(uuid=flaw.uuid).update(impact=impact)
+                        flaw.refresh_from_db()
                     # perform either update or create
                     cvss_score = FlawCVSS.objects.create_cvss(
                         flaw,
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- Properly save updated flaw in NVD collector (OSIDB-3661)
 
 ## [4.5.5] - 2024-11-07
 ### Changed


collectors/nvd/collectors.py
```diff
@@ -327,6 +327,7 @@ def update_cvss_via_flawcvss(self, flaw: Flaw, item: dict) -> bool:
                         flaw.impact = self.calculate_highest_impact(
                             [new_cvss2_score, new_cvss3_score, new_cvss4_score]
                         )
+                        flaw.save(raise_validation_error=False)
```
dmonzonis, line 330: Why not use `update` instead of `save`?

dmonzonis (approved):
LGTM just a small question

collectors/nvd/collectors.py
jobselko: I think that `update` does not change `updated_dt`, so it seems that flaw was not changed.

jobselko: But it is then followed by `save` later, so it should be probably fine.
```
                if self.update_cvss_via_flawcvss(flaw, item):
                    updated_flaws.append(cve_id)
                    # no validation exceptions not to fail here
                    flaw.save(raise_validation_error=False)
```

osoukup: The automatic `updated_dt` change should be solved separately by @MrMarble so you does not have to account for it separately. Whenever `update` is enough, let us use it instead of save.

collectors/nvd/collectors.py
```diff
@@ -324,9 +324,10 @@ def update_cvss_via_flawcvss(self, flaw: Flaw, item: dict) -> bool:
                 if original_cvss != new_cvss:
                     # set impact if it was empty because a validation requires it
                     if not flaw.impact:
-                        flaw.impact = self.calculate_highest_impact(
+                        impact = self.calculate_highest_impact(
                             [new_cvss2_score, new_cvss3_score, new_cvss4_score]
                         )
+                        Flaw.objects.filter(uuid=flaw.uuid).update(impact=impact)
```
JakubFrejlach, line 330: Can we add a test for this?

jobselko: Could you please check if the tests are enough?
https://github.com/RedHatProductSecurity/osidb/blob/7b7aa58ae0e7ac63b1d88ef0c3cc1fa4bf660141/collectors/nvd/tests/test_collectors.py#L294

JakubFrejlach: So the test was already there and did not catch this? :thinking:

jobselko: Yes, do you see any bugs in those tests?

osoukup: Maybe it triggered the `save` in the caller method while the real-life run did not?

JakubFrejlach: @jobselko I am not sure. Since the save of the Flaw was only after the CVSS was created/updated, was it really created/updated? Wasn't there some silent fail? That test needs to be definitely investigated as it passed before and it passes now.

jobselko:
@osoukup I am not sure if I understand you well, `flaw.save(raise_validation_error=False)` is done only in the case when something has changed. Could you please check my solution if it looks okay to you?

osoukup:
> @osoukup I am not sure if I understand you well, `flaw.save(raise_validation_error=False)` is done only in the case when something has changed. Could you please check my solution if it looks okay to you?

Yeah, that is what I mean. The impact changed but we have another DB call which now even followed by a refresh. When the condition in the caller triggers we do another save. Why not just one save or update?

jobselko:
I also disagree with half-baked solutions, but these changes were merged because we need to fix the failing collector in production, and there is no time to do it in a better way.

PR #828, opened 2024-11-12
Osidb 3492

costaconrado:
This PR added a fixture for creating metadata sample and prod-defs sample in root conftest.

This PR also adds a fixture to automatically set envs during VCR recording.
This PR also moved tokens for tests using VCR for root conftest in order to automatically set user token during VCR recording.
This PR also removes tokens variables set in conftests for tests that does not uses external resources + VCRs.

Closes OSIDB-3492.


osoukup (approved):
LGTM a lot!

PR #829, opened 2024-11-13
Introduce purl to Affect

jobselko:
This PR introduces the `purl` field to the `Affect` model and adds a basic validation to check if a `purl` can be parsed into a `PackageURL` object.

Closes OSIDB-3409


osidb/migrations/0174_remove_affect_insert_insert_and_more.py
jobselko, line 1: I am trying to understand why there are operations which lead to the original state. Does anybody have any idea?

JakubFrejlach: I don't think it actually leads to the original state. First the triggers are removed, then affect model is modified by adding the purl field and then the triggers are added back but this time they track also the purl fields (see for example the `OLD."purl" IS DISTINCT FROM (NEW."purl")`.

jobselko: Thanks! I overlooked it.

JakubFrejlach (approved):
LGTM

osidb/tests/endpoints/test_affects.py
```diff
@@ -611,3 +611,96 @@ def test_trigger_affect_flaw_change(self, auth_client, test_api_uri):
...
+    @pytest.mark.parametrize(
+        "purl,should_fail",
+        [
+            ("pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25", False),
+            ("rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25", True),
+        ],
+    )
+    def test_affect_purl_create(self, auth_client, test_api_uri, purl, should_fail):
+        """
+        Test the creation of Affect record only with the correct purl field.
```
osoukup, line 630: nitpick: I think a word is missing in the sentence? As we test two cases and only one is a creation of Affect record with the correct purl field but the failing one is not.

osidb/tests/factories.py
```diff
@@ -300,6 +300,9 @@ class Meta:
     )
     ps_module = factory.sequence(lambda n: f"ps-module-{n}")
     ps_component = factory.sequence(lambda n: f"ps-component-{n}")
+    purl = factory.sequence(
+        lambda n: f"pkg:rpm/fedora/ps-component-{n}@7.50.3-1.fc25?arch=i386&distro=fedora-25"
+    )
```
osoukup, line 305: It is not yet being validated whether the PS component actually corresponds to the PURL but then we must not forget to update this co the PS component and the PURL string corresponds and the factory generates a valid affect. I would suggest to address it already so it is not forgotten but leaving it up to you.

osoukup (approved):
LGTM

osidb/tests/factories.py
JakubFrejlach: Right now this actually holds true as when not specifying these fields will result in the correct affect. Once we add that validaton, there should be probably a purl generated first in the factory and ps_component added via `factory.LazyAttribute` which will parse out the component from the purl via the library.

osidb/tests/endpoints/test_affects.py
jobselko: Would `Test that Affect record is created only in the case when its purl field is correct.` be better?

osoukup: Definitely :+1:

osidb/tests/factories.py
osoukup: Or the other way around if we can generate the correct PURL that way - setting the actually provided PS component to the purl instead of `ps-component-{n}` so it corresponds. I am thinking about this way because there is already a number of tests which specify `ps_component` explicitly and all of those would start failing.

JakubFrejlach: That would be however a little bit counter-intuitive to the fact that we would like to give higher priority tu the purls compared to the ps_components. Then passing just the PURL will then fail as the ps_component will be generated differently and the validation will fail.

osoukup: It could probably be done somehow conditionally both ways. Or it the tests we can for now prioritize the other way as there are tests with component and no PURL and not the other way.

PR #830, opened 2024-11-13
TODO

osoukup:
... not ready for a review


dmonzonis (approved):
Good call making it parametrizable even though it was asked for one specific case!

PR #832, opened 2024-11-20
Validate and set ps_component from purl

jobselko:
This PR adds validation between `ps_component` and `purl` to check they match each other and also enables the automatic setting of `ps_component` if a request contains only `purl`.

Closes OSIDB-3410

### DIFF
--- a/osidb/models/affect.py
+++ b/osidb/models/affect.py
@@ -172,6 +172,22 @@ class Affect(
     def __str__(self):
         return str(self.uuid)
 
+    def save(self, *args, **kwargs):
+        if self.purl and not self.ps_component:
+            try:
+                self.ps_component = PackageURL.from_string(self.purl).name
+            except ValueError:
+                pass
+
+        # TODO: if ps_component is still missing, we must catch an error now
+        #  because the error message in AlertMixin's standard validations is
+        #  not helpful and changing the order of validations is not possible now
+        #  (this should be removed once custom validations run before standard validations)
+        if not self.ps_component:
+            self._validate_purl_and_ps_component()
+
+        super().save(*args, **kwargs)
+
     def _validate_ps_module_old_flaw(self, **kwargs):
         """
         Checks that an affect from an older flaw contains a valid ps_module.
@@ -473,15 +489,32 @@ class Affect(
                     f"{affected_special_consideration_package} is missing statement.",
                 )
 
-    def _validate_purl(self, **kwargs):
+    def _validate_purl_and_ps_component(self, **kwargs):
         """
-        Checks that the purl field can be parsed into PackageURL object.
+        Validate that purl and ps_component comply with one of the following options:
+        * purl is correct and ps_component is not provided
+        * purl is correct and ps_component matches the one included in purl
         """
-        try:
-            if self.purl:
-                PackageURL.from_string(self.purl)
-        except Exception as exc:
-            raise ValidationError(f"Invalid purl '{self.purl}': {exc}")
+        if not self.purl and not self.ps_component:
+            raise ValidationError(
+                f"Affect ({self.uuid}) for {self.ps_module} must have either purl or ps_component."
+            )
+
+        if self.purl:
+            try:
+                ps_component_from_purl = PackageURL.from_string(self.purl).name
+            except ValueError as exc:
+                raise ValidationError(
+                    f"Affect ({self.uuid}) for {self.ps_module} has "
+                    f"an invalid purl '{self.purl}': {exc}."
+                )
+
+            if self.ps_component and self.ps_component != ps_component_from_purl:
+                raise ValidationError(
+                    f"Affect ({self.uuid}) for {self.ps_module} has a ps_component "
+                    "that does not match the one included in purl: "
+                    f"ps_component: {self.ps_component}, purl: {self.purl}."
+                )
 
     @property
     def aggregated_impact(self):
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -1035,6 +1035,11 @@ class AffectSerializer(
     trackers = serializers.SerializerMethodField()
     meta_attr = serializers.SerializerMethodField()
     cvss_scores = AffectCVSSSerializer(many=True, read_only=True)
+    # at least one of ps_component or purl is required
+    ps_component = serializers.CharField(
+        max_length=255, allow_blank=True, allow_null=True, required=False
+    )
+    purl = serializers.CharField(allow_blank=True, allow_null=True, required=False)
 
     @extend_schema_field(
         {
--- a/osidb/tests/factories.py
+++ b/osidb/tests/factories.py
@@ -300,9 +300,6 @@ class AffectFactory(BaseFactory):
     )
     ps_module = factory.sequence(lambda n: f"ps-module-{n}")
     ps_component = factory.sequence(lambda n: f"ps-component-{n}")
-    purl = factory.sequence(
-        lambda n: f"pkg:rpm/fedora/ps-component-{n}@7.50.3-1.fc25?arch=i386&distro=fedora-25"
-    )
     impact = factory.Faker(
         "random_element",
         elements=filter(lambda i: i != "LOW", list(Impact))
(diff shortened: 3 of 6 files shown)


osidb/tests/factories.py
```diff
@@ -300,9 +300,6 @@ class Meta:
     )
     ps_module = factory.sequence(lambda n: f"ps-module-{n}")
     ps_component = factory.sequence(lambda n: f"ps-component-{n}")
-    purl = factory.sequence(
-        lambda n: f"pkg:rpm/fedora/ps-component-{n}@7.50.3-1.fc25?arch=i386&distro=fedora-25"
-    )
```
jobselko, line 305: Since `purl` can be blank, I deleted it from the factory to avoid updating every test using `Affect`. If anybody thinks that keeping it here and updating all tests is a better option, I can do it.

osidb/api_views.py
```diff
@@ -840,6 +840,25 @@ class AffectView(
...
+
+        if data and data.get("purl") and not data.get("ps_component"):
+            new_data = data.copy()
+            purl = data["purl"]
+            try:
+                ps_component = PackageURL.from_string(purl).name
+            except Exception as exc:
+                raise ValidationError(f"Invalid purl '{purl}': {exc}")
+            new_data["ps_component"] = ps_component
+            kwargs["data"] = new_data
```
JakubFrejlach, line 858: I don't think this is the right place to perform such field population or performing any validations. This should be responsibility of either model layer or in this case more probably serializer level. We already perform some field prepopulation on serializer level by overriding the create and update methods there so I would use those as a reference. Also this approach still treats `ps_component` as required field on API which should no longer be true as it will be possible to have it blank in the request no.

JakubFrejlach (changes requested):

osidb/api_views.py
osoukup: I agree that the view is not a good place for it. Even the serializer is not ideal and the business logic should eventually move out of there but at least for now some of it is there. The validation should ideally be in the model validations and setting the component based on the PURL could eg. be in a pre-save signal or some part of the model save procedure.

jobselko: @JakubFrejlach I tried to override `create` in `AffectSerializer`/`AffectPostSerializer`, but without success, `update` itself was okay. If you prefer to use this approach here, I need some help to understand what is wrong with current serializers. I will try to update the API to describe that at least one of `purl` and `ps_component` is required.

JakubFrejlach: @jobselko I am not sure, what that override didn't work for you but as @osoukup wrote, doing that on the model layer, you can avoid this whole hustle. Just setting the ps_component not required at the serializer should be enough for you to successfully pass through serializer and then do the rest on the model layer.

osidb/tests/factories.py
jobselko: `ps_component` is generated from `purl`, so I would have to add `purl` into every mentioned test. I think that generating it oppositely would not make sense.

dmonzonis (approved):
LGTM

osidb/models/affect.py
```diff
@@ -172,6 +172,16 @@ class Meta:
     def __str__(self):
         return str(self.uuid)
 
+    def save(self, *args, **kwargs):
+        if self.purl and not self.ps_component:
+            try:
+                ps_component = PackageURL.from_string(self.purl).name
+            except Exception as exc:
+                raise ValidationError(f"Invalid purl '{self.purl}': {exc}")
```
osoukup, line 180: There should be some human-readable affect identification so when a user creates/updates multiple affects and then sees an error it is clear which one was invalid.

osidb/models/affect.py
```diff
@@ -483,6 +493,20 @@ def _validate_purl(self, **kwargs):
...
             raise ValidationError(f"Invalid purl '{self.purl}': {exc}")
 
+    def _validate_purl_and_ps_component(self, **kwargs):
+        """
+        Checks that ps_component matches the one included in purl if purl exists.
+        """
+        if (
+            self.purl
+            and self.ps_component
+            and self.ps_component != PackageURL.from_string(self.purl).name
```
osoukup, line 503: In the case user provides both but the PURL is invalid the conversion here would result in an exception which would most probably be unclear to the user and without any affect identification. We should catch it and transform into a better one.

osidb/models/affect.py
```diff
@@ -483,6 +493,20 @@ def _validate_purl(self, **kwargs):
...
+        ):
+            raise ValidationError(
+                "ps_component does not match the one included in purl: "
+                f"ps_component: {self.ps_component}, purl: {self.purl}."
```
osoukup, line 507: There should be the full affect identification (also PS module) so the user easily sees which one is incorrect.

osidb/serializer.py
```diff
@@ -1035,6 +1035,13 @@ class AffectSerializer(
     trackers = serializers.SerializerMethodField()
     meta_attr = serializers.SerializerMethodField()
     cvss_scores = AffectCVSSSerializer(many=True, read_only=True)
+    # ps_component is not required because it is created in Affect.save()
```
osoukup, line 1038: ... created out of PURL which is then required. Where do we enforce it?

osoukup (changes requested):
This looks almost OK but I would like the alert messages to be improved. Also it is not clear to me what happens when neither the component (which is no optional) nor PURL is provided. Will uniqueness constrain prevent this to be stored in the database? Will the error be human-readable? I would like to see a test for this case to prevent any potential future regression.

osidb/models/affect.py
```diff
@@ -172,6 +172,26 @@ class Meta:
     def __str__(self):
         return str(self.uuid)
 
+    def save(self, *args, **kwargs):
+        # Check that either purl or ps_component exists
+        if not self.purl and not self.ps_component:
+            raise ValidationError(
+                f"Affect ({self.uuid}) for {self.ps_module} must have either purl or ps_component."
+            )
```
osoukup, line 180: This is not the right place for the validation. A validation method should be defined for it

osidb/models/affect.py
```diff
@@ -172,6 +172,26 @@ class Meta:
...
+
+        # Check that the purl field can be parsed into a PackageURL object
+        if self.purl and not self.ps_component:
+            try:
+                ps_component = PackageURL.from_string(self.purl).name
+            except Exception as exc:
+                raise ValidationError(
+                    f"Affect ({self.uuid}) for {self.ps_module} has an invalid purl "
+                    f"'{self.purl}': {exc}. Provide either a valid one or ps_component."
+                )
```
osoukup, line 190: The same here. This should be a validation

osidb/models/affect.py
osoukup: I was asked to re-review but this does not seem to be addressed. I the PURL string is no parse-able the validation will produce an error message which is unrelated to what is being validated here

osoukup (changes requested):
A bit better but I still do not think that the validations are correct or on the correct place

osidb/models/affect.py
osoukup: I was not referring to the missing tests but to the non-caught PURL parsing exception in the condition

osidb/models/affect.py
```diff
@@ -473,15 +482,31 @@ def _validate_special_consideration_flaw(self, **kwargs):
...
+        elif self.purl and not self.ps_component:
+            try:
+                self.ps_component = PackageURL.from_string(self.purl).name
+            except Exception as exc:
+                raise ValidationError(
+                    f"Affect ({self.uuid}) for {self.ps_module} has an invalid purl "
+                    f"'{self.purl}': {exc}. Provide either a valid one or ps_component."
+                )
+
+        elif self.purl and self.ps_component != PackageURL.from_string(self.purl).name:
```
osoukup, line 505: Here is still a non-caught exception possibly happening during the PURL parsing which would result in potentially non-user-readable error message out of our control.

osoukup (changes requested):
Thank you for moving the validation logic to the validations. The non-caught exception is however still there. This seems to be the very last issue I have here

osidb/mixins.py
```diff
@@ -809,6 +805,10 @@ def validate(self, raise_validation_error=True, dry_run=False):
                         **(e.params or {}),
                     )
 
+        # standard validations
+        # exclude meta attributes
+        self.full_clean(exclude=["meta_attr"])
```
osoukup, line 810: This change should be done in a separate commit as it does not directly relate to PS components or PURLs but with the fact that our custom validations should get priority before the standard ones (as they are more specific and provide better readable and more targeted error messages).

osoukup (approved):
LGTM

PR #833, opened 2024-11-20
no Jira tracker component & version changes on update

osoukup:
We are supposed to only set these on tracker creation as it is common that engineers set them more specifically and the tracker updates then reverts their changes which makes them not totally super-happy.

Closes OSIDB-3677

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -40,6 +40,13 @@ class TrackerQueryBuilder:
         """
         return self.instance
 
+    @property
+    def is_creating(self):
+        """
+        True on tracker creation
+        """
+        return not self.tracker.external_system_id
+
     @cached_property
     def flaws(self):
         """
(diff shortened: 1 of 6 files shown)


apps/trackers/common.py
```diff
@@ -40,6 +40,13 @@ def tracker(self):
         """
         return self.instance
 
+    @property
+    def creation(self):
```
dmonzonis, line 44: How about `creating` or `is_creating`? IMO this way it's clearer that it's a boolean property.

dmonzonis (approved):

apps/trackers/common.py
osoukup: Makes sense. I will rename it

osoukup: The same naming was also used in the BBSync module and for the Bugzilla trackers. I changed the naming throughout to unify it on `is_creating`

PR #834, opened 2024-11-20
Allow closing flaws with no trackers and low/moderate impact

dmonzonis:
See individual commit messages for more info.

Closes OSIDB-3524.


osoukup (approved):
LGTM. I very much like the way you implemented this! I agree that currently it is the correct way and we will most probably need the complex conditions in the future too. This exact condition should IMO eventually be achieved be achieved in a different way which is to specify a different workflow for LOWs and MODERATEs but that will require the workflow framework rework into its former glory :smile: so now this is the way.

docs/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   FlawCVSS, FlawReference, Snippet and Tracker. (OSIDB-3466)
 - Moved metadata creation during tests to root level conftest and
   automatically set envs during VCR recording (OSIDB-3492)
+- Implement complex logic in workflow state requirements and allow moving a flaw to
+  state done if it has no trackers but impact is moderate or low (OSIDB-3524)
```
JakubFrejlach, line 20: I have just a very small nitpick, I would probably split this up and move implementation of the complex condition logic into `Added` section and in `Changed` section I would leave the workflow `DONE` state conditions change. But might be just my personal preference, leaving it up to you.

JakubFrejlach (approved):
Great job! I like the solution very much.

PR #835, opened 2024-11-21
task sync manager preparations

osoukup:
This PR is a move towards async task management. It aims to simplify the stuff a bit and remove some more unnecessary Jira API calls. This time we focus on the calls which sync back the task right after we updated it. We will just assume that when the update is successful the Jira issue will have the values we set. We also set the task last update timestamp on our side which may be a little bit before the actual update in Jira but will be synced by the collector in a short while so no big deal. A number of test changes were necessary along the way.

I added some more parts. I reworked the decision logic regarding the task management and moved some functionality so the execution parts are at the end separated methods which can be called both synchronously and async.

Then I run into an issue that the broken way serializers update the flaws (and other and create but that was put aside) means that the **save** and all the parameters are not used all at once but it is called multiple times with different context - or **tasksync** used to be called completely separately outside of the `save` call. It required quite some serializer rework which is definitely not complete but already brought what I needed. I think that the result is actually simpler and more understandable then it was before but I am glad for any feedback.

### DIFF
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -18,6 +18,8 @@ from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field, extend_schema_serializer
 from pghistory.models import Events
 from rest_framework import serializers
+from rest_framework.serializers import raise_errors_on_nested_writes
+from rest_framework.utils import model_meta
 
 from apps.bbsync.mixins import BugzillaSyncMixin
 from apps.taskman.constants import JIRA_TASKMAN_AUTO_SYNC_FLAW
@@ -350,6 +352,49 @@ class EmbargoedField(serializers.BooleanField):
                 raise serializers.ValidationError(
                     f"Cannot provide access for the LDAP group without being a member: {acl}"
                 )
+
+
+class BaseSerializer(serializers.ModelSerializer):
+    """
+    base serializer class which should be inherited by every serializer
+    of a model which save method requires any additional parameters
+
+    the reason is that the Django ModelSerializer does not provide any way
+    how to pass these parameters through create or update methods and then
+    those need to be called multiple times repeating the same actions and
+    complicating the whole save machinery
+    """
+
+    # TODO rewrite create machinery to use save
+    # def create(self, validated_data, *args, **kwargs):
+
+    def update(self, instance, validated_data, *args, **kwargs):
+        """
+        extended standard Django REST framework update method
+        optionally calling save with additional parameters
+        """
+        raise_errors_on_nested_writes("update", self, validated_data)
+        info = model_meta.get_field_info(instance)
+
+        m2m_fields = []
+        for attr, value in validated_data.items():
+            if attr in info.relations and info.relations[attr].to_many:
+                m2m_fields.append((attr, value))
+            else:
+                setattr(instance, attr, value)
+
+        # the additional arguments to the following save call are the only difference from the original
+        # https://github.com/encode/django-rest-framework/blob/3.15.2/rest_framework/serializers.py#L1018
+        instance.save(*args, **kwargs)
+
+        for attr, value in m2m_fields:
+            field = getattr(instance, attr)
+            field.set(value)
+
+        return instance
+
+    class Meta:
+        abstract = True
 
 
 class ACLMixinSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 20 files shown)


osoukup:
I am not getting any reviews and have to continue working so this PR is growing and growing... Sorry to the future reviewers that it will probably going to take some time but going commit-wise should help understand.

osidb/serializer.py
```diff
@@ -352,7 +354,50 @@ def validate_acl(self, embargoed):
...
+
+        m2m_fields = []
+        for attr, value in validated_data.items():
+            if attr in info.relations and info.relations[attr].to_many:
+                m2m_fields.append((attr, value))
+            else:
+                setattr(instance, attr, value)
+
+        # the additional arguments to the following save call are the only difference from the original
+        # https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L1018
```
JakubFrejlach, line 387: We should reference the exact version (permalink or probably even better tag) as master will change in the future and it will not be clear from exactly where this code derived.

Also by doing this we overwrite some core functionality code in DRF, if this for whatever reason changes in the furure with the new versions, we don't get the changes. I am thinking how can we ensure that we will remember to check this/update this.

osidb/serializer.py
```diff
@@ -892,23 +937,22 @@ def create(self, validated_data):
...
-    def update(self, instance, validated_data):
+    def update(self, instance, validated_data, *args, **kwargs):
         """
         perform the ordinary instance update
         with providing BZ API key while saving
         """
-        skip_bz_sync = validated_data.pop("skip_bz_sync", False)
-        instance = super().update(instance, validated_data)
-        if not skip_bz_sync:
-            instance.bzsync(bz_api_key=self.get_bz_api_key())
```
JakubFrejlach, line 903: Will be `bzsync` still performed after this change?

JakubFrejlach (commented):
I couple of things after first walkthrough but I will probably give it one more go.

osidb/serializer.py
```diff
@@ -352,7 +354,50 @@ def validate_acl(self, embargoed):
...
+    those need to be called multiple times repeating the same actions and
+    complicating the whole save machinery
+    """
+
+    # TODO rewrite create machinery to use save
+    # def create(self, validated_data, *args, **kwargs):
+
+    def update(self, instance, validated_data, *args, **kwargs):
+        """
+        exteded standard Django REST framework update method
```
jobselko, line 373: typo `exteded`

osidb/serializer.py
```diff
@@ -1416,8 +1451,108 @@ def update(self, retrieved_package_instance, validated_data):
         return package_instance
 
 
+# TODO I split the ACLMixinSerializer into two for now
+# as the PackageVersion specifics are not compatible with
+# the fixed way the serializers should work (calling save
+# just once using proper arguments and not multiple times)
+# and the specifics are too complicated and fiting them
```
jobselko, line 1458: typo `fiting`

osidb/serializer.py
osoukup: With the correct parameters it is performed here https://github.com/RedHatProductSecurity/osidb/blob/master/apps/bbsync/mixins.py#L37 through the regular `save` method. And it is more meaningfully done as part of that single `save` call at the end of it without serializer having to care.

osidb/serializer.py
osoukup: Very good point. I am referencing that line in the version 3.15.2 which is the one we use. I am not sure how to make sure we account for any significant change in the source here. I assume it should be pretty rare but we cannot know exactly. For now however it seems to me the right/easiest way how to push the necessary arguments to our custom `save` machinery avoiding the current workarounds in the serializer. I would say that the low risk of a future break is worth fixing the bad serializer-model integration we have.

osidb/serializer.py
osoukup: Fixed. Thank you!

osidb/serializer.py
osoukup: Fixed. Thank you!

JakubFrejlach (approved):
I went through second time and didn't find anything, LGTM! A lot of good work towards simplicity!

osidb/serializer.py
JakubFrejlach: I agree. Lets focus on the serializer offloading in the future.

PR #836, opened 2024-11-21
Set hard limit to paginated results

dmonzonis:
By default each page in our views returns 100 results, but the user can specify another limit using the `limit` keyword. Before this commit, there was no limit to the amonut of records that can be specified this way, which could lead to a DoS if the user specifies lots of records.

To avoid this, a hard limit has been implemented in for every paginated view, which cannot be exceeded. If it is exceeded, it is capped to the hard limit.

Closes OSIDB-643.


osoukup (approved):
LGTM

PR #843, opened 2024-11-25
Add all references as links when creating Jira trackers

dmonzonis:
Instead of only adding references of 'Upstream' type, now references of all types are added as links in Jira trackers.

Closes OSIDB-3733.

### DIFF
--- a/apps/trackers/jira/save.py
+++ b/apps/trackers/jira/save.py
@@ -71,8 +71,8 @@ class TrackerJiraSaver(JiraQuerier):
                 issue_key=issue.key,
                 body=comment,
             )
-        # Add upstream links only on tracker creation
-        for reference in tracker.upstream_references:
+        # Add references only on tracker creation
+        for reference in tracker.references:
             self.add_link(
                 issue_key=issue.key,
                 url=reference.url,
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM. BTW I encountered this PR by accident as no reviewers were set

jobselko:
@dmonzonis Could you please provide/extend some tests? Thanks!

dmonzonis:
@jobselko in the related change https://github.com/RedHatProductSecurity/osidb/pull/801 I added `test_add_link` which tests the method that adds the references as links. Since the links can only be added via an action and not in the query during creation, there is no change to the query builder that needs to be tested.

What do you think needs to be tested further?

jobselko:
> @jobselko in the related change #801 I added `test_add_link` which tests the method that adds the references as links. Since the links can only be added via an action and not in the query during creation, there is no change to the query builder that needs to be tested.
> 
> What do you think needs to be tested further?

Oh, I was not aware of this. Thanks for providing the context!

jobselko (approved):

JakubFrejlach (approved):

PR #844, opened 2024-11-27
Switch out custom fedora image

JakubFrejlach:
This PR:
* switches dependency from old custom fedora image to latest fedora image
* removes old gitlab CI config from the period of transition between Gitlab and Github


osoukup (approved):
LGTM. Thank you for handling this!

.github/workflows/test.yml
```diff
@@ -44,8 +44,50 @@ jobs:
         image: redis
     steps:
       - uses: actions/checkout@v4
+      # This step is required in order to install python-ldap
+      # see https://github.com/python-ldap/python-ldap/issues/432
+      - name: Fix installing python-ldap
+        run: echo "INPUT ( libldap.so )" > /usr/lib64/libldap_r.so
       - name: Install required packages
-        run: dnf install -y openldap-devel openldap-clients python3.9 postgresql-contrib
+        run: >
```
jobselko, line 52: Could the packages be sorted alphabetically?

osoukup: This is actually a good point. Otherwise it is harder for humans to search through the list :smile:

MrMarble (approved):

PR #845, opened 2024-11-27
set security level together with embargo status

osoukup:
Closes OSIDB-3598


osoukup:
@jsvob I have a question. I am reusing the existing Security Level setting logic which was not used for the new vulnerability issue type before and I am just using it as it is. The task is specific about the embargo case. Is it OK to also use it the standard way for the private and public case (here it is actually no difference, the query has it empty). Or should I suppose that if the automation wants a different visibility it will adjust right after but I am following the strictness as requested in the task :smile: setting it the expected most restrictive way initially.

jsvob:
@osoukup 

PR LGTM. See OSIDB-3598.

PR #849, opened 2024-12-03
Remove TODOs from BZSyncManager

MrMarble:
As discussed on today's stand-up, I've removed the TODOs on BZSyncManager class and added a small comment to state why the custom `schedule` implementation is needed

Closes OSIDB-3205


osoukup (approved):
LGTM

PR #855, opened 2024-12-09
disable external user ID queries

osoukup:
Closes OSIDB-2925


osidb/signals.py
```diff
@@ -79,8 +79,10 @@ def auto_create_profile(sender, instance, created, **kwargs):
     if created:
         Profile.objects.create(
             user=instance,
-            bz_user_id=get_bz_user_id(instance.email),
-            jira_user_id=get_jira_user_id(instance.email),
+            # TODO I am disabling the external system queries for now to safe the extra API calls
+            # and dependencies but the full removal requires User-Profile rework and major release
+            # bz_user_id=get_bz_user_id(instance.email),
+            # jira_user_id=get_jira_user_id(instance.email),
```
JakubFrejlach, line 85: I am thinking about the `if created` condition, doesn't that mean that this query should happen only when the User was created? And thus only once per user? I am actually not sure, what is our aim regarding the User profiles, what exact info we would like to store in.

JakubFrejlach (commented):
Before approval I would probably like to discuss briefly what is our aim for profiles. Changes don't seem to be saving much of a traffic, at least from my understanding of the logic.

osidb/signals.py
osoukup: That is the correct understanding as far as I was able to understand the Django User - it does not seem that there would be some lifetime etc. I will be glad to discuss as I am personally also not fully sure what is the right approach here. This all is not used at all and I can remember only that we wanted to use it for the history event authors mapping which should probably by rather a one-time action when we migrate Bugzilla history (and will be only partially successful as not all users have any/clear Kerberos counterpart). Otherwise the Profile seems quite useless at this point. Maybe it could be useful in the future, I am not sure. The full removal (at least from the API) is no possible at this point. I wanted to address the task for now leaving the options open for the future :smile: but maybe this is not the best approach.

JakubFrejlach: We don't use it right away, you are correct, so I am okay with just disabling it now for the sake of time saving and figure out what to do with it later.

JakubFrejlach (approved):

PR #856, opened 2024-12-10
OSIDB-3579: Implement alert versioning to reduce database locks

MrMarble:
- Added new field `last_validated_dt` to the `AlertMixin` and `created_dt` to the `Alert` Model.
- Remove the `alerts.all.delete()` method call to prevent database locks
- Modify the `AlertSerializer` to filter out alerts with a `created_dt` older than the `last_validated_dt` of the object they are related
- Created a `stale_alert_cleanup` celery task to run periodically

Closes OSIDB-3579


collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,42 @@
...
+        still valid by comparing the creation time of the alert
+        with the validation time of the Model.
+
+        If the creation time of the alert is older than
+        the validation time of the Model,
+        the alert is considered stale and will be deleted.
+        """
+        logger.info("Collecting Stale Alerts")
+
+        alerts = Alert.objects.all()
```
costaconrado, line 31: I'm suspecting this will take a long time to finish, we current have ~788k alerts in stage and querying them with all related objects would make this collector interact in almost all entries of the database, I suggest trying querying only for duplicated entries with something like
```
duplicate_alerts = Alert.objects.values('object_id', 'content_object').annotate(
    last_validated_dt_count=Count('last_validated_dt', distinct=True)
).filter(last_validated_dt_count__gt=1)
```

costaconrado (changes requested):
This PR looks mostly good to me, I just requested a change about the collector because I strongly believe it will take more than 1 hour to this collector do finish, but overall, good job!

collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,42 @@
...
+        stale_alerts_count = 0
+
+        for alert in alerts:
+            instance = alert.content_type.get_object_for_this_type(pk=alert.object_id)
```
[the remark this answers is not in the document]
MrMarble: Alert does not extend the AlertMixin, so the last_validated_dt is not added to the model

collectors/osidb_scheduler/collectors.py
MrMarble: Hmm yeah I need to improve my Django filters knowledge 😂 
I'll do some testing to improve it, thanks for the suggestion!

MrMarble: I have updated the collector with a more in depth logic to filter the alerts directly on the database level using subqueries.

I'm nowhere near the 788k alerts in the database, but here are the logs with 45k alerts, it takes 0.9 seconds to complete:
```
2024-12-11 11:34:14,427 [INFO] collectors.osidb_scheduler.collectors: Searching for stale alerts in 2 content types
2024-12-11 11:34:14,951 [INFO] collectors.osidb_scheduler.collectors: Found 1882 stale alerts
2024-12-11 11:34:15,233 [INFO] collectors.osidb_scheduler.collectors: Deleted 1882 stale alerts
```
I'll continue to run tests as the database grows, to see the performance

collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,67 @@
...
+
+        query = Q()
+
+        for content_type in content_types:
+            model_class = content_type.model_class()
+
+            subquery = Subquery(
+                model_class.objects.filter(
+                    pk=Cast(OuterRef("object_id"), output_field=model_class._meta.pk)
+                ).values("last_validated_dt")[:1]
```
costaconrado, line 48: @MrMarble is this subquery sorted by default?

collectors/osidb_scheduler/collectors.py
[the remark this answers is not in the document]
costaconrado: ah, I see now, sorry for overlooking it

collectors/osidb_scheduler/collectors.py
costaconrado: Thank you for addressing it!

collectors/osidb_scheduler/collectors.py
MrMarble: The subqueries are sorted by the default `Model.Meta.ordering` fields, but I think you're right, it should be sorted by the `last_validated_dt` field

costaconrado (approved):
LGTM, thanks!

collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,69 @@
...
+
+logger = get_task_logger(__name__)
+
+
+class StaleAlertCollector(Collector):
+    """
+    Stale Alert Collector
+    """
+
+    def collect(self):
```
JakubFrejlach, line 20: I would use:

```python
from osidb.core import set_user_acls
set_user_acls(settings.ALL_GROUPS)
```
To ensure ACLs are always properly bypassed before doing any collection so nothing is missed.

collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+
+        filtered_queryset = Alert.objects.filter(query)
+
+        if filtered_queryset.count() == 0:
+            return "No Stale Alerts Found"
+
+        logger.info(f"Found {filtered_queryset.count()} stale alerts")
+
+        deleted_alerts_count = filtered_queryset.delete()[0]
```
JakubFrejlach, line 65: I am thinking about atomicity, do we need to somehow enforce it here? :thinking:

collectors/osidb_scheduler/tests/test_collectors.py
```diff
@@ -0,0 +1,104 @@
...
+
+from osidb.mixins import Alert
+from osidb.models import Flaw, FlawSource
+from osidb.tests.factories import AffectFactory, FlawFactory, PsModuleFactory
+
+pytestmark = pytest.mark.unit
+
+
+class TestSlateAlertCollector:
+    @pytest.mark.vcr
```
JakubFrejlach, line 13: Why do we use the `vcr` decorator here? No live calls are made here, right?

collectors/osidb_scheduler/tests/test_collectors.py
```diff
@@ -0,0 +1,104 @@
...
+        # Stale alerts still exist
+        assert alerts.count() == 3
+
+        result = stale_alert_collector.collect()
+        alerts = Alert.objects.filter(object_id=flaw.uuid)
+        # Stale alerts have been deleted
+        assert alerts.count() == 1
+        assert result == "Deleted 2 Stale Alerts"
+
+    @pytest.mark.vcr
```
JakubFrejlach, line 41: Same as above.

collectors/osidb_scheduler/tests/test_collectors.py
```diff
@@ -0,0 +1,104 @@
...
+        # Stale alerts still exist
+        assert alerts.count() == 1
+
+        result = stale_alert_collector.collect()
+        alerts = Alert.objects.filter(object_id=affect.uuid)
+        # Stale alerts have been deleted
+        assert alerts.count() == 0
+        assert result == "Deleted 1 Stale Alerts"
+
+    @pytest.mark.vcr
```
JakubFrejlach, line 63: Same as above.

collectors/osidb_scheduler/collectors.py
MrMarble: Well I'm not sure about it, alerts are not modified once they go stale, so nothing should conflict

collectors/osidb_scheduler/apps.py
```diff
@@ -0,0 +1,11 @@
...
+    OSIDB service
+"""
+
+from django.apps import AppConfig
+
+
+class OSIDBScheduler(AppConfig):
+    """OSIDB scheduler service"""
+
+    name = "collectors.osidb_scheduler"
```
osoukup, line 11: I do not think that `collectors` is the right place. The collectors are meant to bring some data to OSIDB from external sources which does not happen here. I assume that this module should run rather internal tasks like housekeeping here or some checks but not really a data collection.

collectors/osidb_scheduler/collectors.py
```diff
@@ -0,0 +1,73 @@
...
+from django.db.models.functions import Cast
+
+from collectors.framework.models import Collector
+from osidb.core import set_user_acls
+from osidb.mixins import Alert
+
+logger = get_task_logger(__name__)
+
+
+class StaleAlertCollector(Collector):
```
osoukup, line 17: The collector framework is too have for this periodic job. It was designed for the data collection so the model has attributes like collected data status, up until when the data is complete, etc. while nothing like this is necessary here. You just perform full cleanup every time this job runs and that is it. No information to store needed, no status etc. Just a simple periodic Celery task should be enough or not? We do not have any framework for simple internal tasks like this (to eg. report status on API if needed etc.) but also we currently do not have a need for one :smile: so I would think that Celery should be enough.

collectors/osidb_scheduler/collectors.py
osoukup: This job should be the only piece of code performing write operations on the stale alerts so there should be no danger of any conflict. I do not see a reason for handling the atomicity. And as the `delete` is performed on the `queryset` I would expect that it is atomic anyway :smile:

docs/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Use keywords from ps-constants in CVEorg collector (OSIDB-3694)
+- Added last_validated_dt field to AlertMixin and created_dt to alerts;
+  optimized alert handling by matching valid alerts via timestamps
+  and scheduling stale alert removal to reduce database locks. (OSIDB-3579)
```
osoukup, line 12: nitpick (feel free to ignore): The CHANGELOG should be primarily for the user (the developers have Git history) and I am not sure they could have an idea what those attribute names stand for :smile: So describing the reasons or effects of the change is more important - or sometimes the CHANGELOG record could be IMO fully omitted if we do some purely internal changes which do not affect the user or in some totally indirect way.

osoukup (changes requested):
This change is great! Most of it LGTM. I have some complains I think should be addressed regarding the location of the module and that it uses the collector framework.

collectors/osidb_scheduler/apps.py
MrMarble: I've removed the collector and created a periodic celery task instead

docs/CHANGELOG.md
MrMarble: Removed the cange as it does not add value to the users

osidb/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+from django.db.models.functions import Cast
+
+from config.celery import app
+from osidb.core import set_user_acls
+from osidb.mixins import Alert
+
+logger = get_task_logger(__name__)
+
+
+@app.task(autoredtry_for=(Exception,))
```
osoukup, line 14: There seems to be an extra `d` in the parameter name. Also I am thinking what is the reason and consequence of the retry. What errors do we expect? I would probably not expect any exception being thrown during a regular run of the cleanup. If there is some, that might rather be some really serious fatality or a bug. Wouldn't that retry then actually mask the error? Maybe even repeating the failure indefinitely? If so then the retry might actually be undesirable and we should just let it fail so we can see the failure in Flower (not sure about Splunk).

osoukup (changes requested):
This is almost perfect! But :smile: I would like to fix and maybe even remove (or please explain the purpose) the `autoretry_for`

osidb/tasks.py
MrMarble: Honestly, I added it because I used the previous tasks as an example and they had it, but yeah it does not make much sense as it should not throw an error on any case, and even then, since is just removing old data we can afford some occasional errors.

I just removed it

osoukup (approved):
LGTM. Great work! Many thanks for addressing my feedback!

PR #858, opened 2024-12-12
CVEorg and OSV collectors cleanup

jobselko:
This PR changes the following:
- Removes a closing dot in the CVEorg collector error message since the dot is already a part of the nested `exc` variable
- Updates a note regarding invalid CVSS data handling in the OSV collector since the data is already fixed
- Splits `SNIPPET_CREATION_START_DATE` into `CVEORG_START_DATE` and `OSV_START_DATE` as they have different values
- Updates the CVEorg collector to be enabled by default and thus consistent with the OSV collector


collectors/osv/collectors.py
```diff
@@ -35,7 +35,9 @@ class OSVCollector(Collector):
 
     # When the start date is set to None, all snippets are collected
     # When set to a datetime object, only snippets created after that date are collected
-    snippet_creation_start_date = SNIPPET_CREATION_START_DATE
+    snippet_creation_start_date = timezone.datetime(
+        2024, 7, 1, tzinfo=timezone.get_current_timezone()
+    )
```
osoukup, line 40: I am thinking about a situation when one needs to test collection and with this change there is no easy way to customize the data. Would it be possible to make it still settable by an environment variable (the same for CVE.org) but with this date as a default  value? You would not have to specify anything in the OPS repo and nothing would be needed in the local environment file unless one wants to change it.

jobselko:
@osoukup Updated based on our discussion

collectors/osv/constants.py
```diff
@@ -1,4 +1,8 @@
-from osidb.helpers import get_env
+from osidb.helpers import get_env, get_env_date
 
 # Switch to turn the collector on/off
 OSV_COLLECTOR_ENABLED = get_env("OSV_COLLECTOR_ENABLED", default="True", is_bool=True)
+
+SNIPPET_CREATION_START_DATE = get_env_date(
+    "SNIPPET_CREATION_START", default="2024-07-01"
```
MrMarble, line 7: We use the same name for the cveorg collector but why the different default date?
What happens if we set the env `SNIPPET_CREATION_START=2024-07-01`, will the cveorg collector fail to create the snippets?

docker-compose.yml
```diff
@@ -93,7 +93,6 @@ services:
         PRODUCT_DEF_BRANCH: ${PRODUCT_DEF_BRANCH}
         PS_CONSTANTS_URL: ${PS_CONSTANTS_URL}
         SNIPPET_CREATION: ${SNIPPET_CREATION}
-        SNIPPET_CREATION_START: ${SNIPPET_CREATION_START}
```
osoukup, line 96: Why do we remove this line? The variable would be non-settable for the local dev testing without it

collectors/osv/constants.py
jobselko: The collector will not fail, but we agreed with VM team to start the collection as of October.

docker-compose.yml
jobselko: Fixed.

osoukup (approved):
LGTM

collectors/osv/constants.py
jobselko: Split into two variables.

jobselko:
This PR should be merged together with !117 in osidb-ops.

PR #859, opened 2024-12-13
OSIDB-3792: Update certificates

MrMarble:
Certificates expired December 8th, currently the pipeline is failing to install python dependencies with an SSL error.

I've added the updated CA Chain and also migrated the `2015` and `2022` Certificate authorities to the `Current` which is updated automatically with the latest certificates

Closes OSIDB-3792


osoukup (approved):
LGTM

PR #860, opened 2024-12-13
OSIDB-3579: Handle models without manager

MrMarble:
From my testing locally, all content-types had a manager to acces the `model.objects` field, but on stage, this task is failing because there's at least one model without a manager.

I've added a  log message to report the error and continue the loop


costaconrado (approved):
LGTM.

PR #862, opened 2024-12-17
Create new FlawLabel Collector

MrMarble:
Implementation of the flaw_label collector with a base `FlawLabel` model to store the labels on OSIDB.
Added new `FLAW_LABELS_URL` environment variable.

Closes OSIDB-3802


collectors/flaw_labels/core.py
```diff
@@ -0,0 +1,69 @@
...
+    response.raise_for_status()
+
+    try:
+        labels = yaml.safe_load(response.text)
+        return (labels["context_based_labels"], labels["product_family_labels"])
+    except yaml.YAMLError as e:
+        logger.error("Error parsing YAML", exc_info=e)
+
+
+def sync_flaw_labels(context_based: dict, product_family: dict):
```
osoukup, line 32: This should probably be atomic so it cannot happen that we delete the labels and then encounter an error interrupting the creation resulting in having no labels at all.

collectors/flaw_labels/tasks.py
```diff
@@ -0,0 +1,55 @@
+"""
+product definitions collector
```
osoukup, line 2: Copy-paste relic

osoukup (approved):
I already forgot what is the purpose of those include and exclude arrays but I trust that you know what needs to be done here :smile: You can consider introducing a `manage.py` sync command (similar to PS constants etc.). So LGTM

MrMarble:
> I already forgot what is the purpose of those include and exclude arrays but I trust that you know what needs to be done here 😄 You can consider introducing a `manage.py` sync command (similar to PS constants etc.). So LGTM

Those arrays (ps_components, ps_components_exclude...) will be used for automatic flaw labeling when affects are created/deleted.

Example:
The `ansible` label has this list of ps_modules:
   - ansible_engine-2
   - ansible_tower-3
   - ansible_automation_platform-1
   - ansible_automation_platform-2

When an affect is created we search the corresponding label with a queryset to apply it to the Flaw
```python
FlawLabel.objects.filter(
  (
    Q(ps_module__contains=[affect.ps_module]) |
    Q(ps_component__contains=[affect.ps_component])
  ) &
  (
    ~Q(ps_module_exclude__contains=[affect.ps_module]) |
    ~Q(ps_component_exclude__contains=[affect.ps_component])
  )
)
```

collectors/flaw_labels/core.py
MrMarble: done!

JakubFrejlach (approved):

PR #865, opened 2024-12-24
Sync external references to Jira tracker links

dmonzonis:
Before this commit, when filing Jira trackers, external/upstream references were created as links in the Jira issue, but this was only done during tracker creation, and no further changes in the references on OSIDB side were reflected in already created trackers.

With this commit, changes to references on OSIDB will create/update links in existing Jira trackers. Links are kept unique by their URL, so if a reference with a given URL changes its description, it will change the Jira link description as well, but if the URL changes, it will create a new Jira link.

Closes OSIDB-3733.


apps/taskman/tests/cassettes/test_service/TestTaskmanService.test_update_link.yaml
```diff
@@ -0,0 +1,650 @@
...
+        null, "customfield_12317313": null, "creator": {"self": "https://example.com/rest/api/2/user?username=rh-ee-dmonzoni",
+        "name": "rh-ee-dmonzoni", "key": "JIRAUSER227953", "emailAddress": "dmonzoni+stage@redhat.com",
+        "avatarUrls": {"48x48": "https://example.com/secure/useravatar?ownerId=JIRAUSER227953&avatarId=52988",
+        "24x24": "https://example.com/secure/useravatar?size=small&ownerId=JIRAUSER227953&avatarId=52988",
+        "16x16": "https://example.com/secure/useravatar?size=xsmall&ownerId=JIRAUSER227953&avatarId=52988",
+        "32x32": "https://example.com/secure/useravatar?size=medium&ownerId=JIRAUSER227953&avatarId=52988"},
+        "displayName": "Daniel Monzonis", "active": true, "timeZone": "Europe/Madrid"},
+        "subtasks": [], "customfield_12321140": null, "customfield_12320850": null,
+        "reporter": {"self": "https://example.com/rest/api/2/user?username=rh-ee-dmonzoni",
+        "name": "rh-ee-dmonzoni", "key": "JIRAUSER227953", "emailAddress": "dmonzoni+stage@redhat.com",
```
MrMarble, line 416: Should we remove the username?

requirements.txt
```diff
@@ -436,24 +350,16 @@ idna==3.10 \
...
-jeepney==0.7.1 \
-    --hash=sha256:1b5a0ea5c0e7b166b2f5895b91a08c14de8915afda4407fb5022a195224958ac \
-    --hash=sha256:fa9e232dfa0c498bd0b8a3a73b8d8a31978304dcef0515adc859d4e096f96f4f
-    # via
-    #   keyring
-    #   secretstorage
-jira==3.1.1 \
-    --hash=sha256:200c4d19f8be5ae39da3578597b78e21f3db69520c354be782bee9bd8bf21d09 \
-    --hash=sha256:e2fde55d04a421c590cb197cf6b2d01176028004387d3e4cedff653dba408238
+jira==3.8.0 \
```
MrMarble, line 360: Seems that they are having problems with the release process for a couple of months now https://github.com/pycontribs/jira/issues/1922 , the latest version is `v3.9.4` but is not published to PyPi https://github.com/pycontribs/jira/releases/tag/3.9.4

Not sure if this is something we need to consider

MrMarble (approved):

apps/taskman/tests/cassettes/test_service/TestTaskmanService.test_update_link.yaml
dmonzonis: Hmm, I could, but I've noticed this happens in pretty much every cassette that queries Jira, and since it's not really sensitive information I'm not sure if we should be worrying about manually removing any of these.

requirements.txt
dmonzonis: Could be worth upgrading if it gets published soon, although looking at the changelog I don't see any major fixes that would make that a priority. The main reason I upgraded was to be able to use the method to delete remote links.

osidb/serializer.py
```diff
@@ -1615,6 +1616,36 @@ class Meta:
...
+
+        for affect in new_ref.flaw.affects.all():
+            if affect.is_community:
+                continue
+
+            for tracker in affect.trackers.all():
+                if not tracker.is_closed and tracker.type == Tracker.TrackerType.JIRA:
+                    tracker.sync_reference(
+                        jira_token=self.get_jira_token(), reference=new_ref
+                    )
```
osoukup, line 1647: I do not want to over-complicate this right now but I am not convinced that the `serializer` module is a good place for this on-save logic and we should rather continuously migrate it to the model layer. I know that we already backed a lot of it into the `serializer` for various reasons in the past and it is not always trivial to change so I am just leaving a comment and not insisting on the change.

osoukup (approved):
LGTM

apps/taskman/tests/test_service.py
```diff
@@ -102,3 +102,22 @@ def test_add_link(self, jira_token):
...
         )
         assert response2.status_code == 201
+
+    @pytest.mark.vcr
+    def test_update_link(self, jira_token, monkeypatch):
+        """
+        Test that service is able to update remote links in Jira issues.
+        """
+        url = "https://www.redhat.com"
+        flaw = FlawFactory(embargoed=False, uuid="c581c407-5bc3-4a84-83ce-e3688e8fe87c")
```
JakubFrejlach, line 112: Any reason for directly specifying the `uuid`?

JakubFrejlach (approved):
I agree with @osoukup regarding putting things into serializers, but won't insist on changing that right now.

apps/taskman/tests/test_service.py
dmonzonis: No real reason other than it being done in the rest of the tests in that module. But you're right it's not really necessary so might remove them all

collectors/jiraffe/core.py
```diff
@@ -220,6 +226,58 @@ def add_link(self, issue_key, url, title) -> Response:
             )
             return Response(data=response_content, status=e.status_code)
 
+    def update_link(self, issue_key, link_id, url, title) -> Response:
+        """Update a remote link to a task by its internal ID."""
```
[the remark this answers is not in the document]
dmonzonis: Or even better, the more generic `Jira issue` since this is not limited to trackers but can be used for a flaw's task as well.

dmonzonis:
@osoukup @JakubFrejlach I moved the logic to the model layer since in this case it was easy, if you want to re-review

osoukup:
> @osoukup @JakubFrejlach I moved the logic to the model layer since in this case it was easy, if you want to re-review

I would consider the eventual end-goal that serializers do not do more than just serialization while there is still a bit more being done in the overridden `create`/`update` but this is definitely closer to that.

PR #866, opened 2024-12-31
Fix workflow validation with conditional requirements

dmonzonis:
This commit fixes a bug in which the workflow validation fails due to a missing property in conditional requirements, used to populate the list of requirements which are not met.

Closes OSIDB-3524.


MrMarble (approved):

osoukup (approved):
LGTM

PR #874, opened 2025-01-09
moved from docker.io to mirror.gcr.io

costaconrado:
This PR move images from docker.io to mirror.gcr.io making them free from Docker Hub limit download while still publicly available.

Closes OSIDB-3653.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -4,7 +4,7 @@ services:
 
     testldap:
       container_name: testldap
-      image: docker.io/bitnami/openldap:2.5
+      image: mirror.gcr.io/bitnami/openldap:2.5
       ports:
         - '1389:1389'
         - '1636:1636'
@@ -45,7 +45,7 @@ services:
 
     locust:
       container_name: locust
-      image: docker.io/locustio/locust:2.20.1
+      image: mirror.gcr.io/locustio/locust:2.20.1
       ports:
         - '9000:8089'
       volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -112,7 +112,7 @@ services:
       #   retries: 3
 
     osidb-data:
-      image: docker.io/library/postgres:13
+      image: mirror.gcr.io/library/postgres:13
       shm_size: 1gb  # Needed for more consecutive connections
       hostname: osidb-data
       container_name: osidb-data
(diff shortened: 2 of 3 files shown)


osoukup (approved):
The change looks really simple at the end! LGTM

MrMarble:
Sorry I got this late, but I think we shouldn't use the `mirror` directly, you can read the [docs](https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images#pull):
>there is no guarantee that a particular image will remain cached for an extended period of time. Only obtain cached images on mirror.gcr.io by [configuring the Docker daemon](https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images#configure).

The usage of the mirror is for the docker daemon to check first and if it does not exists, fetch from docker hub, but in this case if the image does not exists will just fail, we should just use `gcr.io`
Wdyt?

JakubFrejlach:
I also hadn't change to get to this. I agree with @MrMarble that this seems risky, as there is no guarantee that the images will be always in the registry.

Second thing is that we didn't updated our documentation to use these mirrors, there are still references to docker.io and it instructs developers to make docker.io account.

JimFuller-RedHat:
any reason why quay has not been considered ?

JakubFrejlach:
@JimFuller-RedHat I think it was considered, downside was the need of making mirrors of the images there ourselves, this worked more like a plug and play solution.

costaconrado:
@MrMarble I need to double check, but I believe gcr.io is a service from google that requires authentication and I believe has rate limits that would impose a similar issue that we are facing with docker.io.
 
@JimFuller-RedHat we can store in quay, but the main issue is that since we can´t use docker.io, it would require us to maintain parallel/nom-official dockerfiles to build images based on Red Hat images (e.g. ubi) or copy from docker.io that I think we can't do. We could also copy from the Google cache and it would to improve availability but we still rely on the cache for updating the image in quay.

costaconrado:
Before submitting this PR, I was working on start scripts to support both Red Hat images and docker.io images but it was becoming a clumsy solution, requires a bunch of new environments that I was thinking setting them in the makefile (e.g. user pick registry as a parameter of `make start-local` and the makefile set the variables needed) but it makes more convoluted and hard to understand the overall flow of the app in my opinion.

JakubFrejlach:
> Before submitting this PR, I was working on start scripts to support both Red Hat images and docker.io images but it was becoming a clumsy solution, requires a bunch of new environments that I was thinking setting them in the makefile (e.g. user pick registry as a parameter of `make start-local` and the makefile set the variables needed) but it makes more convoluted and hard to understand the overall flow of the app in my opinion.

Yeah, I once tried a similar approach and in the end it added a lot of complexity.

PR #876, opened 2025-01-10
fix consecutive save calls mid-air collisions

osoukup:
This PR fixes the issue of the possible mid-air collisions between the consecutive `save` calls. The issue is that inside the call the model in DB is being updated but it is not refreshed between the calls. If the top of the second is crosses during the save call, only the TrackingMixin timestamp is propagated but the `.update` one is not and  the mid-air collision occurs.

### DIFF
--- a/collectors/osv/tests/test_collectors.py
+++ b/collectors/osv/tests/test_collectors.py
@@ -2,7 +2,6 @@ import json
 from datetime import datetime, timezone
 
 import pytest
-from freezegun import freeze_time
 from jira.exceptions import JIRAError
 
 from apps.taskman.service import JiraTaskmanQuerier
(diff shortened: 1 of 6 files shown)


osoukup:
I am not sure I can write sane tests under the time pressure we have. I can reliably reproduce the workflow API call error by eg. putting 1 second sleep at https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/models/flaw/flaw.py#L1049 but the test seems to be a bit laborious.

osoukup:
I am analyzing whether there are more problematic places but this should be enough to fix the currently most pressing breakage of the workflow transition functionality. Therefore it is ready for a review.

collectors/osv/tests/test_collectors.py
```diff
@@ -14,7 +13,6 @@
 
 
 class TestOSVCollector:
-    @freeze_time(datetime(2020, 12, 12))  # freeze against top of the second crossing
```
jobselko, line 17: We have more usages of `@freeze_time` in our tests (e.g. also in CVEorg collector), but these can probably wait.

jobselko (approved):

MrMarble (approved):

PR #880, opened 2025-01-13
remove unnecessary test time-freeze

osoukup:
During the recent taskman/mixins rework I found some tests sometimes failing on 1s difference. I considered it a valid race condition between cassette and real-time shift but that was a silly and wrong assumption. The `freeze` was actually covering the real mid-air collision issue. Some of the occurrences were already removed during the recent hotfixes but these remained. However, I was not able to reproduce the previous mid-air collision even with a number of artificial delays on various places in the code so I assume the hotfixes actually fixed all of the mid-air collision issues of this kind.

I accidental passed over one extra historical `freeze` and that one was actually only needed because there was no refresh between multiple flaw DB operations which were modifying the update timestamp. I am fixing that one too.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -356,7 +356,7 @@
         "filename": "osidb/tests/endpoints/flaws/test_package_versions.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
-        "line_number": 141,
+        "line_number": 138,
         "is_secret": false
       }
     ],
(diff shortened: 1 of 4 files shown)


jobselko (approved):
Redundant `@freeze_time` is also in `TestCVEorgCollector`, not sure about other occurrences.

osoukup:
> Redundant `@freeze_time` is also in `TestCVEorgCollector`, not sure about other occurrences.

@jobselko I just grepped for the suspicious ones and this one does not look suspicious in any way :smile: but I will check it and remove if it seems to be unnecessary

PR #881, opened 2025-01-13
Trim Jira task summary if flaw's cve_id and title are too long

jobselko:
This PR adds a check to ensure that the summary in a Jira task is trimmed when the flaw's `cve_id` and `title` exceed the maximum allowed length of 255 characters.

Closes OSIDB-3847

### DIFF
--- a/apps/taskman/service.py
+++ b/apps/taskman/service.py
@@ -17,6 +17,7 @@ from osidb.models import Flaw, Impact, PsProduct
 
 from .constants import (
     JIRA_STORY_ISSUE_TYPE_ID,
+    JIRA_SUMMARY_MAX_LENGTH,
     JIRA_TASKMAN_PROJECT_ID,
     JIRA_TASKMAN_PROJECT_KEY,
     JIRA_TASKMAN_URL,
@@ -209,6 +210,10 @@ class JiraTaskmanQuerier(JiraQuerier):
             summary = flaw.title
         else:
             summary = f"{flaw.cve_id} {flaw.title}"
+
+        if len(summary) > JIRA_SUMMARY_MAX_LENGTH:
+            # Trim the maximum summary length by 3 to account for the triple dots
+            summary = f"{summary[:JIRA_SUMMARY_MAX_LENGTH-3]}..."
 
         data = {
             "fields": {
--- a/apps/taskman/tests/test_service.py
+++ b/apps/taskman/tests/test_service.py
@@ -7,6 +7,7 @@ instance for generating new cassettes.
 
 import pytest
 
+from apps.taskman.constants import JIRA_SUMMARY_MAX_LENGTH
 from apps.taskman.service import JiraTaskmanQuerier
 from apps.workflows.workflow import WorkflowModel
 from osidb.models import Flaw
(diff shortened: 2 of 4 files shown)


apps/taskman/tests/test_service.py
```diff
@@ -120,3 +120,21 @@ def test_update_link(self, jira_token, monkeypatch):
...
+            cve_id="CVE-2024-7142",
+            title="On Arista CloudVision Appliance (CVA) affected releases running on appliances that support "
+            "hardware disk encryption (DCA-350E-CV only), the disk encryption might not be successfully "
+            "performed. This results in the disks remaining unsecured and data on them",
+        )
+        AffectFactory(flaw=flaw)
+        taskman = JiraTaskmanQuerier(token=jira_token)
+
+        response = taskman.create_or_update_task(flaw=flaw)
+        assert response
```
dmonzonis, line 140: This test will succeed even without the change introduced in this commit, so I'm not entirely sure what's actually being tested.

apps/taskman/service.py
```diff
@@ -215,6 +215,8 @@ def _generate_task_data(self, flaw: Flaw):
             summary = flaw.title
         else:
             summary = f"{flaw.cve_id} {flaw.title}"
+        if len(summary) > 255:
```
osoukup, line 218: Instead of magic number use a constant with a descriptive name so it is clear what is the meaning of this

apps/taskman/tests/test_service.py
osoukup: True. The test using a stored cassette is not a good option here as the Jira call will just return the stored success despite the length of the summary. I would suggest to check the generated Jira query to make sure that the title being sent is always below the limit regardless of the actual flaw title length.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Removed `last_validated_dt` from exposed JSON Flaw History data (OSIDB-3814)
+- Trim flaw summary if it is too long (OSIDB-3847)
```
osoukup, line 10: I do not think that this is a good description. The flaw does not have a **summary** attribute so the user reading this would hardly understand. We are actually not even trimming any flaw attribute. The entity in question is the Jira task

osoukup (changes requested):
This needs some polishing

apps/taskman/tests/test_service.py
JakubFrejlach: The tested title is exactly 255 characters long, so it will always succeed.

```python
In [1]: from osidb.tests.factories import FlawFactory

In [2]: flaw = FlawFactory(
   ...:     cve_id="CVE-2024-7142",
   ...:     title="On Arista CloudVision Appliance (CVA) affected releases running on appliances that support "
   ...:     "hardware disk encryption (DCA-350E-CV only), the disk encryption might not be successfully "
   ...:     "performed. This results in the disks remaining unsecured and data on them",
   ...: )

In [3]: flaw.title
Out[3]: 'On Arista CloudVision Appliance (CVA) affected releases running on appliances that support hardware disk encryption (DCA-350E-CV only), the disk encryption might not be successfully performed. This results in the disks remaining unsecured and data on them'

In [4]: len(flaw.title)
Out[4]: 255
```

jobselko: I agree with you, my bad.

apps/taskman/service.py
```diff
@@ -34,6 +34,9 @@
     Impact.NOVALUE: JiraPriority.UNDEFINED,
 }
 
+# The length of summary in a Jira task is restricted to 255
+JIRA_SUMMARY_MAX_LENGTH = 255
```
osoukup, line 38: Is there any reason to put the constant here? Taskman module has `constants.py` file

apps/taskman/service.py
```diff
@@ -210,6 +213,9 @@ def _generate_task_data(self, flaw: Flaw):
         else:
             summary = f"{flaw.cve_id} {flaw.title}"
 
+        if len(summary) > JIRA_SUMMARY_MAX_LENGTH:
+            summary = f"{summary[:252]}..."
```
osoukup, line 217: I do not want to be nitpicky but 252 still sounds a bit magic - it is not clear how does it relate to the rest of the statement and by just looking here one cannot guess that max length is actually 255. What about getting it by subtracting the triple dot length from the constant?

apps/taskman/tests/test_service.py
```diff
@@ -120,3 +120,22 @@ def test_update_link(self, jira_token, monkeypatch):
...
+            title="On Arista CloudVision Appliance (CVA) affected releases running on appliances that support "
+            "hardware disk encryption (DCA-350E-CV only), the disk encryption might not be successfully "
+            "performed. This results in the disks remaining unsecured and data on them",
+        )
+        AffectFactory(flaw=flaw)
+        assert len(flaw.cve_id) + len(flaw.title) > 255
+
+        taskman = JiraTaskmanQuerier(token=jira_token)
+        issue_data = taskman._generate_task_data(flaw)
+        assert len(issue_data["fields"]["summary"]) == 255
```
osoukup, line 140: 255 has not meaning. Use the constant you defined so it is descriptive

osoukup (changes requested):

jobselko:
@osoukup Please review again, sorry for the silly mistakes :)

osoukup (approved):
LGTM. Thank you for addressing my feedback!

PR #883, opened 2025-01-13
Fix fatal error for NoneType values

superbuggy:
Adds a more strict guard conditional for the action to strip `last_validated_dt` field in `serializer.py::HistoricalEventSerializer`.

Closes OSIDB-3814 (once more, with feeling)

### DIFF
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -604,7 +604,10 @@ class HistoricalEventSerializer(serializers.ModelSerializer):
 
     def to_representation(self, instance):
         representation = super().to_representation(instance)
-        if representation["pgh_diff"]:
+        if (
+            isinstance(representation["pgh_diff"], dict)
+            and "last_validated_dt" in representation["pgh_diff"]
+        ):
             representation["pgh_diff"].pop("last_validated_dt")
         return representation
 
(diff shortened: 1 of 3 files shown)


osidb/serializer.py
```diff
@@ -604,7 +604,10 @@ class Meta:
 
     def to_representation(self, instance):
         representation = super().to_representation(instance)
-        if representation["pgh_diff"]:
+        if (
+            isinstance(representation["pgh_diff"], dict)
+            and "last_validated_dt" in representation["pgh_diff"]
+        ):
             representation["pgh_diff"].pop("last_validated_dt")
```
osoukup, line 611: Is it guaranteed that `last_validated_dt` is there? I am not too familiar with the history and whether the record is always a full model or just a subset but I mean that it is a pretty new attribute. Evetually `.pop("last_validated_dt", None)` should make sure it does not result in a `KeyError`.

osoukup (approved):
I am not certain about one thing so I am adding a comment but if you find the concern not to be valid then LGTM

osidb/serializer.py
JakubFrejlach: It is covered above in the condition:
```python
and "last_validated_dt" in representation["pgh_diff"]
```

JakubFrejlach (approved):
LGTM!

osidb/serializer.py
osoukup: No idea why I did not see it :smile: never mind then...

dmonzonis (approved):

MrMarble (approved):

PR #884, opened 2025-01-14
Implement exclusion SLAs and 'in' operator

dmonzonis:
An SLA policy with a null SLA will be used to exclude trackers under certain conditions from getting an SLA and therefore a due date. Exclusion SLAs will take priority over other SLAs, so if a condition is met for a tracker to be excluded, it will not get a due date even if it matches other SLAs.

Additionally, the 'in' operator has been implemented for SLA conditions, allowing to check if an attribute has one of many values described in a YAML list.

An example that showcases both an exclusion SLA and the 'in' operator is the following, which will exclude any trackers for which the `ps_component` of the affect matches one of the values listed:

```yaml
name: Excluded components
description: Components which do not need SLA
conditions:
  affect:
    - PS component in:
      - kpatch
      - kmatch
      - kcatch
sla: null
```

Closes OSIDB-3711.


apps/sla/tests/test_framework.py
```diff
@@ -458,3 +477,77 @@ def test_tracker_sla(self, is_closed):
...
+description: only for very serious cases
+conditions:
+  flaw:
+    - major incident state is approved
+sla:
+  duration: 2
+  start: reported date
+  type: calendar days
+"""
+            flaw = FlawFactory(
```
osoukup, line 525: If I am not overseeing something it seems to me that if the generated flaw is Major Incident the test case of no exclusion may fail on its last `assert`. Or am I wrong?

apps/workflows/checks.py
```diff
@@ -187,3 +210,26 @@ def desc2not_equals(self, check_desc):
...
+
+                def check_inclusion(instance):
+                    field = getattr(instance, attr)
+                    return (field if not callable(field) else field()) in values
+
+                return (doc, check_inclusion)
+
+    def desc2not_in(self, statement, values):
+        if statement.count("_not_in") == 1:
+            result = self.desc2in(statement, values)
```
osoukup, line 229: `_not` should be removed before calling `desc2in` - a test case should uncover that

osoukup (changes requested):
I have a few comments and some need to be addressed or answered. I generally do not like that exclusion and `in` operator are introduced int the same commit but I am OK leaving it as it. Otherwise I think that the implementation is great and goes exactly the right way as I understand it :1st_place_medal:

apps/sla/models.py
```diff
@@ -180,9 +186,16 @@ def __eq__(self, other):
 
     def __lt__(self, other):
         """
-        empty SLA context is greater
+        empty SLA context is greater,
+        exclusion SLA is smaller,
         otherwise compare the end dates
         """
+        # Exclusion takes priority
```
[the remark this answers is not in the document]
osoukup: @dmonzonis here is the fix to the documentation https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/merge_requests/128 - please review.

osoukup:
@dmonzonis you should also accompany this PR with the SLA documentation update in the PS constants repo

apps/sla/tests/test_framework.py
dmonzonis: You're right, I did not remember this was also set randomly. I changed it.

dmonzonis:
@osoukup I addressed your points and split it into two commits, which does make more sense since the changes were independent of each other anyway, thanks for pointing that out.

dmonzonis:
Here is the MR that will update the docs with this new functionality: https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/merge_requests/129

PR #885, opened 2025-01-16
[OSIDB-3804] Add Collaboration labels on flaw promotion

MrMarble:
The original scope of this task was only to trigger Label creation when the Flaw was promoted to `SECONDARY ASSESSMENT` based on the Affects, during testing I found a corner case that we didn't think of, and the VM team decided to make some changes.

- Added new `relevant` field to track if a `product_based` label is removed (Affect is removed from flaw or ps_component/ps_module are updated)
- Set `Flaw.label` field to **read only**
- Expose and endpoint to update `context_based` labels, previously this was done with the `Flaw.labels` field, but the complexity on update was to much to maintain
- Removed `FlawCollaborator.label` Foreign Key in favor of a CharField to allow labels to exists even if the label definition changes
- Added `FlawCollaborator.type` Field for the same reason
- Created management command `sync_flaw_labels` to trigger the synchronization with label definitions

Closes OSIDB-3804


osidb/models/flaw/label.py
```diff
@@ -76,19 +167,40 @@ class FlawCollaboratorState(models.TextChoices):
 
     contributor = models.CharField(max_length=255, blank=True)
 
+    relevant = models.BooleanField(default=True)
```
osoukup, line 170: What does it mean that a label is not relevant? Is such label needed?

osidb/models/flaw/label.py
```diff
@@ -76,19 +167,40 @@ class FlawCollaboratorState(models.TextChoices):
...
     class Meta:
         constraints = [
             models.UniqueConstraint(
                 fields=["flaw", "label"], name="unique label per flaw"
             ),
         ]
 
+    def create(self, *args, **kwargs):
+        self._validate_label()
+        super().create(*args, **kwargs)
```
osoukup, line 183: I assume that this is for some case when we bypass `save` which should better not happen :smile: What is the reason for it? Also explicitly calling validations on various places or raising `ValidationError` directly from `save` seems all together already a bit hacky to me. Could we use `AlertMixin` instead?

osoukup (commented):
I think that this probably OK but I am not sure I understand well how it is supposed to be so I am asking a few questions before I can approve

osidb/models/flaw/label.py
```diff
@@ -42,10 +65,75 @@ class FlawLabelType(models.TextChoices):
...
+
     def __str__(self):
         return self.name
 
 
+class FlawCollaboratorManager(TrackingMixinManager):
+    """collaborator manager"""
+
+    @staticmethod
+    def create_from_affect(affect: Affect):
```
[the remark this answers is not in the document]
MrMarble: > Are they supposed to be created manually or automatically?

Both actually 😆 
`Product based` labels are added automatically to a flaw using the affects ps_modules/ps_components, `Context based` labels are added manually by the users whenever they want. 

> will it be expected that people has to resolve the collaborator states?

The Idea is that if the person working on the Flaw needs help, they can change the status of one of the labels to `REQ`, that will highlight the label in OSIM, so anyone that want to collaborate, can jump in, adding himself as the `contributor` and do whatever is requested. 

The static methods defined in the Manager are helpers to create the `product_family` labels from affects when the workflow triggers, `context_based` labels are handled manually by the users, thats why there aren't methods for it

osidb/models/flaw/label.py
MrMarble: These was one of the changes that was not part of the initial spec.

In OSIM/OSIDB we allow affect deletion and also modifying the ps_components/ps_modules of an affect.
`product_family` labels (the ones created automatically) are marked as `irrelevant` if the affect that added the label was removed or the component/module is not longer valid.

I proposed to just change the state of the label to `SKIP` or even create a new one, but they want to keep the state field just in case it is needed even if the label is not longer relevant.

(the name can be changed, I'm quite bad at naming things 😆 )

osidb/models/flaw/label.py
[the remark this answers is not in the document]
osoukup: OK. So the action is not required by default, only when needed. I worried that for large flaws a lot of extra work would be generated in the sense of resolving automatic labels.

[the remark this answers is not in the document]
MrMarble: If you're worried about performance, I've tried with a flaw with 130 affects and the call to `FlawCollaborator.objects.create_from_flaw(f)` only takes 1ms

osidb/models/flaw/label.py
MrMarble: The reason they gave me is that `nothing should be erased`, and keeping the state field is just in case an analyst has already put some effort into it and wants to move it to `DONE` or `SKIP` when finished, instead of the label being removed or the sate changed. I can ask for a better justification 😆

osidb/models/flaw/label.py
osoukup: OK :thinking: I see! So the source of the weirdness is that the creation is more strict than the update. So probably leave it as it is and we shall see.

osoukup (approved):
I am for some reason a bit skeptical about this but when it is wanted, then let give it to the requestors. I do not want to block it just for that I do not like it. LGTM

JakubFrejlach (approved):
Couple of suggestions/questions, but nothing serious, looking good.

PR #890, opened 2025-01-23
Validate flaw impact and RH CVSSv3 score properly

jobselko:
This PR consolidates the flaw's impact and RH CVSSv3 score validations as they are currently inconsistent and incomplete.

If RH CVSSv3 is present, score and impact should comply with the following:
- RH CVSSv3 score is not zero and flaw impact is set
- RH CVSSv3 score is zero and flaw impact is not set

Closes OSIDB-3738

### DIFF
--- a/apps/trackers/tests/test_jira.py
+++ b/apps/trackers/tests/test_jira.py
@@ -1825,12 +1825,18 @@ class TestTrackerJiraQueryBuilder:
         if flaw_cvss_present:
             assert flaw.cvss_scores.all().count() == 0
             flawcvss = []
-            flawcvss.append(FlawCVSSFactory(flaw=flaw))
+            flawcvss.append(FlawCVSSFactory(flaw=flaw, issuer=FlawCVSS.CVSSIssuer.NIST))
             assert flaw.cvss_scores.all().count() == 1
             if multiscore:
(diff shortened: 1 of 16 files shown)


osidb/tests/factories.py
jobselko, line 1: FYI, I chose not to modify `FlawFactory` to automatically create RH CVSSv3 if necessary, as this change would break many tests and fixing it would be difficult due to the non-deterministic nature of the tests.

apps/trackers/tests/test_jira.py
```diff
@@ -2367,7 +2373,11 @@ def test_generate_query_multiflaw(self):
             source="REDHAT",
             cwe_id="CWE-1",
         )
-        flwcvss = FlawCVSSFactory(flaw=flaw, issuer=FlawCVSS.CVSSIssuer.REDHAT)
+        flwcvss = FlawCVSSFactory(
```
osoukup, line 2376: I guess that this was supposed to be `flawcvss` instead like it is above

osoukup (approved):
LGTM

apps/trackers/tests/test_jira.py
jobselko: This is intentionally, see `flwcvss2`, `flwcvss3`...

PR #895, opened 2025-01-29
make Jira task collector to write only necessary attributes

osoukup:
After finally being able to document an error case in more detail (in the Jira issue reference below) my understanding of the problem is that when a flaw update leading to a consecutive Jira task update happens at the top of the minute, it can happen that the Jira task collector is too fast and fetches the new task changes before the update operation which triggered it could even finish potentially (under circumstances of some unfortunate operation interlacing) reverting part of the changes done by the update.

This is PR mitigates conflict (it will still happen but hopefully with no effect) by making the task collector to write only the task-related flaw attributes. Then, despite the order of the operations we should always end up with the same result.

I am however not able to reproduce the issue and I therefore I am not able to test it or write a test - it is simply too random and I have only a very high-level understanding of what probably happens. Thoughtful review welcome :smile: 

Closes OSIDB-3636

### DIFF
--- a/collectors/jiraffe/convertors.py
+++ b/collectors/jiraffe/convertors.py
@@ -147,9 +147,31 @@ class JiraTaskSaver:
         self.flaw = flaw
 
     def save(self):
-        self.flaw.save(
-            auto_timestamps=False,
-            raise_validation_error=False,
+        # only update the fields which are supposed
+        # to be potentially influenced by the collector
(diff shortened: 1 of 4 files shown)


costaconrado (approved):
I agree with the approach of only updating the fields it actually have, this should improve OSIDB-3636.
LGTM.

PR #897, opened 2025-01-30
filter out empty events from history API resutls

osoukup:
Closes OSIDB-3942


costaconrado (approved):
LGTM.

costaconrado:
Maybe we should consider excluding this field from the audit log since its just for avoiding alert conflicts.

osoukup:
> Maybe we should consider excluding this field from the audit log since its just for avoiding alert conflicts. In any case the PR LGTM.

@costaconrado I was thinking about it but I am not sure. It is a change and maybe having the audit log full might be valuable at some point but hard to tell

PR #900, opened 2025-01-30
Map user name or email to pghistory context, replacing integer user id 

superbuggy:
Replaces the integer user id supplied by `pghistory.middleware.HistoryMiddleware` which reads from `request.user.pk` with a user name or email. Testing this locally however is not feasible since the `backend.py::LDAPRemoteUser` class sources user name and email from kerberos after authentication; therefore UAT testing seems required.

- [ ] Passed tests on UAT

Closes OSIDB-3464


osoukup (commented):
I think I would need to understand more to be able to approve this :smile: So this enriches the **pgh** context with the user being some identifying string and then the record is stored to DB like this? Will it be possible to map the existing DB records to the users?

superbuggy:
> I think I would need to understand more to be able to approve this 😄 So this enriches the **pgh** context with the user being some identifying string and then the record is stored to DB like this? Will it be possible to map the existing DB records to the users?

That is correct, I'm still in the process of developing the logic, triggered when a single flaw is requested, to change a FlawAudit pg_context entry to use a username/email in place of a numerical id.

osidb/middleware.py
```diff
@@ -32,3 +34,41 @@ def __call__(self, request):
...
+        if getattr(request.user, "is_authenticated", False):
+            if getattr(request.user, "email", False):
+                user = request.user.email
+            elif getattr(request.user, "username", False):
+                user = request.user.username
+
+        with pghistory.context(user=user, path=request.path):
+            return get_response(request)
+
+        return get_response(request)
```
dmonzonis, line 72: Is this ever executed?

osidb/middleware.py
```diff
@@ -32,3 +34,41 @@ def __call__(self, request):
             _logger.info(f"ACLs set from middleware for user {request.user.username}")
             set_user_acls(request.user.groups.all())
         return self.get_response(request)
+
+
+def OsidbHistoryMiddleware(get_response):
```
dmonzonis, line 39: Is there a reason why we're defining this middleware as a function while the other is a class? I'm okay with either option but if there's no strong reason to choose one over the other I'd rather keep it consistent at least.

JakubFrejlach, line 39: I am thinking, we do we create a new middleware here which tampers with the `pghistory.middleware.HistoryMiddleware` and `pghistory.middleware.WSGIRequest`. Could we scrape the usage of the HistoryMiddleware and just use the `OsidbHistoryMiddleware`? Looking at the code, there is only the logic which adds user pk there, and we then replace it with the username or email, so why no just skip the pk step? :thinking:

osidb/middleware.py
```diff
@@ -32,3 +34,41 @@ def __call__(self, request):
...
+                    _value = value.username
+        return super(WSGIRequest, self).__setattr__(attr, _value)
+
+    WSGIRequest.__setattr__ = _setattr
+
+    def middleware(request):
+        if not request or not hasattr(request, "user"):
+            return get_response(request)
+
+        user = f"{request.user}"
```
JakubFrejlach, line 62:
```python
user = str(request.user)
```

osidb/middleware.py
```diff
@@ -32,3 +34,41 @@ def __call__(self, request):
...
+    """
+
+    def _setattr(self, attr, value):
+        _value = value
+        if attr == "user":
+            if getattr(value, "is_authenticated", False):
+                if getattr(value, "email", False):
+                    _value = value.email
+                if getattr(value, "username", False):
+                    _value = value.username
```
JakubFrejlach, line 53: In this case `username` has higher preference over `email` because of two separate `if` but in the case bellow in the `middleware` method, the `email` has higher preference over `username` because of `if elif`, shouldn't these two match?

JakubFrejlach (commented):
Couple of questions. Mainly about the mixing the HistoryMiddleware and OsidbHistoryMiddleware.

osidb/middleware.py
superbuggy: Missed this, thank you!

osidb/middleware.py
superbuggy: I had the same thought, it would be less awkward and complicated to ditch the bundled history middleware, I think

osoukup (commented):
AFAIK LGTM. Is there still some issue which needs to be solved or could it be done this way?

JakubFrejlach (approved):
Thanks for addressing all the points. LGTM now.

osoukup (approved):
Good job!

MrMarble (approved):

superbuggy:
> > AFAIK LGTM. Is there still some issue which needs to be solved or could it be done this way?
> 
> The one unfinished task is to update old entries, I opted to do this in `api_views.py` and will have this for review shortly. Conceptually that doesn't quite feel like the right place, but feels appropriate to restrict the update mechanism from triggering when querying multiple flaws and introducing unwanted load/overhead

I opted to create another ticket for this logic with the intention of including it a separate release.

PR #901, opened 2025-01-31
skip non-migrated tracker sync (OSIDB-3966)

osoukup:
for the open Bugzilla trackers which project already migrated to Jira we need to skip the eventual sync not to interrupt the update

### DIFF
--- a/osidb/models/tracker.py
+++ b/osidb/models/tracker.py
@@ -160,13 +160,27 @@ class Tracker(AlertMixin, TrackingMixin, NullStrFieldsMixin, ACLMixin):
         if raise_validation_error:
             self.validate(dry_run=kwargs.get("no_alerts", False))
 
+        # TODO the following condition can be removed when we fully leave Buzilla
+        #
+        # there may still exist open trackers in Bugzilla at the time when the
+        # project their supposed to belong to migrated to Jira already which
+        # breaks the integration and as it is not simple to address but only
+        # temporary we agreed on a workaround of skiping the tracker sync
+        skip_sync = False
+        if affect := self.affects.first():
+            if ps_module := PsModule.objects.filter(name=affect.ps_module).first():
+                # tracker type and BTS mismatch
+                if self.TYPE2BTS[self.type] != ps_module.bts_name:
+                    skip_sync = True
+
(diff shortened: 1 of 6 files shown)


osidb/models/tracker.py
```diff
@@ -160,6 +160,18 @@ def save(self, *args, bz_api_key=None, jira_token=None, **kwargs):
...
+        #
+        # there may still exist open trackers in Bugzilla at the time when the
+        # project their supposed to belong to migrated to Jira already which
+        # breaks the integration and as it is not simple to address but only
+        # temporary we agreed on a workaround of skiping the tracker sync
+        if affect := self.affects.first():
+            if ps_module := PsModule.objects.filter(name=affect.ps_module).first():
+                # tracker type and BTS mismatch
+                if self.TYPE2BTS[self.type] != ps_module.bts_name:
+                    return
```
dmonzonis, line 173: Instead of returning, shouldn't we skip sync and continue with a regular save with `super().save(*args, **kwargs)`? I'm not sure if this could break the save chain and leave other models in an improper state

osoukup: It would mean that we would update the `updated_dt` and desync with Bugzilla. The aim is really cut to the OSIDB-Bugzilla way of sync and only get the updates from there - mostly the eventual closure.

dmonzonis: But do we even care if it updates `updated_dt` and desyncs with Bugzilla? Isn't that the issue to begin with, that they are out of sync anyway so we skip them?

osidb/serializer.py
```diff
@@ -2045,7 +2045,7 @@ def mi_differ(flaw1, flaw2):
                     # If the non-components attributes are unchanged, a tracker update is
                     # necessary only for Vulnerability issuetype Jira trackers because only
                     # those contain components. And this tracker is not Vuln. issuetype.
-                    return
```
costaconrado, line 2048: Nice catch :open_mouth:

costaconrado (approved):
LGTM.

dmonzonis (approved):
LGTM now, thanks for addressing my concerns

PR #905, opened 2025-02-05
custom validations before standard ones

osoukup:
This PR changes the order of the validations. The standard Django validations which were run through **full_clean** are now postponed after the custom ones. The reason is that the custom ones provide more refined error messages and if we run them after the standard ones the user can get an error which is not really the correct one.

This was the reason why there was one affect validation run in the **save** method however that was breaking the concept of the **AlertMixin** where there is an option to suppress the exception and only store the alert. This PR comes as a prerequisite of the orphaned tracker linking where a dummy affects will need to be automatically created and they will possibly be partially invalid due to a missing information. A non-suppressible exception is a blocker.

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -116,7 +116,9 @@ class NullStrFieldsMixin(models.Model):
     # mixin as we would not allow the null values for the Char/Text fields anymore
     def clean(self):
         super().clean()
+        self.convert_to_python()
 
+    def convert_to_python(self):
         str_based_fields = [
             field
             for field in self._meta.get_fields()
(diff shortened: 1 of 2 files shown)


osoukup:
@jobselko FYI we discussed this quite recently. I have to admit that it was not really easy to change the validation order :sweat_smile: but it was necessary so I can later create dummy affects

MrMarble (approved):

osidb/mixins.py
```diff
@@ -831,6 +866,10 @@ def validate(self, raise_validation_error=True, dry_run=False):
                         **(e.params or {}),
                     )
 
+        # standard validations
+        # exclude meta attributes
+        self.full_clean(exclude=["meta_attr"])
```
jobselko, line 871: It seems to me we are validating `ArrayField` twice (the first `self.full_clean` validates `ArrayField` only).

jobselko (approved):

osidb/mixins.py
osoukup: That is true, I can exclude it here

osoukup: Excluded

JakubFrejlach (approved):

PR #906, opened 2025-02-10
Update resolution of Obsolete trackers to Not affected

mprpic:
With new VEX justifications being enabled for the "Not a bug" Jira resolution, we need to be able to reserve a specific tracker resolution for those trackers that were filed erroneously. Since no VEX justification exists for such state, we need to decide on a specific resolution that will be used for this purpose. Obsolete seems like a good fit.

Yes, the current usage of Obsolete is all over the place, but we can clean that up going forward at least a bit and amend our docs to recommend it as a resolution for invalid trackers. If a tracker was opened in error, it is likely that the underlying component/product is not affected, so we should display it as such on the CVE page as well.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Remove time information when validating embargoed flaws (OSIDB-3862)
+- The "Obsolete" tracker resolution is now treated as "Not affected" to allow its use for erroneously filed trackers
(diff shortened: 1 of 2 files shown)


superbuggy (approved):
LGTM (changelog needs updating though)

costaconrado (approved):
@mprpic Please add a description of the change in the docs/CHANGELOG.md so the validations pass, besides that, LGTM.

osoukup (approved):
LGTM. Thank you for the contribution!

MrMarble (approved):

PR #909, opened 2025-02-14
Add not affected justification field to affect

dmonzonis:
Affects now have a new choice field, 'not affected justification', which is mandatory if the affectedness is set to NOTAFFECTED, and auto-removed if the affectedness is set to any other value.

This is a first step toward adding not affected justifications to trackers as well.

Closes OSIDB-3809.


osidb/models/affect.py
```diff
@@ -511,6 +525,21 @@ def _validate_purl_and_ps_component(self, **kwargs):
...
+        If the affectedness is set at a different value than NOTAFFECTED, then the justification should
+        be empty.
+        """
+        if (
+            self.affectedness == Affect.AffectAffectedness.NOTAFFECTED
+            and not self.not_affected_justification
+        ):
+            raise ValidationError(
+                f"Affect ({self.uuid}) is set as NOTAFFECTED but the 'not affected justification' is empty."
+            )
```
osoukup, line 541: I believe that you have a plan for it but just mentioning that the eventual release of this must be coordinated with (at least) OSIM or otherwise nobody will be able to set affects as NOTAFFECTED.

osidb/models/affect.py
```diff
@@ -33,6 +33,16 @@
...
 
 
+class NotAffectedJustification(models.TextChoices):
+    """Valid choices for not affected justifications in affects and trackers."""
+
+    COMPONENT_NOT_PRESENT = "Component not present"
+    INLINE_MITIGATIONS_ALREADY_EXIST = "Inline mitigations already exist"
+    VULN_CODE_CANNOT_BE_CONTROLLED = "Vulnerable code cannot be controlled by adversary"
+    VULN_CODE_NOT_IN_EXECUTE_PATH = "Vulnerable code not in execute path"
+    VULN_CODE_NOT_PRESENT = "Vulnerable code not present"
```
osoukup, line 43: I am looking at some other enums and it seems common to have NOVALUE explicitly defined among the options. I think that it does not have any functional difference but maybe for the sake of readability it may be beneficial to be able to explicitly assign NOVALUE to the field. WDYT?

osoukup (approved):
LGTM

osidb/models/affect.py
dmonzonis: Yes I'm aware, this should not be released until the other parts have been fully implemented

PR #910, opened 2025-02-14
Revert "Map user name or email to pghistory context..." (6139e55b)

superbuggy:
It looks like this change has to be reverted, the kerberos user data is not being picked up and instead we lose the PK of the user who made the change in favor of `AnonymousUser`. 
image
This is not desirable.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -83,7 +83,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
-    "osidb.middleware.OsidbHistoryMiddleware",  # adds request user to audit history context
+    "pghistory.middleware.HistoryMiddleware",  # adds request user to audit history context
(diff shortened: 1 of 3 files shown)


jsvob:
I'm not in OSIDB anymore and followed the reverted change very superficially, so sorry for possible spam. Middleware django stuff AFAIK has a problem that if you read some things from request, it disappears. Is it by chance a case of this problem here? I think I tried originally something like this and gave up for this reason :D

superbuggy:
> I'm not in OSIDB anymore and followed the reverted change very superficially, so sorry for possible spam. Middleware django stuff AFAIK has a problem that if you read some things from request, it disappears for the usual django processing, breaking normal expectations of the usual django stuff like views, etc.. Is it by chance a case of this problem here? I think I tried originally something like this and gave up for this reason :D

This is absolutely it. Apparently django REST framework also introduces some different behavior for the Views layer as well. I wonder if I can solve the issue of the missing kerberos data by extending  the workaround that the pghistory library uses, which was copied/emulated in this feature.

osoukup (approved):
I assume that you are cleaning the **master** but at the end the changes rather need some polishing and not the complete reversion.

PR #911, opened 2025-02-17
Use date instead of datetime when validating embargoed date

MrMarble:
Use **date** instead of **date time** when validating embargoed flaws to allow users to do last minute changes.

This also aligns with the validation in OSIM

Closes OSIDB-3862


dmonzonis (approved):

osoukup (approved):
LGTM

PR #913, opened 2025-02-18
Handle empty queryset before unpacking

MrMarble:
When deleting the last Affect in a Flaw, we were throwing a `not enough values to unpack` error, this change adds a conditional to check emptiness and return early.

It is not a critical issue, since this scenario is not possible, there's a validation in place to force at least 1 affect in `SECONDARY_ASSESSMENT` state, but because of this error, we are throwing an 500 server error instead of the validation error.

Closes OSIDB-3986

### DIFF
--- a/osidb/models/flaw/label.py
+++ b/osidb/models/flaw/label.py
@@ -96,13 +96,14 @@ class FlawCollaboratorManager(TrackingMixinManager):
     def create_from_flaw(flaw: Flaw):
         """Add new labels to the flaw based on the flaw"""
 
-        [ps_modules, ps_components] = list(
-            zip(
-                *Affect.objects.filter(flaw=flaw).values_list(
-                    "ps_module", "ps_component"
-                )
-            )
+        ps_values = Affect.objects.filter(flaw=flaw).values_list(
+            "ps_module", "ps_component"
         )
+
+        if not ps_values:
+            return []
+
+        [ps_modules, ps_components] = list(zip(*ps_values))
         labels = FlawLabel.objects.get_filtered(ps_modules, ps_components)
 
         for label in labels:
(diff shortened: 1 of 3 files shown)


collectors/bzimport/tests/test_convertors.py
```diff
@@ -286,6 +286,7 @@ def test_basic(self):
         assert package_version.version == "version"
         assert package_version.package == package
 
+    @pytest.mark.enable_signals
```
MrMarble, line 289: I don't know why, but for some reason, this test was preventing the `post_delete` signal to trigger and the flaw_collaborator test was failing, enabling the signals here made it work.

I spent like an hour trying to figure why that happens, but I couldn't

osidb/models/flaw/label.py
```diff
@@ -96,13 +96,17 @@ def create_from_affect(affect: Affect):
...
             zip(
                 *Affect.objects.filter(flaw=flaw).values_list(
                     "ps_module", "ps_component"
                 )
             )
         )
+        if not flaw_affects:
+            return []
+
+        [ps_modules, ps_components] = flaw_affects
```
jobselko, line 109: I think the issue is in the unpacking. What about this change?

```
        ps_values = Affect.objects.filter(flaw=flaw).values_list("ps_module", "ps_component")

        if not ps_values:
            return []

        [ps_modules, ps_components] = list(zip(*ps_values))
```

MrMarble: Great! I like your solution more, will change it, thanks!

jobselko: Glad to help! Could you please add a test once the PR is ready for review? Thanks :)

osoukup (approved):
Good that you spotted it! LGTM

jobselko (approved):

PR #914, opened 2025-02-18
Migrate user ids to username/email on Audit entries

superbuggy:
When a single flaw is requested, this PR adds logic during serialization of Flaw that checks its related Audit entries (FlawAudit, AffectAudit, FlawCommentAudit, etc)  replaces numerical ids with the username or email of the corresponding users. 

Closes OSIDB-3988

- [x] Tested on stage

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Make task_key read-only (OSIDB-4080)
 - Split Requests/Responses in API schema
 
+### Changed
+- Migrate user ids on history Audit models to user emails or usernames (OSIDB-3988)
+
 ## [4.9.1] - 2025-03-10
 ### Fixed
 - Prefix PURL-derived ps_components with path in some cases (OSIDB-4071)
--- a/osidb/migrations/0188_map_user_id_to_user_email_or_name.py
+++ b/osidb/migrations/0188_map_user_id_to_user_email_or_name.py
@@ -1,0 +1,38 @@
+from django.db import migrations
+from django.contrib.auth.models import User
+
+
+def forwards_func(apps, schema_editor):
+
+    context_model = next(
+        (
+            model
+            for model in apps.get_models("osidb")
+            if model._meta.model_name == "context"
+        ),
+        None,
+    )
+
+    if not context_model:
+        return
+
+    context_entries = context_model.objects.all().iterator()
+    for entry in context_entries:
+        if "user" in entry.metadata and isinstance(entry.metadata["user"], int):
+            user_id = entry.metadata["user"]
+            user = User.objects.get(id=user_id)
+            if user and (user.email or user.username):
+                entry.metadata["user"] = user.email or user.username
+                entry.metadata["user_id"] = user_id
+                entry.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("osidb", "0187_add_tracker_special_handling"),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards_func, migrations.RunPython.noop),
+    ]


osidb/middleware.py
```diff
@@ -32,3 +35,44 @@ def __call__(self, request):
...
+    """
+    Updates pghistory context when the request.user attribute is updated in order
+    for context to be set before DB User-model lookup happens. User is added to the request
+    by Django in middleware, but added to the request object in the view layer by
+    django-rest-framework.
+    """
+
+    def __setattr__(self, attr, value):
+        _value = get_userstr(value) if attr == "user" else value
+        return super().__setattr__(attr, _value)
```
osoukup, line 60: Why is this different than
https://github.com/RedHatProductSecurity/osidb/pull/912/files#diff-f8968d26d83a0fb2b740bd909eca8e653f90d02b763a7a7b1b77892e195ca29fR58
and what is the purpose of the two PRs?

superbuggy: This is a mistake--I need to rebase again master, **then** the other PR's branch.

Originally, the purpose was to stagger their releases, though they are separate functionalities with separate tickets, we could combine them into one PR

superbuggy: I think I could actually just ditch the other commits in this PR, I just had the others in there for local development reasons.

osidb/serializer.py
```diff
@@ -635,6 +635,29 @@ def __init__(self, *args, **kwargs):
...
+                    user_id = item.pgh_context["user"]
+                    user = User.objects.get(id=user_id)
+                    item.pgh_context["user"] = (
+                        user.email or user.username or user.id or user
+                    )
+                    context_obj = pghistory.models.Context.objects.get(
+                        id=item.pgh_context_id
+                    )
+                    context_obj.metadata = item.pgh_context
+                    context_obj.save()
```
osoukup, line 660: I am not sure that change of the DB data in the serializer GET is a good approach. We are basically migrating the data on the fly.  Would it be possible to create a one time data migration instead?

osoukup: I am thinking about it this way. A migration is probably one time higher complexity task with a bit of risk of failing the deployment but then it is all over. A functionality like this is going to stay in the code base indefinitely and while it is probably less complicated now the complexity will stay with us potentially breaking any time in the future - something changes on the dependency update, someone makes a mistake in a related code etc.

If there is no strong reason why not to do this one time migration, I am really for it.

JakubFrejlach: I think we should definitely go with the migration approach here. Whatever mass change in data should be handled that way.

superbuggy: Is it worth writing a backwards_func for this? it would store the user IDs mapped to the `pgh_context` entries

osoukup: I am not sure it is needed. I see we mostly do not write them. If it is simple then you can, otherwise there is probably no real reason to revert in the future. A **noop** should be enough.

osidb/migrations/0186_map_user_id_to_user_email_or_name.py
```diff
@@ -0,0 +1,38 @@
...
+def forwards_func(apps, schema_editor):
+
+    context_model = next(
+        (
+            model
+            for model in apps.get_models("osidb")
+            if model._meta.model_name == "context"
+        ),
+        None,
+    )
```
osoukup, line 14: I am maybe missing something but this construction to get the model class reference seems too complicated to me. I am looking at some previous migrations and getting the reference seems more straightforward eg.

https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/migrations/0082_migrate_flawacknowledgment_flawmeta.py#L27

Or is there anything special about the **context** model class that it needs a different approach?

osidb/serializer.py
osoukup: This on-the-fly-migrating code in the serializer should go away now when there is the migration.

osoukup: But we should instead somehow tweak the audit creation so it gets the user names/emails going forward.

osoukup (changes requested):
The migration is much better! I think this gets the right direction now but needs some finalization.

osidb/migrations/0186_map_user_id_to_user_email_or_name.py
superbuggy: When I tried `context_model = apps.get_model("osidb", "Context")` where `Context` is the model name, I was getting errors about the model not being registered on django app `"osidb"`, despite the fact that `.get_models` returns it. The other models returned from `.get_models` have `__name__` fields that work with with `.get_model("osidb", <<model_name>>)` but pghistory's context did not, I would guess because it's associated to the app indirectly via the pghistory tracking model decorators

osidb/serializer.py
superbuggy: > But we should instead somehow tweak the audit creation so it gets the user names/emails going forward.

If understand your comment correctly, the middleware I wrote in #912 handles this

osoukup: You are right! I already forgot about this piece :facepalm:

osidb/serializer.py
```diff
@@ -1103,7 +1108,7 @@ class AffectSerializer(
         max_length=255, allow_blank=True, allow_null=True, required=False
     )
     purl = serializers.CharField(allow_blank=True, allow_null=True, required=False)
-    resolved_dt = serializers.DateTimeField(read_only=True)
+    resolved_dt = serializers.DateTimeField(read_only=True, allow_null=True)
```
osoukup, line 1111: This does not seem to be related, probably some rebase artifact

osoukup (commented):
The core is great. There seems to be some cleanup needed

PR #916, opened 2025-02-24
Add not affected justification to trackers, and delegated value for affects

dmonzonis:
Like affects, trackers now have the not affected justification, which in this case is linked to a new Jira field. This justification is collected from the Jira tracker and synced with the OSIDB counterpart.

The delegated not affected justification is computed from related Jira trackers.

If all the trackers related to an affect are closed as 'Not a Bug' and their justifications
are equal, then pick that value. If they are all closed as 'Not a Bug' but the justifications differ, pick 'Component not Present' even if no justification has that value.
If any tracker has a resolution different to 'Not a Bug', then the delegated justification becomes empty.

This field has also been exposed in the API.

Closes OSIDB-3808.
Closes OSIDB-3810.


osoukup (approved):
LGTM

PR #917, opened 2025-02-25
OSIDB-3939: Propagate jira error correctly

MrMarble:
Accesing `.text` property wasn't enough to get the error body from Jira, I've created a small function that returns the `error` object from the response body https://docs.atlassian.com/software/jira/docs/api/REST/9.12.10/#error-responses

There's also `errorMessages` but from my testing was always empty so I decided to leave it out

Closes OSIDB-3939


osoukup (approved):
LGTM

dmonzonis (approved):

PR #918, opened 2025-02-26
Introduce custom command for manual CVEorg run

jobselko:
This PR introduces the collection of a given CVE in the CVEorg collector and exposes this functionality via the `create_cveorg_flaw` custom command. This command is not restricted by the published date and checked keywords.
The PR also adds several small changes to improve the readability and usability of the collector.

Closes OSIDB-3974


collectors/cveorg/collectors.py
```diff
@@ -150,6 +151,60 @@ def get_repo_changes(self) -> tuple[str, timezone.datetime]:
...
+
+        self.clone_repo()
+        self.update_repo()
+        file_path = self.get_cve_file_path(cve)
+
+        with open(file_path) as file:
+            file_content = json.load(file)
+
+        if file_content["cveMetadata"]["state"] != "PUBLISHED":
+            return f"Cannot create '{cve}' because it was rejected by CVE Program."
```
osoukup, line 186: Are there other values above PUBLISHED and REJECTED (or what stands for rejection) ?

osoukup (approved):
LGTM. Nicely documented and split into commits!

collectors/cveorg/collectors.py
jobselko: Only the `PUBLISHED` and `REJECTED` values are public, so the checked value probably does not matter.

PR #919, opened 2025-02-26
Introduce resolved_dt into affect model

costaconrado:
This PR introduces resolved_dt into affect model.
This PR also tries to estimate the resolved_dt for affects resolved in the past during its migration.
This PR also includes an automation of the resolved_dt timestamp when an affect leaves the `NEW:NOVALUE` state.

Closes OSIDB-4058.

### DIFF
--- a/osidb/models/affect.py
+++ b/osidb/models/affect.py
@@ -5,6 +5,7 @@ import pghistory
 from django.contrib.postgres.indexes import GinIndex
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import models
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from packageurl import PackageURL
 from psqlextra.fields import HStoreField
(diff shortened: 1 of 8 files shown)


osidb/models/affect.py
```diff
@@ -158,6 +158,8 @@ class AffectFix(models.TextChoices):
         choices=NotAffectedJustification.choices, max_length=100, blank=True
     )
 
+    resolution_dt = models.DateTimeField(null=True, blank=True)
```
osoukup, line 162: This is pretty minor so feel free to ignore as that might be not wort addressing. I was just looking at other timestamps and see `created_dt`, `updated_dt`, `updated_until_dt`, `task_updated_dt`, `supported_*` `reported_dt`, `shipped_dt` so was thinking that `resolved_dt` would be more consistent with them (all past tense). However, I see we have inconsistencies already in `unembargo_dt`, `major_incident_start` so it does not matter much. Leaving it up to you.

osidb/models/affect.py
```diff
@@ -725,6 +751,14 @@ def bzsync(self, *args, **kwargs):
         self.affect.save(*args, **kwargs)
 
 
+# List of all states an Affect is considered open/not resolved
+AFFECTEDNESS_UNRESOLVED_RESOLUTIONS = {
+    Affect.AffectAffectedness.NEW: [
+        Affect.AffectResolution.NOVALUE,
+    ],
+    Affect.AffectAffectedness.NOVALUE: [],
```
osoukup, line 759: This affectedness option seems to me to never take an effect. Even though the affectedness would be NOVALUE, the resolutions - be it NOVALUE or any non-empty value - could not be in the empty list.

osoukup (approved):
LGTM. Great you went beyond the ask with the audit parsing!

MrMarble (approved):

osidb/models/affect.py
costaconrado: makes sense, I'll remove this variable and move this check directly to the attribute then

osidb/models/affect.py
costaconrado: Makes sense, I've changed resolution_dt to resolved_dt

JakubFrejlach (approved):
LGTM!

PR #920, opened 2025-02-28
Optimize CVE file path retrieval in CVEorg

jobselko:
This PR is a follow-up to #918 and updates the method for retrieving the file path of a given CVE. Instead of looping through all files in the cvelistV5 repository, the file path is now created directly using the known CVE information. Additionally, the CVE collection process exits early if the CVE is invalid.


osoukup (approved):
LGTM

PR #921, opened 2025-03-03
Adjust task collector window

JakubFrejlach:
This PR:
* adjusts task collector window to shift 1 minute in the past so we encounter data reset less frequently

Closes OSIDB-4068


osoukup (approved):
LGTM

PR #925, opened 2025-03-04
Extend Tracker model with resolved date

dmonzonis:
A new field has been added to the Tracker model, `resolved_dt`, which gets the resolved date from both Jira (Resolved) and Bugzilla (Last Closed). This field has also been exposed to the API as read only.

Closes OSIDB-4064.


osoukup (approved):
LGTM

PR #927, opened 2025-03-04
Introduce tracker special handling field

dmonzonis:
Add the special handling field to the Tracker model, which will be
synced to the corresponding field in Jira issues of 'Vulnerability'
type. In Jira, this field is an array which can contain a set of certain
values, so the same has been done in OSIDB.

This field has also been exposed to the API as a read only field.

Closes OSIDB-4065.


dmonzonis:
This has been done on top of #925 to avoid having to rebase the migrations and changelog later. Please review #925 first or ignore the first commit if reviewing this before the other PR is merged.

Note that like #925, when this is released to production, there will be a massive sync to get historical data into current trackers as well.

osoukup (approved):
LGTM. Small nitpick that the commit description sounds like we sync that field to Jira but instead this is rather about syncing it from Jira :smile:

PR #929, opened 2025-03-06
Mark affect resolved_dt nullable in API

JakubFrejlach:
This PR:
* correctly marks `resolved_dt` as nullable in API schema to correctly reflect reality, this bug caused osidb-bindings generate inaccurately and throw errors parsing affects with no `resolved_dt`.

Note: no immediate release of OSIDB is necessary, once this is in master I will simply create new version of osidb-bindings (4.8.1) which will use the corrected openapi schema


osoukup (approved):
LGTM

PR #930, opened 2025-03-06
Split requests/responses in API schema

JakubFrejlach:
This PR enables `COMPONENT_SPLIT_REQUEST` in drf-spectacular option which changes OpenAPI generator heuristic to simply create separated models/components for responses and requests, this creates a lot more accurate API description as for requests you see only fields that are actually needed and also it fixes a nasty bug in `osidb-bindings` where field was marked as readonly and still was being required in request which prevented tracker filing creation after adding `non_affected_justification` for trackers.

With `COMPONENT_SPLIT_REQUEST=False`:
image

with `COMPONENT_SPLIT_REQUEST=True`:
image

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -295,6 +295,7 @@ SPECTACULAR_SETTINGS = {
         "TrackerType": "osidb.models.Tracker.TrackerType",
     },
     "ENUM_GENERATE_CHOICE_DESCRIPTION": False,
+    "COMPONENT_SPLIT_REQUEST": True,
 }
 
 # Execute once a day by default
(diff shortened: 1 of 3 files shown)


osoukup (approved):
LGTM. This is much better!

JakubFrejlach:
Will consult this with @C-Valen before merge regarding OSIM

C-Valen:
> Will consult this with @C-Valen before merge regarding OSIM

I don't see any conflict. Tried it locally and everything still works perfectly.

dmonzonis (approved):

PR #940, opened 2025-03-20
Collect task changes based on history

JakubFrejlach:
This PR:
* removes the Jira Task Collector window 1 minute shift as it is no longer needed, checking the history is much more precise
* implements history check in Jira Task convertor which check Jira history/changelog (given the fact it is provided) and saves only those changes which are new, this prevents overwriting some changes in OSIDB which are not yet synced to Jira.

Closes OSIDB-4015

### DIFF
--- a/collectors/jiraffe/convertors.py
+++ b/collectors/jiraffe/convertors.py
@@ -20,6 +20,7 @@ from ..utils import (
     tracker_parse_update_stream_component,
     tracker_summary2module_component,
 )
+from .constants import JIRA_DT_FULL_FMT, TASK_CHANGELOG_FIELD_MAPPING
 
 logger = logging.getLogger(__name__)
 
@@ -30,6 +31,11 @@ class JiraTaskConvertor:
         task_data,
     ):
         self._raw = task_data
+        self.histories = getattr(
+            getattr(task_data, "changelog", None),
+            "histories",
+            [],
+        )
         # important that this is last as it might require other fields on self
         self.task_data = self._normalize()
         # set osidb.acl to be able to CRUD database properly and essentially bypass ACLs as
@@ -70,7 +76,7 @@ class JiraTaskConvertor:
             "team_id": self.get_field_attr(self._raw, "customfield_12313240", "id"),
             "group_key": self.get_field_attr(self._raw, "customfield_12311140"),
             "task_updated_dt": datetime.strptime(
-                self.get_field_attr(self._raw, "updated"), "%Y-%m-%dT%H:%M:%S.%f%z"
+                self.get_field_attr(self._raw, "updated"), JIRA_DT_FULL_FMT
             ),
         }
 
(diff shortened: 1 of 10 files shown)


collectors/jiraffe/convertors.py
```diff
@@ -111,6 +117,19 @@ def flaw(self):
...
+        for record in self.histories[::-1]:
+            record_dt = datetime.strptime(record.created, JIRA_DT_FULL_FMT)
+            if not flaw.task_updated_dt or record_dt > flaw.task_updated_dt:
+                for item in record.items:
+                    record_changed_fields = TASK_CHANGELOG_FIELD_MAPPING.get(item.field)
+                    if record_changed_fields:
+                        changed_fields.update(record_changed_fields)
+            else:
+                # no more new changes in history
+                break
```
osoukup, line 131: I am not sure whether we go from the most recent or the oldest change. This else-break seems that we go from the most recent. If so could it happen that there are eg. two status changes not synced to OSIDB yet and we rewrite the newest by an older one as we encounter it later in the for loop?

collectors/jiraffe/convertors.py
```diff
@@ -120,23 +139,30 @@ def flaw(self):
...
-            flaw.team_id = self.task_data["team_id"]
-            flaw.owner = self.task_data["owner"]
             flaw.task_key = self.task_data["external_system_id"]
-            flaw.group_key = self.task_data["group_key"]
-            flaw.workflow_name = self.task_data["workflow_name"]
-            flaw.workflow_state = self.task_data["workflow_state"]
-            flaw.task_updated_dt = self.task_data["task_updated_dt"]
+
+            # NOTE: for some unexplainable reason, history record created timestamp
+            #       can actually later in the future (by milliseconds) than Jira issue updated
```
osoukup, line 150: nitpick: missing `be`

osoukup (approved):
LGTM and good job! I have one question but otherwise everything OK

collectors/jiraffe/convertors.py
JakubFrejlach: Yes I go from the most recent changes, but the whole logic of the looking at the history only servers one purpose, which is look at the fields which actually changed and these fields are then populated with the most recent values. I am not taking the values from the changelog, I take them from the current state of the issue, I just only sync those fields which actually change. So for example:

There are two status changelogs, 2025-01-01 8:00 (refinement) and 2025-01-01 8:02 (closed), and the issues was last synced to OSIDB at 2025-01-01 7:50. In this case we go through changelog, we encounter status change to CLOSED, we mark that we should sync the status, then we encounter status change to REFINEMENT, we mark that we should sync the status again (which is actually not necessary, as it is already marked), and then we just sync the status with whatever value the status field in Jira currently holds, this way we get the most recent data and we are not overwriting anything.

osoukup: Oh I see! Then all OK

PR #944, opened 2025-03-27
Implement sync manager postponing

JakubFrejlach:
This PR:
* adds possibility for each sync manager to implement their own `check_conflicting_sync_managers` so if there is any conflict there can be suitable strategy applied to resolve its (revoke, postpone, etc.)
* implements such conflict check for `JiraTaskDownloadManager` so it is not downloading any changes while sync to Jira or workflow transition is pending towards Jira. This should fix some corner cases of Flaw data reset.

Closes OSIDB-4136

### DIFF
--- a/collectors/jiraffe/tests/test_collectors.py
+++ b/collectors/jiraffe/tests/test_collectors.py
@@ -153,8 +153,6 @@ class TestJiraTaskCollector:
 
         # 3 - get the current Jira task and make sure db is in-sync
         issue = jtq.jira_conn.issue(flaw.task_key)
-        last_update = datetime.strptime(issue.fields.updated, "%Y-%m-%dT%H:%M:%S.%f%z")
-        assert last_update == flaw.task_updated_dt
         assert issue.fields.status.name == "New"
 
         # 4 - freeze the issue in time to simulate long queries being outdated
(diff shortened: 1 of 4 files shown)


JakubFrejlach:
@RedHatProductSecurity/osidb-devs honestly I fail to understand currently, what this test actually tests https://github.com/RedHatProductSecurity/osidb/pull/944/commits/3d2d7b2eefd773dbd77ae8926f7714d6c41d29a5#diff-38b1ba11a7384aa0f874960f00bf2e0d864777385183450928efcff592af6bcbR135, it got broken by removal of the `task_updated_dt` guessing, but after fixing it by removing the timestamp checking, I am not sure whether it is still needed. Wdyt?

osoukup:
> @RedHatProductSecurity/osidb-devs honestly I fail to understand currently, what this test actually tests [3d2d7b2#diff-38b1ba11a7384aa0f874960f00bf2e0d864777385183450928efcff592af6bcbR135](https://github.com/RedHatProductSecurity/osidb/commit/3d2d7b2eefd773dbd77ae8926f7714d6c41d29a5#diff-38b1ba11a7384aa0f874960f00bf2e0d864777385183450928efcff592af6bcbR135), it got broken by removal of the `task_updated_dt` guessing, but after fixing it by removing the timestamp checking, I am not sure whether it is still needed. Wdyt?

@JakubFrejlach I am not fully sure. Maybe it is something from the times of synchronous task management - @costaconrado? But anyway if not before, your last change to check the exact timestamps of the changes in Jira should already go much further than any previous time-wise collision safeguard.

PR #945, opened 2025-03-27
Consider affect in AFFECTED/DELEGATED state as not resolved

dmonzonis:
In OSIDB-4058 the affect resolution date was introduced, and only affects in the NEW:NOVALUE state were considered as not resolved. However, affects in AFFECTED:DELEGATED state should also be considered as not resolved and not have a `resolved_dt`.

This commit marks those affects as not resolved as well, and a data migration removes the `resolved_dt` from the affects that already had one in these circumstances.

Closes OSIDB-4137.


osoukup (approved):
LGTM. @JakubFrejlach FYI this one pretty important regarding the SLA data so it should be incorporated in the release with the additional flaw data reset fix.

PR #946, opened 2025-04-10
Add include_history in openapi description

costaconrado:
This PR adds missing include_history query param to openapi documentation.
This is considered a patch fix, the fields were already present in the API, not requiring a minor release.

Closes OSIDB-3960.


JakubFrejlach (approved):
LGTM!

PR #947, opened 2025-04-22
Flaw data reset corner case - part 1

JakubFrejlach:
This PR:
* minor - creates the linkage between the updated model and the sync manager right upon schedule if possible so the conflicting sync managers check won't have a blind spot during the period between the sync managers creation and the first execution
* removes the redundant workflow update after the successful sync into Jira so there are no data overwrites (e.g. transition manager which syncs PRE_SECONDARY_ASSESSMENT finishes earlier than transition manager which syncs TRIAGE), this is unfortunately still imperfect as the Flaw data reset won't now happen directly in OSIDB but can still happen in Jira (e.g. PRE_SECONDARY_ASSESSMENT is written earlier than TRIAGE and TRIAGE then overwrites it) which will eventually be downloaded back to OSIDB, to fix this, we need to implement solution proposed in OSIDB-4175
* moves ACL adjustment upon workflow advancing as it should happen right away when we do advance the workflow in OSIDB and not to wait until we sync it into Jira, this seems like a remaining part from the synchronous task manager

Closes OSIDB-4136

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,6 @@ jobs:
           python3-pytest-cov
           python3.9
           python36
-          python38
           redhat-rpm-config
           rpm-build
           rpmlint
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -134,7 +134,7 @@
         "filename": ".github/workflows/test.yml",
         "hashed_secret": "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3",
         "is_verified": false,
-        "line_number": 93,
+        "line_number": 92,
         "is_secret": false
       }
     ],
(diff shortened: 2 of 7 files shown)


JakubFrejlach:
Question: does it still makes sense to keep a code parts for the synchronous task management? During the debugging and test writing I've encountered a numerous moments when I thought I am using the async version only to find out I am using actually a sync one. A lot of tests are currently written for the synchronous one but in fact we already use async for all our instances.

osoukup:
> Question: does it still makes sense to keep a code parts for the synchronous task management? During the debugging and test writing I've encountered a numerous moments when I thought I am using the async version only to find out I am using actually a sync one. A lot of tests are currently written for the synchronous one but in fact we already use async for all our instances.

Good question. It could probably be removed. The only possible usage would be some local debugging and we do not leave it there just for it. Though I think it may be a non-trivial change adjusting the tests, removing the switch, and probably more.

PR #950, opened 2025-05-02
[OSIDB-4175] Prevent multiple instances of `JiraTaskTransitionManager`

MrMarble:
### Missing container variables
I added some missing variables and sorted them alphabetically: 
- `JIRA_TASKMAN_URL`
-  `JIRA_TASKMAN_PROJECT_ID`
-  `JIRA_TASKMAN_PROJECT_KEY` 
- `JIRA_STORY_ISSUE_TYPE_ID`

### Singleton SyncManager
 To make the system reusable, I implemented the logic in the SyncManager base model, that way having an "exclusive" task is just:
```python
class SomeTaskManager(SyncManager:
	MODE = SyncManager.SyncManagerMode.EXCLUSIVE
	COUNTDOWN = 60
```
Summary of the functionality:
- Running a task once behaves the same as before.
- Trying to run a task **while** there's already one running will schedule the task to run in `COUNTDOWN` seconds.
- Any subsequent tasks will be ignored (not even propagated to Celery) until the scheduled task completes.

This model not only implements a "Singleton" functionality, but will also ignore rapid fired tasks, creating only one tasks that pulls the current state from the database

Closes OSIDB-4175


osidb/sync_manager.py
```diff
@@ -328,11 +365,58 @@ def check_for_reschedules(cls):
...
+    @classmethod
+    def is_scheduled(cls, sync_id):
+        """
+        Check if there's a task already scheduled to run in the future
+        """
+        manager = cls.objects.get(sync_id=sync_id)
+
+        return manager and (
+            manager.planned_start_dt is not None
+            and manager.planned_start_dt > timezone.now()
```
JakubFrejlach, line 411: I am not sure if I understand this logic. We set the `planned_start_dt`  ourselves, so what about the scenario when:
1. task is scheduled first time and `planned_start_dt` is set to e.g. `8:00` but it does not start because the queue is filled up with other tasks
2. meanwhile same task is scheduled in `8:01` for the second time and the first one didn't even start, so based on this logic, it will check whether the `planned_start_dt` (`8:00`) is greater than `now` (`8:01`), which should not be true and thus the `is_scheduled` will be evaluated to `False`

osidb/sync_manager.py
```diff
@@ -85,8 +99,31 @@ def schedule(cls, sync_id, *args, schedule_options=None, **kwargs):
...
+                return
+
+            # If the task is already in progress, schedule it with a countdown
+            if cls.is_in_progress(sync_id):
+                logger.info(
+                    f"{cls.__name__} {sync_id}: Task already in progress"
+                    f", postponing for {cls.COUNTDOWN} seconds"
+                )
+                if "countdown" not in schedule_options:
+                    schedule_options["countdown"] = cls.COUNTDOWN
```
JakubFrejlach, line 121: This could produce log message that the task will be postponed with `cls.COUNTDOWN` value, but different value would be applied instead if `schedule_options` would already contain `countdown`

osidb/sync_manager.py
osoukup: @JakubFrejlach has good points.

I think that we should rather consider a task to be scheduled when it has `last_scheduled_dt` and it has either no `last_started_dt` or is is before `last_scheduled_dt`. I am not sure that `planned_start_dt` matters here as the start is not guaranteed to happen in that moment.

I am not sure this condition is sufficient. It would definitely not account for the cases of the tasks which were scheduled but failed to start for some reason - eg. Celery restart (hopefully rare).

osidb/sync_manager.py
```diff
@@ -328,11 +365,58 @@ def check_for_reschedules(cls):
...
+            and
+            # must NOT have failed since that start
+            not (
+                manager.last_failed_dt is not None
+                and manager.last_failed_dt > manager.last_started_dt
+            )
+            and
+            # and either not finished (-> running) or explicitly in-progress
+            manager.last_finished_dt is None
+        )
```
osoukup, line 400: This is an estimation in the sense that we assume that when started a task must either finish or fail with respect to the sync manager framework but if Celery crashes etc. there will be no finishing or failing. This is probably as much as we can get without going on the Celery level though.

osoukup (changes requested):
I think this needs some polishing but it seems to go the right direction. The approach of never scheduling a task twice within a period sounds promising. I was originally thinking about never executing the same task twice at once approach. That would however need Celery level tweaks. I am looking at SDEngine and it uses `celery-singleton` library so in case this was at the end problematic, we can have a look there.

MrMarble:
@osoukup @JakubFrejlach As you both rightfully pointed out, there are some edge cases that we can't control using SyncManager's logic. We already gave it a try with the BZSync manager, and I tried to add a couple more checks on top, but it's still not enough.

My original idea was to follow Celery's documentation and utilize an external system to implement the singleton logic. In this case, that would be Redis, as we are already using it. When a task is started, a flag is stored in Redis (with an expiration time to prevent blocking forever) and is removed when the task is finished or has failed. Here's a link to an example: https://docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time

This is basically what `celery-singleton` does under the hood, I'll make the change if we agree is a move in the right direction

JakubFrejlach:
> @osoukup @JakubFrejlach As you both rightfully pointed out, there are some edge cases that we can't control using SyncManager's logic. We already gave it a try with the BZSync manager, and I tried to add a couple more checks on top, but it's still not enough.
> 
> My original idea was to follow Celery's documentation and utilize an external system to implement the singleton logic. In this case, that would be Redis, as we are already using it. When a task is started, a flag is stored in Redis (with an expiration time to prevent blocking forever) and is removed when the task is finished or has failed. Here's a link to an example: https://docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time
> 
> This is basically what `celery-singleton` does under the hood, I'll make the change if we agree is a move in the right direction

I would be for going into that direction to ideally completely eliminate the resets and not leave any edge cases behind. Do you plan to implement the singleton on our side or take advantage of the `celery-singleton`?

MrMarble:
> > @osoukup @JakubFrejlach As you both rightfully pointed out, there are some edge cases that we can't control using SyncManager's logic. We already gave it a try with the BZSync manager, and I tried to add a couple more checks on top, but it's still not enough.
> > My original idea was to follow Celery's documentation and utilize an external system to implement the singleton logic. In this case, that would be Redis, as we are already using it. When a task is started, a flag is stored in Redis (with an expiration time to prevent blocking forever) and is removed when the task is finished or has failed. Here's a link to an example: [docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time](https://docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time)
> > This is basically what `celery-singleton` does under the hood, I'll make the change if we agree is a move in the right direction
> 
> I would be for going into that direction to ideally completely eliminate the resets and not leave any edge cases behind. Do you plan to implement the singleton on our side or take advantage of the `celery-singleton`?

I'll implement the singleton on our side. `celery-singleton` seems abandoned, it's been 3 years since the last commit, has a bunch of issues, and also targets Python 3.6 and Celery 4 (we use Celery 5). I also looked at the code, and it is only syntax sugar for the locks, it does not seem to be doing anything on top, basically mimicking the Celery example and providing nice methods to the user. 
My idea is to implement it similarly as what I have already done, with an Enum in the class or maybe a decorator on the `sync_task`

JakubFrejlach:
> > > @osoukup @JakubFrejlach As you both rightfully pointed out, there are some edge cases that we can't control using SyncManager's logic. We already gave it a try with the BZSync manager, and I tried to add a couple more checks on top, but it's still not enough.
> > > My original idea was to follow Celery's documentation and utilize an external system to implement the singleton logic. In this case, that would be Redis, as we are already using it. When a task is started, a flag is stored in Redis (with an expiration time to prevent blocking forever) and is removed when the task is finished or has failed. Here's a link to an example: [docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time](https://docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time)
> > > This is basically what `celery-singleton` does under the hood, I'll make the change if we agree is a move in the right direction
> > 
> > 
> > I would be for going into that direction to ideally completely eliminate the resets and not leave any edge cases behind. Do you plan to implement the singleton on our side or take advantage of the `celery-singleton`?
> 
> I'll implement the singleton on our side. `celery-singleton` seems abandoned, it's been 3 years since the last commit, has a bunch of issues, and also targets Python 3.6 and Celery 4 (we use Celery 5). I also looked at the code, and it is only syntax sugar for the locks, it does not seem to be doing anything on top, basically mimicking the Celery example and providing nice methods to the user. My idea is to implement it similarly as what I have already done, with an Enum in the class or maybe a decorator on the `sync_task`

Sounds great. Maybe we could also made it into a separate python package, so it is potentially reusable among our other projects, but that seems a bit out of scope atm, so just an idea. @Elkasitu did something similar with https://github.com/RedHatProductSecurity/rhubarb which ensures only one instance of the given celery task is always running (collector tasks).

MrMarble:
> > > > > @osoukup @JakubFrejlach As you both rightfully pointed out, there are some edge cases that we can't control using SyncManager's logic. We already gave it a try with the BZSync manager, and I tried to add a couple more checks on top, but it's still not enough.
> > > > > My original idea was to follow Celery's documentation and utilize an external system to implement the singleton logic. In this case, that would be Redis, as we are already using it. When a task is started, a flag is stored in Redis (with an expiration time to prevent blocking forever) and is removed when the task is finished or has failed. Here's a link to an example: [docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time](https://docs.celeryq.dev/en/latest/tutorials/task-cookbook.html#ensuring-a-task-is-only-executed-one-at-a-time)
> > > > > This is basically what `celery-singleton` does under the hood, I'll make the change if we agree is a move in the right direction
> > > > 
> > > > 
> > > > I would be for going into that direction to ideally completely eliminate the resets and not leave any edge cases behind. Do you plan to implement the singleton on our side or take advantage of the `celery-singleton`?
> > > 
> > > 
> > > I'll implement the singleton on our side. `celery-singleton` seems abandoned, it's been 3 years since the last commit, has a bunch of issues, and also targets Python 3.6 and Celery 4 (we use Celery 5). I also looked at the code, and it is only syntax sugar for the locks, it does not seem to be doing anything on top, basically mimicking the Celery example and providing nice methods to the user. My idea is to implement it similarly as what I have already done, with an Enum in the class or maybe a decorator on the `sync_task`
> > 
> > 
> > Sounds great. Maybe we could also made it into a separate python package, so it is potentially reusable among our other projects, but that seems a bit out of scope atm, so just an idea. @Elkasitu did something similar with [RedHatProductSecurity/rhubarb](https://github.com/RedHatProductSecurity/rhubarb?rgh-link-date=2025-05-06T10%3A40%3A38Z) which ensures only one instance of the given celery task is always running (collector tasks).
> 
> I am thinking whether this could actually lead to **rhubarb** extension. It ensures task run exclusion. Here we need a parametrized task run exclusion. I have no checked the lib in depth but maybe it could be used as a base.

My proposed change to rhubarb just got [merged](https://github.com/RedHatProductSecurity/rhubarb/pull/4), I'll proceed and update this pull request by updating to the latest version of rhubarb (once @Elkasitu publish it) and use it for the singleton behavior. @JakubFrejlach

MrMarble:
Here's a summary of the changes in this pull request:

#### Rhubarb

* A new class was added that incorporates a hash of the task arguments into the lock name. This allows the same task to be scheduled multiple times concurrently, provided it's invoked with different arguments (e.g., different `flaw_id`s).
* Introduced an option to specify the lock Time-To-Live (TTL) on a per-task basis using the `lock_ttl` argument. This value overrides the global configuration.

*Note:* These changes alone don't fully resolve the underlying issue. Rhubarb handles the "singleton" locking behavior, but we still need to schedule a separate task to run later for syncing with the latest data.

#### Sync Manager

Both the `is_in_progress` and `is_scheduled` methods have been retained. However, their internal logic has been updated based on previous feedback to address corner cases. The logic has also been made more explicit to improve readability and understanding.

Simplified flow example (where A, B, C... are task instances):

```
A) Starts
   B) Starts
   B) Is rejected because A is already running and C is scheduled
   D) Starts
   D) Is rejected because A is already running
A) Finishes
E) Starts
E) Is rejected because C is already scheduled
C) Starts
...
```

Edge cases that are difficult to control directly, such as a Celery worker crashing mid-task, have also been considered. The `MAX_RUN_LENGTH` constant helps ensure that tasks will eventually run even if a previous instance failed unexpectedly (e.g., by allowing its lock to expire).

osidb/sync_manager.py
```diff
@@ -85,9 +99,25 @@ def schedule(cls, sync_id, *args, schedule_options=None, **kwargs):
...
 
+        if not created and cls.MODE == cls.SyncManagerMode.EXCLUSIVE:
+            # If the task is already scheduled, skip
+            if not cls.is_scheduled(sync_id):
+                logger.info(
+                    f"{cls.__name__} {sync_id}: Task already in progress"
+                    f", postponing for {cls.COUNTDOWN} seconds"
+                )
+                if "countdown" not in schedule_options:
+                    schedule_options["countdown"] = cls.COUNTDOWN
```
JakubFrejlach, line 113: I think this can still produce incorrect message, as in the log message the `cls.COUNTDOWN` is used, but `countdown` can already contained in the `schedule_options` so there should be something like
```python
logger.info(
    f"{cls.__name__} {sync_id}: Task already in progress"
    f", postponing for {schedule_options.get("countdown") or cls.COUNTDOWN} seconds"
)

File: osidb/sync_manager.py [hunk 5]
```diff
@@ -85,9 +99,25 @@ def schedule(cls, sync_id, *args, schedule_options=None, **kwargs):
         if schedule_options is None:
             schedule_options = {}
 
-        cls.objects.get_or_create(sync_id=sync_id)
+        created = cls.objects.get_or_create(sync_id=sync_id)[1]
         cls.objects.filter(sync_id=sync_id).update(last_scheduled_dt=timezone.now())
 
+        if not created and cls.MODE == cls.SyncManagerMode.EXCLUSIVE:
+            # If the task is already scheduled, skip
+            if not cls.is_scheduled(sync_id):
```
JakubFrejlach, line 107: I am not sure I understand this part, the comment says that we skip when the task is already scheduled but the condition is reversed I think, we perform the skip if it is not scheduled? :thinking: So currently we would perform reschedule only if the task is not already scheduled, which I think handles the latter part of the `schedule` method.

JakubFrejlach (changes requested):
One small thing which I think is still present from the last review and one part which I don't fully understands and feels that it works the opposite way than intended so requesting changes just in case.

osidb/sync_manager.py
MrMarble: Oh you're right, 🤔 I remember changing it after your comment, maybe the it got lost after rebasing when I updated rhubarb

osidb/sync_manager.py
MrMarble: Sorry the comment is a bit misleading, is referring that we skip the if itself 😆

The idea is that this `is_scheduled` logic is only used for knowing when we need to schedule a task to run in the future.

JakubFrejlach: Clarified in a meeting, looking good. :+1:

JakubFrejlach (approved):
LGTM! Great job!

PR #952, opened 2025-05-06
add sync_errata command for easier manual linking

costaconrado:
This PR introduces a command called sync_errata that is able to collect and link a single errata for fixing minor data inconsistencies.


osoukup (approved):
LGTM

osidb/management/commands/sync_errata.py
```diff
@@ -0,0 +1,21 @@
...
+        parser.add_argument(
+            "errata_id", nargs="*", type=str, help="errata id list (e.g. 1234 5678)"
+        )
+
+    def handle(self, *args, **options):
+        if not args:
+            raise CommandError("You must provide a list of IDs.")
+        set_user_acls(settings.ALL_GROUPS)
+        erratum_json_list = [get_erratum(errata_id) for errata_id in args]
+        link_bugs_to_errata(erratum_json_list)
```
JakubFrejlach, line 21: This might not be ideal for longer lists as any random failure during the process of getting the errata from ET would just fail the whole syncing process.

PR #955, opened 2025-05-12
Introduce CISA CVSSIssuer type and prevent modification of non-RH CVSS scores

Elkasitu:
This PR introduces the `CVSSIssuer.CISA` type in preparation for collecting CISA-issued CVSS for all Flaws, as well as prevent API users from modifying CVSS scores not issued by Red Hat.

A future PR will provide a v2 API which is more robust and less error prone which will bring with it the deprecation of the v1 `cvss_scores` API.

See individual commits' descriptions for more info.

Closes OSIDB-4203


Elkasitu:
The OpenAPI backwards compatibility check fails due to the fact that the `issuer` field has changed type, but in practice this is a non-issue, the field is simply not required anymore (if supplied -> ignore, if not supplied -> defaults to `"RH"`).

I presume that this is due to how `drf-spectacular` has generated the schema for the field using `allOf`.

osidb/api_views.py
```diff
@@ -742,6 +743,28 @@ class FlawCVSSView(
...
+        cvss: FlawCVSS = self.get_object()
+        if cvss.issuer == FlawCVSS.CVSSIssuer.REDHAT:
+            return super().update(request, *args, **kwargs)
+        return Response(FlawCVSSSerializer(cvss).data)
+
+    @extend_schema(
+        responses={
+            200: {},
+        },
+        parameters=[bz_api_key_param],
```
JakubFrejlach, line 761: I am thinking whether the `bz_api_key_param` is even needed anymore, all communication towards Bugzilla is performed via OSIDB Bugzilla service account with the exception of adding comments and filing trackers. We have this parameter set for a lot of endpoints and we should clean them up properly eventually.

JakubFrejlach (approved):
One minor thing, otherwise LGTM!

JakubFrejlach:
> The OpenAPI backwards compatibility check fails due to the fact that the `issuer` field has changed type, but in practice this is a non-issue, the field is simply not required anymore (if supplied -> ignore, if not supplied -> defaults to `"RH"`).
> 
> I presume that this is due to how `drf-spectacular` has generated the schema for the field using `allOf`.

openapi backwards compatibility check is non-blocking anyway, in the past we have bypassed that upon numerous occasions, for example when fixing something in openapi to match the reality.

osoukup (approved):
LGTM

osidb/api_views.py
Elkasitu: I added it just because the view class has a decorator that adds it to all methods, but when overriding the method it doesn't which led to backwards-incompatible changes in the schema.

Didn't know it was removed though, good to know!

PR #956, opened 2025-05-13
OSIDB-4204: Record CISA CVSS scores from cve.org as CISA's

Elkasitu:
Closes OSIDB-4204
Closes OSIDB-4222


osidb/helpers.py
```diff
@@ -285,3 +287,27 @@ def makeSocket(self, *args, **kwargs):
...
+        * Migrations
+        * Collectors & Scheduled tasks
+        * Non user-driven processes
+    """
+
+    def wrapped(*args, **kwargs):
+        set_user_acls(settings.ALL_GROUPS)
+        f(*args, **kwargs)
+
+    return wrapped
```
osoukup, line 313: Based on the description I would expect the ACLs to be unset to the previous/no ACLs after the call but this seems to simply retain the full access.

osoukup (approved):
LGTM

osidb/helpers.py
Elkasitu: In theory, yes, but it doesn't happen in practice and this is the pattern we use all over OSIDB.

That said, I'll change it as a sanity check

osoukup: That is why I am not insisting on the extension, I am just pointing out that the wrapper description says something little bit different than it is actually done then

JakubFrejlach (approved):
Good job.

collectors/cveorg/collectors.py
```diff
@@ -24,6 +26,7 @@
 from collectors.utils import convert_cvss_score_to_impact, handle_urls
 from osidb.core import set_user_acls
 from osidb.models import FlawCVSS, FlawReference, Snippet
+from osidb.models.flaw.flaw import Flaw
```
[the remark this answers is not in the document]
Elkasitu: I think it was just the autocomplete being autocompletegarbage :D

collectors/cveorg/tests/test_processing.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_updates(self, cisa_cvss_content):
+        collector = CVEorgCollector()
+        collector.keywords_check_enabled = False
+        content = collector.extract_content(cisa_cvss_content)
+        assert Flaw.objects.count() == 0
+
+        # create the flaw per the content
+        collector.save_snippet_and_flaw(content)
+        flaw: Optional[Flaw] = Flaw.objects.first()
```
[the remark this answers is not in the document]
Elkasitu: Not necessarily, I've been using typed Python quite a bit for the past year or so, it's more a habit than anything

PR #957, opened 2025-05-14
update Celery and related dependencies

osoukup:
mainly because of the stability improvements in 5.5.0 which should prevent Celery integration breakage

https://github.com/celery/celery/releases/tag/v5.5.0

Closes OSIDB-4125

### DIFF
--- a/collectors/framework/models.py
+++ b/collectors/framework/models.py
@@ -181,7 +181,7 @@ class CollectorMetadata(NullStrFieldsMixin):
     @property
     def crontab_params(self) -> Optional[Dict[str, str]]:
         if self.crontab:
-            params = re.search(r"<crontab: (.*) \(m/h/d/dM/MY\)>", self.crontab).group(
+            params = re.search(r"<crontab: (.*) \(m/h/dM/MY/d\)>", self.crontab).group(
                 1
             )
             return {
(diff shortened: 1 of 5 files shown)


collectors/framework/models.py
```diff
@@ -181,7 +181,7 @@ class CollectorState(models.TextChoices):
     @property
     def crontab_params(self) -> Optional[Dict[str, str]]:
         if self.crontab:
-            params = re.search(r"<crontab: (.*) \(m/h/d/dM/MY\)>", self.crontab).group(
+            params = re.search(r"<crontab: (.*) \(m/h/dM/MY/d\)>", self.crontab).group(
```
Elkasitu, line 184: What about existing CollectorMetadata objects? I think most of the crontabs we use only use minute/hour but if not, the string representation in the database should be migrated

Elkasitu (approved):

collectors/framework/models.py
osoukup: Good observation, I need to add the migration

osoukup: I theoretically does not matter as we only use hours and minutes which do not change but rather to be safe

osoukup:
@Elkasitu migration added, feel free to re-review

JakubFrejlach (approved):

PR #961, opened 2025-05-26
Avoid upadting cve.org repo unless strictly necessary

Elkasitu:
Since calling said method while a celery worker is running can lead to issues, do not call it during the data migration and instead assume that it's up-to-date with a margin of error of an hour.


osoukup (approved):
LGTM

PR #965, opened 2025-05-27
Properly parse modular components in Jira queries

dmonzonis:
Modular components should follow the module:stream/package format, and in particular we want to extract the component name from the package part when generating the component in a Jira issue.

This commit also makes it not be specific to the RHEL Jira project, components with the modular format will be treated as modular components in any project.

Closes OSIDB-4224.


Elkasitu (approved):
LGTM but maybe wait on a review by one of the more business-seasoned devs in the team before merging 😃

PR #967, opened 2025-05-28
Replace OpenAPI usage of enum in favor of x-extensible-enum

Elkasitu:
This commit introduces a postprocessing hook for drf-spectacular which will convert all enums used in the OSIDB API into x-extensible-enums.

This is a proprietary extension pushed by the Zalando API Guidelines which essentially tells clients that a given enumeration may grow in the future and that said growth does not constitute a breaking change, so clients should be ready to accept any value.

For clients not recognizing the x-extensible-enum extension, the type should become looser (string), so this shouldn't be considered a breaking change either.


Elkasitu:
Closed due to lack of time and low prioritization

PR #970, opened 2025-06-05
OSIDB-4256: Integrate with HCV

Elkasitu:
This is in preparation for working as a proxy to store third-party integration tokens from OSIM.

Closes OSIDB-4256


osidb/integrations.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+        except hvac.exceptions.VaultError as e:
+            _logger.error("Vault AppRole authentication failed.", exc_info=e)
+
+    def upsert_secret(self, subpath: Path, key: str, value: str) -> None:
+        self.client.secrets.kv.v2.create_or_update_secret(
+            path=str(self.BASE_PATH / subpath),
+            secret={key: value},
+            mount_point=self.BASE_MOUNTPOINT,
+        )
```
JakubFrejlach, line 39: Suppose we have a path to secret `apps/osidb-integrations/jira` and this will just create or update the secret with the key/value pair, does it keep all others pairs as well or just creates a new version of this secret with just this submitted pair? I usually managed HCV rather from the UI in terms of updating the secrets and during the update you are always creating a new version which would incline more to the case that it does not keep the previous secrets. :thinking:

By example:
1. `apps/osidb-integrations/jira`  insert `user1: user1_token`
2. `apps/osidb-integrations/jira`  insert `user2: user2_token` -> do we now have both user1 and user2 records or did we just created a new secret version with only user2 record? 

Looking at the HVAC docs maybe the patch is what we need?
* https://python-hvac.org/en/stable/source/hvac_api_secrets_engines.html#hvac.api.secrets_engines.KvV2.patch
* https://python-hvac.org/en/stable/source/hvac_api_secrets_engines.html#hvac.api.secrets_engines.KvV2.create_or_update_secret

JakubFrejlach (changes requested):
Looking good, but I am think the current implementation might cause data being overwritten.

osidb/integrations.py
Elkasitu: Very good catch, indeed it's essentially the difference between `PUT` and `PATCH`!

osoukup (approved):
LGTM

PR #971, opened 2025-06-05
OSIDB-4253: Fix cve.org collector's get_comment_zero and get_references

Elkasitu:
The above closures fail when parsing rejected CVE records from CVE.org, leading to unexpected errors.

Closes OSIDB-4253


osoukup (approved):
LGTM

PR #973, opened 2025-06-06
OSIDB-4244: Provide endpoint for proxying storage of third-party tokens

Elkasitu:
This endpoint accepts (for the time being) JIRA and Bugzilla API tokens for the currently logged-in user and stores them in HashiCorp Vault.

The tokens will eventually be available for consumption from the /whoami endpoint.

Closes OSIDB-4244

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -250,7 +250,7 @@
         "filename": "docker-compose.yml",
         "hashed_secret": "7c6a61c68ef8b9b6b061b28c348bc1ed7921cb53",
         "is_verified": false,
-        "line_number": 126,
+        "line_number": 129,
         "is_secret": false
       }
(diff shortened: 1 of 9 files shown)


Elkasitu:
> LGTM! I am thinking about how will we translate this into `osidb-bindings` library, people just export the env variable with the token/api key there and the library uses it, does it make sense to provide them with access to this endpoint? It will require some changes as this seems like a "special" endpoint and so far the `PATCH` methods were forbidden so bindings are not adjusted to use such method (easy to do so however).

They can use it if they want though I see little benefit: instead of sticking their tokens in an env variable, they'd have to first set it through OSIDB, query `/whoami`, getting the relevant token and setting it for the subsequent request.

PR #977, opened 2025-06-12
OSIDB-4246: Don't require third-party integration tokens in headers

Elkasitu:
While the tokens can still be passed to OSIDB via header (and have precedence), the tokens can now be stored/retrieved using /osidb/integrations.

Closes OSIDB-4246


osoukup (approved):
LGTM

PR #978, opened 2025-06-12
Only match on exact CVE when linking flaw from snippet

superbuggy:
This PR introduces a minor change that only makes possible exact matches of CVE IDs when linking a flaw from a snippet. I wanted to get feedback on the approach while I create unit tests

Closes OSIDB-4168


Elkasitu (commented):
Could we get a regression test?

osoukup (approved):
LGTM. I agree that a test should be added

MrMarble (approved):

MrMarble (approved):

PR #979, opened 2025-06-13
Agnostic podman compose use

Elkasitu:
These are a set of changes to the local environment that make the compose usage vendor-agnostic, meaning both `docker-compose` and `podman-compose` can be used, it should be a drop-in replacement as the `podman compose` subcommand will pick whichever compose vendor is available automagically, if both are available `docker-compose` takes precedence.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -250,7 +250,7 @@
         "filename": "docker-compose.yml",
         "hashed_secret": "7c6a61c68ef8b9b6b061b28c348bc1ed7921cb53",
         "is_verified": false,
-        "line_number": 129,
+        "line_number": 130,
         "is_secret": false
       }
     ],
@@ -475,5 +475,5 @@
       }
     ]
   },
-  "generated_at": "2025-06-12T14:21:50Z"
+  "generated_at": "2025-06-17T07:59:59Z"
 }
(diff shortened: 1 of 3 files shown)


Elkasitu:
Please kindly check whether this works for you locally if you have some time

JakubFrejlach (approved):
Works great with `podman`. Cannot try it with `docker` as I have some problems to run it generally.

osoukup (changes requested):
This does not seem to mostly work for me.

I have both `docker` and `podman` available and my understanding here is that `docker` should be used by default but `which podman` which seems to be used points as `/usr/bin/podman`. This is not the issue however.

When I run `make build` it seems to build everything OK but at the end it fails with
```
Trying to pull localhost/testrunner:latest...                                                                                                                                                              
WARN[0000] Failed, retrying in 1s ... (1/3). Error: initializing source docker://localhost/testrunner:latest: pinging container registry localhost: Get "https://localhost/v2/": dial tcp [::1]:443: connec
t: connection refused
...
```
when I guess it for some reason tries to pull the image from a local registry. However, when I list the images, there are newly built ones as expected. `make start-local` then works OK. I am not sure whether this is in any way serious as it does the work but I would not expect the build to end with an error. I can eventually provide more debugging info, I have not investigated deeper.

osoukup (approved):

PR #980, opened 2025-06-13
Provide workaround for parsing modular components from PURLs

superbuggy:
For compatibility reasons, a special case is handled of deriving a ps_component from a PURL where the format of the ps_component name will be the rpmmod qualifier as namespace for the PURL name.

Closes OSIDB-4292

### DIFF
--- a/osidb/models/tests/test_affect.py
+++ b/osidb/models/tests/test_affect.py
@@ -80,3 +80,33 @@ class TestAffect:
         affect.save()
         assert affect.is_resolved == is_resolved_2
         assert affect.resolved_dt or not is_resolved_2
+
+    @pytest.mark.parametrize(
+        "purl,ps_component",
+        [
+            ("pkg:rpm/redhat/php@7.2?arch=src&rpmmod=php:7.2", "php:7.2/php"),
+            ("pkg:rpm/redhat/squid@4?arch=src&rpmmod=squid:4", "squid:4/squid"),
+            (
+                "pkg:rpm/redhat/gstreamer1-plugins-good@flatpak?arch=src&rpmmod=libreoffice:flatpak",
+                "libreoffice:flatpak/gstreamer1-plugins-good",
+            ),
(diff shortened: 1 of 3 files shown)


osidb/models/tests/test_affect.py
```diff
@@ -80,3 +80,18 @@ def test_resolved_dt_auto_update(
...
+
+    @pytest.mark.parametrize(
+        "purl,ps_component",
+        [
+            (
+                "pkg:rpm/redhat/nginx@1.14.1-9.module+el8.0.0+4108+cba21616?arch=x86_64&rpmmod=mainstream",
+                "mainstream/nginx",
+            )
+        ],
+    )
```
osoukup, line 92: nitpick: parametrizing seems a bit odd when there is only one case to test

osoukup (approved):
LGTM

osidb/models/tests/test_affect.py
JakubFrejlach: Maybe we could test also the other case for the `oci` type and maybe one for normal case when no special handling is applied, then it would make sense to parametrize this test

osidb/models/tests/test_affect.py
```diff
@@ -80,3 +80,18 @@ def test_resolved_dt_auto_update(
...
+    def test_purls(self, purl, ps_component):
```
JakubFrejlach, line 93: I think `test_purls` is a bit vague as the test itself tests the correct inferring of the `ps_component` from `purl`. Maybe `test_ps_component_from_purl` or something like that?

osidb/models/tests/test_affect.py
```diff
@@ -80,3 +80,18 @@ def test_resolved_dt_auto_update(
...
+        affect = AffectFactory(purl=purl, ps_component=None)
+
+        # affect.save()
```
JakubFrejlach, line 96: Leftover comment?

JakubFrejlach (changes requested):
Couple of small things to address.

PR #981, opened 2025-06-16
OSIDB-4284: Update local_updated_dt when flaw-related models are updated

Elkasitu:
Before this commit, some flaw-related models would not trigger an update of their related flaw's local_updated_dt field, this field's freshness is paramount to the correct computation of the changed_{before,after} filters in the API.

Due to this bug, services that poll for updated flaws would miss updates when said updates affected related entities to a flaw but not the flaw itself.

After this commit, said related entities correctly update the local_updated_dt field in the related flaw, thus making the aforementioned API filters correct and reducing missed updates in downstream services.

Closes OSIDB-4284


JakubFrejlach (changes requested):
Missing one related model `FlawCVSS`, otherwise LGTM.

PR #986, opened 2025-06-24
[OSIDB-3846] Flaw data reset in `JiraTaskDownloadManager`

MrMarble:
Latest detected flow with this issue:

1. User rapidly transitions Flaw multiple times, `A => B, B => C`
2. First `JiraTaskTransitionManager` starts
3. Second `JiraTaskTransitionManager` is delayed 60 seconds since there's one already running
4. First `JiraTaskTransitionManager` finishes, writing state `B` in Jira 
5. `JiraTaskDownloadManager` starts and syncs state `B` back to OSIDB, overriding state `C`
6. Second `JiraTaskTransitionManager` pulling `B` from database, writing it again to Jira.
7. State `C` is lost

This pull request introduces a small change in how `JiraTaskDownloadManager` checks conflicting managers to take scheduled tasks into account, the expected flow with this change is:

1. User rapidly transitions Flaw multiple times, `A => B, B => C`
2. First `JiraTaskTransitionManager` starts
3. Second `JiraTaskTransitionManager` is delayed 60 seconds since there's one already running
4. First `JiraTaskTransitionManager` finishes, writing state `B` in Jira 
5. `JiraTaskDownloadManager` is delayed 60 seconds since there are conflicting managers
6. Second `JiraTaskTransitionManager` pulling `C` from database, writing it to jira.
7. `JiraTaskDownloadManager` runs and syncs `C` back to OSIDB
8. Nothing is lost 🤞🏼 

I also added some tests to reproduce this same case.

Closes OSIDB-3846

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Added
-- Trigger Jira tracker syunc when Red Hat's CVSS change (OSIDB-4186)
+- Trigger Jira tracker sync when Red Hat's CVSS change (OSIDB-4186)
 
 ### Fixed
 - Properly handle scheduled tasks collision in Jira Downloader (OSIDB-3846)


osoukup (approved):
LGTM and :crossed_fingers:

osidb/sync_manager.py
```diff
@@ -812,14 +812,9 @@ def check_conflicting_sync_managers(
...
-                        conflicting_sync_manager.last_scheduled_dt is not None
-                        and (
-                            conflicting_sync_manager.last_finished_dt is None
-                            or conflicting_sync_manager.last_finished_dt
-                            < conflicting_sync_manager.last_scheduled_dt
-                        )
-                    )
+                    conflicting_pending_sync = conflicting_sync_manager.is_in_progress(
+                        existing_flaw.uuid
+                    ) or conflicting_sync_manager.is_scheduled(existing_flaw.uuid)
```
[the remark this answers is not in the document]
JakubFrejlach: image

osidb/tests/test_sync_manager.py
```diff
@@ -0,0 +1,173 @@
...
+class TestSyncManager(TestCase):
+    """
+    Test cases for SyncManager and its subclasses.
+    Created to hunt down issues with Flaw data reset.
+    """
+
+    model = SyncManager
+
+    @classmethod
+    def setup_class(self):
```
JakubFrejlach, line 31: Convention for `classmethod` is to use `cls` rather than `self`

osidb/tests/test_sync_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        self.model = ModelBase(
+            f"__TestModel__{self.model.__name__}",
+            (self.model,),
+            {"__module__": self.model.__module__},
+        )
+        with connection.schema_editor() as schema_editor:
+            schema_editor.create_model(self.model)
+
+    @classmethod
+    def teardown_class(self):
```
JakubFrejlach, line 44: Convention for `classmethod` is to use `cls` rather than `self`

osidb/tests/test_sync_manager.py
```diff
@@ -0,0 +1,173 @@
...
+    JiraTaskSyncManager,
+    JiraTaskTransitionManager,
+    SyncManager,
+)
+from osidb.tests.factories import FlawFactory
+
+pytestmark = pytest.mark.unit
+
+
+class TestSyncManager(TestCase):
```
JakubFrejlach, line 22: I am consufed how does this work, django docs mentions `setUpClass`/`tearDownClass`  method but no `setup_class`/`teardown_class` https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.TestCase, haven't really worked much with django TestCases though :smile:

JakubFrejlach (changes requested):
Couple of things/questions for tests, otherwise LGTM!

osidb/tests/test_sync_manager.py
Elkasitu: FWIW, the setup/teardown can be achieved using pytest fixtures and we avoid mixing `pytest` and `unittest`.

Another alternative is defining the test model outside of the test's scope, I think we do this already for a few tests (so you can just reference the model directly and it only exists in the test database).

PR #987, opened 2025-06-25
OSIDB-4326: Set score attribute when updating CVSS scores from cve.org

Elkasitu:
Before this commit, if a CVE scanned by the cve.org collector had a CVSS update, the vector and everything else would be updated but the score wouldn't.

This is because the way the CVSS update is made, the pre_save signal on FlawCVSS which triggers the score computation doesn't run.

This commit explicitly sets the score and introduces a data migration that corrects the score of other CVSS records that might have a wrong score.

Closes OSIDB-4326


osoukup (approved):
LGTM

MrMarble (approved):

PR #990, opened 2025-07-02
[OSIDB-4337] Extend Tracker Special Handling choices with new values

MrMarble:
Tracker model was missing some values in `Special Handling`, this prevented the `JiraTrackerCollector` to skip some updates or entire trackers from syncing back to OSIDB as the value in `special handling` was invalid.

I used a Jira filter to search all the trackers that are using the new values and found 180, we should manually sync those trackers once this is deployed

Closes OSIDB-4337


JakubFrejlach (approved):

Elkasitu (commented):
I think this may break tooling like what happened with the CVSS scores 'CISA' addition, so we should ensure during next release to sync OSIM and osidb-bindings releases at least.

BTW, I think we could just write a data migration to sync the missing trackers, no?

PR #991, opened 2025-07-02
OSIDB-3704: Supply PURL for middleware engineering trackers if available

Elkasitu:
For engineering trackers targeting middleware PsModules we want to populate the JIRA "Downstream component" field with PURLs if available and ps_component otherwise, for middleware the ps_component usually doesn't tell engineering teams much and they desperately need PURLs in order to more efficiently do their job.

Since their dependence on ps_component low, we can safely assume this change won't bring many downsides to them.

Other categories of PsModules may implement the change further down the line.

Closes OSIDB-3704

### DIFF
--- a/apps/trackers/jira/constants.py
+++ b/apps/trackers/jira/constants.py
@@ -1,6 +1,8 @@
 """
 Jira specific tracker constants
 """
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
 from osidb.helpers import get_env
 
 JIRA_SERVER = get_env("JIRA_URL", default="https://issues.redhat.com")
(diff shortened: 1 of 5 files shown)


JakubFrejlach (approved):
LGTM!

Elkasitu:
@osoukup @JakubFrejlach could you re-review? I added a feature flag

apps/trackers/jira/constants.py
```diff
@@ -19,3 +21,11 @@
     "target_release": "customfield_12311240",
     "target_version": "customfield_12319940",
 }
+
+
+class TrackersAppSettings(BaseSettings):
```
JakubFrejlach, line 26: I like the usage of the `pydantic_settings`, maybe we could standardize on that.

JakubFrejlach (approved):
LGTM!

apps/trackers/jira/constants.py
Elkasitu: yeah, it has lots of benefits, I'll bring up a thread in slack

PR #992, opened 2025-07-03
[OSIDB-3835] Add link to Vuln Mgmt documentation in tracker description

Jincxz:
Uses the same logic as the code for OSIDB-4095. A link to Vulnerability Management documentation will be attached to trackers during creation if the environment variable, `VULN_MGMT_INFO_URL`, is defined.

Closes OSIDB-3835.

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -7,7 +7,11 @@ from urllib.parse import urljoin
 
 from apps.bbsync.constants import MAX_SUMMARY_LENGTH, MULTIPLE_DESCRIPTIONS_SUBSTITUTION
 from apps.bbsync.query import summary_shorten
-from apps.trackers.constants import KERNEL_PACKAGES, VIRTUALIZATION_PACKAGES
+from apps.trackers.constants import (
+    KERNEL_PACKAGES,
+    VIRTUALIZATION_PACKAGES,
+    VULN_MGMT_INFO_URL,
+)
 from apps.trackers.jira.constants import TRACKER_FEEDBACK_FORM_URL
 from collectors.bzimport.constants import BZ_URL
 from osidb.helpers import cve_id_comparator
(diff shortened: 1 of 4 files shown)


apps/trackers/common.py
```diff
@@ -366,3 +373,15 @@ def _description_feedback_form(self):
         if TRACKER_FEEDBACK_FORM_URL is None:
             return []
         return [f"Tracker accuracy feedback form: {TRACKER_FEEDBACK_FORM_URL}"]
+
+    def _description_Vuln_Mgmt_info(self):
```
JakubFrejlach, line 377: I would refrain from using capitalized letters in snake case. I think it is even discouraged in PEP.

apps/trackers/tests/test_common.py
```diff
@@ -691,6 +691,39 @@ def test_feedback_form(self, form_url):
             else:
                 assert "feedback form" not in tqb.description
 
+    @pytest.mark.parametrize("info_url", [None, "https://example.doc.com"])
+    def test_VM_Info_link(self, info_url):
```
JakubFrejlach, line 695: Again, I would discourage from using capital leters in snake case and I would make consistent usage of VM vs VulnMgmt. VM is commonly used acronym for Virtual Machine so I would avoid using it in this context. Also I think even the full name `vulnerability management` usage would not be wrong in this case.

apps/trackers/tests/test_common.py
```diff
@@ -691,6 +691,39 @@ def test_feedback_form(self, form_url):
...
+
+        with patch("apps.trackers.common.VULN_MGMT_INFO_URL", info_url):
+            tqb = TrackerQueryBuilder()
+            tqb.instance = tracker
+
+            if info_url is not None:
+                assert info_url in tqb.description
+            else:
+                assert (
+                    "essential vulnerability management information"
```
JakubFrejlach, line 723: I would rather match the URL presence than just this text.

JakubFrejlach (changes requested):
Couple of things. Also once this get merged, we should update the osidb-ops repository to export the form link into env variable.

apps/trackers/tests/test_common.py
Jincxz: In the `else` case, the URL passed to the test URL would be `None` so checking for its presence would not be possible. The intention behind this part of the test is to make sure that the text before the link does not somehow show up if the URL is not provided.

PR #993, opened 2025-07-04
Add AffectCVSS v2 API

Elkasitu:
From an earlier effort to migrate the CVSS APIs to a more robust V2, the V2 API for Flaws was properly created and exposed however the one for Affects was sadly forgotten about.

(It was me, I forgot about it, don't tell Jessie).


osoukup (approved):
LGTM

PR #997, opened 2025-07-09
OSIDB-4293: Add flaw `cve_id` to affect endpoint

MrMarble:
Added `cve_id` field to the affect serializer, this allows OSIM to fetch related flaws from the `/affects` endpoint, improving performance, it was also recently asked by VM team.

Queryset regression grew by 1 as the `flaw` was added to the `prefetch_required` in the view, without this each affect would have added an extra query for the flaw

Closes OSIDB-4293


Elkasitu (approved):
Awesome stuff, thanks!

PR #999, opened 2025-07-09
[OSIDB-4328]  Add available-flaws endpoint

Jincxz:
The endpoint takes in a CVE ID (.../osidb/api/v1/available/flaw-id) and returns a response based on the following criteria:
1) Public / work on flaw is done 
     => 204 No Content
2) Not public and work not done / flaw does not exist
     => 404 Not Found 

Closes OSIDB-4328.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -2520,6 +2520,88 @@ paths:
                     version:
                       type: string
           description: ''
+  /osidb/api/v1/available-flaws/{cve_id}:
+    get:
+      operationId: osidb_api_v1_available_flaws_retrieve
+      description: |-
+        Report whether a flaw is available for public consumption purposes
+        based on the following criteria:
+        1) The work on the flaw is done, or the flaw is public:
+            - 204 status (yes, flaw is available for public consumption)
+        2) The work on the flaw is not done yet, or the flaw doesn't exist in the DB:
+            - 404 status (no, flaw is unavailable for public consumption)
+        3) Invalid CVE ID:
+            - 400 status
+
+        The intention is that this API is consumed by an agent that publishes pages
+        with information about individual CVEs. As long as this API returns 404,
+        the agent waits and doesn't publish the CVE page. Once this API first returns 204,
+        the agent stops polling this API and publishes the CVE page. The consumers of such
+        CVE pages are then informed about the CVE in such a way that the general affectedness
+        ("Does the CVE affect products shipped by the organization that publishes the CVE
+        page, or not?") most likely doesn't change. So this is to prevent public confusion
+        during the early stages of security analysis where the preliminary analysis might
+        switch between "this CVE affects our products" and "this CVE doesn't affect our products".
+      parameters:
+      - in: path
+        name: cve_id
+        schema:
+          type: string
+        required: true
+      tags:
+      - osidb
+      security:
+      - OsidbTokenAuthentication: []
+      - {}
+      responses:
+        '204':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  dt:
+                    type: string
+                    format: date-time
+                  env:
+                    type: string
+                  revision:
+                    type: string
+                  version:
+                    type: string
+          description: ''
+        '400':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  dt:
+                    type: string
+                    format: date-time
+                  env:
+                    type: string
+                  revision:
+                    type: string
+                  version:
+                    type: string
+          description: ''
+        '404':
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  dt:
+                    type: string
+                    format: date-time
+                  env:
+                    type: string
+                  revision:
+                    type: string
+                  version:
+                    type: string
+          description: ''
   /osidb/api/v1/flaws:
     get:
       operationId: osidb_api_v1_flaws_list
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -31,7 +31,12 @@ from rest_framework.decorators import action, api_view, permission_classes
 from rest_framework.permissions import AllowAny, IsAuthenticatedOrReadOnly
 from rest_framework.request import Request
 from rest_framework.response import Response
-from rest_framework.status import HTTP_200_OK, HTTP_204_NO_CONTENT
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_204_NO_CONTENT,
+    HTTP_400_BAD_REQUEST,
+    HTTP_404_NOT_FOUND,
+)
 from rest_framework.views import APIView
 from rest_framework.viewsets import (
     ModelViewSet,
@@ -41,7 +46,9 @@ from rest_framework.viewsets import (
 )
 from rest_framework_simplejwt.views import TokenObtainPairView
 
+from apps.workflows.workflow import WorkflowModel
 from collectors.jiraffe.constants import HTTPS_PROXY, JIRA_SERVER
+from osidb.core import set_user_acls
 from osidb.helpers import get_bugzilla_api_key
 from osidb.integrations import IntegrationRepository, IntegrationSettings
 from osidb.models import Affect, AffectCVSS, Flaw, FlawLabel, Tracker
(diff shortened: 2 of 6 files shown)


osidb/api_views.py
```diff
@@ -811,6 +819,49 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
         return super().destroy(request, *args, **kwargs)
 
 
+class FlawAvailableView(
```
Elkasitu, line 822: nitpick: I think it makes more sense to use a function-based view in this case, you'd avoid a lot of boilerplate and in the end this is a single-method class

osidb/api_views.py
```diff
@@ -811,6 +819,49 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+    def get(self, request, *args, **kwargs):
+        """
+        Report whether a flaw is available based on the following criteria:
+
+        1) Public / work on flaw is done: 200 status (yes)
+        2) Not public and work not done / flaw does not exist: 404 status (no)
+        """
+
+        pk = self.kwargs["flaw_id"]
```
Elkasitu, line 837: Should probably be named `cve_id` and not `flaw_id`, the latter and the variable name strongly imply that this is Flaw's primary key which is a UUID.

osidb/api_views.py
```diff
@@ -811,6 +819,49 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+        """
+        The try-catch block is here so that the response is consistent
+        between criterion (2) and a flaw not being found.
+        """
+        try:
+            # Currently only allow search by CVE ID
+            if CVE_RE_STR.match(pk):
+                set_user_acls(settings.ALL_GROUPS)
+                flaw = get_object_or_404(Flaw, cve_id=pk)
+                self.check_object_permissions(self.request, flaw)
```
Elkasitu, line 848: This is not necessary and saves you the encompassing try/except:

```python
try:
    flaw = Flaw.objects.get(cve_id=pk)
except Flaw.DoesNotExist:
    return Response(status=HTTP_404_NOT_FOUND)
```

`get_object_or_404` is more useful for subviews e.g. `/flaws/<cve-id>/cvss-scores`, as you need the object to complete your task but is not the goal of the task itself.

FYI, while I'm aware we use `check_object_permissions` in OSIDB, we should avoid it as it's essentially useless: we don't use Django permissions in OSIDB, it's cargo cult at this point.

osidb/tests/endpoints/flaws/test_flaws.py
```diff
@@ -1819,3 +1820,64 @@ def test_flaw_history_no_history(self, auth_client, test_api_uri):
...
+        "workflow_status, embargoed, internal",
+        [
+            (WorkflowModel.WorkflowState.NEW, True, False),
+            (WorkflowModel.WorkflowState.NEW, False, False),
+            (WorkflowModel.WorkflowState.NEW, False, True),
+            (WorkflowModel.WorkflowState.DONE, True, False),
+            (WorkflowModel.WorkflowState.DONE, False, False),
+            (WorkflowModel.WorkflowState.DONE, False, True),
+        ],
+    )
```
Elkasitu, line 1834: fun fact: parametrize creates matrices and are thus multiplicative, this is equivalent to:

```python
@pytest.mark.parametrize("workflow_status", (NEW, DONE))
@pytest.mark.parametrize("embargoed, internal", [(True, False), (False, False), (False, True)])
```

osidb/api_views.py
```diff
@@ -811,6 +819,49 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+                set_user_acls([])
+            else:
+                return Response(status=HTTP_400_BAD_REQUEST)
+
+            if (
+                flaw.is_public
+                or flaw.classification["state"] == WorkflowModel.WorkflowState.DONE
+            ):
+                return Response(status=HTTP_200_OK)
```
Elkasitu, line 857: Should be 204

Elkasitu (changes requested):
I agree with what @osoukup said, but the PR as-is was done according to spec so we'll just have to wait on Jakub's input.

To summarize what Ondrej said:

Should 204 mean any flaw regardless of whether it's public/internal/embargoed as long as it's not in a `REJECTED` state?
And 404 mean missing in our records or flaws in `REJECTED` state?

osidb/api_views.py
```diff
@@ -811,6 +818,36 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+        set_user_acls(settings.ALL_GROUPS)
+        flaw = Flaw.objects.get(cve_id=cve_id)
+        set_user_acls([])
+    except Flaw.DoesNotExist:
+        set_user_acls([])
+        return Response(status=HTTP_404_NOT_FOUND)
+
+    if (
+        flaw.is_public
+        or flaw.classification["state"] == WorkflowModel.WorkflowState.DONE
```
jsvob, line 844: What about when the analyst rejects the flaw? I don't understand OSIDB enough anymore, but I think a rejected nonpublic flaw is also considered being done, right? And in such a case we'd want to display it on the portal. If I understand correctly, in the current implementation, not affected flaws would forever stay hidden on the portal, which would go against the idea behind this API.

jsvob (changes requested):

jsvob (changes requested):
@osoukup is correct in the assessment. This API will directly drive the creation/noncreation of CVE pages on portal:

- OSIDB responds 40x on this API: The page is not created.
- OSIDB responds 20x on this API: The page is created. If OSIDB offers the flaw on the public /flaws API, OSIDB's data get used, if OSIDB doesn't offer the flaw on the public /flaws API, the cve page gets created using cvelistV5 repo.

So the 20x case must include all states which we humanly consider as "VM is done, we can publish the cve page, no need to delay it anymore".

Thanks!

osidb/api_views.py
```diff
@@ -811,6 +818,36 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+    1) Public / work on flaw is done: 204 status (yes)
+    2) Not public and work not done / flaw does not exist: 404 status (no)
+    """
+    cve_id = kwargs["cve_id"]
+    if not CVE_RE_STR.match(cve_id):
+        return Response(status=HTTP_400_BAD_REQUEST)
+
+    try:
+        set_user_acls(settings.ALL_GROUPS)
```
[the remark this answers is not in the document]
Elkasitu: Yes, OSIDB sets Django's `ATOMIC_REQUESTS` setting to `True` in all environments.

Furthermore, ACLs are reset to the requesting user's ACLs on each new request, the explicit reset is not really needed here but good practice nevertheless.

osidb/api_views.py
```diff
@@ -811,6 +818,36 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+def flaw_available(request: Request, *args, **kwargs) -> Response:
+    """
+    Report whether a flaw is available based on the following criteria:
+
+    1) Public / work on flaw is done: 204 status (yes)
+    2) Not public and work not done / flaw does not exist: 404 status (no)
+    """
+    cve_id = kwargs["cve_id"]
+    if not CVE_RE_STR.match(cve_id):
+        return Response(status=HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
Elkasitu: Here the 400 is being returned in the case that the client (i.e. SDEngine) submits a malformed CVE id (e.g. "foo"), so yes technically it's not a 1-bit API in this case, but do we want to say "unaffected" when the input is malformed? Shouldn't we return an error to avoid the error possibly propagating into CVE pages?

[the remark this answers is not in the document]
jsvob: Intended consumer's logic described in https://github.com/RedHatProductSecurity/osidb/pull/999#discussion_r2199710261

openapi.yml
```diff
@@ -2520,6 +2520,42 @@ paths:
...
+                  dt:
+                    type: string
+                    format: date-time
+                  env:
+                    type: string
+                  revision:
+                    type: string
+                  version:
+                    type: string
+          description: ''
```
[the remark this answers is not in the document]
Jincxz: I was able to add the response statuses to the documentation but was unable to fill in the description of the response. Attempted to use OpenApiResponse but the description did not register. The description of what each response is currently described in the main endpoint description and the docstring.

Elkasitu (approved):
LGTM, good job!

osidb/api_views.py
```diff
@@ -811,6 +818,59 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+        flaw = Flaw.objects.get(cve_id=cve_id)
+        set_user_acls([])
+    except Flaw.DoesNotExist:
+        set_user_acls([])
+        return Response(status=HTTP_404_NOT_FOUND)
+
+    if (
+        flaw.is_public
+        or flaw.classification["state"] == WorkflowModel.WorkflowState.REJECTED
+        or flaw.classification["state"] == WorkflowModel.WorkflowState.DONE
```
JakubFrejlach, line 867: You can access directly the `workflow_state` field instead of using the `classification` property.

osidb/api_views.py
```diff
@@ -811,6 +818,59 @@ def destroy(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+    during the early stages of security analysis where the preliminary analysis might
+    switch between "this CVE affects our products" and "this CVE doesn't affect our products".
+    """
+
+    cve_id = kwargs["cve_id"]
+    if not CVE_RE_STR.match(cve_id):
+        return Response(status=HTTP_400_BAD_REQUEST)
+
+    try:
+        set_user_acls(settings.ALL_GROUPS)
```
JakubFrejlach, line 857: I think we don't even have to override the ACLs here. `404` should be returned when either the Flaw with such CVE does not exist or when it is not public. If someone with insufficient permissions for anything else than public will hit the API then, it won't find the Flaw because of the ACLs and the result will be exactly the same but we avoid overriding the ACLs, which feels "more safer" IMHO.

osidb/tests/endpoints/flaws/test_flaws.py
```diff
@@ -1819,3 +1820,66 @@ def test_flaw_history_no_history(self, auth_client, test_api_uri):
...
+        assert flaw.save() is None
+
+        # response should return a state with no data
+        response = client.get(f"{test_api_uri}/available-flaws/{flaw.cve_id}")
+        assert response.data is None
+
+        if (
+            flaw.is_public
+            or flaw.classification["state"] == WorkflowModel.WorkflowState.REJECTED
+            or flaw.classification["state"] == WorkflowModel.WorkflowState.DONE
```
JakubFrejlach, line 1871: You can access directly the workflow_state field instead of using the classification property.

JakubFrejlach (changes requested):
Couple of objections. Otherwise LGTM.

osidb/api_views.py
Jincxz: I think I saw `workflow_state`  being accessed elsewhere through `classification` which is why I ended up using it. Are there scenarios where you would choose one access point over the other?

osidb/api_views.py
jsvob: > I think we don't even have to override the ACLs here. 
We have to.

> 404 should be returned when either the Flaw with such CVE does not exist or when it is not public.
No. Please read OSIDB-4328.

>  If someone with insufficient permissions for anything else than public will hit the API then, it won't find the Flaw because of the ACLs and the result will be exactly the same but we avoid overriding the ACLs, which feels "more safer" IMHO.
???

How would you achieve the goal of OSIDB-4328 then? Concretely, this case:

_"A True [20x] answer might mean: 1. The CVE is not switched to public and is in a workflow state that equals to work on it being finished, at least for now. 2. (...)"_

> @jsvob is there a reason for the endpoint to be public? 
Yes. SDEngine.

> In other words, could we potentially rely on SDEngine having the correct authorization?
No. Compliance reasons. And see OSIDB-4328 for an explanation why it doesn't break any embargos or security boundaries. This is why I care so much about the API having as few bits of output information as possible.

jsvob (approved):
LGTM. I like that it has effectively been reverted to a 1bit response (where the 2nd bit is solely for checking whether the request matches a regex, without revealing any DB data). This is the safest approach IMO. Thanks!

osidb/api_views.py
JakubFrejlach: I think initially we had just the `classification` and now we have a dedicated fields for that AFAIK. I think if you can use the field directly, it is usually better.

PR #1000, opened 2025-07-10
OSIDB-4351: Expand Affect unique together constraint to encompass purl

Elkasitu:
With this commit, it is now possible to have two affects with the same flaw, ps_module, ps_component but different PURLs, as the PURLs may have different modifiers that warrant tracking across multiple affects.

Closes OSIDB-4351


JakubFrejlach (changes requested):
I am not sure whether introducing this change will be so straightforward. There might be other things but at the very minimum I see a problem with the tracker linking as it currently relies on the fact that Flaw can only have 1 affect for each `ps_module/ps_component` combination. 

https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/sync_manager.py#L1057

Even tracker automanager will need to be adjusted a bit so the different purls with same ps_component are taken into account. And I am not even sure how this will work out for sdengine and publishing the cve pages because that may break too if they are counting on this.

Requesting changes as we need to think this through a bit more before making this change.

osoukup:
I assume that this will be solved with affects v2

PR #1006, opened 2025-07-21
[OSIDB-4288] Switch linting and formatting to ruff

Jincxz:
Replace flake8, isort, black, and bandit with ruff.

The main changes are in pyproject.toml, tox.ini, .pre-commit-config.yaml, and lint.yml.

### The following rules should replace the original tools:
- flake8: Pyflakes (`F`), pycodestyle (`E`), Warning (`W`), and flake8-django (`DJ`)
- isort: isort (`I`)
- bandit: flake8-bandit (`S`)

black is replaced with ruff format.

Ruff configuration was written to mimic the old configuration as much as possible (see [pyproject.toml](https://github.com/RedHatProductSecurity/osidb/blob/OSIDB-4288-Ruffplacement/pyproject.toml)). The `E`  rules that are ignored are not included by default but were added just in case.

The ruff tool is mostly ran in three parts: ruff-check, ruff-isort, and ruff-format (see [tox.ini](https://github.com/RedHatProductSecurity/osidb/blob/OSIDB-4288-Ruffplacement/tox.ini) and the changed yaml files). The original linting check with flake8 and security check with bandit has been combined into one command (ruff-check) since they share largely check the same directories (osidb, collectors, apps). 

The `S` rules are ignored for files in the test folders. Files in migrations folders are excluded from the ruff rules in general. This is based on configuration from the original tox.ini where migrations are excluded from most of commands ran except for bandit. Will need to adjust ruff to check migrations for `S` rules if needed.

The commit also covers some formatting that the original tools did not format. Most of these are formatting for contents inside models ([DJ012](https://docs.astral.sh/ruff/rules/django-unordered-body-content-in-model/)), whitespace, and other forms of spacing.

flake8 did not detect some missing `__str__` methods for some of the models. I only wanted to add formatting changes for now so I suppressed [DJ008](https://docs.astral.sh/ruff/rules/django-model-without-dunder-str/).

Closes OSIDB-4288.

### DIFF
--- a/apps/bbsync/apps.py
+++ b/apps/bbsync/apps.py
@@ -1,7 +1,7 @@
 """
 
-    BBSync - Bugzilla Backwards Synchronization
-        osidb.bbsync to be precise
+BBSync - Bugzilla Backwards Synchronization
+    osidb.bbsync to be precise
 
 """
 
--- a/apps/bbsync/constants.py
+++ b/apps/bbsync/constants.py
@@ -1,6 +1,7 @@
 """
 BBSync constants
 """
+
 from osidb.helpers import get_env
 
 DATE_FMT = "%Y-%m-%d"
--- a/apps/bbsync/mixins.py
+++ b/apps/bbsync/mixins.py
@@ -38,7 +38,7 @@ class BugzillaSyncMixin(models.Model):
                 *args,
                 bz_api_key=bz_api_key,
                 force_synchronous_sync=force_synchronous_sync,
-                **kwargs
+                **kwargs,
             )
 
     def bzsync(self, *args, bz_api_key, **kwargs):
--- a/apps/bbsync/models.py
+++ b/apps/bbsync/models.py
@@ -1,6 +1,7 @@
 """
 Bugzilla metadata models
 """
+
 from django.contrib.postgres import fields
 from django.db import models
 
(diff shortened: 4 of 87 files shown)


Jincxz:
@osoukup I have split the commits into the core setup changes (https://github.com/RedHatProductSecurity/osidb/pull/1006/commits/a4a6366228634f06f210ce0bf99d8d6afbf17dfc) and the formatting changes (https://github.com/RedHatProductSecurity/osidb/pull/1006/commits/6dbe216591fdb0e435a00262e6693f0221f6b610).

tox.ini
```diff
@@ -92,10 +71,14 @@ commands = /usr/bin/bash -c './scripts/migrations-check.sh'
 allowlist_externals = bash
 commands = /usr/bin/bash -c './scripts/schema-check.sh'
 
-[isort]
-profile = black
-skip = migrations,venv,.tox,src,attic
+[testenv:ruff-check]
+deps = ruff==0.12.3
+commands = ruff check --extend-ignore I osidb collectors apps
```
JakubFrejlach, line 76: We could make use of [substituting values from different section ](https://tox.wiki/en/3.24.0/config.html#substitution-for-values-from-other-sections) to have the `ruff` version defined only in one place.

JakubFrejlach (approved):
One small suggestion, otherwise LGTM!

PR #1011, opened 2025-07-24
OSIDB-4367: Relax PURL and ps_component mismatch validation

Elkasitu:
In the context of backfilling PURLs into OSIDB, it may happen that ps_components extrapolated from PURLs may not match the originally provided ps_component.

This could be due to an error in the assigned ps_component or in the PURL parsing, however this should not be a blocker for such an automatic or pseudo-automatic operation.

This relaxation also allows users to override the ps_component when the PURL parsing does not work properly.

Closes OSIDB-4367

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Changed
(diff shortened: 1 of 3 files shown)


JakubFrejlach (approved):

osoukup:
Could you provide a bit more explanation? We are filling in some historical PURLs (maybe the SDEngine ones?) and they may not match the PS components for some reason? Will we use API for this migration? If not, we can just one time suppress the error. I am concerned about the future data. Relaxing the relation means that going forward I can provide whatever component without any connection to the PURL I provide together with that.

PR #1014, opened 2025-07-29
Avoid dependence on secrets for testing workflow

Elkasitu:
This PR makes it possible for external contributors to run the testing workflow, making it easier for us to review and integrate external contributions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,34 +14,26 @@ jobs:
       # it much simpler to transition from GitLab to GitHub as we use the
       # same image and dependencies / setup as before, it should probably
       # be revisited in the not-so-distant future
-      image: quay.io/fedora/fedora:latest
-      credentials:
-        username: ${{ secrets.QUAY_REGISTRY_USERNAME }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -50,54 +42,33 @@ jobs:
...
-          rsync
-          tox
-          wget
-          which
       - name: Setup PostgreSQL service
         run: psql -f etc/pg/local-dev-app-user.sql postgres://postgres:test@postgres:5432/
       - name: Setup LDAP service
         run: ldapadd -c -H "ldap://testldap:1389" -x -D "cn=admin,dc=redhat,dc=com" -w "adminpassword" -f etc/openldap/local-export.ldif || true
+      - name: Install Python devel dependencies
+        run: pip3 install --no-deps -r devel-requirements.txt
```
JakubFrejlach, line 67: Running the `tox -e ci-osidb` performs the install of both requirements and devel-requirements inside the tox python environment so the only thing necessary should be the tox itself.

JakubFrejlach (approved):
Just minor suggestion regarding the python deps, otherwise LGTM!

PR #1015, opened 2025-07-31
Fix logic for available-flaws to match the original spec

jsvob:
Followup to https://github.com/RedHatProductSecurity/osidb/pull/999/files#r2245509506

Sorry for the misunderstanding.

There's a lot of CVEs that don't exist in OSIDB, I presume because some sort of automatic filtering doesn't let them through to VM. For very random examples, CVE-2024-46451 or CVE-2025-2691. The goal is that these CVEs get processed as "unaffected" ones by the API consumer. Without this PR, the consumer gets no useful information from the API.

Closes OSIDB-4397

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -830,9 +830,9 @@ def flaw_available(request: Request, *args, **kwargs) -> Response:
     """
     Report whether a flaw is available for public consumption purposes
     based on the following criteria:
-    1) The work on the flaw is done, or the flaw is public:
+    1) The work on the flaw is done, or the flaw is public, or the flaw doesn't exist in the DB:
         - 204 status (yes, flaw is available for public consumption)
-    2) The work on the flaw is not done yet, or the flaw doesn't exist in the DB:
+    2) The work on the flaw is not done yet:
         - 404 status (no, flaw is unavailable for public consumption)
     3) Invalid CVE ID:
         - 400 status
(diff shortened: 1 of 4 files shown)


Jincxz:
Would this be instead a change to the original spec where "CVE is not known to OSIDB" is now considered true?

jsvob:
Another mistake I made was that when writing the spec/requirements, I wasn't aware of the `should_create_snippet`-based filtering and assumed OSIDB stores information about all CVEs. This helped the resulting confusion. Explained that in the API to prevent future confusion where a few years later someone wonders why the API contains a seemingly obvious bug of not returning 404 for a non-existent Flaw.

Jincxz (approved):
LGTM

osidb/api_views.py
```diff
@@ -846,6 +846,17 @@ def flaw_available(request: Request, *args, **kwargs) -> Response:
     page, or not?") most likely doesn't change. So this is to prevent public confusion
     during the early stages of security analysis where the preliminary analysis might
     switch between "this CVE affects our products" and "this CVE doesn't affect our products".
+
+    Also an important point is that the client processes CVEs that never get saved to OSIDB's
+    DB (because of internal function `should_create_snippet`), yet the client must
+    publish information about all CVEs. By returning 204 when the flaw doesn't exist in OSIDB's
+    DB, it allows the client to take the output of this API endpoint as actionable advice:
+    When 204, publish the CVE page (either using OSIDB data or using other data), when
+    404, do not publish the CVE page (because the incoming team still works on the CVE).
```
JakubFrejlach, line 855: Nitpick: I am not sure whether the name incoming team still exists in the RH context.

JakubFrejlach (approved):
LGTM.

osidb/api_views.py
jsvob: Do we have a better term that is not internal and intuitively understandable?

osidb/tests/endpoints/flaws/test_flaws.py
```diff
@@ -1888,7 +1888,7 @@ def test_flaw_available(
 
         # check for non-existent flaw (should be the same as criterion (2))
```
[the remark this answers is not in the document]
jsvob: fixed, thx

jsvob:
@Jincxz @JakubFrejlach  I think it should be ready for merging if "incoming" is not an issue but I don't want to make a mess even though I technically still have the rights. Please merge if appropriate. Thank you!

PR #1021, opened 2025-08-04
[OSIDB-4287] Remove dependabot updates

Jincxz:
Adjustment to [PR 1013](https://github.com/RedHatProductSecurity/osidb/pull/1013) to stop dependabot from automatically making PRs for updating packages that are not security updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,0 @@
-version: 2
-
-updates:
-  - package-ecosystem: "uv"
-    directory: "/"
-    schedule:
-      interval: "weekly"


JakubFrejlach:
> This doesn't seem to do what it says it does? 🤔 How does this fix non-security Dependabot updates?
Also these two commits are basically identical to the last two commits of #1013

Jincxz:
The change is that dependabot.yml got removed. I don't think the diff showed because I ended up squashing it into the "github workflow" commit 😅

Will edit to make it clearer.

Elkasitu (approved):

PR #1027, opened 2025-08-06
Fix FlawSource enum in OpenAPI

JakubFrejlach:
This PR:
* fixes the FlawSource enum in OpenAPI schema so its name is fixed rather then being always produced by drf-spectacular name collision resolver, this will help clients like `osidb-bindings` to rely on the enum naming

Closes OSIDB-4401


Elkasitu (approved):

dmonzonis (approved):
LGTM

PR #1030, opened 2025-08-14
Adapt collectors to affects v2

dmonzonis:
Adapt collectors to affects v2, plus some fixes for the migration and core of affects v2.

Closes OSIDB-4044.


osoukup (approved):
Bugzilla collector part might be rather removed right away (OSIDB-4043) but I have not issue with updating it. All together LGTM

PR #1031, opened 2025-08-15
[OSIDB 4202] Python Upgrade

Jincxz:
Updated local, containers, and tests to use Python 3.12.

Code and packages have been updated for deprecated packages/modules (e.g. imp, distutils, and pkg_resources). These changes are primarily in pyproject.toml/uv.lock. setuptools and certifi have also been updated as part of this PR (OSIDB-4307).

Some miscellaneous updates include:
- Corrected make entrypoint name for syncing dependencies in containers. Also adjusted the logic for syncing so that it doesn't error out when the containers are already running.
- Python environment have been moved out of mounted directory to avoid the use of anonymous volumes which prevented updates to container packages.
- Made affect's flaw field un-null-able since it was affecting schema generation (affects should be attached to a flaw). 
- Added requirements.txt (hopefully this is temporarily) because dependabot / github's dependency tree doesn't appear to scan uv.lock yet.

Next steps:
- ~~More testing on stage to check for stability.~~
- Updates to the ops directory and the pipeline as necessary.

Closes OSIDB-4202.
Closes OSIDB-4307.


JakubFrejlach (changes requested):
Placeholder request changes. Would like to go through this before it gets merged. Will try to do it today/tomorrow.

PR #1032, opened 2025-08-20
Use internal ACLs for new flaws

superbuggy:
There are still considerable logic that uses the PUBLIC groups, and conceivably those will need to be adapted. Is it necessary to change to use `settings.INTERNAL_`{READ/WRITE}`_GROUP` everywhere? It seems like it would but it certainly seems to enlarge the impact and scope of this change. Should I abort this approach to preventing the emails? I wonder if there is a more granular, well-scoped approach.

Closes OSIDB-4274

### DIFF
--- a/osidb/mixins.py
+++ b/osidb/mixins.py
@@ -517,25 +517,10 @@ class ACLMixin(models.Model):
                     + ("embargoed" if self.flaw.is_embargoed else "public")
                 )
 
-    def unembargo(self):
+    def set_history_public(self):
         """
-        unembargo the whole instance context internally
-
-        in the Bugzilla world a lot of OSIDB entities are actually parts
-        of the flaw bug and we will update them by a single query afterwards
+        set the history to public
         """
-        # not every related model class has
-        # to have the same visibility logic
-        if not self.is_embargoed:
-            return
-
-        # allow individual model class special handling
-        # which is specifically important in multi-relations
-        if hasattr(self, "can_unembargo") and not self.can_unembargo():
-            return
-
-        # unembargo
-        self.set_public()
         refs = pghistory.models.Events.objects.references(self).all()
         for ref in refs:
             db, model_name = ref.pgh_model.split(".")
@@ -548,6 +533,25 @@ class ACLMixin(models.Model):
                     acl_read=list(self.acls_public_read),
                     acl_write=list(self.acls_public_write),
                 )
+
+    def unembargo(self):
+        """
+        unembargo the whole instance context internally
+
+        in the Bugzilla world a lot of OSIDB entities are actually parts
+        of the flaw bug and we will update them by a single query afterwards
+        """
+        # Since ACLMixin is used across related classes, each must be permitted to implement their own
+        # visibility logic. Bailing out allows for unique visibility handling across such classes
+        isnt_embargoed = not self.is_embargoed
+        cant_unembargo = hasattr(self, "can_unembargo") and not self.can_unembargo()
+        if isnt_embargoed or cant_unembargo:
+            return
+
+        # unembargo
+        self.set_public()
+        self.set_history_public()
+
         kwargs = {}
         if issubclass(type(self), AlertMixin):
             # suppress the validation errors as we expect that during
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -5,6 +5,7 @@ serialize flaw model
 import logging
 import uuid
 from collections import defaultdict
+from enum import Enum
 from typing import Dict, List, Tuple
 
 import pghistory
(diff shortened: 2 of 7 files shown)


osoukup (changes requested):
I stopped the review after spotting a few similar issues as I think there is some misunderstanding. This task is not about changing how embargo works. This task is about changing the ACLs on manual flaw creation from public to internal. No other behaviour should be changed.

osidb/serializer.py
```diff
@@ -429,10 +429,10 @@ def get_acls(self, embargoed):
         generate ACLs based on embargo status
         """
         acl_read = (
-            settings.EMBARGO_READ_GROUP if embargoed else settings.PUBLIC_READ_GROUPS
+            settings.EMBARGO_READ_GROUP if embargoed else settings.INTERNAL_READ_GROUP
         )
         acl_write = (
-            settings.EMBARGO_WRITE_GROUP if embargoed else settings.PUBLIC_WRITE_GROUP
+            settings.EMBARGO_WRITE_GROUP if embargoed else settings.INTERNAL_WRITE_GROUP
```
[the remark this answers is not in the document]
superbuggy: That makes more sense... would I want to hardcode this `ACLMixin` then (feels wrong to do that in `Flaw::save`)?
```py
   def save(self, *args, **kwargs):
        if not self.acl_read and not self.acl_write:
            from osidb.core import generate_acls

            self.acl_read = [
                uuid.UUID(acl) for acl in generate_acls([settings.INTERNAL_READ_GROUP])
            ]
            self.acl_write = [
                uuid.UUID(acl) for acl in generate_acls([settings.INTERNAL_WRITE_GROUP])
            ]
```

This feels kind of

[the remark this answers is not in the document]
superbuggy: Thank you for your helpful thoughts on this PR, @osoukup!

osidb/mixins.py
```diff
@@ -206,6 +206,17 @@ def __init__(self, *args, **kwargs):
         # caching the ACLs and building the group map
         self.acls_all
 
+    def save(self, *args, **kwargs):
+        if not self.acl_read and not self.acl_write:
```
osoukup, line 210: I think this assumes that there are no ACLs set when a user creates the flaw via API. **embargoed** is however a mandatory flaw parameter and it is always translated into either public or embargo ACLs.

osoukup (changes requested):
This solution does not seem correct to me. Anyway, I strongly recommend to create an integration test which will verify that a user-created flaw has internal visibility as intended.

osidb/tests/test_e2e.py
```diff
@@ -644,3 +645,188 @@ def test_cveorg_with_workflows(
...
+        updated_flaw_data = {
+            "title": "Updated Test ACL values",
+            "comment_zero": "This is an updated test for ACL validation",
+            "impact": "IMPORTANT",
+            "components": ["curl"],
+            "source": "REDHAT",
+            "reported_dt": "2024-08-06T00:00:00.000Z",
+            "unembargo_dt": "3339-08-06T00:00:00.000Z",
+            "updated_dt": flaw.updated_dt,
+            "embargoed": True,  # Change to embargoed
```
osoukup, line 788: Changing the visibility to something more strict does not make much sense and ideally/hopefully would be prevented by the validation. The reason is that when something was already visible publicly or internally, it is can be already cached somewhere and trying to hide it is already futile.

osoukup (changes requested):
I have a few comments, I do not want to look to pedantic but I think we still need to polish this. The test is a great addition!

osidb/tests/test_e2e.py
superbuggy: I wondered if it didn't make sense, but then wondered if there are ever edge cases where something needs to become embargoed, but upon looking at it again, a user wouldn't send this type of payload (`"embargoed": True`) with a PUT request. Should I strengthen the validation? This didn't trigger any complaint

osidb/serializer.py
```diff
@@ -428,11 +428,20 @@ def get_acls(self, embargoed):
         """
         generate ACLs based on embargo status
         """
+        internal = self.context["request"].method == "POST"
```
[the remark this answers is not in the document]
superbuggy: My first impulse was to check the pk/uuid (in def `embargo2acls`), but wasn't sure how to reliably access that data so I ended up with this much less felicitous approach here. I was trying stuff like `self.Meta.model.instance` but not seeing what I was expecting to.

superbuggy:
Thank you for thoughtful thoroughness here, @osoukup

osidb/tests/test_e2e.py
osoukup: So there actually is that validation https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/models/flaw/flaw.py#L691 but again only meant for the binary embargo/public world so does not get triggered by internal-to-embargoed change. You may generalize it so it works for any late attempt to restrict the access.

PR #1033, opened 2025-08-21
Update version to 4.15.0

JakubFrejlach:
$title


docs/CHANGELOG.md
```diff
@@ -5,16 +5,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+## [4.15.0] - 2025-08-21
 ### Fixed
 - Fix FlawSource enum reference in OpenAPI schema (OSIDB-4401)
+- Fix the available-flaws API giving a 404 on non-existent flaws instead of 204 (OSIDB-4397).
 
 ### Changed
 - Emit a warning instead of an error when the PURL-derived PsComponent doesn't
   match a user-provided PsComponent (OSIDB-4367)
 
-### Fixed
```
osoukup, line 15: So far we used the English alphabet sorting with C before F and not the weird Czech one where Ch is sort of one single letter. I would be for keeping it this way unless there is a reason :sweat_smile:

osoukup (changes requested):

docs/CHANGELOG.md
osoukup: I see that we do not have it consistent... So I do not insist on it then.

osoukup (approved):

PR #1036, opened 2025-08-21
Adapt sla module tests to affects v2

dmonzonis:
No real change to the sla module is required as it works in an abstract way with both v1 and v2 affects. Only the tests need to be adapted.

Closes OSIDB-4047.


osoukup (approved):
LGTM

PR #1037, opened 2025-08-21
Fix unembargo and adapt workflows tests to affects v2

dmonzonis:
The workflows module is not tied to affects so it works with both affects v1 and v2 out of the box, no changes are required in the module.

However a few tests have been adapted to affects v2. To make all tests pass, the unembargo functionality with the new affect-tracker relationship has also been fixed, which was also in the todo list for affects v2.

Closes OSIDB-4050.


osoukup (approved):
LGTM

PR #1039, opened 2025-08-25
Adapt sync managers to affects v2

Jincxz:
The sync managers were updated to grab affects based on their ps_update_stream rather than the ps_module. There is a test written directly for sync_manager.py but it does not appear to depend on affects so it remains unchanged. Some components uses the sync managers. I believe their dependency on affects v1 will be / has been resolved with their corresponding tickets. 

Please do let me know if I happen to miss something about the sync managers.

Closes OSIDB-4370.

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -569,14 +569,14 @@ class BZTrackerLinkManager(SyncManager):
             for flaw in flaws:
                 try:
                     affect = flaw.affects.get(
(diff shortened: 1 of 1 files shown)


dmonzonis (approved):
LGTM

osoukup (approved):
LGTM. Thank you for incorporating my feedback!

JakubFrejlach (approved):

PR #1040, opened 2025-08-25
Adapt exploits module to use affects v2

Jincxz:
This PR changes the exploits module to filter and perform serialization on a single tracker rather than a list as of the affects v2 update. The tests has also be adapted to use affects v2.

I believe that exploits API might need to be updated too (flaws will get more affects but the exploits API currently only outputs ps_module/product for these affects). Leaving this update for the larger v1 to v2 migration at the suggestion of @dmonzonis.

### DIFF
--- a/apps/exploits/serializers.py
+++ b/apps/exploits/serializers.py
@@ -29,11 +29,11 @@ class TrackerReportDataSerializer(serializers.ModelSerializer):
 
 
 class AffectReportDataSerializer(serializers.ModelSerializer):
-    trackers = TrackerReportDataSerializer(many=True, required=False)
+    tracker = TrackerReportDataSerializer(required=False)
 
     class Meta:
         model = Affect
-        fields = ["ps_module", "ps_component", "affectedness", "resolution", "trackers"]
+        fields = ["ps_module", "ps_component", "affectedness", "resolution", "tracker"]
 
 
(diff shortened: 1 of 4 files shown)


osidb/models/affect.py
```diff
@@ -103,6 +103,39 @@ def fts_search(q):
...
+        from osidb.filters import search_helper
+
+        fields_to_search = (
+            "ps_component",
+            "ps_module",
+            "resolution",
+            "affectedness",
+            "type",
+        )
+        return search_helper(Affect.objects.get_queryset(), fields_to_search, q)
```
dmonzonis, line 134: Why not `AffectV1`?

apps/exploits/serializers.py
```diff
@@ -44,6 +69,16 @@ class Meta:
         fields = ["cve_id", "affects"]
 
 
+class FlawV1ReportDataSerializer(serializers.ModelSerializer):
+    affects = AffectV1ReportDataSerializer(
+        source="v1_affects", many=True, required=False
```
dmonzonis, line 74: As explained above we can just use the reverse relationship.

dmonzonis (changes requested):
Some comments but this is looking much better already!

dmonzonis (approved):
LGTM, thanks for addressing the feedback!

PR #1042, opened 2025-09-02
[OSIDB-4436] Remove `HTTPS_PROXY` environment variable

MrMarble:
`HTTPS_PROXY` variable is directly read by `requests` so every network request made with this library (even third dependencies) goes through this proxy, this broke the `hvac` library used to connect to Hashicorp Vault, as our private instance is not accessible through the proxy, blocking local development.

Since this variable was initially added to access Jira stage, we also had `HTTPS_TASKMAN_PROXY` with the same purpose, I decided to remove `HTTPS_PROXY` so is not globally applied and used the `HTTPS_TASKMAN_PROXY` to be manually added to the Jira client so that one is the one using it

Closes OSIDB-4436


costaconrado (approved):
LGTM

docs/developer/DEVELOP.md
```diff
@@ -82,12 +82,8 @@ TRACKERS_SYNC_TO_JIRA=1
 
 # Run taskman tests behind a proxy (optional)
 # This variable is only necessary if rewriting the taskman cassettes locally and using the Stage Red Hat JIRA instance, which requires a proxy to be accessed.
-# Some tests apply the HTTPS_PROXY variable hardcoded in respective conftest files.
 HTTPS_TASKMAN_PROXY="http://foo.bar"
```
JakubFrejlach, line 85: The naming is a bit confusing now as the first sentence inclines it is only for taskman tests but second sentence also mentions Jira. Maybe we should really just name `HTTPS_JIRA_PROXY` so it feels more general? I don't have a strong preference but somehow this feels weird :smile:

JakubFrejlach (approved):
I think the naming is now a bit confusing, see my comment, but otherwise LGTM. Leaving it up to you.

JakubFrejlach:
@MrMarble also please don't forget to change the example `env_files` in OPs repository afterwards.

PR #1043, opened 2025-09-04
Adapt integration tests to use affects v2 and v2 API

Jincxz:
For this PR I have adjusted the integration tests to use affects v2 and the associated v2 API endpoints. 

I have changed the update function since I was running into an issue saving the affects. The original logic was that the affects were removed from the trackers and the respective change would be propagated to the affects. The new logic does the removal from the affect side.

Closes OSIDB-4041

### DIFF
--- a/apps/taskman/tests/conftest.py
+++ b/apps/taskman/tests/conftest.py
@@ -3,7 +3,7 @@ import uuid
 import pytest
 from django.conf import settings
 
-from osidb.constants import OSIDB_API_VERSION
+from osidb.constants import OSIDB_API_VERSION, OSIDB_API_VERSION_NEXT
 from taskman.constants import TASKMAN_API_VERSION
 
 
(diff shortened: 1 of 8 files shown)


apps/taskman/tests/conftest.py
```diff
@@ -77,5 +77,10 @@ def osidb_api_version():
 
 
 @pytest.fixture
-def test_osidb_api_uri(test_osidb_scheme_host, osidb_api_version):
-    return f"{test_osidb_scheme_host}/api/{osidb_api_version}"
+def test_osidb_api_uri(test_osidb_scheme_host):
+    return f"{test_osidb_scheme_host}/api/{OSIDB_API_VERSION}"
```
osoukup, line 81: I assume that `osidb_api_version` mixing is now unused and could be removed.

osoukup (approved):
LGTM. Good job!

osidb/serializer.py
```diff
@@ -839,19 +839,30 @@ def update(self, tracker, validated_data):
...
+        for affect in removed_affects:
+            affect.tracker = None
+            affect.save()
+            affect.refresh_from_db()
+
+        # Add tracker to newly associated affects
+        for affect in added_affects:
             affect.tracker = tracker
             affect.save()
+            affect.refresh_from_db()
```
dmonzonis, line 858: Why do we need `refresh_from_db` here?

Jincxz: I was doing some testing to check if the trackers for the affects were being updated correctly. Had to refresh since the affects were not synced when I did the check. I don't believe that the refresh is actually necessary for saving the affects. Thoughts on removing them?

dmonzonis: `refresh_from_db` performs a db query as it is essentially the same as doing `Affect.objects.get(pk=affect.pk)`, so it's best if we can avoid the unnecessary calls.

PR #1044, opened 2025-09-08
[OSIDB-4306] Fix `owner` field update from Jira not reflected in OSIDB

MrMarble:
Owner  was missing in the list of fields to synchronize from Jira

Closes OSIDB-4306


osoukup (approved):
LGTM

PR #1048, opened 2025-09-08
Adapt trackers module to affects v2

dmonzonis:
The trackers module has been adapted to the use of affects v2. The changes in the API have been as minimal as possible to avoid breaking changes, so the format of the API calls and responses is the same.

The biggest change is that before the 'modules_components' object was implied to be unique for every module and affect, but with affects v2, there can be many affects for a single module, for each of its streams.

Closes OSIDB-4049.

### DIFF
--- a/osidb/migrations/0203_remove_flaw_download_manager.py
+++ b/osidb/migrations/0203_remove_flaw_download_manager.py
@@ -8,7 +8,7 @@ import pgtrigger.migrations
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('osidb', '0200_affect_v1_view'),
+        ('osidb', '0202_affect_v1_view'),
     ]
 
     operations = [


apps/trackers/product_definition_handlers/base.py
```diff
@@ -18,44 +18,50 @@ def __init__(self) -> None:
...
 
-        active_ps_update_streams = ps_module.active_ps_update_streams.all()
-        if exclude_existing_trackers:
-            active_ps_update_streams = active_ps_update_streams.exclude(
-                name__in=affect.trackers.values_list("ps_update_stream", flat=True)
-            )
+        if (
+            exclude_existing_trackers
+            and affect.tracker is not None
+            and affect.tracker.ps_update_stream == ps_update_stream.name
```
osoukup, line 35: Can there be a tracker linked to an affect but have a different PS update stream?

apps/trackers/api.py
```diff
@@ -108,20 +105,25 @@ def post(self, request, *args, **kwargs):
...
-            }
+
+            if key in targets and offer is not None:
+                targets[key]["streams"].append(offer)
+            else:
+                targets[key] = {
+                    "affect": affect,
+                    "ps_module": affect.ps_module,
+                    "ps_component": affect.ps_component,
+                    "streams": [offer] if offer is not None else [],
```
osoukup, line 123: As an affect is on a stream level, I do not think that this should be left in plural. I know that you probably went the way of less breaking change, when we are changing it anyway I would be for making it as precise as possible.

apps/trackers/product_definition_handlers/base.py
osoukup: I am thinking about it and I think the equality should simply be asserted by a validation. It is a matter of linking between affects and trackers so I would really consider affect with stream 1 linked to a tracker with stream 2 as seriously invalid data. It could have a bad consequences like that affect could seem fixed while the tracker was actually fixing a different stream etc.

I can imagine a situation when somebody modifies an affect stream when it has already a tracker linked. I think that the way the linking works would not result in a de-link. Therefore there should be hard blocking validation. If someone manually modifies the tracker and it syncs back, it should automatically de-link.

apps/trackers/api.py
dmonzonis: Yeah my goal was to not add a breaking change. But if we are going to add a v2 for this endpoint too, then I will also make some other changes to simplify the API response further.

apps/trackers/product_definition_handlers/base.py
dmonzonis: We already have `_validate_tracker_affect_ps_update_stream` in `Tracker` which checks just that, so this might be redundant. I will study if this can be safely removed.

dmonzonis:
Note that the openapi meta check is failing because the serializer for the v1 tracker suggestion API had its serializer renamed, but it has not been otherwise modified.

PR #1049, opened 2025-09-09
Fix not affected justifications from Jira not being collected by OSIDB 

dmonzonis:
The ID of the field has changed within Jira at some point. This commit updates the ID in OSIDB so that it is correctly collected from Jira.

Closes OSIDB-4418.


dmonzonis:
@osoukup the issue that was used in the original convertor test (https://issues.stage.redhat.com/rest/api/latest/issue/RHEL-59004) and which I have just updated to use the new field ID is a good example

PR #1051, opened 2025-09-09
[OSIDB-4440] Remove cve_description validation for major incident

MrMarble:


MrMarble:
> Would `_validate_major_incident_fields` need to be changed too since it has the same blocking logic?

Oh good catch! I missed that one

PR #1056, opened 2025-09-15
Affects v2 rebase

dmonzonis:


osoukup (approved):
There is so much stuff that I am not sure it can be reasonably reviewed :sweat_smile: I had a quick look at the commits and the changes without focussing on any details and have not found anything suspicious. LGTM

PR #1063, opened 2025-09-18
Add OSIDB version update of uv.lock to update_release.sh

Jincxz:
The OSIDB version inside of uv.lock was not updated when running the update_release.sh script. The OSIDB version would  only update from pyproject.toml after running uv lock/sync (or manually changed).

This PR adds the missing uv.lock update to the version update script.


JakubFrejlach (commented):
Can we just update this via `uv lock` as it is supposed to be? :thinking: AFAIK `uv lock` won't automatically touch any package which already adheres to the `pyproject.toml` so e.g. if we have package `foo` on version `1.0.0` and defined in pyproject.toml as `foo>=1.0.0` and `1.2.0` is available, it won't trigger the update as it already satisfies it unless forced to. And for other scenarios (like `bar<=x.x.x`, `bar<x.x.x` or `bar==x.x.x` this will be another safeguard that our pyproject.toml actually corresponds to the uv.lock. I just don't like manually editing files which are supposed to be generated :smile:

PR #1066, opened 2025-09-23
[OSIDB-4296] Adjust `_validate_future_unembargo_date` logic

Jincxz:
The flaw validations `_validate_future_unembargo_date` and `_validate_flaw_without_affect` creates a deadlock in this scenario:
- The set unembargo date is in the past but the flaw is still embargoed.
- The flaw has no affects but is in a workflow state that requires affects.

Users would not be able to update either one individually since the other validation error would block them. OSIM does the updates individually starting with the affects and then the flaw and so a deadlock occurs even when both validations are technically corrected on the UI.

As a fix I have adjusted `_validate_future_unembargo_date` so that it will only perform the validation if the `unembargo_dt` has changed. This should allow other validations to pass without `_validate_future_unembargo_date` blocking them.

Closes OSIDB-4296


osidb/api_views.py
```diff
@@ -1144,7 +1145,23 @@ def bulk_post(self, request, *args, **kwargs):
 
         # Third, proxy the update to Bugzilla
         flaw = Flaw.objects.get(uuid=next(iter(flaws)))
-        flaw.save(bz_api_key=bz_api_key)
+
+        # Temporarily disables the _validate_future_unembargo_date validation
+        # to resolve a deadlock that happens when there flaw has no affects (for certain workflows).
```
osoukup, line 1150: nitpick: there sounds a bit weird in this sentence, maybe the flaw ?

dmonzonis (approved):

PR #1067, opened 2025-09-24
[OSIDB-3552] Remove deprecated field and models

Jincxz:
Deprecated fields and models have been removed for the major release.

Removed:
-  `FlawComment`'s `order` 
- `PsProduct`'s `team`
- some models in `package_versions.py`

I have also refactored the deprecated Bugzilla CC code and swapped them to use the new classes directly (not sure if there is a ticket associated with this? Just thought it would be a good place to do the refactoring).

Please let me know if there are any other deprecated parts that I might have missed. Thank you.

Closes OSIDB-3552


Elkasitu (approved):

Elkasitu:
> For the `order` removal, I am unsure how to handle the TODO in `comment.py` [(L73-L81)](https://github.com/RedHatProductSecurity/osidb/blob/d361f540db42dc21cc9919967758bf333e53d4b5/osidb/models/flaw/comment.py#L73-L81). Any suggestions?

Since we want the order to be reproducible, we want the unique constraint to be the same as the fields in `Meta.ordering`, so it should be `('flaw', 'external_system_id', 'uuid', 'created_dt')`... I would remove `uuid` from ordering and the constraint as uuid4s are not sortable.

I would double-check that comment ordering is kept intact, the `order` field was not formerly deprecated so older bugzilla comments' order may change.

Jincxz:
@Elkasitu any suggestions on how to go about checking the ordering? For example, would it make sense to go and scan through old Bugzilla flaw on OSIM for changes (after this PR gets deployed)?

Elkasitu:
> @Elkasitu any suggestions on how to go about checking the ordering? For example, would it make sense to go and scan through old Bugzilla flaw on OSIM for changes after this PR gets deployed?

Yeah, just take a few samples from stage/prod and check if the order would change given the changes introduced by this PR, if it is we can discuss whether it's a big deal or not

PR #1068, opened 2025-09-24
Unblock edits on flaws w/o affects

superbuggy:
With this PR, OSIDB would permit:
1. Flaws to be promoted past Triage without affects (not sure if this makes sense)
2. Changing CVSS score, impact, text field, etc
3. Instead raising a validation error, an alert is created for Flaws without affects, regardless of workflow state.

Closes OSIDB-4463
Closes OSIDB-4310

### DIFF
--- a/apps/workflows/workflows/default.yml
+++ b/apps/workflows/workflows/default.yml
@@ -29,7 +29,7 @@ states:
     jira_state: To Do
     jira_resolution: null
     requirements:
-      - has affects
+      # - has affects
       - has source
       - has title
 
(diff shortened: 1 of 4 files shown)


Elkasitu (commented):
From the ticket:

> There is a another validation on the Done state that would need to be reviewed which is that in some cases a flaw must have a tracker. If there are no affects, there will be no trackers.

Did you consider this scenario? I would perhaps add a test for which a Flaw without affects gets to the DONE state without issues, though the ticket is vague about which conditions trigger such a validation so I would dig deeper into that.

PR #1070, opened 2025-09-25
OSIDB-4481: Add ability to revert/reset workflows

Elkasitu:
This PR introduces two main methods (and a few helper ones) to the
WorkflowModel class:

* revert(), which does the opposite of promote(), meaning it will go one
  step back in the workflow state machine
* reset(), which regardless of the current workflow will change the
  workflow to the default one and the state to the initial state

These are both mechanisms for cases in which analysts mistakenly
rejected or promoted and need to be reworked on in a previous state.

Closes OSIDB-4481

### DIFF
--- a/apps/workflows/exceptions.py
+++ b/apps/workflows/exceptions.py
@@ -15,6 +15,10 @@ class LastStateException(WorkflowsException):
     """exception for trying to promote further when in the last possible state"""
 
 
+class InitialStateException(WorkflowsException):
+    """exception for trying to revert further when in the first possible state"""
+
+
 class MissingRequirementsException(WorkflowsException):
     """exception for trying to change state without requirements"""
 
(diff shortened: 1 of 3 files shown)


JakubFrejlach (approved):
LGTM!

One question about the internal/public ACLs. I suppose that it is not worth to bother with reverting those as once Flaw was flipped from internal to public it was eventually published via SDEngine and reverting the ACLs seems pointless then?

Elkasitu:
> One question about the internal/public ACLs. I suppose that it is not worth to bother with reverting those as once Flaw was flipped from internal to public it was eventually published via SDEngine and reverting the ACLs seems pointless then?

It can be done if necessary, not sure if worth it

osoukup (approved):
I like this. Well done!

PR #1071, opened 2025-09-25
Update closed trackers if unembargoed

superbuggy:
When closed trackers are unembargoed, with this PR their security levels will be updated instead of being passed over due to being closed.

Closes OSIDB-3729

### DIFF
--- a/apps/trackers/jira/query.py
+++ b/apps/trackers/jira/query.py
@@ -445,6 +445,10 @@ class OldTrackerJiraQueryBuilder(TrackerQueryBuilder):
                 {"name": self.ps_update_stream.version}
             ]
 
+    def generate_only_security(self):
+        self.generate_base()
+        self.generate_security()
+
     def generate_security(self):
         """
         generate the appropriate security level for restricting who can access the Jira
@@ -904,6 +908,21 @@ class TrackerJiraQueryBuilder(OldTrackerJiraQueryBuilder):
             )
 
         self._query["fields"][field_id] = {"value": choice_str}
+
+    def generate_only_embargo_and_security_level(self):
+        """
+        Generate only the embargo status and security level fields.
+
+        The overwriting of _query, which initializes to None,
+        prevents class TrackerQueryBuilder's accessor method `query` from
+        generating the entire query, otherwise it will call `generate`
+        if `_query` is None.
+        """
+
+        self.generate_base()
+        self._query["fields"]["security"] = None
+        self.generate_embargo_status()
+        self.generate_security()
 
     def generate_special_handling(self):
         field_name = "Special Handling"
--- a/apps/trackers/jira/save.py
+++ b/apps/trackers/jira/save.py
@@ -85,8 +85,16 @@ class TrackerJiraSaver(JiraQuerier):
         """
         update an existing representation of tracker model in Jira
         """
-        builder = self.get_builder()
-        query = builder(tracker).query
+        builder = self.get_builder()(tracker)
+
+        if tracker.is_closed:
+            if isinstance(builder, TrackerJiraQueryBuilder):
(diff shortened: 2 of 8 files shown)


Jincxz (changes requested):
LGTM. A suggestion: I think there can be tests written (probably in the `endpoints` directory) to test that we can un-embargo closed trackers.

JakubFrejlach (changes requested):
I don't think this approach will achieve the desired outcome.

Jincxz:
It looks like the JIRA convertor translates the security field to a LDAP group [(link to code)](https://github.com/RedHatProductSecurity/osidb/blob/bc9753786644eeaa18c8ce019280d4953f5d7896/collectors/jiraffe/convertors.py#L445) which I believe dictates whether the tracker is embargoed. The JIRA tracker can be converted into a OSIDB Tracker and tested as such [(example)](https://github.com/RedHatProductSecurity/osidb/blob/bc9753786644eeaa18c8ce019280d4953f5d7896/collectors/jiraffe/tests/test_convertors.py#L62).

osidb/serializer.py
```diff
@@ -1283,9 +1283,11 @@ def update_trackers(self, old_affect, new_affect):
             promote = True  # remember to only regenarate Jira trackers
 
         for tracker in new_affect.trackers.all():
-            # no tracker updates for the closed ones
-            # because we consider these already done
-            if tracker.is_closed:
+            # Update security levels on closed tracker on unembargo
+            is_unembargoing = (
+                old_affect.flaw.is_embargoed and not new_affect.flaw.is_embargoed
```
JakubFrejlach: Will this actually work? :thinking: You have two instances of affect, e.g. the `old_affect` and `new_affect` but you are actually accessing its parent `flaw` only in this function, so for both `old_affect` and `new_affect` the `flaw` is queried just now and thus this will probably almost evaluate to False? :thinking:

superbuggy: It seemed to work in the testing I did, but probably worth looking at more closely to make sure it is doing what I thought it was doing

apps/trackers/jira/save.py
```diff
@@ -86,6 +86,11 @@ def update(self, tracker):
         update an existing representation of tracker model in Jira
         """
         builder = self.get_builder()
+        if tracker.is_closed:
+            # prevent query from being generated with non-security level fields
+            builder._query["fields"] = {"security": None}
```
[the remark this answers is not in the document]
superbuggy: https://github.com/RedHatProductSecurity/osidb/blob/bc9753786644eeaa18c8ce019280d4953f5d7896/apps/trackers/jira/query.py#L199-L205

Correct, it is initialized with `None` and then used as dictionary elsewhere. It is used to restrict how the query is built by avoiding calling `generate` which populates other query fields. 

https://github.com/RedHatProductSecurity/osidb/blob/2b8e92b6e0de857dbb0a7f7a3e6cae0f8ec08375/apps/trackers/common.py#L25-L30

It's a little awkward, and now that you mention it, it does feel to me like this should be handled in a dedicated method in the query building class...

apps/trackers/jira/query.py
```diff
@@ -868,6 +874,21 @@ def generate_embargo_status(self):
...
+        Generate only the embargo status and security level fields.
+
+        The overwriting of _query, which initializes to None,
+        prevents class TrackerQueryBuilder's accessor method `query` from
+        generating the entire query, otherwise it will call `generate`
+        if `_query` is None.
+        """
+
+        self.generate_base()
+        self._query["fields"]["security"] = None
```
JakubFrejlach: This is not necessary since you are now generating the base on the previous line,  `generate_security` will eventually se this if necessary.

PR #1073, opened 2025-09-29
Add denormalized `labels` to `Affect` model

MrMarble:
Users want to see the Labels in the new Affect table in OSIM, this was outside of the scope of the initial implementation of the labels, to prevent extra queries to the database I opted for denormalizing the labels to add them directly as a field in the Affect model.

I've not implemented the update of the labels if the `FlawLabel` changes, as this is the same behavior for `FlawCollaborator` Model. Only if the Affect ps_component/ps_module is updated the labels are regenerated.

I've added a migration to populate the labels for all the affects.

Closes OSIDB-4474


Elkasitu (approved):
Could we get a test to check that the labels are being correctly populated?

PR #1074, opened 2025-09-29
Group v1 affects by flaw, module and component

dmonzonis:
Before this commit the v1 affects which were fabricated from v2 affects were grouped by flaw and module. This leads to a few missing affects in the v1 endpoint when a flaw has more than one affect for the same module, but a different component.


Jincxz (approved):

osoukup (approved):
LGTM

JakubFrejlach (approved):

PR #1076, opened 2025-09-29
[OSIDB-3766] Allow case insensitive CVE ID flaw queries

Jincxz:
This PR adds a new helper function to match string with CVE ID regex in a case insensitive manner.

The function is used to make CVE ID queries for flaws case insensitive.

Closes OSIDB-3766

### DIFF
--- a/apps/workflows/api.py
+++ b/apps/workflows/api.py
@@ -13,10 +13,10 @@ from rest_framework.viewsets import ModelViewSet
 
 from apps.taskman.service import JiraTaskmanQuerier
 from osidb.api_views import RudimentaryUserPathLoggingMixin, get_valid_http_methods
-from osidb.helpers import get_bugzilla_api_key, get_jira_api_key
+from osidb.helpers import get_bugzilla_api_key, get_flaw_or_404, get_jira_api_key
 
 from .exceptions import WorkflowsException
-from .helpers import get_flaw_or_404, str2bool
+from .helpers import str2bool
 from .serializers import (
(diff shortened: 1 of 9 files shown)


osoukup (approved):
LGTM

Elkasitu (commented):
I'm OK with this approach but I think we could make this easier for everyone:

* We often want to fetch a Flaw by PK but we don't care whether it's UUID or CVE id
* If by CVE id, we want it to be accepted regardless of case
* We already have a `get_flaw_or_404` shortcut

Given these facts, I think we could introduce a method like `get_by_identifier` at the `FlawManager` level, so wherever we need to lookup a Flaw by cve_id or uuid we can just pass whatever we have and this method will handle case insensitivity.

```python
# Return obj (assuming obj exists)
Flaw.objects.get_by_identifier("cve-2025-1111")
Flaw.objects.get_by_identifier("CVE-2024-9999")
Flaw.objects.get_by_identifier("d9252120-fe84-4bed-8e52-9cec097b10d5")
# Raises Flaw.DoesNotExist
Flaw.objects.get_by_identifier("CVEE-2025-1111")  # malformed
Flaw.objects.get_by_identifier("foo")  # not a valid identifier
```

We can then simplify `get_flaw_or_404` to simply call `Flaw.objects.get_by_identifier` within a `try/except` and change the callsites of `get_object_or_404` for flaws by `get_flaw_or_404`.

The SDEngine endpoint is the exception, but for that one we can just add `.upper()` and call it a day.

WDYT?

PR #1079, opened 2025-09-30
[OSIDB-4313] Use a field for `is_embargoed`

MrMarble:
The initial idea of this task was to change the computed `is_embargoed` field to a denormalized field in order to reduce database queries. We discussed that the best approach for this should be a [generated field](https://docs.djangoproject.com/en/5.2/ref/models/fields/#generatedfield), but for that we need to update to Django 5.2 which drops support on Postgres 13.

I managed to work around this by utilizing the `embargoed` annotation in the `ACLMixin`, the problem was that the serializer for the `EmbargoedField` was always relying on the computed `is_embargoed` even when the annotation existed, I added a check to use the annotation when available and that made some great improvements.

While testing I found another issue that was easily fixed so I went for it. Since Alerts use a `GenericForeignKey` relation, Django needs to query the `ContentType` of the object that the alert refers to, by prefething this query we where able to reduce the queries as well.

General query count reductions:
  - Flaw list: 84 → 72 queries (-14%)
  - Flaw list filtered: 61 → 58 queries (-5%)
  - Empty flaw: 59 → 58 queries (-2%)
  - Flaw with affects: 67 → 63 queries (-6%)
  - Flaw with affects history: 68 → 64 queries (-6%)
  - Affect list: 60 → 57 queries (-5%)

Closes OSIDB-4313

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Adjust `unembargo_dt` validation for embargoed flaws to only trigger when 
   `unembargo_dt` is changed (OSIDB-4296)
+- Improve performance in Alerts and ACL serialization (OSIDB-4313)
 
 ## [4.16.0] - 2025-09-16
 ### Fixed


osoukup (approved):
Nice!

PR #1080, opened 2025-10-01
OSIDB-4487: Add API endpoints for reverting and resetting flaw workflow

Elkasitu:
Closes OSIDB-4487

### DIFF
--- a/apps/workflows/api.py
+++ b/apps/workflows/api.py
@@ -27,6 +27,22 @@ from .workflow import WorkflowFramework
 logger = logging.getLogger(__name__)
 
 
+jira_api_key_header = OpenApiParameter(
+    name="Jira-Api-Key",
+    type=str,
+    location=OpenApiParameter.HEADER,
+    description="User generated api key for Jira authentication.",
+)
+
+
+bz_api_key_header = OpenApiParameter(
+    name="Bugzilla-Api-Key",
+    type=str,
+    location=OpenApiParameter.HEADER,
+    description="User generated api key for Bugzilla authentication.",
+)
+
+
 class index(RudimentaryUserPathLoggingMixin, APIView):
     """index API endpoint"""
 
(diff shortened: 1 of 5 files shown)


apps/workflows/api.py
```diff
@@ -128,6 +126,88 @@ def post(self, request, flaw_id):
...
+                description="User generated api key for Jira authentication.",
+            ),
+            OpenApiParameter(
+                name="Bugzilla-Api-Key",
+                type=str,
+                location=OpenApiParameter.HEADER,
+                description="User generated api key for Bugzilla authentication.",
+            ),
+        ]
+    )
```
JakubFrejlach, line 145: I see this kind of OpenApi parameters already used several places in this file, can we do something similar as in [`osidb/api_views.py`](https://github.com/RedHatProductSecurity/osidb/blob/master/osidb/api_views.py#L382)?

apps/workflows/api.py
```diff
@@ -128,6 +126,88 @@ def post(self, request, flaw_id):
             return Response({"errors": str(e)}, status=status.HTTP_409_CONFLICT)
 
 
+class RevertWorkflow(RudimentaryUserPathLoggingMixin, APIView):
```
JakubFrejlach, line 129: Should we rename existing API views here to match these newly introduced? :thinking: 
`promote` -> `PromoteWorkflow`
`reject` -> `RejectWorkflow`

apps/workflows/api.py
```diff
@@ -128,6 +126,88 @@ def post(self, request, flaw_id):
...
+                description="User generated api key for Jira authentication.",
+            ),
+            OpenApiParameter(
+                name="Bugzilla-Api-Key",
+                type=str,
+                location=OpenApiParameter.HEADER,
+                description="User generated api key for Bugzilla authentication.",
+            ),
+        ]
+    )
```
JakubFrejlach, line 186: Same as above.

osoukup (approved):
LGTM

PR #1086, opened 2025-10-07
Tighten public logic on workflow change

superbuggy:
Closes OSIDB-4446

As far I can tell, the current logic in `def adjust_acls`  seems ambiguous enough to account for flaws unexpectedly becoming public in some scenarios (where `workflow_state` would be momentarily an unexpected value, probably during a sync task), but I'm still unclear on how `workflow_state` would be `None` or `""` in the cases documented in the ticket.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -432,7 +432,7 @@
         "filename": "osidb/tests/endpoints/test_endpoints.py",
         "hashed_secret": "3c3b274d119ff5a5ec6c1e215c1cb794d9973ac1",
         "is_verified": false,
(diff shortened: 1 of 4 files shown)


osoukup (approved):
This seems to me as a good approach. I agree that explicitly enumerating the public cases  and the positive condition should avoid the side effects of non-public cases with negative condition. LGTM

Elkasitu (approved):
Could you add a test? Basically a Flaw in workflow_state `NOVALUE` after calling `adjust_acls` should remain in internal state

PR #1246, opened 2026-04-10
OSIDB-4878: Automate package discovery and affect creation

Elkasitu:
This experimental feature provisions `newcli` within OSIDB in certain environments.

Whenever a new Flaw is assigned components or its components are changed, newcli is invoked asynchronously in order to find affected packages and streams and then the relevant affects are created and added to the Flaw automatically.

As of this time, the feature is limited to the hummingbird-1 ps_module but can be easily expanded to other ps_modules and enabled/disabled independently thanks to feature flags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,25 @@ RUN uv sync --frozen --no-dev && \
     rm -f /opt/app-root/src/pyproject.toml && \ 
     rm -f /opt/app-root/src/uv.lock
 
+# Optional: GitLab HTTPS clone URL for the newtopia-cli monorepo; when set, install in-repo
+# deptopia_client first, then newtopia_cli (PyPI has no deptopia_client).
+# Optional: branch, tag, or commit for both installs (omit for the remote default branch).
+ARG NEWCLI_REPO_URL=""
+ARG NEWCLI_EXPERIMENTAL_BRANCH=""
+RUN if [ -n "$NEWCLI_REPO_URL" ]; then \
+        if [ -n "$NEWCLI_EXPERIMENTAL_BRANCH" ]; then \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}@${NEWCLI_EXPERIMENTAL_BRANCH}#subdirectory=python/deptopia-client" && \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}@${NEWCLI_EXPERIMENTAL_BRANCH}#egg=newtopia_cli&subdirectory=python/newtopia_cli"; \
+        else \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}#subdirectory=python/deptopia-client" && \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}#egg=newtopia_cli&subdirectory=python/newtopia_cli"; \
+        fi; \
+    fi
+
 # Copy the project into the image
 COPY . /opt/app-root/src
 
--- a/osidb/models/affect.py
+++ b/osidb/models/affect.py
@@ -10,6 +10,7 @@ from django.db import models
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from psqlextra.fields import HStoreField
+from pydantic import Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from apps.bbsync.constants import RHSCL_BTS_KEY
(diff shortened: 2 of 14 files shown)


Dockerfile
```diff
@@ -62,6 +62,25 @@ RUN uv sync --frozen --no-dev && \
...
+                "git+${NEWCLI_REPO_URL}@${NEWCLI_EXPERIMENTAL_BRANCH}#subdirectory=python/deptopia-client" && \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}@${NEWCLI_EXPERIMENTAL_BRANCH}#egg=newtopia_cli&subdirectory=python/newtopia_cli"; \
+        else \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}#subdirectory=python/deptopia-client" && \
+            uv pip install --no-cache --python "$UV_PROJECT_ENVIRONMENT/bin/python" \
+                "git+${NEWCLI_REPO_URL}#egg=newtopia_cli&subdirectory=python/newtopia_cli"; \
+        fi; \
+    fi
```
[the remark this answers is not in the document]
Elkasitu: Not an issue as these are not a secret per-se, just not generally available from outside Red Hat

osidb/tasks.py
```diff
@@ -21,6 +27,92 @@
...
 
+def _newcli_include_modules_csv() -> str:
+    return ",".join(AffectSettings().auto_create_ps_modules)
+
+
+def _flaw_components_for_newcli(flaw: Flaw) -> list[str]:
+    components = [c for c in (flaw.components or []) if c and str(c).strip()]
+    if not components:
+        raise ValueError(f"Flaw {flaw.uuid} has no non-empty components for newcli -s")
+    return components
```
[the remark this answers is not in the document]
Elkasitu: Not really an issue, `newcli` either doesn't care or does the stripping itself, so just checking for emptiness is good enough.

osidb/tasks.py
```diff
@@ -21,6 +27,92 @@
...
+                ps_update_stream=ps_update_stream,
+                purl=purl,
+                acl_read=flaw.acl_read,
+                acl_write=flaw.acl_write,
+                affectedness=Affect.AffectAffectedness.NEW,
+                resolution=Affect.AffectResolution.NOVALUE,
+                impact=flaw.impact or Impact.NOVALUE,
+            )
+            affect.save(raise_validation_error=False)
+            created += 1
```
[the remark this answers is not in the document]
Elkasitu: Not really an issue, `ps_component` is inferred from `purl` and `ps_module` is denormalized through Django signals.

osoukup (approved):
Under the time-sensitive nature of the Hummingbird this LGTM but I think that the long-term solution (when there is Atlas) should rather be a standalone micro-service.

dmonzonis (approved):

osidb/models/affect.py
```diff
@@ -42,6 +43,10 @@ class AffectSettings(BaseSettings):
     model_config = SettingsConfigDict(env_prefix="OSIDB_AFFECTS_")
 
     require_purl_for_middleware: bool = False
+    auto_create: bool = False
+    auto_create_ps_modules: list[str] = Field(
+        default_factory=lambda: ["hummingbird-1"],
+    )
```
JakubFrejlach, line 50: Should we make this more dynamic to cover any new possible Hummingbird module? We could get it from product definitions based on the ps_module which do have a bts_key `HUM` but it would need to be done elsewhere as that is only discoverable during runtime.

osidb/signals.py
```diff
@@ -113,6 +115,49 @@ def populate_cvss_score(sender, instance, **kwargs):
...
+        components_changed = True
+    else:
+        old_list = list(Flaw.objects.get(pk=instance.pk).components or [])
+        components_changed = old_list != new_list
+
+    if not components_changed:
+        return
+
+    if not any(c and str(c).strip() for c in new_list):
+        return
```
JakubFrejlach, line 150: We probably should have done for `Flaw.components` to only allow non empty strings there.

osidb/tasks.py
```diff
@@ -21,6 +27,97 @@
...
+
+            affect = Affect(
+                flaw=flaw,
+                ps_update_stream=ps_update_stream,
+                purl=purl_str,
+                acl_read=flaw.acl_read,
+                acl_write=flaw.acl_write,
+                affectedness=Affect.AffectAffectedness.AFFECTED,
+                resolution=Affect.AffectResolution.DELEGATED,
+                impact=flaw.impact or Impact.NOVALUE,
```
JakubFrejlach: `Flaw.impact` has values restricted to `Impact` values, right? So the default `Impact.NOVALUE` should not be needed? 🤔

JakubFrejlach (approved):
Couple of thoughts but otherwise LGTM!

osidb/models/affect.py
Elkasitu: I can definitely keep this in mind when developing this feature as a microservice, but I think there's no need to do it in OSIDB at this stage.

That said I'm skeptical about tying it to `bts_key`, basing it off of `PsProduct` might work better, but then again this provides enough granularity in case we didn't want to auto-create affects for all ps_modules tied to a product.

osidb/signals.py
Elkasitu: We probably can't now without breaking API, we might consider accepting empty strings and whitespace-padded strings but converting them on save to either be stripped or not stored at all if empty.

PR #1247, opened 2026-04-15
OSIDB-4910: Add schema validations to PURLs

MrMarble:
`packageurl-python` python package only validated purl format, not the actual schema, this differed from the javascript version of the library used in OSIM, so users could create invalid purls in OSIDB by  the API that then would fail to validate in OSIM

Updated to the latest version, `validate` method was introduced in `0.17.6` (December 2025), latest version at the time of writing.

Closes OSIDB-4910


osoukup (approved):
LGTM

osidb/models/fields.py
```diff
@@ -64,19 +64,33 @@ def deconstruct(self):
         return name, path, args, kwargs
 
 
-def to_purl(value: Optional[str | PackageURL]) -> Optional[PackageURL]:
+def validate_purl(purl: PackageURL) -> None:
+    """Validate a PackageURL against its type-specific schema rules."""
+    errors = purl.validate(strict=True)
```
MrMarble: `strict` also validates qualifiers, I've enabled it but I'm not sure if we want to allow arbitrary qualifiers?

osoukup: Ideally not, I hope we do not need anything custom, just standard qualifiers.

MrMarble: Well, I disabled it, `rpmmod` is not a valid qualifier per the spec, so that would give us problems.

Since the library does not raise the error directly but rather return a list of individual issues we could have some logic to ignore some specific qualifiers but I think that goes out of the scope of this task as complicates quite a bit the logic

Elkasitu: Definitely not in-scope for this PR, but we could simply [add it to the spec](https://github.com/package-url/purl-spec/blob/c756cacf766d4bf2711b248b935b3b80d1b1ba2e/types/rpm-definition.json#L28)? Other vendors [seem to be doing it](https://github.com/package-url/purl-spec/pull/818), though it seems that the people responsible for the spec are unable to handle the workload on their plate at the moment, so not sure when it would become part of the spec.

osidb/models/fields.py
```diff
@@ -121,17 +135,20 @@ def get_prep_value(self, value: Optional[str | PackageURL]) -> str:
...
+            validate_purl(purl)
+            return purl.to_string()
         except ValueError as e:
             raise ValidationError(f"Invalid PURL: {e}")
 
     def from_db_value(
         self, value: Optional[str], expression, connection
     ) -> Optional[PackageURL]:
-        return to_purl(value)
+        return to_purl(value, validate=False)
```
Elkasitu, line 154: why do we not validate here? is it just being conservative or should we fix existing data in OSIDB? maybe a comment would help

Elkasitu (approved):
LGTM overall

osidb/models/fields.py
MrMarble: Hey! Sorry for some reason your comment didn't load and I saw the email after the merge...
We probably have invalid purls stored in OSIDB, the cleanup we've been doing was related to Aggregator sync errors, but that only filters public Flaws with CVE, we could have embargoed flaws or flaws without CVE with invalid PURLs, to help with this I only added the validation to the inputs, also not sure what would be the behavior if loading affect with an invalid PURL raises an error.

PR #1248, opened 2026-04-16
OSIDB-4920: Fix JiraProjectFields unique constraint dupliate names

JakubFrejlach:
This PR fixes issue with multiple fields in the same project having the same name which violates unique constraint.


JakubFrejlach:
This issue appears to be currently only problematic for stage, we can possibly workaround it, for Production it is not happening and the proper way to fix it would require a major rework on how we work with the Jira metadata, closing.

PR #1249, opened 2026-04-16
OSIDB-4907: Add index API for flaw ID and updated datetime

Jincxz:
This PR adds a new endpoint to fetch the ID and updated datetime of all of the flaws. The endpoint can be accessed via `api/v2/flaws/index`. 

By default the UUID is returned. The query parameter `id_type` can be set with `cve_id` to return the CVE instead. This is to accommodate the use cases for different services.


jsvob (commented):
If `return Response({"results": list(queryset)})` makes it never use paging, this LGTM based on a cursory review. Didn't try to run it. Thank you!

osidb/api_views.py
```diff
@@ -795,6 +807,24 @@ def create(self, request, *args, **kwargs):
...
+    def index(self, request, *args, **kwargs):
+        """
+        Simple API endpoint to return ID and update time pairs
+        """
+        id_type = request.query_params.get("id_type", "uuid")
+
+        # Default to using UUID unless CVE ID is requested
+        if id_type == "cve_id":
+            queryset = Flaw.objects.exclude(cve_id__isnull=True).values(
+                "cve_id", "updated_dt"
```
osoukup: `updated_dt` is not enough. Either it needs to be changed to `local_updated_dt` or there has to be an option to choose the date field - the first option is probably good enough and definitely simpler. Aggregator needs to know when anything in the flaw context changes, not just flaw. I am not sure it works for stuff like CVSS (if not there is no way to have it complete) but at least for the affects and trackers it should and those are important.

osoukup (changes requested):

osidb/api_views.py
```diff
@@ -795,6 +807,24 @@ def create(self, request, *args, **kwargs):
...
+            )
+        else:
+            queryset = Flaw.objects.values("uuid", "updated_dt")
+
+        return Response({"results": list(queryset)})
```
MrMarble: Currently we have ~105k flaws, `list(queryset)` will load all of them in memory, since there are only two fields is probably fine, have we considere using a `StreamingHttpResponse` and yield the result of the `queryset.iterator()` to not load the whole list in memory?

Its probably fine, I just don't want to create more performance issues in OSIDB, I think this endpoint is going to be called once or twice every minute, so not worth the hassle, just wanted to mention it

Jincxz: I will take a look.

Elkasitu: Another option is to use something like `values_list(flat=True)` which should be about ~8MB of RAM for 100k Flaws

Jincxz: Looked at `StreamingHttpResponse` but it added too much complexity. Decided to use `values_list` as @Elkasitu suggested.

osidb/api_views.py
Jincxz: Updated the date field used. I have also tested the endpoint locally with ~100k flaws. Results came back between 1-3 seconds. Will be looking to test on stage too. 

Does that timing sound sufficient?

osoukup (approved):
LGTM

Elkasitu (approved):

PR #1250, opened 2026-04-23
 OSIDB-4929: Ignore invalid withdrawn osv cves

roduran-dev:
OSV_collector is failing because some CVE's don't have a details field. 
Here are some Cases 
https://rhcorporate.sumologic.com/log-search/create?id=KrKOonAPFb0MuA06kmYGTB4xC4fhOPqVmGBK3aPN

After some investigation all of this cases besides having no `Details`" field had a `withdrawn` field. 

### Resolution
In this [thread](https://redhat-internal.slack.com/archives/C06597XVDQW/p1776943932992759) it was defined that  the item can should be ignored if it has `withdrawn` and comes from **echo** (echohq.com)  

To decrease the possibility of ignoring a valid element it was defined to also add Details field should be empty or missing.  

### Detail: 
The verification of field  of **echo** (echohq.com) was important I had to modify the cassettes url override to add an optional override (so it doesn't override in this case the url) 

Closes OSIDB-4929


collectors/osv/collectors.py
```diff
@@ -148,6 +164,9 @@ def collect(self, osv_id: Union[str, None] = None) -> str:
                 for osv_vuln in self.fetch_osv_vulns_for_ecosystem(ecosystem):
                     try:
                         osv_id, cve_ids, content = self.extract_content(osv_vuln)
+                    except OSVCollectorWithdrawnException as exc:
+                        logger.warning(str(exc))
+                        continue
```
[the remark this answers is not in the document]
roduran-dev: This makes it ignore withdrawn ONLY if they are from echohq and if the details field is an invalid state. Furthermore withdown was not being used until now.

collectors/osv/constants.py
```diff
@@ -4,3 +4,6 @@
 OSV_COLLECTOR_ENABLED = get_env("OSV_COLLECTOR_ENABLED", default="True", is_bool=True)
 
 OSV_START_DATE = get_env_date("OSV_START_DATE", default="2024-07-01")
+
+
+REFERENCES_THAT_CAN_BE_IGNORED = ["echohq.com"]
```
JakubFrejlach: `REFERENCES_THAT_CAN_BE_IGNORED` -> `IGNORABLE_REFERENCES` ?

docs/CHANGELOG.md
```diff
@@ -9,14 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   specific products (OSIDB-4878)
-- Add index API for flaw ID and updated datetime (OSIDB-4738)
-
-### Removed
-- Remove `owner` and `qe_owner` parsing from Jira Tracker and Task download to prevent tracker sync failure(OSIDB-4875)
 
 ### Changed
-- Improve resilience, error-reporting and performance of /bulk create
-  affects endpoint (OSIDB-4748)
+- osv_collector now ignore withdrawn items.
```
JakubFrejlach: Jira issue ID missing

collectors/osv/collectors.py
```diff
@@ -342,3 +358,17 @@ def get_cvss_and_impact(data: dict) -> tuple[list, str]:
...
+
+    def has_reference_to_ignore(self, osv_vuln: dict) -> bool:
+        for ref_to_ignore in REFERENCES_THAT_CAN_BE_IGNORED:
+            for osv_reference in osv_vuln.get("references", [{}]):
+                if ref_to_ignore in osv_reference.get("url", ""):
+                    return True
+        return False
+
+    def is_withdrawn(self, osv_vuln: dict) -> bool:
+        return osv_vuln.get("withdrawn", False)
```
JakubFrejlach: `withdrawn` contains date string, not bool. Technically it works, but it is a bit of a mixture, better would probably be `"withdrawn" in osv_vuln` or some bool conversion

collectors/osv/collectors.py
```diff
@@ -342,3 +358,17 @@ def get_cvss_and_impact(data: dict) -> tuple[list, str]:
...
+
+    def can_be_ignored(self, osv_vuln: dict) -> bool:
+        """Advisories come from some echohq.com site and have been withdrawn can be ignored."""
```
JakubFrejlach: This docstring is very confusing. In `has_reference_to_ignore` list `REFERENCES_THAT_CAN_BE_IGNORED` is being used which implies that the list of the ignorable sites will be extensible yet this docstring specifically talks about `echohq.com`. Also grammar-wise I would rephrase it a bit.

conftest.py
```diff
@@ -130,15 +157,19 @@ def refresh_func():
     return refresh_func
 
 
-@pytest.fixture(scope="session")
-def vcr_config():
+@pytest.fixture(scope="function")
+def vcr_config(request):
```
JakubFrejlach: Why is the scope change needed? 🤔

collectors/osv/collectors.py
```diff
@@ -286,8 +299,11 @@ def get_refs(data: dict) -> list:
             return refs
 
         def get_comment_zero(data: dict) -> str:
-            #  https://ossf.github.io/osv-schema/#summary-details-fields
-            return data.get("details", "")
+            if not (detail := data.get("details", "")) and self.can_be_ignored(
+                osv_vuln
+            ):
+                raise OSVCollectorIgnoredException(osv_vuln["id"])
```
JakubFrejlach: I am skeptical about the check for ignore actually being here. This function used to be pure data extractor and now what it does it can raise exception which needs to be then caught couple of levels up the the trace and creates not really intuitive behavior, would it make more sense to just before the whole process of extracting data process, we perform a check for ignore (which includes presence for details check and ignorable references check and just simply skip there without the need of raising and catching exceptions?

collectors/osv/collectors.py
roduran-dev: I see the point. I didn't went the `"withdrawn" in osv_vuln` because I'm not 100% that withdrawn date could exist without value so I went that way. 

I will convert to bool to leave it better.

collectors/osv/collectors.py
roduran-dev: You're right, I'll move this to pre-check filtering.

I did it this way because the code was breaking when extracting `details`, so handling it at the point of failure seemed natural. But it makes more sense to filter before extraction rather than during it.

conftest.py
roduran-dev: The scope change from `session` to `function` was needed because `vcr_config` now checks for per-test markers (like `@pytest.mark.vcr_override_domain(False)` used in `test_ignore_osv_record_without_details`). Session-scoped fixtures can't access test-specific markers  since they're created once per session, not per test.

This was the only way I could think of to change the behavior for my test as I needed the test to have a specific test. 

The other option was the mock the `has_reference_to_ignore` but IMO it wouldn't as valid because it would be checking the echohq.com case properly as it could be any other domain.

roduran-dev:
@JakubFrejlach Thanks a lot for the review. I asserted your comments. 


To avoid confusion at this point I didn't squash commits but I will a bit later.

collectors/osv/collectors.py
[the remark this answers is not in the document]
roduran-dev: Also. The withdrawn check is new functionality - previously the collector didn't check for withdrawn status at all. The records we're now skipping would have caused collection failures before (missing required fields), so they were never successfully ingested. No reconciliation is needed since this is a forward-looking fix to prevent future collector failures, not a change to how we handle existing data.

collectors/osv/constants.py
roduran-dev: Done.

docs/CHANGELOG.md
roduran-dev: Done

collectors/osv/collectors.py
Elkasitu: According to the [schema](https://ossf.github.io/osv-schema/#withdrawn-field) the presence of the field is enough to confirm that the entry has been withdrawn, so `"withdrawn" in osv_vuln` is likely correct

conftest.py
```diff
@@ -94,28 +94,55 @@ def filter_response(response):
     return response
 
 
+def _vcr_should_override_response_domain(request):
```
Elkasitu: I'm not a big fan of this, a lot of custom logic for a very specific, single edge case.

An alternative to this could be simply monkeypatching `fetch_osv_vuln_by_id` so that the requests.get() returns a pre-determined copy of the payload saved as a JSON file within the tests directory, see test_sync_flaw_affects_from_newcli.py for an example.

Another option would be leveraging dependency injection, we can create a method e.g. `_parse_osv_entry` which accepts the JSON or Python dict and is called by `fetch_osv_vuln_by_id`, and then we can simply write a unit test for `_parse_osv_entry`, no need to mock or record network interactions.

pytest.ini
```diff
@@ -10,6 +10,7 @@ markers =
     enable_signals: enables django signals to run.
     enable_rls: enables row-level-security in the database during testing.
     queryset: marks a database query count test.
+    vcr_override_domain: pass False to keep original domains in recorded response bodies only; request URIs are always normalized (default True).
```
Elkasitu: if we're gonna keep this, I'd follow the existing custom marker naming and do something like `disable_domain_scrubbing` and not take any arguments -- if it's used then we disable domain scrubbing for the body, if it's not then we keep it on

collectors/osv/collectors.py
```diff
@@ -342,3 +357,41 @@ def get_cvss_and_impact(data: dict) -> tuple[list, str]:
...
+    def is_withdrawn(self, osv_vuln: dict) -> bool:
+        return osv_vuln.get("withdrawn", False)
+
+    def can_be_ignored(self, osv_vuln: dict) -> bool:
+        """Check if OSV record can be ignored during collection.
+
+        Records are ignored if they are withdrawn AND contain references
+        from specific sources (see IGNORABLE_REFERENCES).
+        """
+        return self.is_withdrawn(osv_vuln) and self.has_reference_to_ignore(osv_vuln)
```
Elkasitu, line 378: this is not used?

collectors/osv/collectors.py
```diff
@@ -342,3 +357,41 @@ def get_cvss_and_impact(data: dict) -> tuple[list, str]:
...
+        if withdrawn:
+            # Check if details field is missing/empty
+            if not osv_vuln.get("details"):
+                return True, f"withdrawn ({withdrawn}) with no details field"
+
+            # Check if it has ignorable references (e.g., echohq.com)
+            if self.has_reference_to_ignore(osv_vuln):
+                return True, f"withdrawn ({withdrawn}) with ignorable references"
+
+        return False, ""
```
Elkasitu, line 393: this duplicates `is_withdrawn` functionality, either `is_withdrawn` should be removed or it should be used here

collectors/osv/collectors.py
```diff
@@ -286,7 +302,6 @@ def get_refs(data: dict) -> list:
             return refs
 
         def get_comment_zero(data: dict) -> str:
-            #  https://ossf.github.io/osv-schema/#summary-details-fields
```
Elkasitu, line 289: ?

collectors/osv/constants.py
```diff
@@ -4,3 +4,6 @@
 OSV_COLLECTOR_ENABLED = get_env("OSV_COLLECTOR_ENABLED", default="True", is_bool=True)
 
 OSV_START_DATE = get_env_date("OSV_START_DATE", default="2024-07-01")
+
+
+IGNORABLE_REFERENCES = ["echohq.com"]
```
Elkasitu, line 9: I would make this adjustable from the environment just like we do for the two values above, using `pydantic-settings` of course

collectors/osv/collectors.py
roduran-dev: Modify with the ` "withdrawn" in osv_vuln`

collectors/osv/collectors.py
roduran-dev: Changed to

 ```
Check if OSV record can be ignored during collection.

        Records are ignored if they are withdrawn AND contain references
        from specific sources (see IGNORABLE_REFERENCES).
       
```

File: collectors/osv/collectors.py (hunk 11 above)
Comment (roduran-dev) [reply collectors/osv/collectors.py:393]: Totally write. I merge it with the old approach to clean it.

File: collectors/osv/constants.py (hunk 13 above)
Comment (roduran-dev) [reply collectors/osv/constants.py:9]: I don't see a problem with doing this but seems a bit over engineering :thinking:   as this was the first case in years for OSV

File: collectors/osv/collectors.py (hunk 12 above)
Comment (roduran-dev) [reply collectors/osv/collectors.py:289]: added back

File: collectors/osv/collectors.py [hunk 14]
```diff
@@ -342,3 +358,36 @@ def get_cvss_and_impact(data: dict) -> tuple[list, str]:
         }
 
         return osv_id, cve_ids, content
+
+    def has_reference_to_ignore(self, osv_vuln: dict) -> bool:
+        for ref_to_ignore in IGNORABLE_REFERENCES:
+            for osv_reference in osv_vuln.get("references", [{}]):
+                if ref_to_ignore in osv_reference.get("url", ""):
+                    return True
```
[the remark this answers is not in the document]
roduran-dev: I don't see that happening for this case.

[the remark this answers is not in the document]
roduran-dev: Also it would need the missing detail so no problems there

conftest.py
roduran-dev: I see what you mean. I will mock it then for this test

roduran-dev: Mocked

conftest.py
roduran-dev: This is no longer valid (I used a mock assertin @Elkasitu comment.)

pytest.ini
roduran-dev: This is no more valid (I used a mock)  

I don't its a bad feature but it was just for one. If we wanted we could do it in a different PR

roduran-dev:
@Elkasitu @JakubFrejlach Thanks a lot for the reviews. I asserted all your comments.

roduran-dev:
This has been fix in the source. It seem we don't need an edge case set on our side

PR #1251, opened 2026-04-23
OSIDB-4914: Update rescheduler's querying logic in sync manager

Jincxz:
Adjusted the logic behind sync manager's rescheduling. 

The original code checked every sync manager, of a particular class, to determine if they should be rescheduled. The new logic filters out each conditional branch for rescheduling instead. This should minimize the OOM issue and allow the collectors to run. 

The `rescheduled` set is there to prevent re-reschedules in the case that the conditions overlap.

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -8,6 +8,7 @@ from celery.exceptions import Ignore
 from celery.utils.log import get_task_logger
 from django.conf import settings
 from django.db import models, transaction
+from django.db.models import F, Q
 from django.utils import timezone
 from rhubarb.tasks import LockableTaskWithArgs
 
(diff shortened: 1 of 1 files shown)


osidb/sync_manager.py
```diff
@@ -306,97 +307,102 @@ def check_for_reschedules(cls):
...
-            if (
-                sync_manager.last_scheduled_dt is None
-                and sync_manager.last_started_dt is not None
-            ):
-                logger.info(
-                    f"{sync_manager.__class__.__name__} {sync_manager.sync_id}: "
-                    f"Started but not scheduled, this should NEVER happen"
-                )
-                continue
+        sync_managers = SyncManager.objects.filter(name=cls.__name__)
```
JakubFrejlach, line 328: The move from Python-level filtering to DB queries is a solid improvement, but I think there's still a potential bottleneck: if many sync managers match the reschedule conditions, we're still calling `reschedule()` for each one individually (multiple DB queries + a broker call per sync manager). On stage, where we're seeing OOM kills due to a large volume of failed sync managers, this could still be problematic.

Would it make sense to cap each reschedule check to a reasonable batch size? Since most collectors run every minute, we'd eventually process all pending reschedules across successive runs.

To prevent starvation (the same rows getting picked every time while others wait indefinitely), we could order candidates by `last_scheduled_dt` ascending, oldest-waiting items get rescheduled first, and once rescheduled their timestamp moves to the back of the queue.

JakubFrejlach (approved):
Change LGTM! Just a thought about another possible bottleneck.

osidb/sync_manager.py
Jincxz: That sounds like a good idea. Scheduling in batches makes sense. Will convert the rescheduler into a capped queue-like system.

Jincxz: Added a cap for each check and sorted each one with their respective datetimes. Thoughts on the limit/sorting order?

JakubFrejlach (approved):

Elkasitu (approved):

PR #1253, opened 2026-04-23
OSIDB-4923 - Add helper function to pull cpe list

darakian:
Adds a helper function to extract nvd cpe data and adds it to the list of data returned Partial implementation for OSIDB-4923

I didn't see where the function `response2result` is being tested, so I didn't add a test, but happy to do so if that can be pointed out.

This PR does not store the data elsewhere. The eventual location will be one of the new fields discussed in OSIDB-4923

### DIFF
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -70,7 +70,7 @@ class NVDQuerier:
             Return a list of CPEs from the CVE `data`
             """
             cpe_list = []
-            if "cpe" in data and len(data.cpe) > 0:
+            if hasattr(data, "cpe"):
                 for entry in data.cpe:
                     cpe_list.append(entry.criteria)
 
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Union
+from typing import List, Union
 
 import nvdlib
 from celery.utils.log import get_task_logger
@@ -65,7 +65,7 @@ class NVDQuerier:
         filtering out everything unnecessary and simplifying
         """
 
-        def get_cpe_list(data: CVE) -> Optional[List[str]]:
+        def get_cpe_list(data: CVE) -> List[str]:
             """
             Return a list of CPEs from the CVE `data`
             """
--- a/collectors/nvd/collectors.py
+++ b/collectors/nvd/collectors.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import List, Optional, Union
 
 import nvdlib
 from celery.utils.log import get_task_logger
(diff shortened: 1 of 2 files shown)


collectors/nvd/collectors.py
```diff
@@ -65,6 +65,20 @@ def response2result(self, vulnerabilities: list) -> list:
         filtering out everything unnecessary and simplifying
         """
 
+        def get_cpe_list(data: CVE) -> Optional[List[str]]:
+            """
+            Return a list of CPEs from the CVE `data`
+            """
+            cpe_list = []
+            if "cpe" in data and len(data.cpe) > 0:
```
Jincxz: This can be reduced to `hasattr(data, "cpe")` since we are checking if the field exists. An empty `data.cpe` will just mean that the following loop wouldn't run.

collectors/nvd/collectors.py
```diff
@@ -65,6 +65,20 @@ def response2result(self, vulnerabilities: list) -> list:
...
+                for entry in data.cpe:
+                    cpe_list.append(entry.criteria)
+
+            if len(cpe_list) > 0:
+                return cpe_list
+            else:
+                return None
```
Jincxz: I think it should be fine with just returning the `cpe_list` as is (potentially an empty list). The new field will likely have a default of an empty list and a `None` value might produce an unexpected result when writing to it. Will make adjustments later if that is no longer the case.

Jincxz (changes requested):
Some comments regarding the helper function. 

There aren't tests directly for `response2result` but there are tests for the NVD collector at https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/nvd/tests/test_collectors.py

You can add tests for the `NVDQuerier`, and `response2result`, there if you would like. Testing the collectors and the data being written to flaw(s) will have to come after the field is added.

collectors/nvd/collectors.py
Elkasitu: `cpe` doesn't seem to be a valid top-level attribute per NVD's Vulnerability API [schema](https://csrc.nist.gov/schema/nvd/api/2.0/cve_api_json_2.0.schema), I would add a test or two as Jin suggested.

darakian: Oh good call. Changed that in https://github.com/RedHatProductSecurity/osidb/pull/1253/commits/fa056a019a3a98467e2e9bde8b406b3db4236739

collectors/nvd/collectors.py
darakian: Fair enough. Updated in https://github.com/RedHatProductSecurity/osidb/pull/1253/commits/bb2d409453e198c5b09c1ee793a479223507154c

collectors/nvd/tests/test_collectors.py
```diff
@@ -236,6 +236,20 @@ def test_cvss4(self):
         flaw = Flaw.objects.get(cve_id=cve_id)
         assert flaw.cvss_scores.filter(version=FlawCVSS.CVSSVersion.VERSION4)
 
+    @pytest.mark.vcr
+    def test_cpe_load(self):
```
JakubFrejlach, line 240: This test won't work without the actual cassette recorded. https://github.com/RedHatProductSecurity/osidb/blob/master/docs/developer/DEVELOP.md#testing-external-apis

darakian:
Closing this out as the work has been de-prioritized. NVDs reduced scope and unclear SLA make this less viable for our goals.

PR #1257, opened 2026-04-29
OSIDB:4865-replace JiraUserMapping with Profile.kerberos_to_cloud_id.

svelamak:
Replace `JiraUserMapping` with `Profile.kerberos_to_cloud_id` for resolving Atlassian Cloud IDs from Kerberos IDs. The resolved IDs are cached on the `Profile` model to avoid repeated Jira API calls.
-Added `atlassian_cloud_id` field and `kerberos_to_cloud_id` classmethod to `Profile` that checks DB cache first, calls Jira API only on miss.
-Added `get_jira_cloud_id` helper in `helpers.py`.
-Replaced `JiraUserMapping` with `Profile.kerberos_to_cloud_id` in `query.py` and `service.py`
-New test file `osidb/models/tests/test_profile.py`
-Updated existing tests in `test_service.py`, `test_e2e.py`, `test_jira.py` to use `Profile`.
Closes:OSIDB-4865


osidb/helpers.py
```diff
@@ -470,3 +471,29 @@ def get_jira_api_email(request: Request) -> str:
...
+    """
+    from collectors.jiraffe.constants import HTTPS_PROXY
+
+    email = f"{kerberos_id}@redhat.com"
+    auth_token = get_env("JIRA_AUTH_TOKEN")
+    jira_url = get_env("JIRA_URL", "https://redhat.atlassian.net")
+    try:
+        jira_api = JIRA(
+            options={"server": jira_url, "check_update": False},
+            token_auth=auth_token,
```
JakubFrejlach: `token_auth` is not working for the atlassian cloud instance, you need to use `basic_auth` with email+token.

osidb/helpers.py
```diff
@@ -470,3 +471,29 @@ def get_jira_api_email(request: Request) -> str:
...
+            get_server_info=False,
```
JakubFrejlach: this needs to be set to `True` otherwise Jira lib is not able to determine whether it is talking to Jira Cloud instance or datacenter and some of the methods are not available

osidb/helpers.py
```diff
@@ -470,3 +471,29 @@ def get_jira_api_email(request: Request) -> str:
...
+            proxies={"https": HTTPS_PROXY},
+        )
+        users = jira_api.search_users(email)
```
JakubFrejlach: This passes the mail to `username` arg which is not available do to a GDPR restrictions. You need to use `search_users(query=email)`.

osidb/helpers.py
```diff
@@ -470,3 +471,29 @@ def get_jira_api_email(request: Request) -> str:
...
+    email = f"{kerberos_id}@redhat.com"
+    auth_token = get_env("JIRA_AUTH_TOKEN")
+    jira_url = get_env("JIRA_URL", "https://redhat.atlassian.net")
+    try:
+        jira_api = JIRA(
+            options={"server": jira_url, "check_update": False},
+            token_auth=auth_token,
+            get_server_info=False,
+            proxies={"https": HTTPS_PROXY},
+        )
```
JakubFrejlach: There is already a working Jira connector with all the necessary setup https://github.com/RedHatProductSecurity/osidb/blob/master/collectors/jiraffe/core.py#L26

docs/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve resilience, error-reporting and performance of /bulk create
   affects endpoint (OSIDB-4748)
 
+### Fixed
+- Replace JiraUserMapping with direct Jira API user lookup and cache in Profile model(OSIDB-4865)
```
JakubFrejlach: I don't think this should be in `Fixed` category but rather changed.

osidb/models/profile.py
```diff
@@ -11,10 +14,27 @@ class Profile(models.Model):
...
+    def kerberos_to_cloud_id(cls, kerberos_id: str) -> str:
+        """
+        return Atlassian Cloud ID for a user
+        """
+        try:
+            profile = cls.objects.get(user__username=kerberos_id)
+        except cls.DoesNotExist:
+            raise JiraUserMappingException(
+                f"No profile found for Kerberos ID:{kerberos_id}"
+            )
```
JakubFrejlach: This is not sufficient, currently the profiles in OSIDB are created for users interacting with OSIDB, but the cloud mapping is also needed for users which are for example CCed to tracker, etc. These might not be in the OSIDB. Rather than fail, we need to create those profiles.

osidb/helpers.py
```diff
@@ -470,3 +471,29 @@ def get_jira_api_email(request: Request) -> str:
...
+        users = jira_api.search_users(email)
+    except Exception:
+        logger.error(f"Failed to fetch Jira cloud ID for {kerberos_id}")
+        return ""
+    if users:
+        return users[0].accountId
+    return ""
```
JakubFrejlach: I don't think we should be returning empty string when we are unable to get the ID, when the ID cannot be found in the Jira it means there is a problem somewhere which needs to be fixed.

JakubFrejlach (changes requested):
Some changes are needed, mainly we need the profiles to be created when they are missing as we will need mapping also for users which are not users of OSIDB. Also I think part of this should be removing the previous Jira Mapping + the collector for it, as it should not longer be needed.

osidb/helpers.py
svelamak: Fixed, now using `basic_auth` with email+token via `JiraConnector`

osidb/helpers.py
svelamak: Fixed using `JiraConnector`.

osidb/helpers.py
svelamak: Fixed.

osidb/helpers.py
svelamak: Fixed

docs/CHANGELOG.md
svelamak: Fixed

osidb/helpers.py
svelamak: Fixed, now raises `JiraUserMappingException`

osidb/models/profile.py
svelamak: Fixed, now using `get_or_create` for both `User` and `Profile`.

svelamak:
Removed entire `jira_user_mapping` collector and model. 
Removed `sync_jira_user_mapping.py`, `JiraUserMappingFactory` from `factories.py` and also references in `config/settings.py` and `models/__init__.py`.
Added a migration.

apps/trackers/jira/query.py
```diff
@@ -531,15 +527,13 @@ def generate_cc(self):
...
-                    for un in cc_list
+                    {"accountId": Profile.kerberos_to_cloud_id(un)} for un in cc_list
                 ]
             elif inv_field_obj := JiraProjectFields.objects.filter(
                 project_key=self.ps_module.bts_key, field_name="Involved"
             ).first():
                 self._query["fields"][inv_field_obj.field_id] = [
-                    {"accountId": JiraUserMapping.kerberos_to_cloud_id(un)}
-                    for un in cc_list
+                    {"accountId": Profile.kerberos_to_cloud_id(un)} for un in cc_list
```
Elkasitu: you can pre-calculate the cloud ids so that you only query once

osidb/models/profile.py
```diff
@@ -11,10 +13,24 @@ class Profile(models.Model):
...
     def username(self):
         return self.user.username
+
+    @classmethod
+    def kerberos_to_cloud_id(cls, kerberos_id: str) -> str:
+        """
+        return Atlassian Cloud ID for a user
+        """
+        user, _ = User.objects.get_or_create(username=kerberos_id)
+        profile, _ = cls.objects.get_or_create(user=user)
```
Elkasitu, line 31: I'm not a big fan of this, we're creating user accounts for people that might never use OSIDB, for people that potentially don't have and should not have access to OSIDB and/or for people who might use OSIDB in the future, we need to make sure that this user object won't clash with the user object of a future OSIDB user.

Elkasitu: There's also a risk that if someone's off-boarded from JIRA, OSIDB will still keep the `atlassian_cloud_id` as there's no updates

JakubFrejlach: Does the account actually provide any sort of access to OSIDB? :thinking: Data are provided based on the LDAP groups membership AFAIK.  I agree with the clash point and thus we should make sure that newly created users are potentially merged with the existing account?

Regarding the offboarding, only real concern is that there will be obsolete data, right?

svelamak: To avoid this, we could skip creating User accounts for non OSIDB users and instead query Jira directly for them without caching. Would that work?

JakubFrejlach: > To avoid this, we could skip creating User accounts for non OSIDB users and instead query Jira directly for them without caching. Would that work?

this does not really play well with the Tracker CC lists, there are some CC lists which can even contain like 20-30 users AFAIK and without caching we are essentially doing 20-30 requests to Jira every time we want to file a new tracker.

Elkasitu: > Does the account actually provide any sort of access to OSIDB?

Probably not, but it's just useless data to have

> Regarding the offboarding, only real concern is that there will be obsolete data, right?

If the user has been offboarded from Jira but not from product-definitions creating trackers or whatever will still blow up as the user won't be found in Jira

apps/trackers/jira/query.py
svelamak: Pre-calculated the cloud ID's so that it is only queried once.

osidb/models/profile.py
svelamak: Double checked that `django_auth_ldap` source uses `get_or_create_user` internally when a user logs in, so if a user already exists, ldap will just reuse it.

Elkasitu (approved):

osidb/helpers.py
```diff
@@ -470,3 +470,28 @@ def get_jira_api_email(request: Request) -> str:
...
+
+    email = f"{kerberos_id}@redhat.com"
+    try:
+        connector = JiraConnector()
+        # filter for exact email match
+        users = [
+            u
+            for u in connector.jira_conn.search_users(query=email)
+            if u.emailAddress == email
+        ]
```
JakubFrejlach: I realized one thing, user can actually have his email hidden in the atlassian cloud, in that case `search_users` will still return him but the `emailAddress` field will be empty, so I think we should tweak this logic a bit. If only one user is returned, then just go with it as there is no ambiguity, if multiple users are returned, try to find the one which matches the email and if not, then we fail because we cannot reliable find out which account is it supposed to be.

JakubFrejlach (commented):
Looking good! Thank you for addressing all of my concerns. I've realized one more thing, see my comment, should be easy to change and after that is taken care of, we can merge.

osidb/helpers.py
svelamak: Updated - If only one user is returned we use it directly, if multiple are returned we filter by exact email match and if no match is found we raise an exception.

JakubFrejlach (approved):
LGTM! Thank you for addressing my feedback!

PR #1258, opened 2026-05-01
Add CME IDs to Affect model

stuwrtlttle:
## Summary
- Adds a `cme_ids` ArrayField to the Affect model, allowing defensive security controls from the [CME (Common Mitigation Enumeration)](https://github.com/stuwrtlttle/cme) taxonomy to be associated with individual affects
- Each CME ID (e.g. `CME-601`) references a specific mitigation control (ASLR, SELinux, seccomp, etc.) that attenuates CVSS impact when active on a product stream
- Enables deterministic environmental CVSS scoring per-product rather than a single global score

## Changes
- **`osidb/validators.py`** — `validate_cme_id` validator enforcing `CME-NNN` or `CME-NNNN` format
- **`osidb/models/affect.py`** — `cme_ids` ArrayField (Postgres array of validated strings, defaults to empty list)
- **`osidb/serializer.py`** — `cme_ids` exposed as read-write field in `AffectSerializer`
- **`osidb/filters.py`** — `cme_ids` (contains lookup) and `cme_ids__isempty` filters on `AffectFilter`
- **`osidb/migrations/0232_affect_cme_ids.py`** — Non-destructive `AddField` migration
- **`osidb/models/tests/test_affect.py`** — Unit tests for default value, storage, and validator (10 parametrized cases)

## API Usage
```bash
# Associate CME mitigations with an affect
PUT /osidb/api/v2/affects/{uuid}
{"cme_ids": ["CME-101", "CME-301", "CME-601"]}

# Query affects by active mitigation
GET /osidb/api/v2/affects/?cme_ids=CME-601

# Find affects with no mitigations recorded
GET /osidb/api/v2/affects/?cme_ids__isempty=true
```

## Test plan
- [ ] Verify migration applies cleanly on a fresh database
- [ ] Verify existing affects default to empty `cme_ids`
- [ ] Test creating/updating affects with valid CME IDs via API
- [ ] Test that malformed CME IDs are rejected (e.g. `CME-10`, `cme-101`, `CME101`)
- [ ] Test `cme_ids` and `cme_ids__isempty` filters return correct results
- [ ] Run existing affect test suite to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)


osidb/models/tests/test_affect.py
```diff
@@ -363,3 +364,37 @@ def test_validate_purl_and_ps_component(self, purl, ps_component, should_raise):
             assert "does not match user-provided ps_component" in str(exc_info.value)
         else:
             affect.save()
+
+    def test_cme_ids_default_empty(self):
+        """Test that the cme_ids field defaults to an empty list"""
+        affect = AffectFactory()
+        assert affect.cme_ids == []
```
JakubFrejlach: This test does not really test anything meaningful, it is Django responsibility to assign default value `[]` for this field.

osidb/models/tests/test_affect.py
```diff
@@ -363,3 +364,37 @@ def test_validate_purl_and_ps_component(self, purl, ps_component, should_raise):
...
+
+    def test_cme_ids_stores_valid_ids(self):
+        """Test that valid CME IDs can be stored on an affect"""
+        affect = AffectFactory(cme_ids=["CME-101", "CME-601", "CME-1001"])
+        affect.refresh_from_db()
+        assert affect.cme_ids == ["CME-101", "CME-601", "CME-1001"]
```
JakubFrejlach: This test does not really test anything meaningful, the `CME` format validation is tested in the next test so this test essentially just test whether you can store values in the field which is Django/Postgres responsibility.

osidb/migrations/0232_affect_cme_ids.py
```diff
@@ -0,0 +1,27 @@
...
+                base_field=models.CharField(
+                    max_length=20,
+                    validators=[osidb.validators.validate_cme_id],
+                ),
+                blank=True,
+                default=list,
+                size=None,
+            ),
+        ),
+    ]
```
JakubFrejlach: Unless there is a special reason for that (custom logic, data migrations) a DB migration should be autogenerated via Django (see `make migrations-create`), this one lacks the Django timestamps (like `# Generated by Django 5.2.12 on 2026-05-04 10:07`) and apparently it is missing something because migration check pipeline is failing.

JakubFrejlach (changes requested):
Couple of things and also pipelines need fixing:
* migrations -> see one of my comments
* schema -> you are adding new fields and query parameters and thus the Open api schema needs to be updated (see `make update-schema`)
* tests -> should probably fix itself once the migrations are settled
* changelog -> this is a user facing change, please add a record to the `docs/CHANGELOG.md`

JakubFrejlach:
@stuwrtlttle also is there a Jira tracking this change?

osoukup:
Why do we need CMEs? Was that discussed with the OSIDB development team and stakeholders? I read that this has something to do with environmental metrics and I do not really remember an agreement on supporting them.

stuwrtlttle:
Thanks for the input Jakub! I'll work on those corrections.  @osoukup CME is a new taxonomy that I've put together and shared with Mitre and First.org. Vince has given the greenlight for Red Hat to take the lead in showcasing/using the taxonomy. We're a test subject so to speak.

osidb/migrations/0232_affect_cme_ids.py
stuwrtlttle: @JakubFrejlach I think I worked through all the errors with claude; but admittedly I don't understand the code that well, so if you want to double check ... I won't be offended.

osidb/models/tests/test_affect.py
JakubFrejlach: ^ @stuwrtlttle this test should go away

osidb/models/tests/test_affect.py
JakubFrejlach: ^ @stuwrtlttle this test should go away

osidb/models/tests/test_affect.py
stuwrtlttle: Test is gone now

osidb/models/tests/test_affect.py
stuwrtlttle: Test is gone now!

znovacek:
I'm rejecting this one for now as if CME becomes a priority (which is not for Q3 atm) we will need a proper spike to be run first in order to cover the following gaps:

Generally this does not consider complexity of the tooling and consequences/impacts it would have so this cannot be easily merged nor reworked before doing a proper research.
The idea is that we will provide customers with this new piece of data but it is not thought about the way we should do it - VEX will not easily adopt this, CVE pages are not machine readable and ready for adopting it either.
Additionally it is not clear how analysts are going to deal with it from process perspective and how they would manage additional task from capacity perspective. Given the situation with growing CVE backlog this would represent additional risk.

PR #1260, opened 2026-05-04
OSIDB-2995-update drf-spectacular to generate OpenAPI 3.1 schema

svelamak:
Update drf-spectacular to generate OPenAPI 3.1 schema.
-Added OAS version to SPECTACULAR_SETTINGS to enable OpenAPI 3.1 schema generation.
-Regenerated openapi.yml
Closes:OSIDB-2995

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,4 +1,4 @@
-openapi: 3.0.3
+openapi: 3.1.0
 info:
   title: OSIDB API
   version: 5.9.0
@@ -14002,19 +14002,19 @@ components:
           type: string
           readOnly: true
         ps_component:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
           default: ''
-          maxLength: 255
         impact:
           oneOf:
           - $ref: '#/components/schemas/ImpactEnum'
           - $ref: '#/components/schemas/BlankEnum'
         tracker:
-          allOf:
+          oneOf:
           - $ref: '#/components/schemas/Tracker'
+          - type: 'null'
           readOnly: true
-          nullable: true
         delegated_resolution:
           type: string
           readOnly: true
@@ -14024,8 +14024,9 @@ components:
             $ref: '#/components/schemas/AffectCVSS'
           readOnly: true
         purl:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
           default: ''
         subpackage_purls:
           type: array
@@ -14041,10 +14042,11 @@ components:
           - $ref: '#/components/schemas/DelegatedNotAffectedJustificationEnum'
           - $ref: '#/components/schemas/BlankEnum'
         resolved_dt:
-          type: string
+          type:
+          - string
+          - 'null'
           format: date-time
           readOnly: true
-          nullable: true
         labels:
           type: array
           items:
@@ -14128,17 +14130,18 @@ components:
           type: string
           minLength: 1
         ps_component:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
           default: ''
-          maxLength: 255
         impact:
           oneOf:
           - $ref: '#/components/schemas/ImpactEnum'
           - $ref: '#/components/schemas/BlankEnum'
         purl:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
           default: ''
         subpackage_purls:
           type: array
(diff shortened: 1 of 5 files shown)


JakubFrejlach (changes requested):
Requesting changes as a placeholder. Will need to check the compatibility of https://github.com/RedHatProductSecurity/osidb-bindings package with this change, will let you know tomorrow if all is good, we can proceed.

openapi.yml
```diff
@@ -14002,19 +14002,19 @@ components:
           type: string
           readOnly: true
         ps_component:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
           default: ''
-          maxLength: 255
```
JakubFrejlach, line 14092: `maxLenght` is being removed because there seems to be a [bug](https://github.com/tfranzel/drf-spectacular/issues/1302#event-15484108353) in `drf-spectacular==0.27.2` which we are using currently.

It was fixed in this [commit](https://github.com/tfranzel/drf-spectacular/commit/8156c0bef161c44e4451be3ffd22fe0266d26a50) and based on the [full changelog](https://github.com/tfranzel/drf-spectacular/compare/0.27.2...0.28.0) released as part of `drf-spectacular==0.28.0` so we should update to the newer version of drf-spectacular to remove this problem. Latest seems to be `drf-spectacular==0.29.0` so probably latest since we are already need to perform the update.

openapi.yml
```diff
@@ -14002,19 +14002,19 @@ components:
           type: string
           readOnly: true
         ps_component:
-          type: string
-          nullable: true
+          type:
+          - string
+          - 'null'
```
JakubFrejlach, line 14091: This is expected based on https://learn.openapis.org/upgrading/v3.0-to-v3.1.html#change-schema-example-to-examples

openapi.yml
JakubFrejlach, line 16929: Based on https://learn.openapis.org/upgrading/v3.0-to-v3.1.html#change-schema-example-to-examples this should be converted into standard JSON Schema `examples` keyword from OpenAPI-specific singular `example`, maybe the newer version of `drf-spectacular` has this covered. I've tried to render it via Swagger UI as it is and seems to be working, so if `drf-spectacular` update won't cover this, we might just proceed and maybe file an issue for https://github.com/tfranzel/drf-spectacular

JakubFrejlach (commented):
Overall looking good, but there is a few things which need attention, see my comments.

Tested with https://github.com/RedHatProductSecurity/osidb-bindings and it generates the package without any problems, but I will need to check again once we update to the new `drf-spectacular` version as mentioned in the comments.

Also there is a handy migration checklist for migrating from OpenAPI 3.0 to 3.1:
https://learn.openapis.org/upgrading/v3.0-to-v3.1.html#migration-checklist

for us probably the most important are these points (but might be worth checking others as well):
* Replace `nullable: true` with type arrays (`type: ["string", "null"]`) -> already covered
* Change schema `example` to `examples` (as an array) -> should be hopefully covered by drf-spectacular update

openapi.yml
svelamak: Updated drf-spectacular to >= 0.29.0 and regenerated the schema.
`maxLength` is now back.

openapi.yml
svelamak: After updating the `drf-spectacular`, it is still `example` not `examples`.

JakubFrejlach: @svelamak Ok, I did some digging and it seems that even though the `example` is "deprecated" in the OpenAPI 3.1, it does not mean it is not supported. Most tools which we do use (openapi-python-client for osidb-bindings package, or Swagger for rendering) will use the `example` just fine. Might be worth to report it to `drf-spectacular` but definitely not a prio.

JakubFrejlach (approved):
LGTM! Rebase and address conflicts and we are good to go.

svelamak:
@coderabbitai checked - Those nullable fields  come from DRF’s built-in pagination class `(next/previous)` not our serializers. We do have a custom `HardLimitOffsetPagination` but it just extends DRF's base class without overriding the schema. Marking as a drf-spectacular limitation. And updated the version pin  `drf-spectacular==0.29.0` as suggested.

svelamak:
@coderabbitai review

openapi.yml
```diff
@@ -14141,7 +14144,9 @@ components:
           type: string
           readOnly: true
         assist_meta:
-          nullable: true
+          oneOf:
+          - {}
+          - type: 'null'
```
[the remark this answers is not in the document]
svelamak: @coderabbitai
`assist_meta` is a nullable `JSONField` and `pgh_context` comes from `pghistory,` both are auto-generated by `drf-spectacular.`
`oneOf: [{}, {type: 'null'}]` pattern is how `drf-spectacular` currently handles nullable JSON Fields.

PR #1262, opened 2026-05-05
Add is_hummingbird flag for new SLA policy

costaconrado:
This PR add is_hummingbird flag for affects for the condition on the new 1-day SLA policy.

Closes OSIDB-4884.

### DIFF
--- a/osidb/models/affect.py
+++ b/osidb/models/affect.py
@@ -852,6 +852,12 @@ class Affect(
             name=self.ps_module, ps_product__business_unit="Community"
         ).exists()
 
+    def is_hummingbird(self) -> bool:
+        """
+        check and return whether the given affect is targeting HUM project
+        """
+        return PsModule.objects.filter(name=self.ps_module, bts_key="HUM").exists()
+
     @property
     def is_notaffected(self) -> bool:
         """
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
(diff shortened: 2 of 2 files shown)


JakubFrejlach (approved):

osidb/models/affect.py
```diff
@@ -852,6 +852,12 @@ def is_community(self) -> bool:
             name=self.ps_module, ps_product__business_unit="Community"
         ).exists()
 
+    def is_hummingbird(self) -> bool:
+        """
+        check and return whether the given affect is targeting HUM project
+        """
+        return PsModule.objects.filter(name=self.ps_module, bts_key="HUM").exists()
```
Elkasitu, line 859: This suffers from the N+1 problem

Elkasitu (commented):
I guess this has a Summit deadline or similar, so please don't take my observations as blocking:

* The implementation seems a bit hackish as we rely on the bts_key, why not on Product or maybe this should be added to product-definitions as `is_hummingbird`, then VM / Hummingbird has total control over this
* The task and the contents of this PR do not match, how does adding a method change SLA policy? It doesn't seem like the contents of the method are being exposed via API or used anywhere?

JakubFrejlach:
> I guess this has a Summit deadline or similar, so please don't take my observations as blocking:

Yes you are very right :smile: 
> 
>     * The implementation seems a bit hackish as we rely on the bts_key, why not on Product or maybe this should be added to product-definitions as `is_hummingbird`, then VM / Hummingbird has total control over this

Product would be probably the best but currently SLA policies are not really working with Products because products do usually have spaces in their names and there is bug with that, see OSIDB-4956 but yeah ideally Product check would be the best.

> 
>     * The task and the contents of this PR do not match, how does adding a method change SLA policy? It doesn't seem like the contents of the method are being exposed via API or used anywhere?

SLA policy is able to access models fields and properties (Flaw, Affect, etc.) so you do not really need to expose it via API in this case.

costaconrado:
I'm merging this, I mapped this concern on OSIDB-4964 so we revisit this conversation in the future after summit.

PR #1265, opened 2026-05-05
OSIDB-4810: Add validation to require PURLs

Jincxz:
This PR adds a validation to require PURLs for affects along with affects context for error messages related to PURLs. Tests have been adjusted to handle the new requirement (e.g. the affect factory implicitly generates a PURL if not specified). The tests were adapted with help from Cursor.

In the `PURLField`, schema validation doesn't carry the affect context since `get_prep_value` doesn't carry the object instance. Any ideas on how I could add the context there (or if I even should?)


osidb/models/affect.py
```diff
@@ -702,49 +709,36 @@ def _validate_purl_and_ps_component(self, **kwargs):
...
+    def _validate_purl_existence(self, **kwargs):
         """
-        Validate that new affects related to middleware products have a PURL.
-        Also disallows removing PURL from existing middleware affects.
+        Validate that a PURL exists for affects. This does not include
+        affects for services or community affects.
         """
-        if not AffectSettings().require_purl_for_middleware:
-            return
+        from osidb.constants import SERVICES_PRODUCTS
```
costaconrado: nitpick: I took a look on osidb.constants it doesn't seem to import models generating a circular import, please check and if it works, consider moving this to the top of the file.

costaconrado (approved):
LGTM.

osidb/models/affect.py
```diff
@@ -702,49 +709,36 @@ def _validate_purl_and_ps_component(self, **kwargs):
...
 
         ps_module = PsModule.objects.filter(name=self.ps_module).first()
         if not ps_module:
             return
 
-        if not ps_module.is_middleware:
+        ps_product = ps_module.ps_product
+        if ps_product.is_community or ps_product.short_name in SERVICES_PRODUCTS:
```
Elkasitu: I think we could implement a property on `PsProduct` e.g. `is_service` that does this check

Elkasitu (approved):

PR #1266, opened 2026-05-05
OSIDB-4965: Include builds in auto-created affects

Elkasitu:
Before this commit, auto-creation of affects by querying newcli would only take into account the dependencies of the search term but not the search term itself.

E.g. when creating affects for the component ostree, the bootc component would be created (dependency) but not ostree (build).

This commit includes the builds when auto-creating affects.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -455,6 +455,34 @@
         "is_secret": false
       }
     ],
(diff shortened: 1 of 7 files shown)


MrMarble (approved):

dmonzonis (approved):
LGTM

PR #1270, opened 2026-05-08
OSIDB-4956: Fix check parser replaces spaces with underscores in equality check values

roduran-dev:
### problem 
CheckParser replaced all spaces with underscores, causing equality checks with multi-word values to fail. 

For example, "PS product is Red Hat Enterprise Linux" compared "Red_Hat_Enterprise_Linux"
  instead of "Red Hat Enterprise Linux".

  Impact: SLA exclusion policies (like Kpatch) and workflow conditions with multi-word values silently failed to match.

### solution
Modified desc2equals() to restore spaces in values for non-choices fields while preserving uppercase transformation for enum fields.

Closes OSIDB-4956

### DIFF
--- a/apps/sla/tests/test_models.py
+++ b/apps/sla/tests/test_models.py
@@ -992,6 +992,112 @@ class TestSLOPolicy:
                 == accepted
             )
 
+        @pytest.mark.parametrize(
+            "true_case, false_case",
+            [
+                ("Red Hat Enterprise Linux", "Red_Hat_Enterprise_Linux"),
+                ("Red_Hat_Enterprise_Linux", "Red Hat Enterprise Linux"),
+                ("Red_Hat Enterprise_Linux", "Red Hat_Enterprise Linux"),
+                ("Red Hat_Enterprise_Linux", "Red_Hat_Enterprise_Linux"),
+            ],
+        )
+        def test_policy_with_spaces(self, true_case, false_case):
+            """
+            Test the SLA policy with multi-word product name
+            (regression test for OSIDB-4956)
+
+            This mirrors and actually policy from ps-constants with spaces in the value:
+            """
+            policy_desc = {
+                "name": "testCase",
+                "description": "Magical test case",
+                "conditions": {
+                    "affect": [
+                        f"PS product is {true_case}",
+                        "PS component is magic",
+                    ],
+                },
+                "slo": None,  # Exclusion policy - no SLO applies
+            }
+
+            policy = SLOPolicy.create_from_description(policy_desc)
+
+            ps_product = PsProductFactory(short_name="rhel", name=true_case)
+            ps_module = PsModuleFactory(bts_name="bugzilla", ps_product=ps_product)
+            ps_update_stream = PsUpdateStreamFactory(ps_module=ps_module)
+            flaw = FlawFactory(embargoed=False)
+
+            # Test 1: Affect that SHOULD match the policy (Magic on RHEL)
+            magic_affect = AffectFactory(
+                flaw=flaw,
+                affectedness=Affect.AffectAffectedness.AFFECTED,
+                resolution=Affect.AffectResolution.DELEGATED,
+                ps_module=ps_module.name,
+                ps_component="magic",
+                ps_update_stream=ps_update_stream.name,
+                impact=Impact.MODERATE,
+            )
+            magic_tracker = TrackerFactory(
+                affects=[magic_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=ps_update_stream.name,
+                type=Tracker.BTS2TYPE[ps_module.bts_name],
+            )
+
+            # Should accept because both conditions match
+            assert policy.accepts(
+                TemporalContext(flaw=flaw, affect=magic_affect, tracker=magic_tracker)
+            ), f"Ps product is {true_case} should match component"
+
+            # Test 2: Affect that should NOT match (different component)
+            other_affect = AffectFactory(
+                flaw=flaw,
+                affectedness=Affect.AffectAffectedness.AFFECTED,
+                resolution=Affect.AffectResolution.DELEGATED,
+                ps_module=ps_module.name,
+                ps_component="kernel",
+                ps_update_stream=ps_update_stream.name,
+                impact=Impact.MODERATE,
+            )
+            other_tracker = TrackerFactory(
+                affects=[other_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=ps_update_stream.name,
+                type=Tracker.BTS2TYPE[ps_module.bts_name],
+            )
+
+            # Should NOT accept because component doesn't match
+            assert not policy.accepts(
+                TemporalContext(flaw=flaw, affect=other_affect, tracker=other_tracker)
+            ), "Magic policy should not match non-magic components"
+
+            # Test 3: Affect on different product (magic but not RHEL)
+            other_product = PsProductFactory(short_name="fedora", name=false_case)
+            other_ps_module = PsModuleFactory(
+                bts_name="bugzilla", ps_product=other_product
+            )
+            other_ps_update_stream = PsUpdateStreamFactory(ps_module=other_ps_module)
+            fedora_affect = AffectFactory(
+                flaw=flaw,
+                affectedness=Affect.AffectAffectedness.AFFECTED,
+                resolution=Affect.AffectResolution.DELEGATED,
+                ps_module=other_ps_module.name,
+                ps_component="magic",  # Same component
+                ps_update_stream=other_ps_update_stream.name,
+                impact=Impact.MODERATE,
+            )
+            fedora_tracker = TrackerFactory(
+                affects=[fedora_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=other_ps_update_stream.name,
+                type=Tracker.BTS2TYPE[other_ps_module.bts_name],
+            )
+
+            # Should NOT accept because product doesn't match
+            assert not policy.accepts(
+                TemporalContext(flaw=flaw, affect=fedora_affect, tracker=fedora_tracker)
+            ), "Magic policy should not match kpatch on non-RHEL products"
+
     class TestContext:
         """
         test SLOPolicy context determination
--- a/apps/sla/tests/test_models.py
+++ b/apps/sla/tests/test_models.py
@@ -943,8 +943,12 @@ class TestSLOPolicy:
             [
                 ("in", "kpatch", True),
                 ("in", "coffee", False),
+                ("in", "red hat", True),
+                ("in", "red_hat", False),
                 ("not in", "kpatch", False),
                 ("not in", "coffee", True),
+                ("not in", "red hat", False),
+                ("not in", "red_hat", True),
             ],
         )
         def test_in_condition(self, operator, component, accepted):
@@ -956,7 +960,14 @@ class TestSLOPolicy:
                 "description": "Test for the in/not in operator",
                 "conditions": {
                     "affect": [
-                        {f"PS component {operator}": ["kpatch", "kmatch", "kcatch"]}
+                        {
+                            f"PS component {operator}": [
+                                "kpatch",
+                                "kmatch",
+                                "kcatch",
+                                "red hat",
+                            ]
+                        }
                     ],
                 },
                 "slo": {
@@ -993,7 +1004,14 @@ class TestSLOPolicy:
             )
 
         @pytest.mark.parametrize(
-            "true_case, false_case",
+            "operator, accepted",
+            [
+                ("is", True),
+                ("is not", False),
+            ],
+        )
+        @pytest.mark.parametrize(
+            "valid_case, invalid_case",
             [
                 ("Red Hat Enterprise Linux", "Red_Hat_Enterprise_Linux"),
                 ("Red_Hat_Enterprise_Linux", "Red Hat Enterprise Linux"),
@@ -1001,20 +1019,17 @@ class TestSLOPolicy:
                 ("Red Hat_Enterprise_Linux", "Red_Hat_Enterprise_Linux"),
             ],
         )
-        def test_policy_with_spaces(self, true_case, false_case):
+        def test_policy_with_spaces(self, operator, accepted, valid_case, invalid_case):
             """
             Test the SLA policy with multi-word product name
             (regression test for OSIDB-4956)
-
-            This mirrors and actually policy from ps-constants with spaces in the value:
             """
             policy_desc = {
                 "name": "testCase",
                 "description": "Magical test case",
                 "conditions": {
                     "affect": [
-                        f"PS product is {true_case}",
-                        "PS component is magic",
+                        f"PS product {operator} {valid_case}",
                     ],
                 },
                 "slo": None,  # Exclusion policy - no SLO applies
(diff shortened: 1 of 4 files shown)


apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,107 @@ def test_in_condition(self, operator, component, accepted):
                 == accepted
             )
 
+        def test_kpatch_exclusion_policy_with_spaces(self):
+            """
+            Test the Kpatch SLA exclusion policy with multi-word product name
+            (regression test for OSIDB-4956)
+
+            This mirrors the actual Kpatch policy from ps-constants:
+            https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/blob/master/data/sla_policies.yml
```
JakubFrejlach: Do not put internal Gitlab links into the public code please.

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,107 @@ def test_in_condition(self, operator, component, accepted):
...
+                affects=[fedora_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=other_ps_update_stream.name,
+                type=Tracker.BTS2TYPE[other_ps_module.bts_name],
+            )
+
+            # Should NOT accept because product doesn't match
+            assert not policy.accepts(
+                TemporalContext(flaw=flaw, affect=fedora_affect, tracker=fedora_tracker)
+            ), "Kpatch policy should not match kpatch on non-RHEL products"
```
JakubFrejlach: This is internal SLA/SLO policy which should not be publicly disclosed, use some test policy instead please.

apps/workflows/checks.py
```diff
@@ -184,6 +182,11 @@ def choices_field(model, name):
                     or attr in self.TEXT_CHOICES_PROPERTIES
                 ):
                     value = value.upper()
+                else:
+                    # For non-choices fields, restore spaces that were replaced with underscores
+                    value = value.replace("_", " ")
```
JakubFrejlach: What about this scenario:
1. original value to check is `foo bar_foo`
2. in `_parse_dict` this gets converted to `foo_bar_foo`
3. here it gets converted to `foo bar foo`
4. `foo bar_foo` != `foo bar foo`

JakubFrejlach (changes requested):

JakubFrejlach:
Also mark as technical/add changelog entry to fix the failing `Changelog check`

apps/workflows/checks.py
roduran-dev: I see, I will add this other scenario :+1:

docs/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Add validation to require PURLs (OSIDB-4810)
```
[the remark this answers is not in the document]
roduran-dev: nop, that's correct and is from another PR

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,110 @@ def test_in_condition(self, operator, component, accepted):
...
+            )
+            other_ps_update_stream = PsUpdateStreamFactory(ps_module=other_ps_module)
+            fedora_affect = AffectFactory(
+                flaw=flaw,
+                affectedness=Affect.AffectAffectedness.AFFECTED,
+                resolution=Affect.AffectResolution.DELEGATED,
+                ps_module=other_ps_module.name,
+                ps_component="kpatch",  # Same component
+                ps_update_stream=other_ps_update_stream.name,
+                impact=Impact.MODERATE,
```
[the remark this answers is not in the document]
roduran-dev: I changed so it doesn't have a direct relation with prod configs.

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,110 @@ def test_in_condition(self, operator, component, accepted):
...
+                ("Red Hat Enterprise Linux", "Red_Hat_Enterprise_Linux"),
+                ("Red_Hat_Enterprise_Linux", "Red Hat Enterprise Linux"),
+            ],
+        )
+        def test_policy_with_spaces(self, value, false_case):
+            """
+            Test the SLA exclusion policy with multi-word product name
+            (regression test for OSIDB-4956)
+
+            This mirrors and actually policy from ps-constants with spaces in the value:
```
JakubFrejlach: `and actually` ?

apps/workflows/checks.py
```diff
@@ -155,19 +159,22 @@ def has_element(instance):
...
+        original_value = original_value[total_chars:]
+        return original_value
+
+    def desc2equals(self, check_desc, original_value=None):
         """
         attribute to literal value equality check
 
-        currently only supports string attributes
-        which values do not contain any spaces
+        supports string attributes including those with spaces in values
```
JakubFrejlach, line 175: I think `desc2not_equals` needs to be addressed as well

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,110 @@ def test_in_condition(self, operator, component, accepted):
...
+                affects=[fedora_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=other_ps_update_stream.name,
+                type=Tracker.BTS2TYPE[other_ps_module.bts_name],
+            )
+
+            # Should NOT accept because product doesn't match
+            assert not policy.accepts(
+                TemporalContext(flaw=flaw, affect=fedora_affect, tracker=fedora_tracker)
+            ), "Magic policy should not match kpatch on non-RHEL products"
```
JakubFrejlach: ps_component is `magic` not `kpatch`

apps/workflows/checks.py
```diff
@@ -84,7 +85,10 @@ def _parse_string(self, check_desc):
             self.desc2equals,
             self.desc2in,
         ]:
-            result = func(check_desc)
+            if func == self.desc2equals:
+                result = func(check_desc, original_check_desc)
+            else:
+                result = func(check_desc)
```
JakubFrejlach: This seems fragile towards decorators for example and also less readable, I think better approach would be to just add `**kwargs` param to all the funcs and for the specific ones which actually need the original_check_desc you can have a named param as we currently do have in `desc2equals`. This way you don't need to perform this specific check for `desc2equals`.

apps/workflows/checks.py
```diff
@@ -155,19 +159,22 @@ def has_element(instance):
 
                 return (doc, has_element)
 
-    def desc2equals(self, check_desc):
+    def get_original_value(self, value_after_underscore, original_value):
+        total_chars = len(value_after_underscore) * -1
+        original_value = original_value[total_chars:]
+        return original_value
+
+    def desc2equals(self, check_desc, original_value=None):
```
JakubFrejlach: original value of what? Why not name it so it actually describes what it is -> original_check_desc

apps/workflows/checks.py
```diff
@@ -155,19 +159,22 @@ def has_element(instance):
 
                 return (doc, has_element)
 
-    def desc2equals(self, check_desc):
+    def get_original_value(self, value_after_underscore, original_value):
+        total_chars = len(value_after_underscore) * -1
+        original_value = original_value[total_chars:]
+        return original_value
```
JakubFrejlach: Get original value from `original_value`, of what? This is really hard to understand. What is actually happening here is that we are getting original value from the original check description.

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,110 @@ def test_in_condition(self, operator, component, accepted):
...
+        @pytest.mark.parametrize(
+            "value, false_case",
+            [
+                ("Red Hat Enterprise Linux", "Red_Hat_Enterprise_Linux"),
+                ("Red_Hat_Enterprise_Linux", "Red Hat Enterprise Linux"),
+            ],
+        )
+        def test_policy_with_spaces(self, value, false_case):
+            """
+            Test the SLA exclusion policy with multi-word product name
```
JakubFrejlach: This does not necessarily needs to be an exclusion policy, right? Such check for `Red Hat Enterprise Linux` can be used in any other policy.

apps/sla/tests/test_models.py
```diff
@@ -992,6 +992,110 @@ def test_in_condition(self, operator, component, accepted):
                 == accepted
             )
 
+        @pytest.mark.parametrize(
+            "value, false_case",
+            [
+                ("Red Hat Enterprise Linux", "Red_Hat_Enterprise_Linux"),
+                ("Red_Hat_Enterprise_Linux", "Red Hat Enterprise Linux"),
```
JakubFrejlach, line 1017: Can we add a test case with name which has some combination of spaces and underscores? -> `Red_Hat Enterprise Linux`

JakubFrejlach (commented):
This is much better but there is still couple of things which needs some attention.

apps/sla/tests/test_models.py
roduran-dev: yep, good catch,

apps/sla/tests/test_models.py
roduran-dev: Sure

roduran-dev: Added :+1:

apps/sla/tests/test_models.py
roduran-dev: Fixed

apps/workflows/checks.py
roduran-dev: ok, I will clarify beter.

apps/workflows/checks.py
roduran-dev: ok, make sense

apps/sla/tests/test_models.py
JakubFrejlach: Still not addressed.

apps/sla/tests/test_models.py
JakubFrejlach: Still not addressed.

JakubFrejlach (commented):
Lot of my previous comments were not addressed so I am not going to review further.

apps/sla/tests/test_models.py
```diff
@@ -943,8 +943,12 @@ def test_conditions_multiple_flaws(self):
             [
                 ("in", "kpatch", True),
                 ("in", "coffee", False),
+                ("in", "red hat", True),
```
roduran-dev, line 946: Just to be sure I added this edge case to the "in"/"no in" test

apps/workflows/checks.py
roduran-dev: changed variables to a more descriptive name.

apps/workflows/checks.py
roduran-dev: I also added "is not" cases to the test.

apps/sla/tests/test_models.py
roduran-dev: Done :+1:

apps/sla/tests/test_models.py
roduran-dev: I changed the test to be more direct and specific to the actual case. I removed kpatch and the ps_component (now its only ps product <is/is not> <Multi word description>

roduran-dev:
> Lot of my previous comments were not addressed so I am not going to review further.

Hi @JakubFrejlach  sorry last time, I click the re review by mistake.  I just finish asserting all your comments.

apps/workflows/checks.py
```diff
@@ -84,7 +85,7 @@ def _parse_string(self, check_desc):
             self.desc2equals,
             self.desc2in,
         ]:
-            result = func(check_desc)
+            result = func(check_desc, original_check_desc=original_check_desc)
```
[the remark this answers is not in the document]
roduran-dev: But this is from way before this PR. And I didn't change the orignals.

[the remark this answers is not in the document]
roduran-dev: @JakubFrejlach I think here is another BUG that was here for a few months. (missing values desc2in) 

Created https://redhat.atlassian.net/browse/OSIDB-5025 for the fix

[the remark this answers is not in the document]
JakubFrejlach: @roduran-dev actually I am not sure whether the `desc2in` should even be in that list in the first place, if you look at the method [`_parse_dict`](https://github.com/RedHatProductSecurity/osidb/blob/993e8bdcd9c91aea26a08db7bb32920027dcd6fb/apps/workflows/checks.py#L95C9-L95C20) `desc2in` and even `desc2not_in` (which is just a negated counterpart for the `desc2in`) are processed there. Also if you look in method [`parse`](https://github.com/RedHatProductSecurity/osidb/blob/993e8bdcd9c91aea26a08db7bb32920027dcd6fb/apps/workflows/checks.py#L66) you can see that there is actually a check which in case of `dict` calls the `_parse_dict` and otherwise `_parse_string`.

apps/sla/tests/test_models.py
```diff
@@ -992,6 +1003,112 @@ def test_in_condition(self, operator, component, accepted):
...
+                impact=Impact.MODERATE,
+            )
+            valid_tracker = TrackerFactory(
+                affects=[valid_affect],
+                embargoed=flaw.embargoed,
+                ps_update_stream=ps_update_stream.name,
+                type=Tracker.BTS2TYPE[ps_module.bts_name],
+            )
+
+            # # Should accept because condition match
```
JakubFrejlach: minor nitpick: double hashtag

apps/workflows/checks.py
```diff
@@ -184,24 +191,33 @@ def choices_field(model, name):
...
 
                 return (doc, compare_element)
 
-    def desc2not_equals(self, check_desc):
+    def desc2not_equals(self, check_desc, **kwargs):
         """
         negative attribute to literal value comparison check
 
         currently only supports string attributes
         which values do not contain any spaces
```
JakubFrejlach: stale docstring

apps/workflows/checks.py
```diff
@@ -155,19 +156,25 @@ def has_element(instance):
...
+
+    def desc2equals(self, check_desc, **kwargs):
         """
         attribute to literal value equality check
 
-        currently only supports string attributes
-        which values do not contain any spaces
+        supports string attributes including those with spaces in values
         """
+        original_check_desc = kwargs.get("original_check_desc")
```
JakubFrejlach: `original_check_desc` should be an explicit parameter on `desc2equals` and `desc2not_equals` rather than pulled from `**kwargs`, these methods actually depend on it, so the signature should reflect that. The other `desc2*` methods can keep `**kwargs` to absorb it from the dispatch loop.

```python
def desc2equals(self, check_desc, original_check_desc=None, **kwargs):
```

JakubFrejlach (commented):
Great improvement, I have a few more things which should be simple to handle.

apps/sla/tests/test_models.py
roduran-dev: fixed.

apps/workflows/checks.py
roduran-dev: changed :+1:

JakubFrejlach (approved):
LGTM now!

apps/workflows/checks.py
Elkasitu: comment still unaddressed

apps/workflows/checks.py
```diff
@@ -68,6 +68,7 @@ def parse(self, check_desc):
         return self._parse_string(check_desc)
 
     def _parse_string(self, check_desc):
+        original_check_desc = check_desc
         check_desc = check_desc.replace(" ", "_")
```
Elkasitu, line 72: did you consider not converting spaces to underscores in the first place? if we're looking for tokens having spaces or otherwise should not matter

maybe @osoukup knows why we did it like this in the first place?

Elkasitu (approved):
I think this can be merged into a single commit, have a few nitpicks but LGTM otherwise

osoukup (approved):
LGTM

apps/workflows/checks.py
osoukup: I think at this point it is mostly historical reasons and if we could get rid of it. I think that originally the aim was to prioritize properties and constants over the free text and those are always underscored. We could however achieve that other ways and this was probably just a simple one. The side effect of not parsing the free text with spaces well was not taken into account.

PR #1273, opened 2026-05-12
Collect affected block from cve record

darakian:
We want to have the affected block available for processing. This change adds the affected block to the dict returned by the cve collector and does not try to normalize as there is no structure enforced by the cve schema see:
https://github.com/CVEProject/cve-schema/blob/ce5f5c865f14dc40a6548d36b74751abca1c588a/schema/CVE_Record_Format.json#L801 and
https://github.com/CVEProject/cve-schema/blob/ce5f5c865f14dc40a6548d36b74751abca1c588a/schema/CVE_Record_Format.json#L107


collectors/cveorg/collectors.py
```diff
@@ -470,6 +470,9 @@ def get_refs(data: dict) -> list:
         def get_title(data: dict) -> str:
             return data["containers"]["cna"].get("title", "From CVEorg collector")
 
+        def get_affected(data: dict) -> dict:
+            return data["containers"].get("affected")
+
```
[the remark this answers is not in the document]
darakian: Good bot

JakubFrejlach:
@darakian What is the reason for this change? Is there any Jira which tracks this?

darakian:
Hey @JakubFrejlach, the goal of this change is to make some zero latency product info available for processing. There's no jira for it yet, but it's related to ~ARGUS-193~ ARGUS-253.

JakubFrejlach:
> Hey @JakubFrejlach, the goal of this change is to make some zero latency product info available for processing. There's no jira for it yet, but it's related to ~ARGUS-193~ ARGUS-253.

I am sorry, but there needs to be a relevant Jira (story in this case) for that in OSIDB project first. Also looking at ARGUS-253, is there a relevant AEGIS counterpart for this? AEGIS won't automatically start to use new fields added to Flaw.

JakubFrejlach (changes requested):
Requesting changes because there needs to be a OSIDB Jira first so this change is justified and thought through.

darakian:
Gotcha. OSIDB-4995 is up for you 👍

JakubFrejlach:
> Gotcha. OSIDB-4995 is up for you 👍

Thank you!

osidb/models/flaw/flaw.py
```diff
@@ -248,6 +248,9 @@ class FlawRequiresCVEDescription(models.TextChoices):
     # from cveorg collector
     mitre_cve_description = models.TextField(blank=True)
 
+    # from cveorg collector
+    cve_affected_block = models.JSONField(default=dict, blank=True)
```
JakubFrejlach: This should not be a direct `Flaw` field but rather field on the new `UpstreamData` model which was introduced in https://github.com/RedHatProductSecurity/osidb/pull/1274, see reasoning in this [discussion](https://github.com/RedHatProductSecurity/osidb/pull/1274#discussion_r3239719682)

collectors/cveorg/collectors.py
```diff
@@ -470,6 +470,9 @@ def get_refs(data: dict) -> list:
         def get_title(data: dict) -> str:
             return data["containers"]["cna"].get("title", "From CVEorg collector")
 
+        def get_affected(data: dict) -> dict:
+            return data["containers"]["cna"].get("affected") or []
```
JakubFrejlach, line 474: Return type annotation is `dict` but can return list, what is the correct type here?

osidb/models/flaw/flaw.py
JakubFrejlach: Similar question to the one above, what is the correct default type for this field? List or dict? Or can it be both? (if so, that would need some preprocessing so we can count having such data in one format.

osoukup (changes requested):
I had a look as I promised last 👼 I have a few comments. First, the commits should be consolidated - the changes seem on the same topic and are small so it should be simple. Also the new field should be moved to the `UpstreamDate` model to align with https://github.com/RedHatProductSecurity/osidb/pull/1274. Finally, there does not seem to be any usage of the new field and it is not being exposed on the API so in the current form the change does not seem to be very useful.

collectors/cveorg/collectors.py
darakian: Good catch. Based on the cve schema this should be a list
https://github.com/CVEProject/cve-schema/blob/ce5f5c865f14dc40a6548d36b74751abca1c588a/schema/CVE_Record_Format.json#L801-L805

Updated in https://github.com/RedHatProductSecurity/osidb/pull/1273/commits/9a838dc33501b0aee3cc8e21c5c51cdb05bf3f4c

osidb/models/flaw/flaw.py
darakian: Same as above, this should be a list. I altered the data type in https://github.com/RedHatProductSecurity/osidb/pull/1273/commits/9a838dc33501b0aee3cc8e21c5c51cdb05bf3f4c but this will likely get refactored to target the `UpstreamData` type. I need to read up on that before refactoring.

darakian:
@osoukup 

> I had a look as I promised last 👼 I have a few comments. First, the commits should be consolidated - the changes seem on the same topic and are small so it should be simple. Also the new field should be moved to the UpstreamDate model to align with https://github.com/RedHatProductSecurity/osidb/pull/1274.

Can do. I'll drop the commits that are related to the `Flaw` model and re-target

> Finally, there does not seem to be any usage of the new field and it is not being exposed on the API so in the current form the change does not seem to be very useful.

I was under the assumption that the data would be accessible via existing method to access the flaw class. I suppose that's not the case. Looking at 
https://github.com/RedHatProductSecurity/osidb/pull/1274/changes
it seems like the new model is available via api. Extending that model should maintain that availability correct?

osoukup:
> Looking at https://github.com/RedHatProductSecurity/osidb/pull/1274/changes it seems like the new model is available via api. Extending that model should maintain that availability correct?

Yes, you are correct :wink:

darakian:
Ok, I'm gonna remake this PR with a new branch. I should not have merged the master branch in. Sorry for the noise 🤦

PR #1274, opened 2026-05-12
OSIDB-4923, OSIDB-4953, OSIDB-4954: Add fields and collection for upstream data

Jincxz:
Add `UpstreamData` model attached to `Flaw` to hold data on upstream purls.  This data is currently gathered via the OSV collector.

Structure: `upstream_purls`: {purl, name, ecosystem, range, version} per affected

There are two fields that is unused but are added in as comments for potential future use at (the [OSIDB-4953-OSIDB-4954-upstream-fields](https://github.com/RedHatProductSecurity/osidb/tree/OSIDB-4953-OSIDB-4954-upstream-fields) branch):

`upstream_description`: a list of string(s) (each string is a description)
`upstream_severities`: either {severity, db_severity} OR {affect_severity, affect_db_severity} per affected

See fields at https://ossf.github.io/osv-schema/#field-details
Tests, model, and de-duplication added with help from cursor

_**Note that, for this implementation, data is only filled for new vulnerabilities collected (a.k.a ones that create new snippets). Updates from OSV are not currently reflected in the model.**_


osidb/models/flaw/flaw.py
```diff
@@ -293,6 +293,14 @@ class FlawRequiresCVEDescription(models.TextChoices):
         choices=FlawNistCvssValidation.choices, max_length=20, blank=True
     )
 
+    upstream_purls = models.JSONField(default=list, blank=True)
+
+    upstream_descriptions = fields.ArrayField(
+        models.TextField(blank=True), default=list, blank=True
+    )
+
+    upstream_severities = models.JSONField(default=list, blank=True)
```
osoukup: Could we add an object `upstream` with these attributes inside that so it is better separated? This seems to me as polluting the Flaw model.

osoukup (changes requested):
Could you please first explain how is this going to work as a process with Aegis? Aegis does not process CVE-less flaws - which is the extra value of OSV above CVE.org so there will be no extra flaws to analyses. If the OSV flaw gets CVE ID, it was first assigned in CVE.org so we already scrapped that flaw and Aegis already analysed that. What will be the extra value and why it is claimed to be for Aegis when it goes completely out of its workflow?

JakubFrejlach:
Adding to @osoukup comment. kdudka left a comment in OSIDB-4953 where he points out that AEGIS should be able to access this data once AEGIS-368 is done and AEGIS is able to read external references.

osidb/models/flaw/flaw.py
Elkasitu: We did discuss having some sort of `UpstreamRecord` or `UpstreamData` object as this will be implemented both for CVE.org and OSV and each implement their own fields which kinda do the same and would end up polluting the model. Kind of like when we had rh_cvss, nist_cvss, etc. fields (good times 😄 )

osidb/models/flaw/upstream.py
```diff
@@ -0,0 +1,81 @@
...
+
+    upstream_purls = models.JSONField(default=list, blank=True)
+
+    upstream_descriptions = fields.ArrayField(
+        models.TextField(blank=True), default=list, blank=True
+    )
+
+    upstream_severities = models.JSONField(default=list, blank=True)
+
+    source = models.CharField(max_length=10, blank=True)
```
osoukup: I would expect the sources to be pretty limited (enum) based on what collectors we have.

osoukup (approved):
The benefit of this change was clarified (not Aegis but newcli query improvement based on the upstream PURL version) in the Jira so I am not going to block this longer.

costaconrado (approved):
LGTM.

PR #1275, opened 2026-05-12
Upgrade Django to latest security revision

Elkasitu:
For some reason Dependabot failed to upgrade this dependency so I had to do it manually.


osoukup (approved):
LGTM

PR #1277, opened 2026-05-14
Workflow framework resurrection

osoukup:
This PR introduces a resurrection of the original workflow framework philosophy. I know it is huge and I apologise to the reviewers. Unfortunately, most of the changes do not make sense separately. There are a few considerations.

1. The classification (**workflow:status**) is a pure function the flaw data.
     - The old workflow is purely based on the user clicking and some transitions are not underlined by data at all. 
Most specifically the rejection. Therefore, I decided to base it on the presence of **rejected** label.
     - The workflow manipulation endpoints are clearly in contradiction with this and must be deprecated.
     - It does not matter in which workflow or state the flaw was, the new classification is always only based on the data - eg. DONE is not a state of the process, it is state of the data and it can be undone.
     - For the time being there is going to be human review at the final stage so I decided to use **approved** label as a way of confirming the right status of the data and proceeding to the DONE state.
2. There must be no external or cyclic classification dependencies.
     - This specifically means that the classification must not be synchronised from any external source. It would effectively rewrite the computed classification.
     - The old sync from Jira must be discontinued as that is external and moreover cyclic dependency. Only one-way sync to Jira is acceptable and will be persisted.
4. The classification is auto-adjusted on **save()**.
     - This is especially needed when we now shift to automated|agentic approach when multiple different actors collectively process the flaw. OSIDB ensure the workflow transitions when one actor is done allowing the next one to follow - they are not supposed to click buttons in any way, just do the actual work.
     - Classification auto-adjustion based on the data can lead to no change, progression through the states, but also regression if the previous requirements are unfulfilled and in the workflow change if conditions change.
     - This must also account for the changes in (at least some important) related models.
5. The state requirements should cover as much validations as possible.
     -  Validations are naturally stateless applied all the time but most of the data requirements do not have to hold always. Non-triaged flaw in the NEW state can rightfully miss attributes. Also we do not care for REJECTED flaws but validations effectively block the rejection of crappy flaws.
     - Unfortunately a lot of validations are quite complex and their description by the workflow checks is beyond its current capabilities. For now I am adding the straightforward ones and we can iterate on this.
6. The workflows are cumulative.
     - In state N+1 the requirements of all previous N states hold on top of the requirements of the N+1th state.
     - There can never be a requirement set in one state and contradicted in a consecutive state - unless the intention would be to block  (eg. to change the workflow instead of progressing forward - see EMBARGOED workflow).
7. There can be multiple different workflows with different conditions. They must have different priorities so the first one with the fulfilled conditions is applied. There must exist a DEFAULT workflow with empty conditions and the lowest priority.
8. The set of allowed states is fixed. Only a subset can be used for a particular workflow.
     - Not sure if this needs to hold in general in the future.
     - NEW is assumed to represent an initial point in the workflow while DONE the final one but that is a soft requirement. It is mostly for a convenient querying.
     - REJECTED is not a state, it is a workflow. Rejected flaw goes there and when done process-wise (which is immediately there) it is classified as REJECTED:DONE. This can be a breaking change for some integrations which do not check the workflow but the state.
10. Each workflow must have the first state without requirements so a flaw can always classify into some state there if it falls into the workflow based on the conditions and priority.
11. The workflow definition files are intended to serve as a source of truth about the workflows for both humans and machines and should therefore be well human-readable but also machine-parsable.
     - The documentation always desync with the implementation. When there is no implementation, only interpretation, so the source is only one, it can never desync.
     - These YAMLs should ideally live probably in PS constants and be co-owned by the PS BUs.

So these are the general ideas. The commits should demonstrate the though process if you want to follow it. There are also some minor changes and nice-to-haves like demonstrative graphical pages.

Note: I was using README.md as a bit of checklist so it is not necessary to review it in the individual commits but it is enough to check the final version. It should pretty much describe the considerations above plus a bit more.

**Demo:** https://drive.google.com/file/d/1JzksE-5sslGvdZ_9YaaYSljKFfILhwpU/view?usp=sharing


osoukup:
I am not sure why there are four extra commits being shown. Locally I do not see them in the git history between the two branches...

apps/workflows/serializers.py
```diff
@@ -122,22 +123,74 @@ def _accepts(self, instance):
...
 
 class RejectSerializer(serializers.Serializer):
-    """Task rejection serializer"""
+    """
+    Task rejection serializer (DEPRECATED)
+
+    Used by the deprecated reject endpoint. Will be removed when the endpoint is removed.
+    """
 
     reason = serializers.CharField()
```
dojalvo: Can/should we modify this to use the new endpoint alongside deprecating it?

osidb/api_views.py
```diff
@@ -1111,7 +1114,7 @@ def flaw_available(request: Request, *args, **kwargs) -> Response:
 
     if (
         flaw.is_public
-        or flaw.workflow_state == WorkflowModel.WorkflowState.REJECTED
+        or flaw.workflow_name == "REJECTED"
```
dojalvo, line 1117: This should probably still use the variable WorkflowModel.WorkflowState.REJECTED or something tracked in a variable

apps/workflows/workflow.py
dojalvo, line 214: Why not bail immediately when save is false?

openapi.yml
```diff
@@ -5765,8 +5762,12 @@ paths:
...
         workflow promotion API endpoint
 
-        try to adjust workflow classification of flaw to the next state available
-        return its workflow:state classification or errors if not possible to promote
+        DEPRECATED: Workflow classification is now automatic based on flaw data.
+        This endpoint no longer performs any action - it only returns the current
+        computed classification. This endpoint will be removed in a future version.
+
+        To change workflow state, update the flaw data directly (assign owner, create
+        affects, file trackers, etc.). Classification will update automatically.
```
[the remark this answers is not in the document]
osoukup: Thanks for the opinion but these are to be deprecated. Nobody should build any new stuff around these endpoints and the existing stuff does not need documentation. Let us leave it as it is.

openapi.yml
```diff
@@ -14050,6 +14182,14 @@ components:
...
           readOnly: true
+        created_by:
+          type: string
+          readOnly: true
+        updated_by:
+          type: string
+          readOnly: true
+        assist_meta:
+          nullable: true
         embargoed:
```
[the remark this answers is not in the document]
osoukup: This is from some previous commit not introduced by me here.

openapi.yml
```diff
@@ -17874,6 +18042,85 @@ components:
...
+      required:
+      - alerts
+      - created_dt
+      - embargoed
+      - flaw
+      - source
+      - updated_dt
+      - upstream_purls
+      - uuid
+      - visibility
```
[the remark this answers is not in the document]
osoukup: This is from some previous commit not introduced by me here.

osidb/models/affect.py
```diff
@@ -246,6 +246,16 @@ class AffectFix(models.TextChoices):
     # Stored as list of label names: ["label1", "label2", ...]
     labels = models.JSONField(default=list, blank=True)
 
+    # Who originally created this affect (user email, username, or system identifier)
+    created_by = models.CharField(blank=True)
+
+    # Who last updated this affect
+    updated_by = models.CharField(blank=True)
```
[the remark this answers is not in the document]
osoukup: This is from some previous commit not introduced by me here.

osidb/models/flaw/upstream.py
```diff
@@ -0,0 +1,87 @@
...
+    source = models.CharField(choices=Source.choices, max_length=10, blank=True)
+
+    objects = UpstreamDataManager.from_queryset(CustomQuerySetUpdatedDt)()
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["flaw", "source"],
+                name="unique_upstream_data_of_a_flaw",
+            ),
```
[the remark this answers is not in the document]
osoukup: This is from some previous commit not introduced by me here.

osidb/api_views.py
osoukup: Actually I am making REJECTED to be no more state but workflow. The state there is DONE to denote that the flaw is fully processed. The workflow then tells us how it is processed.

Based on some feedback from the demo I made yesterday on this, we might even drop the fixed set of states and leave the analysis teams to define their states and workflows as they like. There is not really a reason to make it fixed, it is just convenience - like DONE means done, NEW means unseen etc.

apps/workflows/workflow.py
osoukup: You mean at the top of the method?

osidb/serializer.py
```diff
@@ -2132,8 +2132,15 @@ class Meta:
...
+        # Alias and workflow labels are always relevant
+        if validated_data.get("type") in (
+            FlawLabel.FlawLabelType.ALIAS,
+            FlawLabel.FlawLabelType.WORKFLOW,
+        ):
             validated_data["relevant"] = True
+            # Workflow labels are always in DONE state
+            if validated_data.get("type") == FlawLabel.FlawLabelType.WORKFLOW:
+                validated_data["state"] = FlawCollaborator.FlawCollaboratorState.DONE
             return super().create(validated_data)
```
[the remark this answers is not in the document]
osoukup: Interesting point but we do not plan to implement any more granular ACLs on this now.

costaconrado (approved):
LGTM!!

Great job, few points I was concerned and paid special attention but you nailed it
- the change on workflows name could break some hard coded places but I checked and you have updated everything
- publishing flaws past triage could affect new embargoed workflow, but the condition there seems enough to prevent it
- I've also tried the endpoints for workflow and internal / embargoed are not accessible unauthenticated (seems to respect ACLs)

the one thing that I'm missing (not in this PR, but in the overall process we're building) is a way for vm team check whether all trackers are present before closing the flaw, since they will not fill trackers manually anymore it would be helpful to the reviewer there to know if any required tracker has failed (e.g. mismatch components with prod-defs / jira project, missing fields on jira) before he can close the flaw, those errors are currently stateless, maybe we should think in a label TrAm can add or some blocking alerts at flaw level for this

apps/workflows/workflow.py
dojalvo: Ah, I see that self.classification is being set now. Is the expectation that it would be saved later?

osoukup: It depends on the caller, this method just offers both. If it is called just to adjust the classification, then we can save it right away. However, this case I think does not even happen in reality. The classification adjustion is typically a part of a more complex save operation and it is not desired to `save()` multiple times during it so the caller will only ask for a re-classification but will save to DB later together with other changes.

apps/workflows/serializers.py
```diff
@@ -122,24 +123,82 @@ def _accepts(self, instance):
...
+                (
+                    ClassificationCheckSerializer(req, context=self.context).data
+                    if isinstance(req, Check)
+                    else ClassificationConditionSerializer(
+                        req, context=self.context
+                    ).data
+                )
+                for req in instance.requirements
+            ],
+        }
```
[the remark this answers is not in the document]
osoukup: This is not super important but maybe to be explicit the verbose output might have `selected` attribute so it is clear on the first look.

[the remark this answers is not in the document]
osoukup: I tried the `selected` approach but it is a bit noisy. Implicit classification information on what is selected and what not seems enough.

apps/workflows/tests/test_endpoints.py
```diff
@@ -104,15 +143,45 @@ def test_workflows_uuid_non_existing(self, auth_client, test_api_uri):
...
+
+    def test_graph_classification(self, auth_client, test_api_uri):
+        """test graph classification endpoint highlights accepting states"""
+        flaw = FlawFactory(embargoed=False, task_key="TASK-GRAPH")
+        response = auth_client().get(f"{test_api_uri}/graph/workflows/{flaw.uuid}")
+        assert response.status_code == 200
+        content = response.content.decode()
+        assert "mermaid" in content
+        assert "fill:#dbeafe" in content  # classified state highlighted in blue
+        assert "fill:#fee2e2" in content  # non-accepting states in red
```
JakubFrejlach: This test seems strange, it just asserts whether the color which was defined in the HTML template is actually used which feels a bit like testing whether constant has a value it was given in the code.

apps/workflows/tests/test_endpoints.py
```diff
@@ -104,15 +143,45 @@ def test_workflows_uuid_non_existing(self, auth_client, test_api_uri):
...
+
+    # graph/workflows
+    def test_graph_workflows(self, auth_client, test_api_uri):
+        """test graph workflows endpoint renders HTML with Mermaid diagrams"""
+        response = auth_client().get(f"{test_api_uri}/graph/workflows")
+        assert response.status_code == 200
+        content = response.content.decode()
+        assert "mermaid" in content
+        assert "flowchart LR" in content
+        assert "DEFAULT" in content
```
JakubFrejlach: This is essentially a smoke test, it only checks that the endpoint returns 200 and that the response contains hardcoded fragments from the template. It doesn't verify that the correct workflows are rendered or that the diagram content is meaningful.

osidb/serializer.py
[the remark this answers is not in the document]
osoukup: No. If it is needed in the future, we will deal with it.

apps/workflows/checks.py
```diff
@@ -135,6 +137,58 @@ def desc2not_property(self, check_desc):
                     lambda instance: not func(instance),
                 )
 
+    def desc2parameterized_method(self, check_desc):
+        """
+        parameterized method check
+        """
```
JakubFrejlach, line 162: Could we add some more detailed docstring on how this works and add known limitations? There is expected one parameter if I read the implementation correctly and also the parameter argument needs to have no spaces nor underscores in its values, otherwise the parsing logic will not handle that correctly.

apps/workflows/checks.py
```diff
@@ -135,6 +137,58 @@ def desc2not_property(self, check_desc):
...
+
+        doc = f"check that {self.model.__name__}.{method_name}({parameter!r}) returns True"
+
+        def call_parameterized(instance):
+            return func(instance, parameter)
+
+        return (doc, call_parameterized)
+
+    def desc2not_parameterized_method(self, check_desc):
+        """negative parameterized method check"""
```
JakubFrejlach: Same as for `desc2parameterized_method`.

apps/workflows/serializers.py
```diff
@@ -122,24 +123,81 @@ def _accepts(self, instance):
...
+        if not flaw or not instance.accepts(flaw):
+            return None
+        return instance.classify(flaw).name
+
+
+class ClassificationResultSerializer(serializers.Serializer):
+    """Serializer for the workflow:state classification result"""
+
+    workflow = serializers.CharField()
+    state = serializers.CharField()
```
JakubFrejlach, line 192: This could be `ChoiceField` with `WorkflowModel.WorkflowState` as choices.

docs/CHANGELOG.md
```diff
@@ -14,13 +14,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- auto-adjust workflow:state on Flaw.save
 
 ### Fixed
 - Exclude streams with no offer from tracker suggestions response (OSIDB-4977)
 
+### Removed
+- deprecate workflow manipulation endpoints
+- disconnect OSIDB workflow:state from the Jira
+- stop classifying legacy flaws
+
```
JakubFrejlach: These should go into unreleased section.

apps/workflows/api.py
```diff
@@ -5,33 +5,55 @@
...
 
+DEPRECATION_MESSAGE = (
+    "Workflow classification is now automatic based on flaw data. "
+    "This endpoint no longer performs any state changes. "
+    "To change workflow state, update the flaw data (owner, affects, trackers, etc.). "
+    "Use GET /workflows/api/v2/workflows/{id} to view computed classification."
+)
+
+
+class DeprecatedWorkflowMixin:
```
JakubFrejlach, line 34: In addition to this I would mark the endpoints as deprecated as well via OpenAPI schema. You can just use the [`extend_schema_view`](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema_view)  and feed it with the [`extend_schema(deprecated=True)`](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema) for each respective HTTP method, or you can decorate the methods directly with `extend_schema(deprecated=True)`.

apps/workflows/urls.py
```diff
@@ -16,12 +22,25 @@
...
+        graph_classification.as_view(),
+    ),
+    # V2 API
+    path(f"api/{WORKFLOWS_API_VERSION_V2}/workflows", workflows.as_view()),
+    path(
+        f"api/{WORKFLOWS_API_VERSION_V2}/workflows/<str:pk>", classification.as_view()
+    ),
+    path(f"api/{WORKFLOWS_API_VERSION_V2}/graph/workflows", graph_workflows.as_view()),
+    path(
+        f"api/{WORKFLOWS_API_VERSION_V2}/graph/workflows/<str:pk>",
```
JakubFrejlach: Do we even need `v2` endpoints here? 🤔 They seems identical to the `v1` endpoints with the exception of missing `/adjust`, but since we are not removing it but only deprecating it, it seems fine to me to just basically leave this as it is on `v1`.

JakubFrejlach (commented):
Great job! This seems very solid! I have a couple of comments, but nothing really serious.

jvaught-rh:
I had a similar comment as @costaconrado. I know the workflow definitions are outside of this PR, but from a user workflow and according to the [reqs](https://gitlab.cee.redhat.com/product-security/argus-p3/-/blob/main/docs/automated-flaw-analysis-workflow.md?ref_type=heads#state-secondary_assessment), after trackers are filed, the workflow state should update to secondary_assessment where a human must review the entire flaw. The user then updates/fixes anything and should take some action to approve or reject the flaw so that it i can finally be closed. 

So would that manual assessment step be a part of the definition of the workflow? Or do we need to consider anything special?

osoukup:
> LGTM!!
> 
> Great job, few points I was concerned and paid special attention but you nailed it
> 
>     * the change on workflows name could break some hard coded places but I checked and you have updated everything
> 
>     * publishing flaws past triage could affect new embargoed workflow, but the condition there seems enough to prevent it
> 
>     * I've also tried the endpoints for workflow and internal / embargoed are not accessible unauthenticated (seems to respect ACLs)
> 
> 
> the one thing that I'm missing (not in this PR, but in the overall process we're building) is a way for vm team check whether all trackers are present before closing the flaw, since they will not fill trackers manually anymore it would be helpful to the reviewer there to know if any required tracker has failed (e.g. mismatch components with prod-defs / jira project, missing fields on jira) before he can close the flaw, those errors are currently stateless, maybe we should think in a label TrAm can add or some blocking alerts at flaw level for this

@costaconrado `has trackers` condition of the `SECONDARY_ASSESSMENT` actually checks whether all affects where we expect the trackers have them. The flaw will not reach this state (where the manual review/approval happens) until there are trackers filed for all those affects.

apps/workflows/tests/test_endpoints.py
osoukup: Good point! I have not checked the tests for these optional APIs very much. I will ask Claude to fix it.

apps/workflows/tests/test_endpoints.py
osoukup: Same here.

apps/workflows/checks.py
osoukup: Will do.

apps/workflows/serializers.py
osoukup: This is a good point but the feedback I got from the demo is that it would be good to leave the states to be freely defined so in the next iteration I plan to change it to just `CharField` so I would leave it as it is.

docs/CHANGELOG.md
osoukup: Oh! Definitely. This is a complete oversight. I will fix it.

apps/workflows/api.py
osoukup: Good point. I will add it. I think change is so significant it would deserve a major release at the end so hopefully they will go away completely.

apps/workflows/urls.py
osoukup: You are correct. I am no more sure why I introduced them :thinking: I will consolidate on v1

osoukup: I probably first wanted to really remove them and later decided to only deprecate.

osoukup:
> I had a similar comment as @costaconrado. I know the workflow definitions are outside of this PR, but from a user workflow and according to the [reqs](https://gitlab.cee.redhat.com/product-security/argus-p3/-/blob/main/docs/automated-flaw-analysis-workflow.md?ref_type=heads#state-secondary_assessment), after trackers are filed, the workflow state should update to secondary_assessment where a human must review the entire flaw. The user then updates/fixes anything and should take some action to approve or reject the flaw so that it i can finally be closed.
> 
> So would that manual assessment step be a part of the definition of the workflow? Or do we need to consider anything special?

The workflow definitions in this PR do not contain the newly designed automation labels yet, only the manual ones (the auto ones will be added in the followup). However, the manual requirements are already there. To answer your question, DONE state contains `has label approved` requirement which ensures the human review.

apps/workflows/workflows/default.yml
```diff
@@ -1,54 +1,54 @@
...
     description: >
-      Queue captain has taken a ticket off the incoming queue and will
-      evaluate if it should be rejected or qualified for further steps in
-      the workflow.
+      Queue captain has taken the flaw off the incoming queue and will evaluate
+      whether it should be rejected or analysed further.
     jira_state: Refinement
     jira_resolution: null
     requirements:
       - has owner
```
oaubrey, line 21: We will need to add another alternative requirement here to account for the automation process since https://redhat.atlassian.net/browse/OSIDB-4950 can not currently be done. The requirement should allow for the state to transition to the `TRIAGE` state if the flaw has an 'auto-affects' label.

apps/workflows/workflows/default.yml
```diff
@@ -1,54 +1,54 @@
...
+          - has cve_description
 
   - name: SECONDARY_ASSESSMENT
-    description: An analyst assigned the task to themselves.
+    description: The trackers were filed. The flaw is ready for the review.
     jira_state: In Progress
     jira_resolution: null
     requirements:
-      - has owner
+      - has trackers
```
oaubrey, line 46: Do we need a secondary check here in the requirements for if the impact is low/ cvss score is < 7.0 or if the status is anything other than AFFECTED:DELEGATED? Not sure if that is covered.

osoukup: Status is covered. I do not think that anything else is needed. If it is low, has low CVSS score, or whatever special case not requiring trackers, simply **do not set** AFFECTED:DELEGATED. We have other resolutions like DEFER so if people don't abuse DELEGATED, we do not need anything complicated.

apps/workflows/workflows/default.yml
osoukup: You are correct, that will come with the next iteration.

PR #1279, opened 2026-05-14
OSIDB-4951, OSIDB-4978: Add auditing of tool-assisted, auto-created affects

Elkasitu:
This commit adds various fields to the Affect model which will be used to track which affects were created by which tools and by whom.

These should be able to provide enough audit information to replicate behavior, analyse and improve said behavior as well as know where to adjust or who to contact in case anything goes awry.


osidb/models/affect.py
```diff
@@ -246,6 +246,16 @@ class AffectFix(models.TextChoices):
     # Stored as list of label names: ["label1", "label2", ...]
     labels = models.JSONField(default=list, blank=True)
 
+    # Who originally created this affect (user email, username, or system identifier)
+    created_by = models.CharField(blank=True)
```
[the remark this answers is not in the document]
Elkasitu: I caught you slippin', `max_length` is not required for PostgreSQL backends since 4.2 https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.CharField.max_length

Jincxz (approved):
LGTM

osidb/models/affect.py
osoukup, line 250: Isn't this just duplicating the already existing history audit? The same for the update except that it only covers the last one while the history covers all.

osoukup (approved):
LGTM

osidb/models/affect.py
Jincxz: The fields do semi-duplicate the history audit but they allow affects that are directly created to have more context. For example, the history won't show that affects are created by the AffectCreationEngine since that is being done by OSIDB and not an external user. 

Are there potential concerns / considerations with somewhat duplicating the data? 
(side note: I have been asked by Jessie to help merge the PR for Adrian)

PR #1280, opened 2026-05-15
OSIDB-4977: Fix tracker suggestions with empty offer

JakubFrejlach:
This PR fixes trackers suggestions for affects v2. Currently there was a problem that if the affect was not eligible to receive a tracker `offer` field in the tracker suggestions response had `null` (`None`) value, which however is in contradict of what the OpenAPI schema says as this field is not nullable.


osoukup (approved):
LGTM

Jincxz (approved):

PR #1283, opened 2026-05-19
Separate ACLs changes of models from history

costaconrado:
This PR decouple the logic from updating ACLs on models from pghistory ACLs, making the later asynchronous.
This PR also adds a weekly task for fixing inconsistencies on affects and trackers.

I've expend some time creating tests and checking this but please pay extra attention in this review since it will impact ACLs of ~12k affects and ~3300 trackers on production at the first run of the task.

Closes OSIDB-4992.


osidb/sync_manager.py
```diff
@@ -1061,3 +1064,92 @@ def sync_task(task, tracker_id, **kwargs):
             JiraTrackerDownloadManager.failed(tracker_id, e)
         else:
             JiraTrackerDownloadManager.finished(tracker_id)
+
+
+class ACLHistorySyncManager(SyncManager):
+    """
+    Sync manager class for cascading ACL changes from model instances to their
+    corresponding pghistory audit entries. This allows ACL updates to happen
+    asynchronously without blocking users during operations like publicizing models.
```
osoukup, line 1073: nitpick: `publicizing` is a bit weird word 😅 I would say `publishing`

osidb/sync_manager.py
```diff
@@ -1061,3 +1064,92 @@ def sync_task(task, tracker_id, **kwargs):
...
+
+        :param instance: Model instance that inherits from ACLMixin
+        """
+        if not isinstance(instance, ACLMixin):
+            raise ValueError(
+                f"Instance must inherit from ACLMixin, got {type(instance).__name__}"
+            )
+
+        # Use the instance's UUID as the sync_id and store model info in kwargs
+        sync_id = str(instance.uuid)
```
osoukup, line 1092: This is probably highly unrealistic scenario but if the instances are of different classes I think there is a chance that the UUIDs could collide. Within the same class they are primary keys so they the uniqueness is ensured.

osidb/sync_manager.py
```diff
@@ -1061,3 +1064,92 @@ def sync_task(task, tracker_id, **kwargs):
...
+        super().schedule(
+            sync_id,
+            *args,
+            model_label=model_label,
+            **kwargs,
+        )
+
+    @staticmethod
+    @app.task(name="sync_manager.acl_history_sync", bind=True)
+    def sync_task(task, sync_id, model_label, **kwargs):
```
osoukup, line 1105: I do not recall the details but I have a memory of that the Celery task actually cannot have more then one parameter. I think the extra ones are simply being ignored. Are you sure that this works? If not, we could make `sync_id` to be something like class name plus UUID concatenated into a single string.

osidb/tasks.py
```diff
@@ -127,3 +129,84 @@ def async_send_email(**kwargs) -> int:
...
+        Q(acl_read=F("flaw__acl_read")) & Q(acl_write=F("flaw__acl_write"))
+    ).select_related("flaw")
+
+    for affect in affects_to_fix:
+        if affect.flaw:
+            with transaction.atomic():
+                # Update ACLs to match the parent Flaw
+                affect.acl_read = affect.flaw.acl_read
+                affect.acl_write = affect.flaw.acl_write
+                affect.save(raise_validation_error=False, auto_timestamps=False)
```
osoukup, line 159: I am thinking about the disabled `auto_timestamps` 🤔 If we assume a case when a flaw was already public but we failed to publish an affect, it can be therefore missing in the VEX or CVE pages. If we fix the ACLs and publish it but without the timestamp update, the downstream tools like SDEngine and Aggregator (if flaw was embargoed) will not get any information about the change and the downstream data will not fix.

osoukup (changes requested):
I have a few comments. Most importantly I think the Celery task parameters does not work with more than one. I may be wrong or there was some update but I would at least like us to double check that. I consider the cleanup job as a workaround which would not be needed if we do not fail with proper ACL changes when they are supposed to happen. I understand that currently we fail with that and the cleanup job is probably acceptable workaround but we should check time to time whether it still makes some updates or the discrepancies stop happening with async publishing.

PR #1285, opened 2026-05-19
OSIDB-4918: Update gunicorn keepalive time to match worker timeout

dojalvo:
Summary:
- Modify keepalive to match the worker timeout
- Add Emacs backup/autosave patterns to .gitignore


.gitignore
```diff
@@ -37,3 +37,12 @@ osidb_data_backup_dump*
...
 pyrightconfig.json
+
+# Emacs backup and autosave files
+*~
+\#*\#
+.\#*
+
+# Claude Code local state
+.claude/
+CLAUDE.md
```
Elkasitu: Instead of this I would suggest we pool up resources and come together with a single, shared CLAUDE.md... Feel free to contribute your own CLAUDE.md or ./claude rules in a separate PR

gunicorn_config.py
```diff
@@ -9,17 +11,26 @@
...
 # the gunicorn default for worker_tmp_dir is /tmp which may not reliably
 # exist in deployment environments, setting to shm filesystem avoids this
 worker_tmp_dir = "/dev/shm"
 
-if get_execution_env() in ["stage", "prod", "ci"]:
+if execution_env in ["stage", "prod", "ci"]:
     preload_app = True
     graceful_timeout = 800  # if a restart must happen then let it be graceful
-    keepalive = 60  # specifically this should be a value larger then nginx setting
+    keepalive = 60  # specifically this should be a value *smaller* then nginx setting
```
Elkasitu: `keepalive` != `timeout`, we also don't use nginx so we should use `reverse-proxy` or `load-balancer`.

the actual timeout is at the beginning of the file and it matches exactly the value set at the reverse-proxy level.

Elkasitu (changes requested):
OSIDB-5005 is IMHO a non-issue, allowing arbitrary X-Forwarded headers is the reverse-proxy's responsibility, not the application's, our current deployment of OSIDB has the reverse-proxy replace these headers if they're provided by the client, so there's no risk.

Furthermore, we don't have full control of the underlying network so statically setting these IPs can lead to issues down the line if the IPs change unbeknownst to us.

As for OSIDB-4918 I think the description is maybe not the best, but the problem is that some downstream clients get a "Connection Closed" error, which can happen due to many things (it could be the timeout, graceful_timeout, keepalive, HAProxy timeout, rate-limiting, etc.) so we may need to adjust those values.

.gitignore
dojalvo: I'm not opposed. I've been preferring to use CLAUDE.md has a local AI instance then using the other *.md files to store relevant information for the project. But, I'm seeing different patterns emerge.

gunicorn_config.py
dojalvo: I see. I think I misunderstood. Keepalive has to do on waiting for a response from the client , not the workers. This keepalive actually seems to be pretty long. Without a load-balancer, the timeout is only 1-5 seconds. Should a connection be open an additional 60 seconds if the client just stops responding?

https://gunicorn.org/reference/settings/#keepalive

dojalvo:
If there's no way to determine the HAProxy IP address at run and/or deploy time, then we probably can't do this, but this is a pretty useful defense-in-depth measure in case something goes wrong.

.gitignore
```diff
@@ -37,3 +37,12 @@ osidb_data_backup_dump*
 .DS_Store
 # pyright local configuration file
 pyrightconfig.json
+
+# Emacs backup and autosave files
```
osoukup, line 41: nitpick: This does not seem to be related to GUnicorn so I would put it in a separate commit.

gunicorn_config.py
```diff
@@ -9,17 +11,26 @@
 reuse_port = True
 
 errorlog = "-"
-# Make sure wsgi.url_scheme gets set to HTTPS, by trusting the X_FORWARDED_PROTO header set by the proxy
-forwarded_allow_ips = "*"
+
+# Set GUNICORN_FORWARDED_ALLOW_IPS environment variable to the HAProxy IP/subnet.
+# Examples: "10.0.1.12", "10.0.0.0/8", "127.0.0.1"
+execution_env = get_execution_env()
+if execution_env in ["stage", "prod", "uat"]:
```
osoukup: Why do we have special handling for certain environments? Wouldn't it be enough to just set the env var appropriately in each of them (OPS repo) and read it?

gunicorn_config.py
```diff
@@ -9,17 +11,26 @@
...
+else:
+    # Local/dev/ci environments: allow all for development convenience
+    forwarded_allow_ips = getenv("GUNICORN_FORWARDED_ALLOW_IPS", "*")
 
 # the gunicorn default for worker_tmp_dir is /tmp which may not reliably
 # exist in deployment environments, setting to shm filesystem avoids this
 worker_tmp_dir = "/dev/shm"
 
-if get_execution_env() in ["stage", "prod", "ci"]:
+if execution_env in ["stage", "prod", "ci"]:
```
osoukup: There is yet another hard-coded set of environments. I would be for just making the values configurable by env vars with some reasonable defaults so the environments with special needs can configure the values themselves however they need.

.gitignore
```diff
@@ -42,7 +42,3 @@ pyrightconfig.json
 *~
 \#*\#
 .\#*
-
-# Claude Code local state
-.claude/
-CLAUDE.md
```
osoukup: This should be squashed if we add it in one commit and remove in the next one.

gunicorn_config.py
```diff
@@ -30,7 +34,7 @@
 if execution_env in ["stage", "prod", "ci"]:
     preload_app = True
     graceful_timeout = 800  # if a restart must happen then let it be graceful
-    keepalive = 60  # specifically this should be a value *smaller* then nginx setting
+    keepalive = 300  # Must be >= HAProxy's timeout to prevent race conditions (OSIDB-5005)
```
osoukup: This value is being changed through multiple commits. Could that be just done once?

docs/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Exclude streams with no offer from tracker suggestions response (OSIDB-4977)
 
 ### Changed
-- Improve Gunicorn security configuration by restricting forwarded_allow_ips to specific proxy IPs in production environments via GUNICORN_FORWARDED_ALLOW_IPS environment variable
+- Extend Gunicorn keepalive time to 300 seconds to prevent race conditions (OSIDB-4918)
```
osoukup: This is also changed through multiple commits. Squash it before the review please.

osoukup (changes requested):
Could you please first squash the changes so they are not added and removed left and right so I can understand what is the though process behind them?

osoukup (approved):
LGTM, nice and clear!

Elkasitu (approved):
LGTM, please squash into two commits (emacs + keepalive) before merging

PR #1286, opened 2026-05-21
OSIDB-4927-fix: exclude orphaned trackers from v1 flaw and tracker endpoints

svelamak:
Fixed orphaned trackers leaking into the v1 flaw endpoint when filtering by tracker_ids. 
-Query real Affect model instead of AffectV1 view in `tracker_ids_filter` and `get_tracker_uuids_from_affect_v1_filter`.
-Added test case.
Closes: OSIDB - 4927.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed v1 API endpoints to properly exclude orphaned trackers when filtering by `tracker_ids`.

* **New Features**
  * Added filter parameters for affects: `is_tool_created`, `amended`, `created_by`, and `updated_by` with `icontains` lookup support.


osidb/tests/endpoints/flaws/test_flaws.py
```diff
@@ -1951,6 +1951,47 @@ def test_list_flaws_tracker_ids(self, auth_client, test_api_v2_uri):
...
+        refresh_v1_view()
+
+        # Close tracker first, then delete the affect to make tracker orphaned
+        # but affect_v1 view may still have stale data
+        tracker.status = "Closed"
+        tracker.save(raise_validation_error=False)
+        affect.delete()
+
+        response = auth_client().get(
+            f"{test_api_uri}/flaws?tracker_ids={tracker.external_system_id}"
```
costaconrado: the API you're using to validate is the v2, not the v1 that you actually want to test.

osidb/filters.py
```diff
@@ -990,12 +990,9 @@ def affect_trackers_errata_datetime_filter(self, queryset, name, value):
         return queryset.filter(uuid__in=flaw_ids)
 
     def tracker_ids_filter(self, queryset, name, value):
-        tracker_uuids = Tracker.objects.filter(
-            external_system_id__in=value
-        ).values_list("uuid", flat=True)
 
         flaw_ids = (
-            AffectV1.objects.filter(all_tracker_ids__overlap=list(tracker_uuids))
+            Affect.objects.filter(tracker__external_system_id__in=value)
```
costaconrado, line 994: although this would solve the REST API issue for this filter, this answer don't solves the issue at the AffectsV1 level, please further investigate why the read-only table is being malformed and try to solve on the root source so other possible filters also get fixed in the process. (e.g. will the orphaned trackers appear if I query `/v1/affects/<id>?include_fields=trackers` ?)

costaconrado (changes requested):

osidb/tests/endpoints/flaws/test_flaws.py
svelamak: Updatd test to use v1 endpoint

osidb/filters.py
svelamak: Replaced all `all_tracker_ids` references from the stale `AffectV1` materialized view with real `Affect` FK traversal in both `serializers.py` and `filters.py`. This covers all related filter methods.

osoukup (approved):
LGTM

costaconrado (approved):
LGTM, thanks for addressing my concerns

PR #1288, opened 2026-05-29
OSIDB-4949-Add-automation-label-types-for-automated-analysis

svelamak:
Adds 4 new label types (`manual_triage, auto_affects, auto_rejected, potential_rejection`) to support the automated analysis pipeline:
-Added 4 new FlawLabelType choices also added AUTOMATION_LABEL_TYPES constant on FlawCollaborator along with `_validate_automation_prerequisites()` and `_enforce_mutual_exclusivity()` and updated `save()` in `label.py`
-Updated `FlawLabelView.destroy()` to block deletion of automation labels in `api_views.py`
-Added automation label types in `serializer.py` and validations to block manual POST of automation labels
-Migration file to register labels in FlawLabel catalog
-Added tests in `test_labels.py` and `test_flaw_collaborator.py`
-Added entry in `CHANGELOG.md`
-Updated `openapi.yml` with new label types
Closes:OSIDB-4949


osidb/models/flaw/label.py
```diff
@@ -36,6 +37,10 @@ class FlawLabelType(models.TextChoices):
         CONTEXT_BASED = "context_based"
         PRODUCT_FAMILY = "product_family"
         BU = "bu"
+        MANUAL_TRIAGE = "manual_triage"
+        AUTO_AFFECTS = "auto_affects"
+        AUTO_REJECTED = "auto_rejected"
+        POTENTIAL_REJECTION = "potential_rejection"
```
osoukup: These sounds like individual labels, not label types.

osidb/models/flaw/label.py
```diff
@@ -191,6 +206,48 @@ def save(self, *args, **kwargs):
...
+                {"flaw": "Flaw must have components to apply automation labels."}
+            )
+        if flaw.workflow_state != WorkflowModel.WorkflowState.NEW:
+            raise ValidationError(
+                {"flaw": "Flaw must be in a NEW state to apply automation labels."}
+            )
+        if flaw.embargoed:
+            raise ValidationError(
+                {"flaw": "Automation labels cannot be applied to embargoed flaws."}
+            )
```
osoukup: The state-less validations are quite against the state-aware workflow requirements. I think this should be first re-thinked in the context of the workflow framework redesign. Are we clear on who/what will be adding which label and under what conditions? Why should they only be applied in the NEW state? Why not to embargoed flaws - are they even visible by the automation in question? Shouldn't CVE ID restriction rather be a workflow requirement?

osidb/models/flaw/label.py
```diff
@@ -191,6 +206,48 @@ def save(self, *args, **kwargs):
...
+        if self.type == FlawLabel.FlawLabelType.POTENTIAL_REJECTION:
+            return
+        FlawCollaborator.objects.filter(
+            flaw=self.flaw,
+            type__in=[
+                FlawLabel.FlawLabelType.MANUAL_TRIAGE,
+                FlawLabel.FlawLabelType.AUTO_AFFECTS,
+                FlawLabel.FlawLabelType.AUTO_REJECTED,
+            ],
+        ).delete()
```
osoukup, line 234: This looks pretty confusing to me. So if a flaw is eg. auto-rejected, I can set it potentially rejected to make it not be auto-rejected any more? Why not to do that by removing the auto-rejection? Also when affectes were auto-created and then the flaw was auto-rejected, does that mean that the affects are no more auto-created? What if the triage was manual, the affects must not be auto-created? I think that we should just use the state requirements if we want some process guardrails and not state-less validations.

osoukup (changes requested):
Having a special type for every single label does not make sense to me. Type should be something general. More over, why not to just use the `workflow` label type introduced at https://github.com/RedHatProductSecurity/osidb/pull/1277 ? I do not think that we should introduce a new label type for every label use-case we can think about. Personally, I do not even see the BU labels a great addition but I came late, my fault.

Safeguarding the workflow by state-less validations is wrong. Those should be used for the data integrity, not to drive the workflow. We should think about how to achieve similar effect by the workflow framework. If we see some limitations, we can of course extend that but going with the validations instead is a step back.

osidb/models/flaw/label.py
svelamak: You’re right, these are individual labels not types. I’ll remove the 4 new `FlawLabelType` values and use `WORKFLOW` type with name strings instead.

osidb/models/flaw/label.py
svelamak: `auto-affects`, `auto-rejected`, and `manual-triage` are mutually exclusive, `potential-rejection` is additive alongside `auto-affects` only and cannot coexist with `auto-rejected`. I’ll just rewrite it to filter by `type=WORKFLOW` and label name string.

osidb/models/flaw/label.py
svelamak: The labels are applied by an automated pipeline service account. The NEW state restriction and CVE/embargoed checks came from the spec for this ticket, the reasoning being that the pipeline only processes flaws that have just arrived and are unembargoed. That said, you raise a fair point about whether embargoed flaws are even visible to the automation, and whether CVE and state should be workflow requirements rather than model guards. I’ll remove `_validate_automation_prerequisites()` from the model and move these conditions into `default.yml`. Could you let me know if that's the right place to make the change?

osidb/models/flaw/label.py
osoukup: The workflow type is not restricted to any specific label names so we do not have to define them individually, the users can just add any labels however they want.

osidb/models/flaw/label.py
osoukup: I need to revisit OSIDB-4949 and try to put there some suggestions on how to connect it with the workflow framework.

osoukup (changes requested):
Is this PR still relevant? It seems updated to me. I do not think that we need any extra logic regarding the new labels. The workflow framework enables generic introduction of any labels we want and the logic around them should rather be defined through the workflow YAMLs so it is state-aware. Also any API restrictions are a misunderstanding, these labels must be allowed to be normally operated through the API as these automations will live outside of OSIDB.

svelamak:
Closing as obsolete. Workflow framework redesign handles label management.

PR #1290, opened 2026-06-03
OSIDB-4992 - Fix inconsistent transactions

costaconrado:
This PR asserts atomic changes on ACLs preventing inconsistencies.
This PR also centralize ACLs fixture to root conftest so we don't have duplicated fixtures across the codebase.

Note that I've separated the cleanup from this task, now it's mapped on OSIDB-5037 and the perfomance issues, that is on OSIDB-4999.

Closes OSIDB-4992.

### DIFF
--- a/osidb/tests/test_acl_consistency.py
+++ b/osidb/tests/test_acl_consistency.py
@@ -1,0 +1,159 @@
+"""
+Test ACL consistency and propagation.
+
+Django's ATOMIC_REQUESTS=True ensures that all HTTP requests are atomic.
+These tests verify that ACL updates propagate correctly to all nested entities.
+"""
+
+from datetime import datetime, timezone
+from unittest.mock import patch
+
+import pytest
+
+from osidb.models import Affect, Flaw, Tracker
+from osidb.tests.factories import (
+    AffectFactory,
+    FlawFactory,
+    PsModuleFactory,
+    PsUpdateStreamFactory,
+    TrackerFactory,
+)
+
+
+class TestAtomicRollbackOnFailure:
+    """Verify transactions rollback correctly when failures occur during ACL updates."""
+
+    @pytest.mark.django_db
+    def test_rollback_when_recursive_unembargo_fails(
+        self,
+        embargoed_read_groups,
+        embargoed_write_groups,
+    ):
+        """When recursive unembargo of nested entity fails, all changes rollback."""
+        ps_module = PsModuleFactory()
+        ps_stream = PsUpdateStreamFactory(ps_module=ps_module)
+
+        flaw = FlawFactory(
+            embargoed=True,
+            acl_read=embargoed_read_groups,
+            acl_write=embargoed_write_groups,
+        )
+
+        affect = AffectFactory(
+            flaw=flaw,
+            ps_module=ps_module.name,
+            ps_update_stream=ps_stream.name,
+            ps_component="test-component",
+            affectedness=Affect.AffectAffectedness.AFFECTED,
+            acl_read=embargoed_read_groups,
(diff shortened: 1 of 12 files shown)


osoukup (approved):
LGTM

apps/workflows/workflow.py
```diff
@@ -407,10 +407,11 @@ def adjust_acls(self, save=True):
...
-            self.set_public_nested()
-            self.set_history_public()
+            with transaction.atomic():
+                self.set_public()
+                # updates ACLs of all related objects except for snippets
+                self.set_public_nested()
+                self.set_history_public()
 
         if save:
             self.save()
```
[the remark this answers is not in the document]
Elkasitu: @costaconrado I think this comment is valid, we should not add the transaction block here but instead at the call-sites, kind of like how you did in `unembargo()`

osidb/tests/test_acl_consistency.py
```diff
@@ -0,0 +1,411 @@
...
+from osidb.tests.factories import (
+    AffectFactory,
+    FlawFactory,
+    PsModuleFactory,
+    PsUpdateStreamFactory,
+    TrackerFactory,
+)
+
+
+class TestACLPropagation:
```
Elkasitu: Did any of these tests fail before and succeed now? Otherwise I'm not convinced that they add anything over existing tests

osidb/tests/test_acl_consistency.py
```diff
@@ -0,0 +1,411 @@
...
+class TestAtomicRequestConfiguration:
+    """Verify Django's ATOMIC_REQUESTS is enabled for consistency."""
+
+    def test_atomic_requests_enabled(self, settings):
+        """Django ATOMIC_REQUESTS setting ensures all HTTP requests are atomic."""
+        assert settings.DATABASES["default"]["ATOMIC_REQUESTS"] is True, (
+            "ATOMIC_REQUESTS must be True to ensure ACL update consistency. "
+            "This wraps every HTTP request in a database transaction, "
+            "guaranteeing all-or-nothing behavior for ACL updates."
+        )
```
Elkasitu: This test is a bit pedantic

osidb/tests/test_acl_consistency.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+class TestAtomicRollbackOnFailure:
+    """Verify transactions rollback correctly when failures occur during ACL updates."""
+
+    @pytest.fixture
+    def embargoed_flaw_with_nested_entities(
+        self, embargoed_read_groups, embargoed_write_groups
+    ):
```
Elkasitu: AIs have a bad habit of creating fixtures in the same place as the test, maybe we should update the rules / CLAUDE.md to direct it to always create fixtures in `conftest.py`

osidb/tests/test_acl_consistency.py
costaconrado: yes, both test_set_history_public_failure_during_workflow_transition and test_rollback_when_recursive_unembargo_fails prior the changes, I'll remove the other ones.

osidb/tests/test_acl_consistency.py
costaconrado: test removed.

osidb/tests/test_acl_consistency.py
costaconrado: test removed.

osoukup (approved):
LGTM

costaconrado:
@coderabbitai resolve

I've addressed the acl test, but conftest and changelog requests are not part of this issue nor this PR, also the transaction includes jira request on purpose since jira is the authoritative source of this data.

PR #1292, opened 2026-06-04
OSIDB-4776: Include comment_zero in community tracker description

svelamak:
Add comment_zero to the community tracker template. 
-Appended flaw title and `comment_zero` for each flaw in `common.py`
-Updated `test_common.py `
Closes:OSIDB-4776

## Release Notes

* **New Features**
  * Community tracker descriptions now include additional flaw comments, providing users with more comprehensive context about tracked issues.

### DIFF
--- a/apps/trackers/common.py
+++ b/apps/trackers/common.py
@@ -263,6 +263,8 @@ class TrackerQueryBuilder:
             "best effort basis. Package maintainers are required to ascertain if the flaw indeed "
             "affects their package, before starting the update process."
         )
+        for flaw in self.flaws:
+            description_parts.append(flaw.comment_zero)
         return description_parts
 
     def _description_embargoed(self):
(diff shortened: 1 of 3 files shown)


apps/trackers/common.py
```diff
@@ -263,6 +263,9 @@ def _description_community(self):
             "best effort basis. Package maintainers are required to ascertain if the flaw indeed "
             "affects their package, before starting the update process."
         )
+        for flaw in self.flaws:
+            description_parts.append(f"{flaw.title}")
```
JakubFrejlach: Why is `title` being added here? Jira issue only talks about adding the `comment_zero` here and also the `title` is already included in the tracker summary (see  `def summary(self):`)

svelamak: Removed the title, Good catch, Thanks for the feedback.

JakubFrejlach (approved):
LGTM now! Also do the same changes in the Tracker AutoManager tool.

PR #1300, opened 2026-06-10
workflow framework improvements

osoukup:
This PR at the end evolved into the second (and hopefully last) part of the workflow framework resurrection. The base was already established in the previous part and this is more about adjusting it according to the Argus requirements - though I am still pending some feedback so some details may change, like shifting a requirement to another state etc.

This PR also introduces some new functionality. Most notable is the shift of the automatic visibility changes from being hard-coded to being defined by the workflows, then introducing some extra workflow endpoint params to get the next steps and classification related audit log - this is mainly to provide human understandable way of presenting the past workflow changes as with automatic agents and auto-classification, it may not be clear on the first look what actually happened.

OSIDB-5049


osoukup:
**make lint** does not seem to be enough to make lint :thinking:

dojalvo (commented):
I strongly recommend having something (strenums, pre-defined variables, etc)  to store the previous Workflow state options and not use strings everywhere. Otherwise maintenance is going to be pretty painful.

osoukup:
> I strongly recommend having something (strenums, pre-defined variables, etc) to store the previous Workflow state options and not use strings everywhere. Otherwise maintenance is going to be pretty painful.

I understand the point but I am thinking what kind of maintenance is going to be an issue. There could be some desync with the YAML definitions and the stored states - like we remove an existing name entirely and provide no data migration. The enum will not prevent it either. Or you mean we will loose the sense of meaning of the names? Already the tooling has little knowledge about it. There are some hard-coded conditions but the only actual sense of meaning is defined in the YAMLs themselves and again making it predefined in the code will I think not help.

The aim is to get eventually get the definitions out of the tooling entirely the similar way as we already did with the SLA/SLO policies (eg. https://gitlab.cee.redhat.com/prodsec-dev/ps-constants/-/blob/master/data/sla_policies.yml). There are two important reasons. First is to give the real ownership to the BUs as the DEV team should not be defining how they work. Second is to make two different sources of the truth - documentation and code - which always desync be joint in one both human and machine readable. I think the policies are good example where it works well and they have no hard-coded counterpart, the tooling just interprets the YAML definitions.

apps/bbsync/query.py
```diff
@@ -199,8 +199,7 @@ def generate_component(self):
         component = "vulnerability"
 
         if (
-            self.flaw.workflow_state == Flaw.WorkflowState.NEW
-            or self.flaw.workflow_name == "REJECTED"
+            self.flaw.workflow_state == "NEW" or self.flaw.workflow_name == "REJECTED"
         ) and self.flaw.meta_attr.get("bz_component") in [None, "vulnerability-draft"]:
```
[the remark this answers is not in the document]
osoukup: This is a good comment but I intentionally left the Bugzilla related part as it is since it is to be deprecated as soon as we can so it does not need any more maintenance.

osidb/api_views.py
```diff
@@ -1112,11 +1111,7 @@ def flaw_available(request: Request, *args, **kwargs) -> Response:
     except ValidationError:
         return Response(status=HTTP_400_BAD_REQUEST)
 
-    if (
-        flaw.is_public
-        or flaw.workflow_name == "REJECTED"
-        or flaw.workflow_state == WorkflowModel.WorkflowState.DONE
-    ):
+    if flaw.is_public or flaw.workflow_state == "DONE":
```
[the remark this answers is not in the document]
osoukup: This is a good comment but I left this here intentionally. This piece of code is specifically used by a legacy client integration which should hopefully be soon deprecated so this whole functionality would go away.

osidb/mixins.py
```diff
@@ -302,6 +371,34 @@ def is_internal(self):
...
+        target = ACLMixinVisibility(target)
+        if self.visibility >= target:
+            return
+
+        if target == ACLMixinVisibility.PUBLIC:
+            self.set_public()
+            self.set_public_nested()
+            self.set_history_public()
+        elif target == ACLMixinVisibility.INTERNAL:
+            self.set_internal()
```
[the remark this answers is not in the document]
osoukup: I will generalize the methods to work in both cases.

osidb/models/tests/test_flaw_collaborator.py
```diff
@@ -162,12 +150,12 @@ def test_create_from_flaw(self):
...
-            WorkflowModel.WorkflowState.PRE_SECONDARY_ASSESSMENT,
-            WorkflowModel.WorkflowState.SECONDARY_ASSESSMENT,
-            WorkflowModel.WorkflowState.DONE,
+            "",
+            "NEW",
+            "TRIAGE",
+            "PRE_SECONDARY_ASSESSMENT",
+            "SECONDARY_ASSESSMENT",
+            "DONE",
         ],
```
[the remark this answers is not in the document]
osoukup: I do not think this matter much. It can always have a better coverage.

osidb/signals.py
```diff
@@ -159,35 +158,15 @@ def updated_local_updated_dt_affectcvss(sender, instance, **kwargs):
...
-        == WorkflowModel.WorkflowState.PRE_SECONDARY_ASSESSMENT
-    ):
-        if instance._state.adding:
-            FlawCollaborator.objects.create_from_affect(instance)
-        else:
-            FlawCollaborator.objects.mark_irrelevant(instance.flaw)
+    if instance._state.adding:
+        FlawCollaborator.objects.create_from_affect(instance)
+    else:
+        FlawCollaborator.objects.mark_irrelevant(instance.flaw)
```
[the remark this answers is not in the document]
osoukup: I will fix this.

apps/workflows/api.py
```diff
@@ -291,6 +303,20 @@ def get(self, request, pk):
...
+        if next_param is not None:
+            if str2bool(next_param, "next"):
+                state_index = workflow.states.index(state)
+                if state_index + 1 < len(workflow.states):
+                    next_state = workflow.states[state_index + 1]
+                    response["next"] = ClassificationStateSerializer(
+                        next_state, context={"flaw": flaw}
+                    ).data
+                else:
+                    response["next"] = None
```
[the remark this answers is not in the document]
osoukup: Should be resolved now.

roduran-dev:
> **make lint** does not seem to be enough to make lint 🤔

I had the same issue, seems like ruff behaves different depending on the environment but I couldn't find what was different.

roduran-dev (approved):
Awesome change! now this should allow to go to the latest state automatically.

From what I see this will make the system to set the workflow state automatically not base on the current state but in the state general flaw. So now a flaw can end up in a prior state (like moving from `SECONDARY_ASSESSMENT` to `TRIAGE`)

I'm not a fan of changing variables with strings (its usually the other way around :rofl: ) but I understand is to avoid  maintenance  issues

apps/workflows/serializers.py
```diff
@@ -99,9 +99,25 @@ class WorkflowSerializer(serializers.Serializer):
...
+            "conditions": [
+                (
+                    CheckSerializer(cond).data
+                    if isinstance(cond, Check)
+                    else ConditionSerializer(cond).data
+                )
+                for cond in instance.conditions
+            ],
+            "states": [StateSerializer(state).data for state in instance.states],
+        }
```
JakubFrejlach: This seems a bit overkill, basically we are only interested in `conditions` serialization and yet we need to redo all other fields, why not just use  `SerializerMethodField`  and then it would look like this:

```python

  conditions = serializers.SerializerMethodField()

  def get_conditions(self, instance):
      return [
          CheckSerializer(cond).data if isinstance(cond, Check)
          else ConditionSerializer(cond).data
          for cond in instance.conditions
      ]
```

apps/workflows/serializers.py
```diff
@@ -174,10 +190,35 @@ def to_representation(self, instance):
...
+                    ).data
+                )
+                for cond in instance.conditions
+            ],
+            "states": [
+                ClassificationStateSerializer(state, context=self.context).data
+                for state in instance.states
+            ],
+            "classified_state": self.get_classified_state(instance),
+        }
```
JakubFrejlach: Same question [as above](https://github.com/RedHatProductSecurity/osidb/pull/1300/changes#r3460371345).

osidb/migrations/0241_merge_workflow_improvements.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("osidb", "0237_delete_jirataskdownloadmanager"),
+        ("osidb", "0240_workflow_state_free_form"),
+    ]
+
+    operations = []
```
JakubFrejlach, line 13: What is this migration purpose?

osidb/mixins.py
```diff
@@ -302,6 +371,35 @@ def is_internal(self):
...
+    @property
+    def visibility(self):
+        if self.is_embargoed:
+            return ACLMixinVisibility.EMBARGOED
+        if self.is_internal:
+            return ACLMixinVisibility.INTERNAL
+        return ACLMixinVisibility.PUBLIC
+
+    @visibility.setter
+    def visibility(self, target):
```
JakubFrejlach, line 383: Just a sanity check, won't this clash with the `visbility` annotation in the `ACLMixinManager` which is needed for performing queries using the `visibility` fields?

osidb/serializer.py
```diff
@@ -447,29 +447,16 @@ class ACLMixinSerializer(BaseSerializer):
             "modifies the ACLs but is mandatory as it controls the access to the resource."
         ),
     )
-    visibility = serializers.SerializerMethodField(
-        help_text="The visibility level of the resource based on ACL read groups."
+    visibility = serializers.CharField(
+        read_only=True,
+        help_text="The visibility level of the resource based on ACL read groups.",
     )
```
JakubFrejlach: There is a regression in the API schema, visibility is now only string and not enum with `ACLMixinVisibility` values, serializers.ChoiceField(choices=ACLMixinVisibility)` needs to be used instead.

osidb/serializer.py
```diff
@@ -1822,29 +1809,16 @@ class FlawPackageVersionACLMixinSerializer(serializers.ModelSerializer):
             "modifies the ACLs but is mandatory as it controls the access to the resource."
         ),
     )
-    visibility = serializers.SerializerMethodField(
-        help_text="The visibility level of the resource based on ACL read groups."
+    visibility = serializers.CharField(
+        read_only=True,
+        help_text="The visibility level of the resource based on ACL read groups.",
```
JakubFrejlach: There is a regression in the API schema, visibility is now only string and not enum with ACLMixinVisibility values, serializers.ChoiceField(choices=ACLMixinVisibility)` needs to be used instead.

JakubFrejlach (changes requested):
There is a lot of great work here! Very nice improvements. I've noticed couple things which should be changed and I've raised couple of questions, but nothing really serious.

apps/workflows/serializers.py
osoukup: Will change

osidb/migrations/0241_merge_workflow_improvements.py
osoukup: Discussed in person.

osidb/mixins.py
osoukup: It should not collide and the reason is 1) they compute the same way 2) annotation is used in the query sets where the property is not yet available and the property overrides it otherwise.

osidb/serializer.py
osoukup: Will change

JakubFrejlach (approved):
LGTM now!

PR #1301, opened 2026-06-11
OSIDB-4734-skip non-DB  fields without raising FieldDoesNotExist

svelamak:
Non-DB fields cause `FieldDoesNotExist` to be raised on every request.
-Fix in `filters.py` check field names against `get_fields()` before calling `get_fields()`.
-Added test case.
Closes: OSIDB-4734

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the `/flaws` API endpoint incorrectly raised errors when using non-database declared fields in filter parameters. The endpoint now properly accepts and processes these fields without raising exceptions. This improves the flexibility of the filtering system for API consumers who need to work with extended field sets.


osoukup (commented):
It took me a bit of staring before I understood the issue 😅 but LGTM

osoukup (approved):

PR #1303, opened 2026-06-15
feat: add Datadog APM integration for distributed tracing

AlejandroMinaya:
Integrates Datadog APM for distributed tracing across Django, Celery workers, and Celery
  beat. See docs/developer/DATADOG_APM.md for details.


docker-compose.yml
```diff
@@ -291,5 +317,23 @@ services:
...
+      image: gcr.io/datadoghq/agent:7
+      environment:
+        - DD_API_KEY=${DD_API_KEY}
+        - DD_DOGSTATSD_NON_LOCAL_TRAFFIC=true # enables agent to receive custom metrics from other containers
+        - DD_APM_ENABLED=true # enables tracing
+        - DD_APM_NON_LOCAL_TRAFFIC=true # enables agent to receive traces from other containers
+        - DD_AGENT_HOST=datadog-agent
+        - DD_SITE=datadoghq.com # determines datadog instance to send data to (e.g change to datadoghq.eu for EU1)
+      volumes:
+        - ${HOME}/.local/share/containers:/var/lib/containers:ro
```
AlejandroMinaya, line 330: Datadog Agent needs access to either the container dir (for rootless containers) or to the Podman socket (for rootful containers), i.e. `/run/podman/podman.sock`

I'm not sure if we should include both since we don't know how people plan to deploy it or if rootless is generally preferred

osoukup (changes requested):
I do not know much about Datadog and would this do so I will start with a general question. **ddtrace-run** with the default **datadoghq.com** server sends all the logs to this external URL? OSIDB contains embargoed information and we can never know what ends up in the logs (eg. due to a bug) so we need this to be safe.

Also I know in the past with the Splunk we had to use some side-car pods, looks like we need it no more. I am not an expert on this (@JakubFrejlach would know better), maybe it was Splunk specifics, but I think the reason was to be able to send the data a safe and infosec-approved way. Changing that to some other mechanism would I think require consulting either infosec or some security experts in ProdSec so we know we are not shooting ourselves in a foot.

AlejandroMinaya:
> I do not know much about Datadog and would this do so I will start with a general question. **ddtrace-run** with the default **datadoghq.com** server sends all the logs to this external URL? OSIDB contains embargoed information and we can never know what ends up in the logs (eg. due to a bug) so we need this to be safe.

Good catch! Datadog documentation makes a distinction between APM traces and logs, although I am not certain that it would not accidentally keep track of embargoed information. I will look into how we can have a finer-grain control and research what is advised by infosec. Thank you!

osoukup:
> > I do not know much about Datadog and would this do so I will start with a general question. **ddtrace-run** with the default **datadoghq.com** server sends all the logs to this external URL? OSIDB contains embargoed information and we can never know what ends up in the logs (eg. due to a bug) so we need this to be safe.
> 
> Good catch! Datadog documentation makes a distinction between APM traces and logs, although I am not certain that it would not accidentally keep track of embargoed information. I will look into how we can have a finer-grain control and research what is advised by infosec. Thank you!

I do not think we can reliably categorize the sensitivity of the data spitted by a random traceback. We should therefore treat it as potentially the worst case - which is embargoed data. Therefore all the logs/traces of OSIDB should be handled confidentially and ensuring only explicitly allowed people have access to it and that holds throughout the whole log retrieval and storage pipeline. Having an infosec approved way would be ideal.

AlejandroMinaya:
Since we figured out a way to carve out performance improvement work without having to add Datadog APM I am closing this PR. It is worth revisiting in the future but it is no longer a prerequisite

PR #1305, opened 2026-06-15
OSIDB:5038 - Wrap bulk PUT external errors with affect context

svelamak:
Wraps external errors in the bulk affect PUT endpoint with meaningful context.
- Updated `bulk_put` to collect per-item errors when affect failed in `api_views.py` 
- Updated existing test and added new test case
Closes: OSIDB-5038

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -1447,16 +1447,38 @@ class AffectView(
         # Second, save the updated affects to the database, but not sync with BZ.
         actor = self._actor(request.user)
         ret = []
+        errors = []
         for serializer in validated_serializers:
             # Make the serializer skip the sync for each affect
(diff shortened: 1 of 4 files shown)


osidb/api_views.py
```diff
@@ -1431,7 +1431,7 @@ def bulk_put(self, request, *args, **kwargs):
                 error_detail = (
                     e.message_dict
                     if hasattr(e, "message_dict")
-                    else {"non_filed_errors": e.messages}
+                    else {"non_field_errors": e.messages}
```
osoukup, line 1461: nitpick: This should be squashed to the commit where the typo was introduced.

osoukup (approved):
It looks OK, I would probably just squash it to a single commit - the last one shows no changes anyway and the secrets update is a result of the previous changes.

PR #1307, opened 2026-06-16
OSIDB-5108: Change HashiCorp Vault integration to opt-in

AlejandroMinaya:
## Summary

Makes HashiCorp Vault integration **credential-based opt-in**: Vault is enabled ONLY when all three credentials (OSIDB_VAULT_ADDR, OSIDB_ROLE_ID, OSIDB_SECRET_ID) are provided. If any credential is missing, Vault integration is gracefully disabled.

**Change in Direction**: Initial implementation used environment-based detection (disabled for local, enabled otherwise). This was changed to credential-based opt-in to make Vault truly optional without coupling to specific environments.

### Core Implementation
- **`osidb/integrations.py`**:
  - Made Vault credentials optional (empty string defaults)
  - `is_vault_enabled()` now checks if all three credentials are provided (not environment)
  - Removed `validate_credentials()` method (no longer needed)
  - Modified `IntegrationRepository.__init__()` to skip Vault client initialization when credentials missing
  - Guards in `read_secret()` and `upsert_secret()` return None/no-op when client is None

- **`osidb/api_views.py`**:
  - `integration_tokens()` endpoint returns 503 when Vault is disabled
  - Updated error message to reflect credential-based disable

### Test Updates
- **`osidb/tests/test_hcv_integration.py`**:
  - Replaced environment-based tests with credential-based tests
  - 8 new tests covering all credential combinations
  - Tests verify disabled behavior when credentials missing
  - Tests verify enabled behavior when all credentials provided

- **`osidb/tests/conftest.py`** & **`apps/workflows/tests/conftest.py`**:
  - `set_hvac_test_env_vars` fixture sets test credentials with cleanup
  - Tests opt-in to credentials via fixture parameter
  - Removed environment mocking from `patch_hvac_client`

### Documentation
- **`.env`**: Updated comments to clarify credential-based opt-in
- **`config/settings_local.py`**: Added note about credential-based behavior

## Behavior

- **Vault enabled** (all credentials provided):
  - Works in any environment (local, prod, stage, ci)
  - Requires `OSIDB_VAULT_ADDR`, `OSIDB_ROLE_ID`, `OSIDB_SECRET_ID`

- **Vault disabled** (any credential missing):
  - Works in any environment (local, prod, stage, ci)
  - Integration tokens fall back to HTTP headers (`Bugzilla-Api-Key`, `JIRA-Taskman-Api-Key`)
  - No Vault connection attempted

## Test Plan

- ✅ All existing Vault integration tests pass
- ✅ New credential-based tests verify disabled state behavior
- ✅ Endpoint tests work with credential fixture
- ✅ Manual verification: Local dev works without credentials, Vault activates when credentials set

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## New Features
* HashiCorp Vault integration is now credential-based opt-in—provide `OSIDB_VAULT_ADDR`, `OSIDB_ROLE_ID`, and `OSIDB_SECRET_ID` environment variables to enable.
* API endpoints return `503 Service Unavailable` with clear messaging when Vault integration is disabled.

## Documentation
* Updated API documentation to reflect optional Vault integration behavior.

### DIFF
--- a/osidb/integrations.py
+++ b/osidb/integrations.py
@@ -14,9 +14,18 @@ _logger = logging.getLogger(__name__)
 class IntegrationSettings(BaseSettings):
     model_config = SettingsConfigDict(env_prefix="OSIDB_")
 
-    vault_addr: str = Field(default=...)
-    role_id: str = Field(default=...)
-    secret_id: str = Field(default=...)
+    vault_addr: str = Field(default="")
+    role_id: str = Field(default="")
+    secret_id: str = Field(default="")
+
+    def is_vault_enabled(self) -> bool:
+        """
+        Determine if Vault should be enabled based on credentials.
+
+        Vault is enabled only when ALL THREE credentials are provided:
+        OSIDB_VAULT_ADDR, OSIDB_ROLE_ID, OSIDB_SECRET_ID
+        """
+        return bool(self.vault_addr and self.role_id and self.secret_id)
 
 
 class IntegrationRepository:
@@ -24,6 +33,16 @@ class IntegrationRepository:
     BASE_MOUNTPOINT = "apps"
 
     def __init__(self, settings: IntegrationSettings):
+        self.settings = settings
+
+        if not settings.is_vault_enabled():
+            _logger.info(
+                "Vault integration is disabled because required credentials are not provided"
+            )
+            self.client = None
+            return
+
+        # Credentials are guaranteed to be present by is_vault_enabled()
         self.client = hvac.Client(url=settings.vault_addr)
         try:
             self.client.auth.approle.login(
@@ -33,6 +52,10 @@ class IntegrationRepository:
             _logger.error("Vault AppRole authentication failed.", exc_info=e)
 
     def upsert_secret(self, subpath: Path, key: str, value: str) -> None:
+        if self.client is None:
+            _logger.debug(f"Vault is disabled, cannot write secret at {subpath}/{key}")
+            return
+
         self.client.secrets.kv.v2.patch(
             path=str(self.BASE_PATH / subpath),
             secret={key: value},
@@ -40,6 +63,10 @@ class IntegrationRepository:
         )
 
     def read_secret(self, subpath: Path, key: str) -> Optional[str]:
+        if self.client is None:
+            _logger.debug(f"Vault is disabled, cannot read secret at {subpath}/{key}")
+            return None
+
         r = self.client.secrets.kv.v2.read_secret_version(
             path=str(self.BASE_PATH / subpath),
             mount_point=self.BASE_MOUNTPOINT,
--- a/tox.ini
+++ b/tox.ini
@@ -83,3 +83,4 @@ commands = uvx {[ruff]ruff_version} check --fix --diff --select I .
 [testenv:ruff-format]
 commands_pre = 
 commands = uvx {[ruff]ruff_version} format --check .
+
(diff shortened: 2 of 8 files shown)


tox.ini
```diff
@@ -82,4 +82,4 @@ commands = uvx {[ruff]ruff_version} check --fix --diff --select I .
 
 [testenv:ruff-format]
 commands_pre = 
-commands = uvx {[ruff]ruff_version} format --check .
\ No newline at end of file
+commands = uvx {[ruff]ruff_version} format --check .
```
osoukup, line 85: nitpick: Does not seem to be a real change worth committing.

osidb/integrations.py
```diff
@@ -14,16 +14,35 @@
...
+
+    def is_vault_enabled(self) -> bool:
+        """
+        Determine if Vault should be enabled based on execution environment.
+
+        Vault is disabled for local development (get_execution_env() == "local")
+        and enabled for all other environments (prod, stage, uat, ci, etc.)
+        """
+        execution_env = get_execution_env()
+        return execution_env != "local"
```
osoukup: I am not in favor of having this hard-coded here. It should be based on a settings constant and ideally allowed to be overridden by an environmental variable. I would make it turned off by default and only enable it where it makes sense - prod, stage, UAT.

osoukup (changes requested):
Mostly LGTM but I would generalize the setup. The most important reason I see now is the focus on better OSIDB open-sourcing and HashiCorp integration is probably very specific for our internal RH usage not really useful to anyone else. That should be turned off by default and configurable through the settings.

AlejandroMinaya:
Agreed. That makes a lot of sense. Is the expected default behavior just the headers fallback?

Thanks for the review btw!

osoukup:
> Agreed. That makes a lot of sense. Is the expected default behavior just the headers fallback?
> 
> Thanks for the review btw!

Yeah, I think at this point it should be just the fallback. Hopefully soon we will be able to remove the tracker write operations from OSIDB when Tracker AutoManager takes this responsibility and then we could deprecate the API key proxying completely. At this point I would at least like to make it as optional as possible.

tox.ini
AlejandroMinaya: Understandable. It is just to appease the pre-commit hooks

osoukup: I mean I am not against, feel free to leave it here. I though it is just a random unrelated change but if it has a reason then OK. I would then just put it in a separate commit with description of why it is needed.

config/settings_local.py
```diff
@@ -3,6 +3,10 @@
 
 from .settings import *
 
+# NOTE: Vault integration is automatically disabled for local development.
+# Integration tokens (JIRA, Bugzilla) will fall back to HTTP headers
+# (Bugzilla-Api-Key, JIRA-Taskman-Api-Key) when Vault is unavailable.
```
osoukup: Is this comment still necessary when the integration is disabled everywhere by default unless you provide all the necessary credentials?

osoukup (approved):
LGTM. I would just prefer a bit consolidated commits so the related changes are introduced together either for the easier review or future git log. Eg. in the case of the API docs if you find out that you forgot something, update the previous commit which introduced it and not create a new one. Similar with enhancing the tests. Also if you introduce something and remove it in a followup commits (some tests here), it is confusing for the reviewer and has no value for the future git history, that should be just squashed.

PR #1308, opened 2026-06-17
Add CRA_regulatory_reporting for atomated reviews in this proyect

roduran-dev:
# Context

For the CRA regulatory reporting project we create a specific branch that will be our base for de moment. The coderabbit only triggers a review for project its master or release. 

# change
This will make code rabbit to also run automatically for PR's CRA_regulatory_reporting as a base

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -40,6 +40,8 @@ reviews:
       - "main"
       - "master"
       - "release-\\d+\\.\\d+\\.\\d+(-prep)?$"
+      # Special branch for CRA regulatory reporting project (remove after merge)
+      - "CRA_regulatory_reporting"
 
   path_instructions:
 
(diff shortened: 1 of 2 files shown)


osoukup (changes requested):
Why is this necessary? You can always trigger rabbit manually if you need. I do it for the workflow-framework branches also eg. at https://github.com/RedHatProductSecurity/osidb/pull/1300. I am not a fan a adding any special cases into the config.

roduran-dev:
> Why is this necessary? You can always trigger rabbit manually if you need. I do it for the workflow-framework branches also eg. at #1300. I am not a fan a adding any special cases into the config.

I know we can do the trigger the process manually but I wanted to allow for now to be able to run it automatically. 
It's just for the duration of this project. That's why I added to remove after merge and even if we forgot it won't affect the process at all.

osoukup (changes requested):
So add it to your branch and to master. I do not agree with any temporary changes to master because of some project branch.

roduran-dev:
> So add it to your branch and to master. I do not agree with any temporary changes to master because of some project branch.

Ohh, I see what you mean. Now it will be merge into the standard branch. :+1:

roduran-dev:
I will close this PR and do a new one. 

CRA branch need to merge a lot from master so now the changes look huge. (I will create a separated PR for each)

PR #1313, opened 2026-06-19
OSIDB 5141 add option to run the whole test suite in parallel

AlejandroMinaya:
We have `pytest-xdist` installed but are not leveraging this plugin to run our test in parallel. This PR adds a command, but also modifies some tests that inadvertently share states with unrelated suites. For example, the queryset regression tests can yield more or less queries depending on whether the user was fetched from the database on a previous test. In the case of the row-level security suite, I added a fixture to ensure the clean ACL state of each test.

I believe there is a case to be made to use `pytest-xdist` in our CI pipeline, but it should hinge on the adoption and success of this new command.


osoukup (approved):
LGTM

PR #1311, opened 2026-06-18
Make integrations Vault base path configurable

hseljenes:
## Summary
- Add `integrations_base_path` setting to `IntegrationSettings` (env var: `OSIDB_INTEGRATIONS_BASE_PATH`), defaulting to `/osidb-integrations`
- Replace the hardcoded `BASE_PATH` class variable on `IntegrationRepository` with an instance variable built from the new setting
- Follows the existing pydantic_settings pattern used for other vault credentials

## Test plan
- [ ] Verify default behavior is unchanged (no env var set → uses `/osidb-integrations`)
- [ ] Verify setting `OSIDB_INTEGRATIONS_BASE_PATH` to a custom value changes the Vault path accordingly
- [ ] Run existing integration tests to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1315, opened 2026-06-22
fix the PR permissions so the forks work

osoukup:
To enable

https://github.com/RedHatProductSecurity/osidb/pull/1311


costaconrado (approved):
LGTM

PR #1318, opened 2026-06-22
Use simple changelog notifications for fork PRs

costaconrado:


osoukup (approved):
LGTM

.github/workflows/meta.yml
```diff
@@ -32,5 +32,6 @@ jobs:
         with:
           fileName: docs/CHANGELOG.md
           noChangelogLabel: technical
+          checkNotification: Simple
```
JakubFrejlach, line 35: Is the `checkNotification: Detailed` which seems to be the default value any beneficial compared to `Simple`? If so we could do something like this:
```
checkNotification: ${{ github.event.pull_request.head.repo.fork && 'Simple' || 'Detailed' }}
```

Also looking at the code of the github action, they actually do have [builtin fallback for Forks](https://github.com/Zomzog/changelog-checker/blob/master/src/service/Checks.ts#L60), is it really necessary?

JakubFrejlach (approved):
I am not against this change, just raising awareness whether it is really needed and whether at least not keep the detailed for non-fork stuff, but no strong preference.

PR #1323, opened 2026-06-24
OSIDB-5067: Mark flaw as srp reportable when criteria are met

roduran-dev:
It adds a signal so when a flaw gets saved it will create a SRPRerport or get it and modify it with the new info from the Flaw


roduran-dev:
Duplicate

PR #1324, opened 2026-06-24
Make LDAP group names configurable via environment variables

hseljenes:
## Summary
- Replace hardcoded LDAP group names in `settings_stage.py` and `settings_prod.py` with `get_env()` calls
- Allows deployments to customize group names via environment variables without code changes
- Defaults preserve the existing behavior — no change for current deployments

## New environment variables
| Variable | Default (stage) | Default (prod) |
|---|---|---|
| `OSIDB_PUBLIC_READ_GROUPS` | `["osidb-stage-public-read", "red-hat-product-security"]` | `["osidb-prod-public-read", "red-hat-product-security"]` |
| `OSIDB_PUBLIC_WRITE_GROUP` | `osidb-stage-public-write` | `osidb-prod-public-write` |
| `OSIDB_EMBARGO_READ_GROUP` | `osidb-stage-embargo-read` | `osidb-prod-embargo-read` |
| `OSIDB_EMBARGO_WRITE_GROUP` | `osidb-stage-embargo-write` | `osidb-prod-embargo-write` |
| `OSIDB_INTERNAL_READ_GROUP` | `osidb-stage-internal-read` | `osidb-prod-internal-read` |
| `OSIDB_INTERNAL_WRITE_GROUP` | `osidb-stage-internal-write` | `osidb-prod-internal-write` |
| `OSIDB_SERVICE_MANAGE_GROUP` | `osidb-stage-manage` | `osidb-prod-manage` |

## Motivation
Downstream deployments that use different Rover/LDAP group naming conventions (e.g. prefixed with their project name) currently cannot override the group names without forking the settings files.

## Test plan
- [ ] Verify default behavior is unchanged (no env vars set → uses existing group names)
- [ ] Verify setting env vars overrides the group names
- [ ] Existing tests pass — test suite uses `settings_ci.py` which is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/settings_prod.py
+++ b/config/settings_prod.py
@@ -1,6 +1,7 @@
 import ssl
 
 import ldap
+from django.core.exceptions import ImproperlyConfigured
 from django_auth_ldap.config import GroupOfUniqueNamesType, LDAPSearch, LDAPSearchUnion
 
 from .settings import *
@@ -18,6 +19,8 @@ PUBLIC_READ_GROUPS = get_env(
     default='["osidb-prod-public-read", "red-hat-product-security"]',
     is_json=True,
 )
+if not PUBLIC_READ_GROUPS:
+    raise ImproperlyConfigured("OSIDB_PUBLIC_READ_GROUPS must be a non-empty list")
 PUBLIC_WRITE_GROUP = get_env("OSIDB_PUBLIC_WRITE_GROUP", default="osidb-prod-public-write")
 EMBARGO_READ_GROUP = get_env("OSIDB_EMBARGO_READ_GROUP", default="osidb-prod-embargo-read")
 EMBARGO_WRITE_GROUP = get_env("OSIDB_EMBARGO_WRITE_GROUP", default="osidb-prod-embargo-write")
--- a/config/settings_prod.py
+++ b/config/settings_prod.py
@@ -13,22 +13,16 @@ SECRET_KEY = get_env("DJANGO_SECRET_KEY")
 USE_X_FORWARDED_HOST = True
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
-# Minimal group for read access of public flaws in OSIDB
-# TODO: In the future we might simply use a proxy group in which
-# membership is based off of one or more LDAP groups
-# e.g. (|(memberOf=group-a)(memberOf=group-b))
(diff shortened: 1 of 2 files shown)


osoukup:
@hseljenes thank you for the contribution! I have to admit that this way of hard-coded is not a good design and another very ProdSec-customised piece of OSIDB. However, I am not sure your change moves it well in the right direction as it would basically created another dependency (your OSIDB instance) on these configs.

I would like to remove them and move to our OPS repo. Would it be OK for you for now to just create your own config file (and probably just importing `settings.py` there) and provide it through eg. `DJANGO_SETTINGS_MODULE=config.your_settings` ?

hseljenes:
> @hseljenes thank you for the contribution! I have to admit that this way of hard-coded is not a good design and another very ProdSec-customised piece of OSIDB. However, I am not sure your change moves it well in the right direction as it would basically created another dependency (your OSIDB instance) on these configs.
> 
> I would like to remove them and move to our OPS repo. Would it be OK for you for now to just create your own config file (and probably just importing `settings.py` there) and provide it through eg. `DJANGO_SETTINGS_MODULE=config.your_settings` ?

It looks setup to just use the existing values you had in the repo, but allows them to be customized via env vars. I don't see how it now links to the Lightwell one?

osoukup:
> > @hseljenes thank you for the contribution! I have to admit that this way of hard-coded is not a good design and another very ProdSec-customised piece of OSIDB. However, I am not sure your change moves it well in the right direction as it would basically created another dependency (your OSIDB instance) on these configs.
> > I would like to remove them and move to our OPS repo. Would it be OK for you for now to just create your own config file (and probably just importing `settings.py` there) and provide it through eg. `DJANGO_SETTINGS_MODULE=config.your_settings` ?
> 
> It looks setup to just use the existing values you had in the repo, but allows them to be customized via env vars. I don't see how it now links to the Lightwell one?

I am sorry but I am not sure I understand this comment :sweat: I meant that having prod, stage, and UAT settings in the code repo is not a good design from our side and I would like to remove them instead of make them configurable. I have a change on the way https://github.com/RedHatProductSecurity/osidb/pull/1327/changes (not completely done) so in the meantime could you please just provide settings you need through the Django variable?

EDIT: The other PR removing the evn-specific settings files is ready.

Elkasitu:
@hseljenes OSIDB LDAP groups are now completely configurable as of #1327

PR #1329, opened 2026-06-25
fix pghistory trigger blocking migration

costaconrado:
This PR temporarily disables the triggers that blocks deleting events on pghistory during the migration 0238_exclude_last_validated_dt_from_history, once the migration is done it re-enables it.


osoukup (approved):
LGTM

PR #1331, opened 2026-06-25
OSIDB-5068: trigger srp milestone creation on reportability change

roduran-dev:
Implements automatic creation of SRP reports milestones when flaws are marked as actively exploited (KEV) or severe incidents, enabling EU Cyber Resilience Act compliance.

This PR follows the logic that if Flaw is saved it will create the 3 main milestones. 
If the SRP is created then updates attached milestones ACL's 

OSIDB-5068

### DIFF
--- a/regulatory_reporting/models/srp_report_milestone.py
+++ b/regulatory_reporting/models/srp_report_milestone.py
@@ -59,7 +59,7 @@ class SRPReportMilestone(SRPReportBase):
     MILESTONE_DURATION_BY_TYPE = {
         MilestoneType.LEVEL_24H: timedelta(hours=24),
         MilestoneType.LEVEL_72H: timedelta(hours=72),
-        MilestoneType.LEVEL_FINAL: timedelta(days=14),
+        MilestoneType.LEVEL_FINAL: None,  # Duration is calculated based on the reportable event type
         MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE: timedelta(days=30),
     }
 
@@ -129,19 +129,62 @@ class SRPReportMilestone(SRPReportBase):
 
     @property
     def due_at(self):
+        """
+        Calculate milestone due date.
+
+        For LEVEL_FINAL: duration depends on event type:
+        - KEV (actively_exploited_vulnerability): 14 days
+        - Severe Incident (severe_incident): 30 days
+        - Additional Information Request (additional_information_request): 30 days from the request received
+        """
+        if (
+            self.milestone_type
+            == self.MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE
+        ):
+            if not self.request_received_at:
+                return None
+            return self.request_received_at + timedelta(days=30)
+
         if not self.srp_report.timer_started_at:
             return None
 
-        return (
-            self.srp_report.timer_started_at
-            + self.MILESTONE_DURATION_BY_TYPE[self.milestone_type]
-        )
+        if self.milestone_type == self.MilestoneType.LEVEL_FINAL:
+            # Check parent report's event type
+            if (
+                self.srp_report.reportable_event_type
+                == SRPReport.ReportableEventType.ACTIVELY_EXPLOITED_VULNERABILITY
+            ):
+                duration = timedelta(days=14)
+            elif (
+                self.srp_report.reportable_event_type
+                == SRPReport.ReportableEventType.SEVERE_INCIDENT
+            ):
+                duration = timedelta(days=30)
+            else:
+                raise ValidationError("Invalid reportable event type")
+        else:
+            # Use static duration for 24h, 72h, etc.
+            duration = self.MILESTONE_DURATION_BY_TYPE[self.milestone_type]
+
+        return self.srp_report.timer_started_at + duration
 
     def __str__(self):
         return f"{self.milestone_type} - {self.srp_report.flaw.cve_id or self.srp_report.flaw.uuid}"
 
     @validator
     def _validate_due_at_required(self, **kwargs):
-        """Due date must be set for all milestones"""
+        """
+        Due date must be set for all milestones.
+
+        Exception: LEVEL_ADDITIONAL_INFORMATION_RESPONSE milestones can have
+        None due_at if request_received_at is not yet set.
+        """
         if not self.due_at:
+            # Allow None for additional info milestones without request time
+            if (
+                self.milestone_type
+                == self.MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE
+                and not self.request_received_at
+            ):
+                return  # Valid state - waiting for request
             raise ValidationError("due_at must be set for all milestones")
--- a/regulatory_reporting/models/srp_report_milestone.py
+++ b/regulatory_reporting/models/srp_report_milestone.py
@@ -161,7 +161,7 @@ class SRPReportMilestone(SRPReportBase):
             ):
                 duration = timedelta(days=30)
             else:
-                raise ValidationError("Invalid reportable event type")
+                return None
         else:
             # Use static duration for 24h, 72h, etc.
             duration = self.MILESTONE_DURATION_BY_TYPE[self.milestone_type]
@@ -179,6 +179,16 @@ class SRPReportMilestone(SRPReportBase):
         Exception: LEVEL_ADDITIONAL_INFORMATION_RESPONSE milestones can have
         None due_at if request_received_at is not yet set.
         """
+        if (
+            self.milestone_type == self.MilestoneType.LEVEL_FINAL
+            and self.srp_report.reportable_event_type
+            not in {
+                SRPReport.ReportableEventType.ACTIVELY_EXPLOITED_VULNERABILITY,
+                SRPReport.ReportableEventType.SEVERE_INCIDENT,
+            }
+        ):
+            raise ValidationError("Invalid reportable event type")
+
         if not self.due_at:
             # Allow None for additional info milestones without request time
             if (
--- a/regulatory_reporting/signals.py
+++ b/regulatory_reporting/signals.py
@@ -71,7 +71,6 @@ def update_srp_report_milestones(srp_report: SRPReport):
         )
 
 
-@receiver(post_save, sender=Flaw)
 def create_srp_report(sender, instance: Flaw, created: bool, **kwargs):
     """
     Auto-create SRP Report and milestones when Flaw is marked as KEV or Major Incident approved.
@@ -119,13 +118,10 @@ def create_srp_report(sender, instance: Flaw, created: bool, **kwargs):
         )
 
 
-@receiver(post_save, sender=Flaw)
 def check_upstream_notifiable(sender, instance, **kwargs):
     """
     On Flaw save, check criteria for upstream maintainer notification.
     """
-    if not settings.CRA_NOTIFICATIONS_ENABLED:
-        return
     if not is_flaw_upstream_notifiable(instance):
         return
 
--- a/regulatory_reporting/signals.py
+++ b/regulatory_reporting/signals.py
@@ -24,6 +24,15 @@ REPORTABLE_EVENT_TYPE_MAP = {
 
 
 def create_srp_report_milestones(srp_report: SRPReport):
+    """
+    Create the required SRP milestones (24h, 72h, final) for a new SRP Report.
+
+    Does NOT create additional_information_response milestones - those are
+    created on-demand when requests are received.
+
+    Args:
+        srp_report: The SRP Report to create milestones for
+    """
     milestone_types = [
         SRPReportMilestone.MilestoneType.LEVEL_24H,
         SRPReportMilestone.MilestoneType.LEVEL_72H,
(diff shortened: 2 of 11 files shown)


roduran-dev:
@coderabbit do a full review

regulatory_reporting/models/srp_report_milestone.py
```diff
@@ -129,10 +129,40 @@ class Meta:
...
+        - KEV (actively_exploited_vulnerability): 14 days
+        - Severe Incident (severe_incident): 30 days
+        - Additional Information Request (additional_information_request): 30 days from the request received
+        """
+        if (
+            self.srp_report.reportable_event_type
+            == SRPReport.ReportableEventType.ADDITIONAL_INFORMATION_REQUEST
+        ):
+            duration = timedelta(days=30)
+            return self.request_received_at + duration
```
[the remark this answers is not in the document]
roduran-dev: ok, I added a validation, now it will return None

regulatory_reporting/tests/test_milestone_creation.py
```diff
@@ -0,0 +1,466 @@
...
+import pytest
+from django.core.exceptions import ValidationError
+from django.utils import timezone
+
+from osidb.models import Flaw
+from osidb.tests.factories import FlawFactory
+from regulatory_reporting.models import SRPReport, SRPReportMilestone
+from regulatory_reporting.models.abstracts import SRPReportBase
+
+pytestmark = pytest.mark.unit
```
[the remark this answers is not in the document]
roduran-dev: In OSIDB, unit vs integration is not “no DB” vs “with DB”. From pytest.ini:

unit — standard in-process tests (DB, factories, signals OK)
integration — tests that need a running environment (external Jira/BZ, live server, etc.)

These tests hit the DB and opt into signals via @pytest.mark.enable_signals, but they do not call external services — same pattern as regulatory_reporting/tests/test_signals.py and regulatory_reporting/tests/test_models.py, and most of osidb/tests/.

Marking them integration would move them into the wrong tox env (integration-tests) and they would not run with the normal uni

regulatory_reporting/tests/test_milestone_creation.py
```diff
@@ -0,0 +1,466 @@
...
+
+        srp_report = SRPReport.objects.get(flaw=flaw)
+        milestone_24h = SRPReportMilestone.objects.get(
+            srp_report=srp_report,
+            milestone_type=SRPReportMilestone.MilestoneType.LEVEL_24H,
+        )
+
+        milestone_str = str(milestone_24h)
+        assert "24h" in milestone_str
+        assert "CVE-2024-9999" in milestone_str
```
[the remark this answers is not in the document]
roduran-dev: Added test

regulatory_reporting/models/srp_report_milestone.py
```diff
@@ -129,16 +129,62 @@ class Meta:
...
+        if not self.srp_report.timer_started_at:
+            return None
+
+        if (
+            self.milestone_type
+            == self.MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE
+        ):
+            if not self.request_received_at:
+                return None
+            return self.request_received_at + timedelta(days=30)
```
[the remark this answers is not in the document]
roduran-dev: Fixed.

regulatory_reporting/signals.py
```diff
@@ -58,16 +104,19 @@ def create_srp_report(sender, instance: Flaw, created: bool, **kwargs):
...
+        srp_report.title = instance.title or f"SRP Report for {instance.uuid}"
         srp_report.acl_read = instance.acl_read
         srp_report.acl_write = instance.acl_write
         srp_report.timer_started_at = instance.major_incident_start_dt
         srp_report.save()
-        logger.info(f"Updated SRP Report {srp_report.uuid} for {instance.uuid}")
+        update_srp_report_milestones(srp_report)
+        logger.info(f"Updated SRP Report {srp_report.uuid} for Flaw {instance.uuid} ")
     else:
         create_srp_report_milestones(srp_report)
```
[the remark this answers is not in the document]
roduran-dev: But when a SRPreport is updated I ust want to update its milestones, not create new ones.

regulatory_reporting/models/srp_report_milestone.py
```diff
@@ -129,16 +129,62 @@ class Meta:
...
+                == SRPReport.ReportableEventType.ACTIVELY_EXPLOITED_VULNERABILITY
+            ):
+                duration = timedelta(days=14)
+            elif (
+                self.srp_report.reportable_event_type
+                == SRPReport.ReportableEventType.SEVERE_INCIDENT
+            ):
+                duration = timedelta(days=30)
+            else:
+                raise ValidationError("Invalid reportable event type")
```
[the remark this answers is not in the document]
roduran-dev: Ok, I move validations as asked.

regulatory_reporting/tests/test_milestone_creation.py
```diff
@@ -0,0 +1,575 @@
...
+        srp_report = SRPReport.objects.get(flaw=flaw)
+        initial_count = SRPReportMilestone.objects.filter(srp_report=srp_report).count()
+        assert initial_count == 3
+
+        # Act - save flaw again without changes
+        flaw.save()
+
+        # Assert no duplicates
+        final_count = SRPReportMilestone.objects.filter(srp_report=srp_report).count()
+        assert final_count == 3, "Should not create duplicate milestones on flaw resave"
```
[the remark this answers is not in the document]
roduran-dev: ok, check now (I change the name to test_milestones_refreshed_in_place_on_flaw_resave)

regulatory_reporting/signals.py
```diff
@@ -24,32 +24,78 @@
...
+        milestone.acl_write = srp_report.acl_write
+        milestone.save()
+        logger.info(
+            f"Updated {milestone.milestone_type} milestone for SRP Report {srp_report.uuid} "
+            f"for Flaw {srp_report.flaw.uuid}, created at {milestone.created_dt}, due at {milestone.due_at}"
+        )
 
 
 @receiver(post_save, sender=Flaw)
 def create_srp_report(sender, instance: Flaw, created: bool, **kwargs):
```
JakubFrejlach, line 70: There should be some mechanism how to disable this signal, currently it will always fire and cannot be disabled. In `check_upstream_notifiable` we do have env variable `CRA_NOTIFICATIONS_ENABLED` check and if it evaluates to `False` the signal just returns, which is not ideal as the signal still fires, it just always returns. Better would be to conditionally register these signals during the app `ready` method execution, something like:

```python
# apps.py
  class RegulatoryReportingConfig(AppConfig):
      def ready(self):
          if settings.CRA_NOTIFICATIONS_ENABLED:
              from django.db.models.signals import post_save
              from osidb.models import Flaw
              from .signals import check_upstream_notifiable, create_srp_report
              post_save.connect(check_upstream_notifiable, sender=Flaw)
              post_save.connect(create_srp_report, sender=Flaw)
```

regulatory_reporting/signals.py
JakubFrejlach: See me [suggestion above](https://github.com/RedHatProductSecurity/osidb/pull/1331/changes#r3520222851).

JakubFrejlach (changes requested):
Looking good, but we should really make all the CRA stuff optional in OSIDB, see my comments.

regulatory_reporting/signals.py
roduran-dev: I was thinking adding something like this but in settiings.py to decide if the app should be loaded or not. 

But this idea seem better :thinking:

roduran-dev: I'm apply the change here but I feel this should be in its own PR. (it overlaps with sam changes)

roduran-dev:
@JakubFrejlach I agree that setting signals from the app.py. 

The change makes a bit more complex the testing because now the modules are never imported.

regulatory_reporting/tests/conftest.py
```diff
@@ -0,0 +1,54 @@
+import pytest
```
roduran-dev, line 1: @costaconrado I remember you asked to just use the base conftest but I had to reintroduced to add 2 fixture and some markers exclusive to `regulatory_reporting/` I though it was better than changing an OSIDB core base file.

regulatory_reporting/signals.py
```diff
@@ -58,25 +103,25 @@ def create_srp_report(sender, instance: Flaw, created: bool, **kwargs):
...
+            f"Created SRP Report {srp_report.uuid} for Flaw {instance.uuid}, event type {event_type}"
+        )
 
 
-@receiver(post_save, sender=Flaw)
 def check_upstream_notifiable(sender, instance, **kwargs):
     """
     On Flaw save, check criteria for upstream maintainer notification.
     """
-    if not settings.CRA_NOTIFICATIONS_ENABLED:
```
roduran-dev, line 78: @svelamak I removed because now it will be decided in the app level. (check[ this comment](https://github.com/RedHatProductSecurity/osidb/pull/1331#discussion_r3520222851))

svelamak: Makes sense, thanks for clarifying!

regulatory_reporting/signals.py
JakubFrejlach: Probably does not really matter in which PR this is going to be,  but in this PR we are adding similar functionality which is guarded in a new way so probably no problem to make the change here also for the other signal.

regulatory_reporting/models/srp_report_milestone.py
```diff
@@ -129,16 +129,72 @@ class Meta:
...
+        - Severe Incident (severe_incident): 30 days
+        - Additional Information Request (additional_information_request): 30 days from the request received
+        """
+        if (
+            self.milestone_type
+            == self.MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE
+        ):
+            if not self.request_received_at:
+                return None
+            return self.request_received_at + timedelta(days=30)
```
JakubFrejlach, line 146: Why is `MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE` key listed in `MILESTONE_DURATION_BY_TYPE` if that value is never used?

regulatory_reporting/signals.py
```diff
@@ -24,32 +24,77 @@
...
             acl_read=srp_report.acl_read,
             acl_write=srp_report.acl_write,
         )
+        logger.info(
+            f"Created {milestone_type} milestone for SRP Report {srp_report.uuid} "
+            f"for Flaw {srp_report.flaw.uuid}, created at {milestone.created_dt}, due at {milestone.due_at}"
+        )
+
+
+def update_srp_report_milestones(srp_report: SRPReport):
```
JakubFrejlach, line 52: A really broad function name, basically it just copies ACLs from srp_report to related milestones, but function just states it updates. I don't even think this needs to be a function, you can just easily inline the logic there.

regulatory_reporting/signals.py
JakubFrejlach: Leftover imports?

regulatory_reporting/tests/test_services.py
```diff
@@ -25,7 +31,6 @@ def test_cveorg_source_is_not_notifiable(self):
         assert is_flaw_upstream_notifiable(flaw) is False
 
 
-@pytest.mark.enable_signals
 @pytest.mark.django_db
 class TestUpstreamNotificationSignal:
     @override_settings(CRA_NOTIFICATIONS_ENABLED=True)
```
JakubFrejlach: Are these overrides still needed with the refactor of the signals opt-in/opt-out?

regulatory_reporting/tests/test_milestone_creation.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        flaw = FlawFactory(
+            major_incident_state="EXPLOITS_KEV_REQUESTED",  # Not approved yet
+            major_incident_start_dt=timezone.now(),
+        )
+        assert SRPReport.objects.filter(flaw=flaw).count() == 0
+        assert SRPReportMilestone.objects.filter(srp_report__flaw=flaw).count() == 0
+
+    @pytest.mark.no_cra_reporting
+    def test_milestones_not_created_when_cra_reporting_is_disabled(self):
```
JakubFrejlach, line 427: CRA report is generated when:

```
 Triggers on:
    - EXPLOITS_KEV_APPROVED → Creates ACTIVELY_EXPLOITED_VULNERABILITY report
    - MAJOR_INCIDENT_APPROVED → Creates SEVERE_INCIDENT report
```
and major incident is set to `EXPLOITS_KEV_REQUESTED` so no report is generated regardless of the cra reporting being enabled/disabled.

JakubFrejlach (commented):
regulatory_reporting opt-in implementation looks good. Found couple of more things.

regulatory_reporting/signals.py
roduran-dev: yep :sweat:  

I wonder why this did not trigger ruff check error :thinking: ?

regulatory_reporting/models/srp_report_milestone.py
roduran-dev: Is not used in the report creation but it can be used later.

MilestoneType.LEVEL_ADDITIONAL_INFORMATION_RESPONSE is a special milestone to be use when CRA ask for more information. This one will require manual creation. 

Basically when CRA ask for more information someone will generate a new milestone of this type.

regulatory_reporting/signals.py
roduran-dev: I set it in a function because most likely after the first version this will get more complex. More steps to change to avoid changing if something happens

regulatory_reporting/tests/test_services.py
roduran-dev: It's true this is not needed anymore.  I will remove them.

regulatory_reporting/tests/test_milestone_creation.py
roduran-dev: Good catch :+1:

regulatory_reporting/tests/test_services.py
roduran-dev: Removed

regulatory_reporting/tests/test_milestone_creation.py
roduran-dev: Fixed

regulatory_reporting/signals.py
roduran-dev: removed

regulatory_reporting/signals.py
roduran-dev: In Yesterday meeting we talked about SRP reports updates so now I know for sure this function will get a bit more complex later

roduran-dev:
@JakubFrejlach Thanks for the second review, the new not attach signals if the variable is not set bring a lot of new code so the second review was spot on and you didn't miss anything in the first one :+1: 

@svelamak I rebased your resent merged PR and refactored base on the new pattern where signals are not attached instead of early living if global variables are not set.

regulatory_reporting/models/srp_report_milestone.py
JakubFrejlach: IMHO it creates a false impression that it controls something, but will leave that up to you.

JakubFrejlach (approved):
LGTM now, thanks for addressing my feedback.

PR #1338, opened 2026-06-29
Add relative datetime queries for API

AlejandroMinaya:
## Summary
- Added `RelativeDateTimeField` that accept relative time expressions (e.g., `-1d`, `+2h`, `-30m`) and absolute time expressions (e.g. `2026-07-02 00:00:00`)
- Update filters to use this relative field for datetime values (excluding deprecated V1 filters)

### DIFF
--- a/osidb/filters.py
+++ b/osidb/filters.py
@@ -2,9 +2,11 @@
 Implement filters for OSIDB REST API results
 """
 
+from copy import deepcopy
 from datetime import timedelta
 from typing import Union
 
+from django import forms
 from django.contrib.postgres.search import (
     SearchQuery,
     SearchRank,
@@ -15,6 +17,7 @@ from django.core.exceptions import ValidationError
 from django.core.validators import EMPTY_VALUES
 from django.db import models
 from django.db.models import Max, Min, Q
+from django.utils import timezone
 from django_filters.rest_framework import (
     BaseInFilter,
     BooleanFilter,
@@ -48,6 +51,7 @@ from osidb.models import (
 )
 from osidb.sync_manager import SyncManager
 
+from .datetime_utils import parse_relative_datetime
 from .djangoql import FlawQLSchema
 from .mixins import ACLMixinVisibility, Alert
 
@@ -79,6 +83,71 @@ class UUIDInFilter(BaseInFilter, UUIDFilter):
     """
 
     pass
+
+
+class RelativeDateTimeField(forms.DateTimeField):
+    """
+    DateTimeField that accepts both absolute datetime values (default behavior)
+    and relative datetime strings like "-1d", "+2h", "-30m", "-6M", "+1y", etc.
+
+    Relative format: [+/-]<number><unit>
+    where unit is one of: s (seconds), m (minutes), h (hours), d (days),
+    w (weeks), M (months), y (years)
+
+    Examples:
+        -1d     -> 1 day ago
+        +2h     -> 2 hours from now
+        -30m    -> 30 minutes ago
+        -6M     -> 6 months ago (calendar arithmetic)
+         1y     -> 1 year from now
+
+    If the value doesn't match the relative format, it falls back to standard
+    absolute datetime parsing.
+
+    Notes:
+        - Month/year arithmetic uses calendar dates, not fixed durations
+        - If target day doesn't exist (e.g., Jan 31 + 1M), uses last day
+          of target month (Feb 28/29)
+        - Unit 'M' (uppercase) = months; 'm' (lowercase) = minutes
+    """
+
+    def strptime(self, value, format):
+        """
+        Parse a datetime string using the given format.
+
+        First tries standard strptime parsing with the given format.
+        If that fails, attempts to parse as a relative datetime.
+
+        This override is necessary because Django's DateTimeField.to_python()
+        internally calls strptime() in a loop for each input_format. Without this,
+        relative datetime strings would fail validation since they don't match
+        standard datetime formats.
+        """
+        try:
+            return super().strptime(value, format)
+        except (ValueError, TypeError):
+            parsed = parse_relative_datetime(value, timezone.now())
+            if parsed is not None:
+                return parsed
+            raise
+
+
+class RelativeDateTimeFilter(DateTimeFilter):
+    field_class = RelativeDateTimeField
+
+
+FILTER_FOR_DBFIELD_DEFAULTS = deepcopy(FilterSet.FILTER_DEFAULTS)
+FILTER_FOR_DBFIELD_DEFAULTS.update(
+    {
+        models.DateTimeField: {
+            "filter_class": RelativeDateTimeFilter,
+        },
+    }
+)
+
+
+class OSIDBFilterSet(FilterSet):
+    FILTER_DEFAULTS = FILTER_FOR_DBFIELD_DEFAULTS
 
 
 class PURLFilter(CharFilter):
@@ -208,7 +277,7 @@ class NullForeignKeyFilter(BooleanFilter):
         return method(query)
 
 
-class InFilterSet(FilterSet):
+class InFilterSet(OSIDBFilterSet):
     """
     Adds 'in' support for ChoiceFilter, CharFilter, and UUIDFilter fields for Meta fields.
     Does not add 'in' for date and numeric comparisons.
@@ -247,7 +316,7 @@ class InFilterSet(FilterSet):
         return filters
 
 
-class DistinctFilterSet(FilterSet):
+class DistinctFilterSet(OSIDBFilterSet):
     """
     Custom FilterSet which enforces the distinct for field names that starts
     with specified prefixes so filtering on related models would not cause
@@ -324,7 +393,7 @@ class DistinctOrderingFilter(OrderingFilter):
         return queryset
 
 
-class SparseFieldsFilterSet(FilterSet):
+class SparseFieldsFilterSet(OSIDBFilterSet):
     def _preprocess_fields(self, value):
         """
         Converts a comma-separated list of fields into an ORM-friendly format.
@@ -517,10 +586,10 @@ class FlawFilter(
     cve_id = CharInFilter(field_name="cve_id")
     components = CharInFilter(field_name="components", lookup_expr="contains")
 
-    changed_after = DateTimeFilter(
+    changed_after = RelativeDateTimeFilter(
         field_name="updated_dt", method="changed_after_filter"
     )
-    changed_before = DateTimeFilter(
+    changed_before = RelativeDateTimeFilter(
         field_name="updated_dt", method="changed_before_filter"
     )
 
--- a/osidb/tests/test_relative_datetime_field.py
+++ b/osidb/tests/test_relative_datetime_field.py
@@ -1,0 +1,78 @@
+"""
+Unit tests for RelativeDateTimeField in osidb.filters
+"""
+
+from datetime import timedelta
+
+import pytest
+from django.core.exceptions import ValidationError
+from django.utils import timezone as django_timezone
+
+from osidb.filters import RelativeDateTimeField
+
+pytestmark = pytest.mark.unit
+
+
+class TestRelativeDateTimeField:
+    """
+    Unit tests for RelativeDateTimeField class - Django integration only
+    """
+
+    def test_required_field_with_empty_value(self):
+        """Test that required field raises ValidationError for empty values"""
+        field = RelativeDateTimeField(required=True)
+
+        with pytest.raises(ValidationError):
+            field.clean(None)
+
+        with pytest.raises(ValidationError):
+            field.clean("")
+
+    def test_optional_field_with_empty_value(self):
+        """Test that optional field accepts empty values"""
+        field = RelativeDateTimeField(required=False)
+
+        # Should not raise exception
+        result = field.clean(None)
+        assert result is None
+
+        result = field.clean("")
+        assert result is None
+
+    def test_clean_method_with_absolute_datetime(self):
+        """Test that clean() method properly processes absolute datetimes"""
+        field = RelativeDateTimeField()
+
+        result = field.clean("2024-06-15T14:30:00Z")
+
+        assert result is not None
+        assert result.year == 2024
+        assert result.month == 6
+        assert result.day == 15
+
+    def test_strptime_called_internally_during_clean(self):
+        """Test that strptime is called during field.clean() and handles relative datetimes"""
+        field = RelativeDateTimeField(input_formats=["%Y-%m-%d"])
+
+        # This should trigger strptime internally with the input format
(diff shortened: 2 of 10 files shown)


osidb/tests/test_relative_date_field.py
```diff
@@ -0,0 +1,349 @@
...
+        field = RelativeDateField()
+
+        result = field.to_python("0d")
+
+        # 0 days should be today
+        expected_date = django_timezone.now().date()
+        assert result == expected_date
+
+    def test_leap_year_relative_values(self):
+        """Test that leap year calculations work (366 days ago)"""
```
osoukup: How do we know this or previous year was a leap one? I do not see any year pinning.

osidb/tests/test_relative_date_field.py
```diff
@@ -0,0 +1,349 @@
...
+        field = RelativeDateField()
+
+        # Test with 366 days (leap year scenario)
+        result = field.to_python("-366d")
+
+        expected_date = django_timezone.now().date() - timedelta(days=366)
+        assert result == expected_date
+
+    def test_extreme_past_value_exceeds_date_min(self):
+        """Test that values exceeding date.min raise OverflowError"""
```
osoukup: Is it necessary to care for some boundaries? I do not see any issue in letting the user to query for -5000d or whatever day want. Sure -5000y probably makes no sense but is there any technical reason not to allow it?

osidb/tests/test_relative_date_field.py
```diff
@@ -0,0 +1,585 @@
...
+    def test_lowercase_m_is_minutes_for_date_field(self):
+        """-1m (minutes) when applied to date field"""
+        # Note: This tests edge case behavior - minutes don't make semantic sense
+        # for date-only fields, but we should handle it gracefully
+        field = RelativeDateField()
+        result = field.to_python("-1m")
+        # When relativedelta adds minutes to a date base, it returns datetime at 23:59 previous day
+        # For DateField, this is converted back to just the date part
+        assert isinstance(result, date)
+        assert result == date(2024, 6, 14)
```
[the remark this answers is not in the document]
AlejandroMinaya: Since our users may not be able to tell the difference between a DateTime and a Date field, it would create confusion if we raise an error for certain date params and not for others. The feedback is valid, but rejecting the request doesn't seem like a reasonable path forward

osidb/forms.py
```diff
@@ -0,0 +1,130 @@
...
+from django.utils import timezone
+
+# Module-level constants for relative date/datetime parsing
+RELATIVE_TIME_UNIT_MAP = {
+    "s": "seconds",
+    "m": "minutes",
+    "h": "hours",
+    "d": "days",
+    "w": "weeks",
+}
```
osoukup: Why no bigger period values?

osidb/forms.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            return super().strptime(value, format)
+        except (ValueError, TypeError):
+            parsed = _parse_relative_value(value, timezone.now())
+            if parsed is not None:
+                return parsed
+            raise
+
+
+class RelativeDateField(forms.DateField):
```
osoukup: I am not sure I understand the split between `Date` and `DateTime`, could you please explain why both are needed and where?

osidb/tests/test_relative_date_field.py
```diff
@@ -338,6 +324,256 @@ def test_mixed_positive_and_negative_weeks(self):
         expected_pos = django_timezone.now().date() + timedelta(weeks=3)
         assert result_pos == expected_pos
 
+    # Month and year tests
+    @freeze_time("2024-06-15")
+    def test_one_month_ago(self):
+        """Test -1M returns date 1 month ago"""
+        field = RelativeDateField()
+        result = field.to_python("-1M")
```
osoukup: I saw somewhere that a moth is denoted by `mo` 🤔 Each has an advantage and disadvantage

osoukup (commented):
I though that this is already ready for a review (was not a draft) but now I see I was probably incorrect and the PR is still being shaped. It seems to go well in the right direction.

osidb/forms.py
AlejandroMinaya: V1 Filters are using DateFilter instead of DateTimeFilter. In order to keep compatibility I created a RelativeDateFilter. I agree that it would be better to DateTimeFilter for everything, but I do not know if that change will break any underlying logic

osidb/tests/test_relative_date_field.py
AlejandroMinaya: Yes! I saw it too. I am okay using `mo` but I do not know what would be better for our current UX. I can argue for and against either, but ultimately it will come down to what our users find more intuitive.

osidb/forms.py
AlejandroMinaya: Since supporting months and years meant considering more irregularities (28 vs. 30 vs. 31 days; leap years), I decided to first address the changes at this scale and figure out the support for larger periods later on. This is addressed in a later commit

osidb/tests/test_relative_date_field.py
osoukup: I would probably maybe just discus that inside the DEV team and dictate one or the other way to the users :sweat_smile: not to make the process overly complicated. I can imagine that some would prefer one way and some the other but there would probably be no clear winner from the factual standpoint.

AlejandroMinaya: I am okay leaving 'M' until requested otherwise

osoukup: I probably do not have a strong preference here so you can just choose :wink:

osidb/tests/test_relative_datetime_filtering.py
```diff
@@ -0,0 +1,858 @@
...
+        results = response.json()["results"]
+        uuids = {r["uuid"] for r in results}
+
+        assert str(affects["affect_5_hours_ago"].uuid) in uuids
+        assert str(affects["affect_2_hours_ago"].uuid) in uuids
+        assert str(affects["affect_30_min_ago"].uuid) not in uuids
+
+    # Note: The affects__created_dt filter uses the AffectV1 view which may not
+    # reflect timestamp changes made with auto_timestamps=False in tests.
+    # Commenting out this test as it requires investigation of the view refresh behavior.
```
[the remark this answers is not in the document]
AlejandroMinaya: I originally wanted to include this test for thoroughness but I wasn't aware at the moment that it depended on the AffectsV1 Materialized View. I think we can safely drop it since in the end it is using the same django-filter FilterSet and coupling this test with an AffectsV1 refresh is not worth it

osidb/tests/test_relative_date_field.py
AlejandroMinaya: I still have to perform a cleaning sweep through the tests, but you're right to call this out. I don't think we need this test. Thank you

osidb/tests/test_relative_date_field.py
AlejandroMinaya: I wanted to make sure that we handled correctly relative queries out of the date.min and date.max bounds but in either case it should behave the same as if you query for an absolute date outside this range. Do you think it is worth it to raise it as a ValidationError instead?

AlejandroMinaya:
@coderabbitai full-review

AlejandroMinaya:
@coderabbitai help

osidb/tests/test_relative_datetime_filtering.py
```diff
@@ -0,0 +1,822 @@
...
+
+        response = auth_client().get(f"{test_api_uri}/flaws?created_dt__lte=-4M")
+        assert response.status_code == status.HTTP_200_OK
+        results = response.json()["results"]
+        uuids = {r["uuid"] for r in results}
+
+        assert str(flaws["flaw_6_months_ago"].uuid) in uuids
+        assert str(flaws["flaw_1_month_ago"].uuid) not in uuids
+        assert str(flaws["flaw_1_week_ago"].uuid) not in uuids
+
```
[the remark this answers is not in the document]
AlejandroMinaya: This is no longer necessary since this test was for V1 filters and these are deprecated

AlejandroMinaya:
@coderabbitai full-review

osidb/tests/test_relative_datetime_filtering.py
```diff
@@ -0,0 +1,263 @@
...
+        assert response.status_code == status.HTTP_200_OK
+        results = response.json()["results"]
+        uuids = {r["uuid"] for r in results}
+
+        assert str(trackers["tracker_3_days_ago"].uuid) in uuids
+        assert str(trackers["tracker_1_day_ago"].uuid) in uuids
+        assert str(trackers["tracker_2_hours_ago"].uuid) not in uuids
+
+    # Note: affects__trackers__created_dt filtering has a bug in the filter configuration
+    # (uses method name instead of field name). Skipping this test until the filter is fixed.
```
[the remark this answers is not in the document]
AlejandroMinaya: V1 filters are deprecated

osidb/forms.py
```diff
@@ -0,0 +1,166 @@
...
+    sign, amount, unit = match.groups()
+    amount = int(amount) if sign != "-" else -int(amount)
+
+    # Case-sensitive lookup to distinguish 'm' (minutes) from 'M' (months)
+    delta_unit = RELATIVE_TIME_UNIT_MAP.get(unit)
+    if not delta_unit:
+        return None
+
+    delta_kwargs = {delta_unit: amount}
+    return base_value + relativedelta(**delta_kwargs)
```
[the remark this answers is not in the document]
AlejandroMinaya: This has been addressed

osidb/tests/test_relative_datetime_field.py
```diff
@@ -0,0 +1,338 @@
...
+    def test_parse_invalid_format_no_sign(self):
+        """Test that missing sign but valid number returns datetime (defaults to +)"""
+        # Note: Based on regex, this should actually work (sign is optional)
+        field = RelativeDateTimeField()
+
+        result = field.to_python("5h")
+        # The datetime should be approximately 1 day ago
+        expected_time = django_timezone.now() + timedelta(hours=5)
+
+        # Allow 1 second tolerance for test execution time
```
Elkasitu: For these sort of tests we like to use `freeze_time`, this makes comparing times directly easy and reliable.

osidb/filters.py
```diff
@@ -81,6 +82,26 @@ class UUIDInFilter(BaseInFilter, UUIDFilter):
...
+    RelativeDateTimeFilter that accepts both absolute datetime values (default behavior)
+    and relative datetime strings like "-1d", "+2h", "-30m", etc.
+
+    Relative format: [+/-]<number><unit>
+    where unit is one of: s (seconds), m (minutes), h (hours), d (days), w (weeks)
+
+    Examples:
+        -1d     -> 1 day ago
+        +2h     -> 2 hours from now
+        -30m    -> 30 minutes ago
```
Elkasitu, line 100: we should add an example without signs, the format just above hints at the possibility but it's not immediately clear, seems like we push for explicit sign usage

osidb/forms.py
```diff
@@ -0,0 +1,96 @@
+"""
+Custom form fields for OSIDB
```
Elkasitu: Why? we don't use forms and the only place in which we use the logic herein is in `filters.py`, why not have this logic directly there?

osidb/filters.py
```diff
@@ -791,6 +791,11 @@ class Meta:
             "references__url": ["exact"],
             "references__uuid": ["exact"],
         }
+        filter_overrides = {
+            models.DateTimeField: {
+                "filter_class": RelativeDateTimeFilter,
+            },
+        }
```
Elkasitu: Instead of copy-pasting this for every `FilterSet`, we should either create a base class which each FilterSet inherits from or override it in `DJANGO_FILTERS.FILTERSET_BASE_CLASS`.

osidb/tests/test_relative_datetime_filtering.py
```diff
@@ -0,0 +1,620 @@
...
+    # reflect timestamp changes made with auto_timestamps=False in tests.
+    # Commenting out this test as it requires investigation of the view refresh behavior.
+
+
+class TestRelativeDateTimeFilteringTrackers:
+    """
+    Test relative datetime filtering on Tracker endpoints
+    """
+
+    @pytest.fixture
```
Elkasitu: we tend to prefer fixtures in `conftest.py`

osidb/filters.py
Elkasitu: We might also want to replace explicit field declarations e.g. `created_dt = DateTimeFilter` for `created_dt = RelativeDateTimeFilter` as the `filter_overrides` only works for generated fields, it seems like you eventually realized and replaced `changed_after` and `changed_before` but we should do it for all explicitly declared `DateTimeFilter`

osidb/tests/test_relative_datetime_field.py
```diff
@@ -0,0 +1,388 @@
...
+        assert abs(result - expected_time) < timedelta(seconds=1)
```
AlejandroMinaya: I'm dropping this test since it is a carry-over from initial iterations

osidb/tests/test_relative_datetime_field.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def test_leap_year_relative_values(self):
+        """Test that leap year calculations work (366 days ago)"""
+        field = RelativeDateTimeField()
+
+        # Test with 366 days (leap year scenario)
+        result = field.to_python("-366d")
+
+        expected_time = django_timezone.now() - timedelta(days=366)
+        assert abs(result - expected_time) < timedelta(seconds=1)
```
AlejandroMinaya: Dropping this test too

osidb/tests/test_relative_datetime_field.py
```diff
@@ -145,30 +145,6 @@ def test_leap_year_relative_values(self):
...
-        field = RelativeDateTimeField()
-
-        # Calculate days that would exceed date.max (year 9999)
-        now = django_timezone.now()
-        days_to_max = (date.max - now.date()).days + 1000  # Go beyond max
-
-        # This should raise OverflowError when timedelta calculation exceeds range
-        with pytest.raises(OverflowError):
-            field.to_python(f"+{days_to_max}d")
-
```
Elkasitu: The commit message says that instead of `OverflowError` we raise `ValidationError` instead, this is not reflected in the tests

Elkasitu (changes requested):
Heading in the right direction but requires some polish still, please ensure all commits are squashed into a single one before merging

osoukup (approved):
Looks well, good job!

AlejandroMinaya:
In order to support OSIM Advanced Search I need to create a relative date field for DjangoQL. I will be pushing those changes later today

Elkasitu (approved):

PR #1349, opened 2026-07-06
workflow framework

osoukup:
The workflow framework changes needs to get to `master` as otherwise we cannot test the Argus fully. The master have ACE changes, relative time filters, and probably more. This branch has also a lot. We could go the other way round making it easier for us if we need and urgent release, on the other hand, `workflow-framework` branch is already hard to maintain and I literary spending ours to rebase it and make all the checks passing with every change :sweat_smile: so I propose this and just go out with it.

I am not sure how much review this needs. It is a lot and everything already got reviewed in specific PRs.

### Added
- document workflow framework
- introduce EMBARGOED workflow
- introduce workflow labels

### Changed
- align existing workflows with workflow framework concepts
- auto-adjust workflow:state on Flaw.save
- unrestrict workflow state names (OSIDB-5049)
- move automatic visibility change handling specification into the workflow definitions
- align workflow definition YAMLs with the Argus requirements (OSIDB-5050)
- introduce true labels and deprecate collaborator-centric pseudo-labels (OSIDB-5191)

### Removed
- deprecate workflow manipulation endpoints
- disconnect OSIDB workflow:state from the Jira
- stop classifying legacy flaws


dmonzonis (approved):

Elkasitu (approved):

roduran-dev (approved):
Just in case as I'm not a 100% about this change. 

ACE apps/ace/tasks.py in _apply_label still writes to legacy FlawCollaborator, but workflow classification reads from labels_v2 via Flaw.has_label(). 
This means that labels like auto-rejected and manual-triage written by ACE won't trigger reclassification via the label post_save → flaw re-save chain.

Otherwise LGTM.

PR #1350, opened 2026-07-07
Update ACE task to use new workflow framework

dmonzonis:
The new workflow framework uses `WorkflowLabels` instead of the old labels. ACE will rely on this framework so that flaws can be auto-rejected under certain conditions, using the new labels, so the ACE code has been updated accordingly.


osoukup (approved):
LGTM

PR #1353, opened 2026-07-08
Cra regulatory reporting test1

roduran-dev:
test openapi


roduran-dev:
test showed that it failed,

PR #1354, opened 2026-07-08
Cra regulatory reporting test2

roduran-dev:
Test to see after a second rebase.


roduran-dev:
Test showed that updating the file to how it was in master fixed the problem.

PR #1359, opened 2026-07-09
Add worktree make commands for runtime isolation

AlejandroMinaya:
## Summary
- Added make commands for running OSIDB in git worktrees with runtime isolation
- Auto-generates unique ports per worktree to avoid conflicts with main development environment
- Adds worktree-specific docker-compose overrides

## Changes
- New `mk/worktree.mk` with commands: `worktree-start`, `worktree-dev-env`, `worktree-migrate`, etc.
- New `scripts/generate-worktree-ports.sh` for automatic port allocation based on worktree name
- New `docker-compose.worktree.yml` and `docker-compose.test.worktree.yml` override files
- Updated main docker-compose files to support configurable ports via environment variables

### DIFF
--- a/docker-compose.worktree.yml
+++ b/docker-compose.worktree.yml
@@ -1,0 +1,31 @@
+version: '3'
+
+# Worktree isolation override
+# Usage: docker-compose -f docker-compose.yml -f docker-compose.worktree.yml up
+# Set COMPOSE_PROJECT_NAME to namespace containers/volumes per worktree
+
+services:
+  osidb-service:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-service
+
+  osidb-data:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-data
+
+  redis:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-redis
+
+  flower:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-flower
+
+  celery_beat:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-celery-beat
+
+  celery-fifo-1:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-celery-fifo-1
+
+  celery-fifo-2:
+    container_name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-celery-fifo-2
+
+volumes:
+  pg-data:
+    name: ${COMPOSE_PROJECT_NAME:?COMPOSE_PROJECT_NAME is required - run make worktree-generate-env first}-pg-data
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -68,7 +68,7 @@ services:
       stdin_open: true
       tty: true
       ports:
-        - "8000:8000"
+        - "${OSIDB_PORT:-8000}:8000"
       environment:
         BBSYNC_SYNC_TO_BZ: ${BBSYNC_SYNC_TO_BZ}
         BBSYNC_SYNC_FLAWS_TO_BZ: ${BBSYNC_SYNC_FLAWS_TO_BZ}
@@ -144,7 +144,7 @@ services:
         - ${PWD}/etc/pg/local-server.crt:/var/lib/postgresql/local-server.crt:z
         - ${PWD}/etc/pg/local-server.key:/var/lib/postgresql/local-server.key:z
       ports:
-        - "5432"
+        - "${POSTGRES_PORT:+${POSTGRES_PORT}:}5432"
       command: >
           -c ssl=on
           -c ssl_cert_file=/var/lib/postgresql/local-server.crt
@@ -269,7 +269,7 @@ services:
       pull_policy: never
       command: ./scripts/run-flower.sh
       ports:
-        - "5555:5555"
+        - "${FLOWER_PORT:-5555}:5555"
       volumes:
         - ${PWD}:/opt/app-root/src:z
       environment:
@@ -287,7 +287,7 @@ services:
       hostname: redis
       image: mirror.gcr.io/library/redis:6
       ports:
-        - "6379"
+        - "${REDIS_PORT:+${REDIS_PORT}:}6379"
 
 volumes:
   pg-data:
--- a/mk/worktree.mk
+++ b/mk/worktree.mk
@@ -1,0 +1,169 @@
+############################################################################
+## Worktree-isolated development
+############################################################################
+
+# Load worktree-specific environment if it exists
+-include .env.worktree
+
+
+#***********************************
+### Generate worktree environment file with unique ports
+#***********************************
+.PHONY: worktree-generate-env
+worktree-generate-env:
+	@./scripts/generate-worktree-ports.sh
+
+
+#***********************************
+### Ensure .env.worktree exists (generate if absent)
+#***********************************
+.PHONY : worktree-ensure-env
+worktree-ensure-env:
+	@[ -f .env.worktree ] || $(MAKE) worktree-generate-env
+
+
+#***********************************
+### Build worktree-isolated docker images
+#***********************************
+.PHONY: worktree-build
+worktree-build: worktree-ensure-env
+	@echo ">building docker images for worktree: $(COMPOSE_PROJECT_NAME)"
+	$(podman) compose --env-file=.env.worktree -f docker-compose.yml -f docker-compose.worktree.yml -f docker-compose.test.yml -f docker-compose.test.worktree.yml down
+	$(podman) compose --env-file=.env.worktree -f docker-compose.yml -f docker-compose.worktree.yml build
+	$(podman) compose --env-file=.env.worktree -f docker-compose.yml -f docker-compose.worktree.yml pull
+	$(podman) compose --env-file=.env.worktree -f docker-compose.test.yml -f docker-compose.test.worktree.yml build
+	$(podman) compose --env-file=.env.worktree -f docker-compose.test.yml -f docker-compose.test.worktree.yml pull
+
+
+#***********************************
+### Build worktree dev env: images, venv, do checks
+#***********************************
+.PHONY: worktree-dev-env
+worktree-dev-env:
+	@echo -n "This will check installed RPMs, create certificate files, build podman images, and create a python venv for this worktree. Continue? [y/N]" && read ans && [ $${ans:-N} = y ]
+	$(MAKE) worktree-generate-env
+	$(MAKE) dev-rpm-install
+	$(MAKE) check-reg
+	$(MAKE) generate_local_pg_tls_cert
+	$(MAKE) worktree-build
+	$(MAKE) venv
+	$(MAKE) check-venv
+
+
+#***********************************
+### Start containers for worktree-isolated development
+#***********************************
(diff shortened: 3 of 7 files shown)


osoukup (commented):
What is the use case here? To have multiple parallel local OSIDB instances? Or to switch more easily? I started to read the code but stopped and would first like to understand. Should the worktrees share .env file, config, DB?

AlejandroMinaya:
> What is the use case here? To have multiple parallel local OSIDB instances? Or to switch more easily?

I believe that it would allow us to switch context more easily. Ideally, this would not be needed, but if you want to checkout changes from somebody else or maybe want to knock out a quick bug fix while working on a task, this allows you to do that without having to disrupt your setup.

I know that for agentic parallel development it may be useful, but that would be a happy side-effect, not necessarily my goal.

> Should the worktrees share .env file, config, DB?

They shouldn't. You do need an .env file to apply modifications to, but I included a script that appends specific ports for that specific worktree directory

AlejandroMinaya:
@coderabbitai review

docker-compose.worktree.yml
```diff
@@ -0,0 +1,31 @@
...
+
+  celery-fifo-1:
+    container_name: ${COMPOSE_PROJECT_NAME:-osidb}-celery-fifo-1
+
+  celery-fifo-2:
+    container_name: ${COMPOSE_PROJECT_NAME:-osidb}-celery-fifo-2
+
+volumes:
+  pg-data:
+    name: ${COMPOSE_PROJECT_NAME:-osidb}-pg-data
```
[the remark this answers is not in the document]
AlejandroMinaya: Solved

scripts/generate-worktree-ports.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Auto-generate unique ports for worktree based on directory name
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+IFS=$' \t\n'
+set -euo pipefail
+
+WORKTREE_NAME=$(basename "$PWD" | tr '[:upper:]' '[:lower:]' | sed 's/^[^a-z0-9]*//' | sed 's/[^a-z0-9_-]/-/g')
```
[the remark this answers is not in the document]
AlejandroMinaya: Since the name is extracted from the PWD, wouldn't this be unlikely?

scripts/generate-worktree-ports.sh
```diff
@@ -0,0 +1,67 @@
...
+    tries=0
+    while _has_port_conflict "$PORT_OFFSET"; do
+        PORT_OFFSET=$(( (PORT_OFFSET % 999) + 1 ))
+        tries=$((tries + 1))
+        [ "$tries" -lt 999 ] || { echo "ERROR: Cannot find a non-conflicting port allocation" >&2; exit 1; }
+    done
+fi
+
+# Copy base .env to .env.worktree
+cp .env .env.worktree
```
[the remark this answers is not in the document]
AlejandroMinaya: This is a valid concern for parallel agentic development, but we are not there yet. Will skip for now

mk/worktree.mk
```diff
@@ -0,0 +1,145 @@
...
+# Load worktree-specific environment if it exists
+-include .env.worktree
+
+
+#***********************************
+### Generate worktree environment file with unique ports
+#***********************************
+.PHONY: worktree-generate-env
+worktree-generate-env:
+	@./scripts/generate-worktree-ports.sh
```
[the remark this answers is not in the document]
AlejandroMinaya: Solved

scripts/generate-worktree-ports.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Auto-generate unique ports for worktree based on directory name
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+IFS=$' \t\n'
+set -euo pipefail
+
+WORKTREE_NAME=$(basename "$PWD" | tr '[:upper:]' '[:lower:]' | sed 's/^[^a-z0-9]*//' | sed 's/[^a-z0-9_-]/-/g')
+PORT_OFFSET=$(echo "$WORKTREE_NAME" | cksum | cut -d' ' -f1)
+PORT_OFFSET=$(( (PORT_OFFSET % 999) + 1 ))
```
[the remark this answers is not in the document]
AlejandroMinaya: Done

Elkasitu:
>I believe that it would allow us to switch context more easily. Ideally, this would not be needed, but if you want to checkout changes from somebody else or maybe want to knock out a quick bug fix while working on a task, this allows you to do that without having to disrupt your setup.

What exactly is so disrupting about the current setup? If not using worktrees, a git checkout will usually trigger a hot reload of the server, at most you may have to nuke the database and run `make start-local`.

I haven't used worktrees in a while, but wouldn't something like the following snippet restart the server with the code from the other branch?

```bash
make stop-local
cd ../my-worktree-feature-branch
make start-local
```

Why do you need instances of OSIDB running in parallel? A concrete example would help.

mk/worktree.mk
```diff
@@ -0,0 +1,161 @@
+############################################################################
+## Worktree-isolated development
+############################################################################
+
+# Load worktree-specific environment if it exists
+-include .env.worktree
```
JakubFrejlach, line 6: `.env.worktree` should be gitignored

docker-compose.yml
```diff
@@ -144,7 +144,7 @@ services:
         - ${PWD}/etc/pg/local-server.crt:/var/lib/postgresql/local-server.crt:z
         - ${PWD}/etc/pg/local-server.key:/var/lib/postgresql/local-server.key:z
       ports:
-        - "5432"
+        - "${POSTGRES_PORT:-5432}:5432"
```
JakubFrejlach: It has been some time since we did the compose files, but I think the intention here was for devs to not have the postgres default port mapped by default so it is free if some other postgres is being run as well.

docker-compose.yml
```diff
@@ -287,7 +287,7 @@ services:
       hostname: redis
       image: mirror.gcr.io/library/redis:6
       ports:
-        - "6379"
+        - "${REDIS_PORT:-6379}:6379"
```
JakubFrejlach: Same as for the postgres port.

AlejandroMinaya:
> I haven't used worktrees in a while, but wouldn't something like the following snippet restart the server with the code from the other branch?

Yes!

> Why do you need instances of OSIDB running in parallel? A concrete example would help.

I initially didn't think it would be necessary either. The scenario that pushed me to do this was having setup `django-silk` with all its migrations for the optimization ticket and wanting to work on another task while waiting for a review. 

I didn't want to compromise the DB state since I was comparing metrics with the same data but different commits, but for the other ticket I needed to mutate the data. Instead of juggling the desired DB state if I needed to switch tasks, I figured that we could leverage containerization and spin up different instances.

scripts/generate-worktree-ports.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Auto-generate unique ports for worktree based on directory name
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
JakubFrejlach: Do we need the `PATH` override here?

scripts/generate-worktree-ports.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+# Copy base .env to .env.worktree
+cp .env .env.worktree
+
+# Append worktree-specific overrides
+cat >> .env.worktree <<EOF
+
+# === Worktree-specific overrides below ===
+# Auto-generated - DO NOT EDIT - regenerate with: make generate-worktree-env
```
JakubFrejlach: Mismatch in make target `worktree-generate-env` vs `generate-worktree-env`

mk/worktree.mk
```diff
@@ -0,0 +1,161 @@
...
+	@echo ">Use 'make worktree-stop' to stop containers"
+
+
+#***********************************
+### Stop worktree containers without deleting
+#***********************************
+.PHONY : worktree-stop
+worktree-stop: worktree-ensure-env
+	@echo ">stopping worktree env without deleting containers"
+	$(podman) compose --env-file=.env.worktree -f docker-compose.yml -f docker-compose.worktree.yml -f docker-compose.test.yml -f docker-compose.test.worktree.yml stop || $(podman) compose --env-file=.env.worktree -f docker-compose.yml -f docker-compose.worktree.yml -f docker-compose.test.yml -f docker-compose.test.worktree.yml stop testrunner testldap flower redis celery_beat celery celery-fifo-1 celery-fifo-2 locust osidb-service osidb-data
```
JakubFrejlach: this is really long command, can you please break it up via `\` for readability?

JakubFrejlach (commented):
I am generally not against this change. I personally do have a similar workflow to @Elkasitu  and I do not use worktrees, so I cannot say this would solve an issue for me currently. That said, I can see the value for developers who do work with multiple branches simultaneously.

JakubFrejlach:
> I initially didn't think it would be necessary either. The scenario that pushed me to do this was having setup `django-silk` with all its migrations for the optimization ticket and wanting to work on another task while waiting for a review.
> 
> I didn't want to compromise the DB state since I was comparing metrics with the same data but different commits, but for the other ticket I needed to mutate the data. Instead of juggling the desired DB state if I needed to switch tasks, I figured that we could leverage containerization and spin up different instances.

Ok, this one is valid and I agree, experienced similar frustration when working with django-silk some time ago. Damn, people just tend to forget negative stuff :joy:

Elkasitu:
> I initially didn't think it would be necessary either. The scenario that pushed me to do this was having setup `django-silk` with all its migrations for the optimization ticket and wanting to work on another task while waiting for a review.
> 
> I didn't want to compromise the DB state since I was comparing metrics with the same data but different commits, but for the other ticket I needed to mutate the data. Instead of juggling the desired DB state if I needed to switch tasks, I figured that we could leverage containerization and spin up different instances.

Thanks for the added context, please correct me if I'm wrong, but I think there's two takeaways from this:

1. we regularly need django-silk for performance testing
2. we need to fill the database with test data in order to perform meaningful performance testing

So I think that the disruptions you run into are more of a symptom than the core issue, regardless of git workflow as pointed out by Jakub, and I think that the core issue is that you need meaningful data in the local DB in order to do performance testing. You don't really care about having one or N instances of OSIDB running, you just really don't wanna lose the test data you've created by switching context.

I think we can address both gaps in the following way:

* add `django-silk` to the dev dependencies once and for all
* add meaningful test data by leveraging something like [Django fixtures](https://docs.djangoproject.com/en/5.2/topics/db/fixtures/), whether this is done ad-hoc or we commit a set of fixtures to the repository for everyone to use is up to you, either one works for me

With this approach, you could simply restart OSIDB, load the test data and resume testing, WDYT?

If we really want to support parallel OSIDB instances, I suggest we simply parametrize the relevant ports and get rid of the explicit `container_name` declarations. This should make the existing compose files worktree-compatible without needing new make commands and/or compose files, as by default `docker-compose` namespaces containers based on a project label that defaults to the directory name, which would be different per branch in a worktree setting.

AlejandroMinaya:
I like the idea of adding `django-silk` as a dev dependency. I believe that we're all using it in our own way and could benefit of having a shared configuration.

I also like the idea of having dummy fixtures for testing or even just populating the database. You are right to point out that what I bring up is not a need to actually run OSIDB in parallel, but rather a need to have consistent testing data.

Regardless, I am glad that this has served as a conversation starter for how we would go about supporting parallel development workflows

AlejandroMinaya:
Having said that, other than the change to the `docker-compose.yml` and `docker-compose.test.yml` I think this is a pretty inoffensive addition. Would it be okay to merge this commands until we address adding `django-silk` and alternatives for parallel workflows? @Elkasitu

osoukup (commented):
I am not sure about this. It adds a layer of complexity I am not sure I will ever use. Each of us has some development habits and we probably invented multiple different ways how to deal with the issue of changing OSIDB instances/branches. This PR seems to me as yet another personal preference. However, it might be beneficial for a new-comers to have something prepared so they do not have to re-invent yet another shape of a wheel. That would however only make sense if properly documented which does not seem to be the case here.

PR #1363, opened 2026-07-09
OSIDB - 5083- Upstream maintainer notification endpoints

svelamak:
Add Upstream Maintainer notification endpoints with create, view, filter and send emails.
- Add `UpstreamNotificationView` in `regulatory_reporting/views.py` and also `send-email.`
- Added embargo check in `regulatory_reporting/services.py`.
- Added celery callbacks in `regulatory_reporting/tasks.py`.
- Added `paylod_text` and `QUEUED` status in `models/upstream.py` and migrations.
- Added new email templates `upstream_notification.txt` and `upstream_notification.html`.
- Registered view in `osidb/urls.py`.
- Added tests for embargo check in `test_services.py`.
- Added tests for endpoints in `test_views.py`.
- Generated `openapi.yml`
- Entry in `CHANGELOG.md`
Closes: OSIDB-5083


osidb/urls.py
```diff
@@ -103,6 +104,11 @@
 vnext_router.register(r"flaws", FlawView)
 vnext_router.register(r"affects", AffectView, basename="affectsv2")
 vnext_router.register(r"trackers", TrackerView)
+vnext_router.register(
+    r"notifications/upstream",
+    UpstreamNotificationView,
+    basename="upstreamnotifications",
+)
```
JakubFrejlach: I think this should not be registered by default, but rather opt-in via variable?

regulatory_reporting/tests/test_views.py
```diff
@@ -0,0 +1,101 @@
...
+import pytest
+
+from osidb.models.flaw import FlawSource
+from osidb.tests.factories import FlawFactory
+from regulatory_reporting.models.upstream import UpstreamNotification
+
+from .factories import UpstreamNotificationFactory, UpstreamProjectFactory
+
+
+@pytest.mark.django_db
```
JakubFrejlach: You can create `regulatory_reporting/tests/conftest.py` file and put an autouse fixture which just enables tests access to database by default so you do not need to spam this pytest mark:
 ```python
@pytest.fixture(autouse=True)
def enable_db_access_for_all_tests(db):
    pass
```

File: regulatory_reporting/views.py [hunk 3]
```diff
@@ -1,3 +1,107 @@
...
+    status = ChoiceFilter(
+        field_name="status",
+        choices=UpstreamNotification.NotificationStatus.choices,
+    )
+    method = ChoiceFilter(
+        field_name="method",
+        choices=UpstreamNotification.NotificationMethod.choices,
+    )
+    upstream_project = CharFilter(field_name="upstream_project__uuid")
+    flaw = CharFilter(field_name="flaw__uuid")
```
JakubFrejlach: filters should ideally go into its own `filters.py` file

openapi.yml
```diff
@@ -12718,6 +12718,369 @@ paths:
                   version:
                     type: string
           description: ''
+  /osidb/api/v2/notifications/upstream:
```
JakubFrejlach, line 12721: There should be ideally no changes to this file as all the CRA related endpoints should be opt in and thus only displayed in dynamically generated OpenAPI schema by the instance of OSIDB (producation presumably) which will have it turned on.

regulatory_reporting/views.py
```diff
@@ -1,3 +1,107 @@
...
+        text_body = render_to_string("email/upstream_notification.txt", context=context)
+        html_body = render_to_string(
+            "email/upstream_notification.html", context=context
+        )
+
+        payload = {
+            "subject": f"Security notification for {flaw_id}",
+            "to": [upstream_project.security_contact],
+            "body": text_body,
+            "reply_to": [upstream_project.security_contact],
```
JakubFrejlach: Will this be a mailing list or direct email? If latter then it does not make sense to be `reply_to`

osidb/templates/email/upstream_notification.html
```diff
@@ -0,0 +1,27 @@
...
+    <p><strong>Known exploitation state:</strong> {{ exploitation_state }}</p>
+    {% endif %}
+    {% if embargo_notice %}
+    <p>{{ embargo_notice }}</p>
+    {% endif %}
+    {% if corrective_measure %}
+    <p><strong>Corrective or mitigating measure:</strong> {{ corrective_measure }}</p>
+    {% endif %}
+
+    <p>Sent by Red Hat Product Security under applicable regulatory reuirements.</p>
```
JakubFrejlach: typo: `reuirements` -> `requirements`

osidb/templates/email/upstream_notification.txt
```diff
@@ -0,0 +1,20 @@
...
+Known exploitation state: {{ exploitation_state }}
+{% endif %}
+{% if embargo_notice %}
+{{ embargo_notice }}
+{% endif %}
+{% if corrective_measure %}
+Corrective or mitigating measure: {{ corrective_measure }}
+{% endif %}
+
+Sent by Red Hat Product Security under applicable regulatory reuirements.
```
JakubFrejlach: typo: `reuirements` -> `requirements`

svelamak: fixed

osidb/templates/email/upstream_notification.html
svelamak: fixed

osidb/urls.py
svelamak: Fixed. Wrapped the router registraion.

regulatory_reporting/tests/test_views.py
svelamak: Added `conftest.py` with autouse db fixture as suggested and removed marks from test files..

regulatory_reporting/views.py
svelamak: Moved to its own `filters.py`

openapi.yml
svelamak: This got resolved.

regulatory_reporting/views.py
svelamak: Thanks for catching this! Checked model, migrations and CRA doc.
This is still an open/undecided question, Will discuss with the team and update on it.

.secrets.baseline
```diff
@@ -548,6 +548,32 @@
...
+      },
+      {
+        "type": "Hex High Entropy String",
+        "filename": "regulatory_reporting/migrations/0004_remove_upstreamnotification_insert_insert_and_more.py",
+        "hashed_secret": "4a3dbaada8feb2f977c0e79768d6e47572165912",
+        "is_verified": false,
+        "line_number": 117,
+        "is_secret": true
+      }
+    ],
```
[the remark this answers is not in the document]
svelamak: These 3 entries are SHA1 hashes of pgtrigger-generated SQL, not real secrets. Yes, they are false positives. CI secret check passes with them as well. I looked into flipping them via detect-secret audit, but 1.4.0 reports "nothing to audit" for entries already recorded in baseline and a full rescan would re-audit all 52 findings. Leaving as is for now, can revisit as a separate cleanup if preferred.

svelamak:
@coderabbitai
 Fixed. Wrapped the enqueue in try/except block
 Added test_send_email_wrong_method_fails for non-email gate and added an assertion in the success test.

regulatory_reporting/views.py
costaconrado: I believe you're right, that was a design mistake I made, we don't need to support multiple emails at this time

svelamak: Thanks Conrado! 
Removed `reply_to` since its a single email.

regulatory_reporting/views.py
```diff
@@ -1,3 +1,110 @@
...
+from .filters import UpstreamNotificationFilter
+from .models.upstream import UpstreamNotification
+from .serializers.upstream import UpstreamNotificationSerializer
+from .tasks import mark_upstream_notification_failed, mark_upstream_notification_sent
+
+
+@include_exclude_fields_extend_schema_view
+class UpstreamNotificationView(RudimentaryUserPathLoggingMixin, ModelViewSet):
+    """
+    API endpoint for listing, creating, retrieving, and updating upstream maintainer notifications.
```
JakubFrejlach, line 33: Is UpstreamNotification deletion intended? Because the docstring does not mentions it but `ModelViewSet` implements it.

tox.ini
```diff
@@ -10,6 +10,7 @@ setenv =
     OSIDB_DEBUG = 1
     DJANGO_SETTINGS_MODULE=config.settings_local
     DJANGO_SECRET_KEY = local
+    CRA_NOTIFICATIONS_ENABLED = true
```
JakubFrejlach, line 13: I think this should be enabled for a specific set of tests rather than globally for all tests.

tox.ini
```diff
@@ -53,6 +54,7 @@ setenv =
     OSIDB_DEBUG = 1
     DJANGO_SETTINGS_MODULE=config.settings_ci
     DJANGO_SECRET_KEY = ci
+    CRA_NOTIFICATIONS_ENABLED = true
```
JakubFrejlach: I think this should be enabled for a specific set of tests rather than globally for all tests.

tox.ini
```diff
@@ -70,6 +72,8 @@ allowlist_externals = /usr/bin/bash
 commands = /usr/bin/bash -c './scripts/migrations-check.sh'
 
 [testenv:schema-check]
+setenv =
+    CI = true
```
JakubFrejlach: What is this for?

tox.ini
svelamak: Fixed.

tox.ini
svelamak: Fixed.

regulatory_reporting/views.py
svelamak: Confirmed that no use case for deleting notifications. Switched to explicit mixins.

tox.ini
svelamak: I added `CI = true ` to schema-check tox env as the CI check was failing by default. Thought this fixes the old venv check
Now removed as it is sorted.

regulatory_reporting/views.py
```diff
@@ -1,3 +1,117 @@
...
+    """
+
+    queryset = UpstreamNotification.objects.all()
+    serializer_class = UpstreamNotificationSerializer
+    filter_backends = (DjangoFilterBackend,)
+    filterset_class = UpstreamNotificationFilter
+    lookup_url_kwarg = "notification_uuid"
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
costaconrado, line 44: This view should use `get_valid_http_methods` like the rest of OSIDB viewsets. Otherwise PATCH remains enabled and `send-email` can still be called when OSIDB is in read-only mode. Some thing like:

```
from osidb.api_views import get_valid_http_methods
...
class UpstreamNotificationView(...):
    http_method_names = get_valid_http_methods(viewsets.GenericViewSet, excluded=["delete"])
```

regulatory_reporting/tasks.py
```diff
@@ -0,0 +1,24 @@
+from celery import shared_task
+
+from .models.upstream import UpstreamNotification
+
+
+@shared_task
```
costaconrado: I wonder if we shouldn't use regular celery task to be consistent with the rest of OSIDB code, is there any specific reason for using `shared_task` instead of `task`?

regulatory_reporting/tasks.py
```diff
@@ -0,0 +1,24 @@
...
+def mark_upstream_notification_sent(result, notification_uuid):
+    """
+    Success callback once async_send_email completes successfully.
+    """
+    UpstreamNotification.objects.filter(uuid=notification_uuid).update(
+        status=UpstreamNotification.NotificationStatus.SENT,
+    )
```
costaconrado, line 18: usually tasks that runs on Celery will not have ACL set automatically, this might need something like `set_user_acls(settings.ALL_GROUPS)` before being able to update the entries, please double check and feel free to ignore if this works with RLS enabled.

costaconrado (changes requested):
I think regulatory notifications should be enabled by default in tox/CI. In this repo, merged code should validate optional modules as if they are mandatory (I believe we want to have regulatory notification endpoints on the openapi), with separate checks only for the disabled-mode behavior.

Right now `ci-osidb` sets `CRA_NOTIFICATIONS_ENABLED=true`, but then runs:
```
pytest --ignore=regulatory_reporting
```

So the env var is enabled, but the notification endpoint tests are skipped. If we want these tests covered by default, I'd remove the ignore and run the full suite from `ci-osidb`.

You can shape it as you prefer but I suggest something like this for the tox:
```
    [testenv]
    passenv = *
    basepython = python3.12
    setenv =
        OSIDB_DEBUG = 1
        DJANGO_SETTINGS_MODULE = config.settings_local
        DJANGO_SECRET_KEY = local
        CRA_NOTIFICATIONS_ENABLED = true
    runner = uv-venv-runner

    [testenv:ci-osidb]
    setenv =
        {[testenv]setenv}
        DJANGO_SETTINGS_MODULE = config.settings_ci
        DJANGO_SECRET_KEY = ci
    commands =
        pytest
```

With that, the current GitHub workflow can stay as:
```
    run: uvx tox -e ci-osidb
```

For OpenAPI, schema generation also needs `CRA_NOTIFICATIONS_ENABLED=true`, because the route is registered at URLConf import time:

    if settings.CRA_NOTIFICATIONS_ENABLED:
        vnext_router.register(...)

That means a per-test marker is too late for schema generation. The flag needs to be set in tox before `manage.py spectacular` imports the URLs.

Suggested schema-check env:
```
    [testenv:schema-check]
    setenv =
        {[testenv]setenv}
        CRA_NOTIFICATIONS_ENABLED = true
    allowlist_externals = /usr/bin/bash
    commands =
        /usr/bin/bash -c './scripts/schema-check.sh'
```

That way CI will fail if this PR adds `/notifications/upstream` routes but does not update `openapi.yml`.


`CRA_REPORTING_ENABLED` should not be required for this endpoint. That flag is for the SRP/reporting flow. For this PR, one flag should control upstream notification routes/tests.

For the `CI=true` question: I'd avoid adding `CI=true` to tox just to bypass `make check-venv-active`. That works, but it uses CI as a proxy for "running under tox". A cleaner fix is to check `TOX_ENV_NAME`.

Example `scripts/schema-check.sh`:
```
    #!/usr/bin/env bash
    if [[ -z "${TOX_ENV_NAME}" ]]; then
        make check-venv-active
    fi
    ./manage.py spectacular --file openapi.yml --settings=config.settings_local &> /dev/null && git diff --quiet openapi.yml &> /dev/null
```

Or update `mk/checks.mk`:
```
    @([[ ! -z "${CI}" ]] && exit 0) || ([[ ! -z "${TOX_ENV_NAME}" ]] && exit 0) || ([[ ! -z "${VIRTUAL_ENV}" ]] || { echo "venv is not active"; exit 1 ; })
```

either way is fine by me.

regulatory_reporting/tasks.py
svelamak: Swapped both matching pattern in `osidb/tasks.py`

regulatory_reporting/views.py
svelamak: Added, using existing pattern from osidb/api_view.py.

regulatory_reporting/tasks.py
svelamak: Added `acl_read/acl_write` to get_or_create() and also added `set_user_acls(settings.ALL_GROUPS)` in both celery tasks.

svelamak:
Removed `--ignore=regulatory_reporting` from `testenv:ci-osidb' and moved `CRA_NOTIFICATIONS_ENABLED` into base `testenv` per your suggestions. 
Switched to checking `TOX_ENV_NAME` in `scripts/schema-check.sh` instaed, no more `CI=true` in `tox.ini`.

regulatory_reporting/views.py
```diff
@@ -1,3 +1,121 @@
...
+            "flaw_id": flaw_id,
+            "vulnerability_summary": flaw.title,
+            "upstream_component": upstream_project.component_name,
+            "severity": flaw.impact,
+            "exploitation_state": "",
+            "embargo_notice": "This flaw is currently embargoed."
+            if flaw.is_embargoed
+            else "",
+            "corrective_measure": "",
+        }
```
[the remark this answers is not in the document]
svelamak: This is intentional. Not an oversight. We block embargoed flaws from automatically creating notifications but manually `send_email` is left to vmteam to do their own discretion.

osidb/templates/email/upstream_notification.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
```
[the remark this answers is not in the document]
svelamak: Fixed

regulatory_reporting/views.py
```diff
@@ -1,3 +1,121 @@
...
+from .models.upstream import UpstreamNotification
+from .serializers.upstream import UpstreamNotificationSerializer
+from .tasks import mark_upstream_notification_failed, mark_upstream_notification_sent
+
+
+@include_exclude_fields_extend_schema_view
+class UpstreamNotificationView(
+    RudimentaryUserPathLoggingMixin,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
```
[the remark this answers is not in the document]
svelamak: Fixed

costaconrado (approved):
LGTM, plase squash the changes that are being made and reverted in the same PR before merging.

PR #1370, opened 2026-07-17
Replace ACL queryset annotations with GeneratedFields

AlejandroMinaya:
## Summary
- Replace runtime `embargoed`/`visibility` queryset annotations with persisted `GeneratedField` columns
- Introduce `ACL` enum abstraction, remove `ACLMixinManager`
- Simplify serializer ACL logic with `_apply_acls` and `_unembargo_if_requested` helpers

Query performance: annotation vs GeneratedField (100 iterations, local DB: 79,807 flaws / 20 affects
  / 34,376 trackers)
```
  ┌────────────────────────────────┬──────────┬────────────────┬─────────┐
  │           Operation            │ Baseline │ GeneratedField │ Speedup │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Flaw.objects.get(pk=...)       │ 0.410s   │ 0.145s         │ 2.8x    │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Flaw.objects.all()[:20]        │ 11.506s  │ 11.310s        │ 1.0x    │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Affect.objects.all()[:20]      │ 0.624s   │ 0.283s         │ 2.2x    │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Tracker.objects.filter(pk=...) │ 0.098s   │ 0.044s         │ 2.2x    │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Affect.objects.get(pk=...)     │ 0.139s   │ 0.099s         │ 1.4x    │
  ├────────────────────────────────┼──────────┼────────────────┼─────────┤
  │ Tracker.objects.get(pk=...)    │ 0.203s   │ 0.037s         │ 5.5x    │
  └────────────────────────────────┴──────────┴────────────────┴─────────┘
 ```

  Flaw.objects.all()[:20] shows no improvement because the query is already expensive (many columns,
  large table) — the annotation overhead is proportionally negligible. Single-row lookups show the real
  win since the CASE/WHEN cost is a larger fraction of the total query time.

## Review follow-ups
- Removed `ACLMixinListSerializer`: it was dead code, bulk endpoints validate items one-by-one so the
  `many=True` list-level `validate()` never ran; the equivalent embargoed-consistency check in
  `AffectView.bulk_post` is the one that's actually exercised.
- Deduplicated the `current_acl`/`is_embargoed`/`is_internal`/`is_public` block that was copy-pasted
  into `AffectV1` (can't inherit `ACLMixin`, it's a view model) by adding `ACL.from_acl_read()` and
  having both `ACLMixin` and `AffectV1` call it.
- `Snippet.save()` and a test workaround in `test_query_regresion.py` now use `set_acls(ACL.INTERNAL)`
  instead of hand-setting `acl_read`/`acl_write`.

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### CODE REVIEW

File: osidb/api_views.py [hunk 1]
```diff
@@ -1720,6 +1721,23 @@ def bulk_post(self, request, *args, **kwargs):
             ).select_related("ps_module")
         }
 
+        # All items in a bulk request must share the same embargoed value
+        embargoed_values = {
+            d.get("embargoed") for d in request.data if isinstance(d, dict)
+        }
```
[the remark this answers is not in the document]
AlejandroMinaya: Wouldn't this kind of type error be caught by the individual level serializer before reaching validated_data for the create/update?

osidb/migrations/0247_affect_v1_add_embargoed.py
```diff
@@ -0,0 +1,112 @@
...
+),
+-- Group all cvss scores from v2 affects into the v1 affect
+affect_cvss_ids AS (
+    SELECT
+        affect_id,
+        array_agg(uuid) AS all_cvss_score_ids
+    FROM
+        osidb_affectcvss
+    GROUP BY
+        affect_id
```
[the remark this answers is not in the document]
AlejandroMinaya: Addressed

AlejandroMinaya:
@coderabbitai What comments am I missing for the approve?

AlejandroMinaya:
@coderabbitai The `uuid5` hash is not for security but stable hashing identification, so it shouldn't be an issue

AlejandroMinaya:
@coderabbitai In this case the weak-crypto is okay, so no need to block the merge

AlejandroMinaya:
@coderabbitai Could you give me the approve then?

osidb/migrations/0262_replace_acl_annotations_with_generated_fields.py
```diff
@@ -0,0 +1,240 @@
...
+        migrations.AddField(
+            model_name='upstreamdata',
+            name='embargoed',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=[uuid.UUID('b5a7e618-aac7-559f-9465-58cc9e9ed439')], then=True), default=False, output_field=models.BooleanField()), output_field=models.BooleanField()),
+        ),
+        migrations.AddField(
+            model_name='upstreamdata',
+            name='visibility',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=[uuid.UUID('b5a7e618-aac7-559f-9465-58cc9e9ed439')], then=models.Value('EMBARGOED')), models.When(acl_read=[uuid.UUID('08e4d24f-8d50-535b-a7b5-d9bdf0cbb337')], then=models.Value('INTERNAL')), models.When(acl_read=[uuid.UUID('e62dac33-2b05-5be1-9417-e3d301ff57da')], then=models.Value('PUBLIC')), default=models.Value('PUBLIC'), output_field=models.CharField()), output_field=models.CharField(max_length=10)),
+        ),
```
costaconrado, line 277: using hard-coded ACL UUIDs is very dangerous, each environment has it's unique ACL UUID groups and had-coding this may miss classify data in some environments

osidb/migrations/0246_replace_acl_annotations_with_generated_fields.py
```diff
@@ -0,0 +1,240 @@
...
+        ),
+        migrations.AddField(
+            model_name='affect',
+            name='embargoed',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=[uuid.UUID('b5a7e618-aac7-559f-9465-58cc9e9ed439')], then=True), default=False, output_field=models.BooleanField()), output_field=models.BooleanField()),
+        ),
+        migrations.AddField(
+            model_name='affect',
+            name='visibility',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=[uuid.UUID('b5a7e618-aac7-559f-9465-58cc9e9ed439')], then=models.Value('EMBARGOED')), models.When(acl_read=[uuid.UUID('08e4d24f-8d50-535b-a7b5-d9bdf0cbb337')], then=models.Value('INTERNAL')), models.When(acl_read=[uuid.UUID('e62dac33-2b05-5be1-9417-e3d301ff57da')], then=models.Value('PUBLIC')), default=models.Value('PUBLIC'), output_field=models.CharField()), output_field=models.CharField(max_length=10)),
```
costaconrado: this is defaulting to public with hard-coded uuids, in different environments we have different uuids, this pottentially unembargoes data wrongly

osidb/mixins.py
```diff
@@ -281,65 +282,46 @@ class ACLMixinVisibility(ComparableTextChoices):
...
+            models.When(
+                acl_read=ACL.PUBLIC.uuid_read,
+                then=models.Value(ACLMixinVisibility.PUBLIC),
+            ),
+            default=models.Value(ACLMixinVisibility.PUBLIC),
+            output_field=models.CharField(),
+        ),
+        output_field=models.CharField(max_length=10),
+        db_persist=True,
+    )
```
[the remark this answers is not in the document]
costaconrado: this is a valid concern that needs to be addressed.

osidb/migrations/0250_affect_v1_add_embargoed.py
```diff
@@ -0,0 +1,168 @@
+from django.db import migrations
+
+
+CREATE_AFFECT_V1_MATERIALIZED_VIEW = """
+DROP MATERIALIZED VIEW IF EXISTS affect_v1;
+CREATE MATERIALIZED VIEW affect_v1 AS
```
costaconrado: the materialized view doesn't seem to have RLS being created/enabled, I believe this need to be enabled in order to respect embargoed/internal values on v1

docs/CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - ACE: Add Go stdlib special CVE workflow (OSIDB-5033)
 
 ### Changed
+- Replace `ACLMixinManager` with `ACL` enum abstraction for ACL handling
+- Replace `embargoed` and `visibility` queryset annotations with persisted `GeneratedField` columns
+- Simplify serializer ACL logic: introduce `_apply_acls` and `_unembargo_if_requested` helpers
```
costaconrado: please add the issue key (OSIDB-5236) to the CHANGELOG entries to make the release notes easier.

osidb/mixins.py
```diff
@@ -281,65 +282,46 @@ class ACLMixinVisibility(ComparableTextChoices):
...
 
+    embargoed = models.GeneratedField(
+        expression=models.Case(
+            models.When(acl_read=ACL.EMBARGO.uuid_read, then=True),
+            default=False,
+            output_field=models.BooleanField(),
+        ),
+        output_field=models.BooleanField(),
+        db_persist=True,
+    )
```
costaconrado, line 391: since this is now a field, I believe we can update the osidb/djangoql.py::FlawEmbargoedField so we don't annotate the embargoed status

osidb/serializer.py
```diff
@@ -418,29 +417,47 @@ class Meta:
...
+    validated_data["acl_read"] = acl.uuid_read
+    validated_data["acl_write"] = acl.uuid_write
+
+
+def _unembargo_if_requested(validated_data: dict, instance) -> None:
+    validated_data["acl_read"] = instance.acl_read
+    validated_data["acl_write"] = instance.acl_write
+    if validated_data.get("embargoed") is False and instance.is_embargoed:
+        _apply_acls(validated_data, False)
+
```
costaconrado: I'm not sure I understand the meaning of this method, doesn't this if means it only unembargoed things if embargoed is already false? Is it for internal only?

costaconrado (dismissed):

osidb/serializer.py
AlejandroMinaya: This only works for updates when the original object was embargoed but the incoming request has embargoed set to false. 

In this case `instance` refers to the object before applying the update and `validated_data` a dictionary with the new incoming changes.

osidb/migrations/0250_affect_v1_add_embargoed.py
AlejandroMinaya: In the previous migrations there is no explicit rule for RLS. Since it is a view I believe that the RLS rules are applied by the underlying `osidb_affects` table

osidb/migrations/0262_replace_acl_annotations_with_generated_fields.py
AlejandroMinaya: These values are no longer harcoded but if the ACL groups change, this migration needs to be applied again. Since this is translated into the PostgreSQL generated column, in the end the values are technically hardcoded in that definition.

The alternative is having a database table instead of a settings configuration for the ACL groups but I don't know if they change frequently. This is something I spoke about with @Elkasitu. I don't mind taking that approach but my understanding is that this should suffice.

osidb/mixins.py
AlejandroMinaya: Addressed in a new commit

osidb/acls.py
```diff
@@ -0,0 +1,42 @@
...
+        self.write = write_list
+        self.uuid_read = self.generate_acl_uuids(read_list)
+        self.uuid_write = self.generate_acl_uuids(write_list)
+
+    def __lt__(self, other: "ACL") -> bool:
+        if self.__class__ is not other.__class__:
+            return NotImplemented
+
+        members = list(self.__class__)
+        return members.index(self) < members.index(other)
```
Elkasitu, line 83: lacks some unit tests

osidb/mixins.py
```diff
@@ -371,32 +355,30 @@ def is_internal(self):
...
+        if target == ACLMixinVisibility.EMBARGOED:
+            self.set_embargoed()
         elif target == ACLMixinVisibility.INTERNAL:
             self.set_internal()
+        elif target == ACLMixinVisibility.PUBLIC:
+            self.set_public()
 
+        self._visibility_dirty = True
         self.set_acls_nested()
         self.set_acls_history()
```
Elkasitu: I'm not very keen on managing state / relying on in-memory values at all now that it's a generated field, can't we just compare against the database value instead of the in-memory value? Are there any actual cases where we might change the underlying value in-memory multiple times before doing this check?

Elkasitu: if we can achieve this then we can also get rid of all the `is_x` properties and the `_current_visibility` method

osidb/serializer.py
```diff
@@ -336,43 +333,45 @@ class EmbargoedField(serializers.BooleanField):
...
+    def to_internal_value(self, data):
+        embargoed = super().to_internal_value(data)
+        instance = self.parent.instance if self.parent else None
+        if embargoed:
+            acl = ACL.EMBARGO
+        elif instance is None:
+            acl = self.non_embargo_acl
+        elif instance.is_embargoed:
+            acl = ACL.PUBLIC
```
Elkasitu: huh?

osidb/serializer.py
```diff
@@ -336,43 +333,45 @@ class EmbargoedField(serializers.BooleanField):
...
+        elif instance.is_internal:
+            acl = ACL.INTERNAL
```
Elkasitu: the original implementation only allowed for users to set resources as either embargoed or public, not internal, this changes the behavior of this field

osidb/serializer.py
```diff
@@ -336,43 +333,45 @@ class EmbargoedField(serializers.BooleanField):
...
-        self.validate_acl(data)
-        raise serializers.SkipField()
-
-    def validate_acl(self, embargoed):
-        acl_read = (
-            settings.EMBARGO_READ_GROUPS if embargoed else settings.PUBLIC_READ_GROUPS
-        )
-        acl_write = (
-            settings.EMBARGO_WRITE_GROUPS if embargoed else settings.PUBLIC_WRITE_GROUPS
-        )
```
Elkasitu, line 337: why not keep the original implementation? also the original implementation is very explicit about not passing the validated data down to the model, is it still the case with your implementation?

Elkasitu (commented):
Good progress, but I think it can still be improved on

osidb/mixins.py
AlejandroMinaya: I agree with you and it wasn't my first choice. The thing is that we check for visibility in transient states, i.e. when we haven't saved it in the database yet.

This raises a DoesNotExist error since it automatically tries to fetch the information for a row that is not there.

I wanted to replace the `is_*` methods as you mentioned but we have the same issue.

What would be a good alternative?

osidb/serializer.py
AlejandroMinaya: I realize this is confusing and I would like to improve on that based on your feedback.


`embargoed` is the incoming value after being parsed into a boolean.

`instance` is the already present database record. 

If `embargoed` is `True`, then it sets the ACL to embargoed, no more questions asked.

If `embargoed` is `False` and `instance.embargo` is `True`, that means that the incoming request wants to lift the embargo.

The fact that this didn't translate immediately is enough to refactor it, but I would like to know what was your initial interpretation and if it was a variable naming issue or something else.

osidb/serializer.py
AlejandroMinaya: You're right to point this out. I may be confused as to the intended purpose of that behavior.

As an internal user, with no embargo permissions, the objects I create can only be public?

Maybe we could discuss having a specific test to cover this intended behavior

osidb/mixins.py
AlejandroMinaya: You're right that the consecutive calls are not a plausible case, but if introduced by mistake it wouldn't fail and I believe that it would be a hard bug to catch.

I'm happy to discuss what would be the better DX, but it hinges on whatever we decide on the previous matter.

osidb/serializer.py
AlejandroMinaya: It gets passed down to the model but this is ignored. Before we had a test making sure that we get an error if the model received an embargoed parameter but the framework drops generated fields anyway.

The test now checks that even when setting this value directly for a model instance, it is ignored. 

This effectively turns this field into read-only (at the model level) without having to keep in mind not to pass this parameter into initializers.

osidb/acls.py
AlejandroMinaya: Added unit tests on the first commit. Marking this as resolved

osidb/mixins.py
Elkasitu: Let's look at call-sites for this property/method:

It only gets called from the workflow framework, there's two possible scenarios here:

- A Flaw is created as embargoed by an analyst, and through the workflow process becomes un-embargoed (public).
- A Flaw is collected from a public source and starts as internal, and through the workflow process becomes public once it has been processed and is ready for end-user consumption.

There's no realistic scenario in which a Flaw gets created with a given set of ACLs which are then changed before it's saved.

Let's assume the scenario can happen: if as part of a process you create an object with `acls = foo` and before saving you call `set_visibility(bar)`, then the `set_visibility` call should disregard the in-memory state -- the object doesn't exist yet so setting any visibility should be valid IMO.

For updates, we should always base it on the database-generated value.

Lastly, we probably should just remove the `set_visibility` altogether (and the accompanying methods / properties) and just have something specific in workflows -- not in `ACLMixin`.

Does that make sense?

AlejandroMinaya: I much rather add this directly in workflows. I am okay with removing it from the ACLMixin

osidb/serializer.py
Elkasitu: I think it's confusing because it's doing more than it should: `to_internal_value` really should just care about converting a serialized value into a python value and not the business logic behind it, this may be why I prefer the original implementation.

I think here you should just convert `True` or `False` in input to the correct `ACL` value, then in the mixin's  create/update (where you have easy access to the instance) you can compare and set if needed.

osidb/serializer.py
Elkasitu: > As an internal user, with no embargo permissions, the objects I create can only be public?

Yes, "internal" is mostly system-managed, the only cases when a user can create internal objects are:

1. Comments, in which case internal really just means comments in the JIRA task
2. Flaw creation, if no value is provided for `embargoed`

My original assessment might be wrong, but it's hard to be certain which is why I think this method should simply handle the conversion of `True/False` to `ACL.EMBARGO/ACL.PUBLIC` and nothing more.

The actual business logic should live in the relevant `create/update` methods.

osidb/serializer.py
```diff
@@ -336,43 +333,45 @@ class EmbargoedField(serializers.BooleanField):
     """The embargoed boolean attribute is technically read-only as it just indirectly
     modifies the ACLs but is mandatory as it controls the access to the resource."""
 
+    def __init__(self, *args, non_embargo_acl=ACL.PUBLIC, **kwargs):
+        self.non_embargo_acl = non_embargo_acl
+        super().__init__(*args, **kwargs)
```
Elkasitu: I think we should also get rid of the override, this field should simply translate `True/False` to the relevant acl_read / acl_write.

osidb/serializer.py
AlejandroMinaya: I see what you mean. Here the `acl` variable is only set so that it can be used below on the permissions validation. The serializer performed the permissions check in the original implementation too, but I am not against moving this validation to a deeper layer.

osidb/serializer.py
AlejandroMinaya: Sounds good to me. Will do

AlejandroMinaya:
@coderabbitai review

osidb/migrations/0262_replace_acl_annotations_with_generated_fields.py
```diff
@@ -0,0 +1,265 @@
...
+        migrations.AddField(
+            model_name='affect',
+            name='embargoed',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=ACL.EMBARGO.uuid_read, then=True), default=False, output_field=models.BooleanField()), output_field=models.BooleanField()),
+        ),
+        migrations.AddField(
+            model_name='affect',
+            name='visibility',
+            field=models.GeneratedField(db_persist=True, expression=models.Case(models.When(acl_read=ACL.EMBARGO.uuid_read, then=models.Value('EMBARGOED')), models.When(acl_read=ACL.INTERNAL.uuid_read, then=models.Value('INTERNAL')), models.When(acl_read=ACL.PUBLIC.uuid_read, then=models.Value('PUBLIC')), default=models.Value('PUBLIC'), output_field=models.CharField()), output_field=models.CharField(max_length=10)),
```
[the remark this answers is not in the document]
AlejandroMinaya: Validators already prevent unknown ACLs; visibility is informational, RLS handles real access control. Adding UNKNOWN cascades across enum, model, migration, and filters for a scenario that can't happen through normal code paths.

osidb/mixins.py
```diff
@@ -371,32 +355,30 @@ def is_internal(self):
...
 
-    @visibility.setter
-    def visibility(self, target):
-        """
-        Monotonically widen visibility to the given target level.
-
-        ACLMixinVisibility can only be widened (EMBARGOED -> INTERNAL -> PUBLIC),
-        never narrowed. If the current level is already at or wider than
-        the target, this is a no-op.
-        """
```
osoukup, line 487: Why is this comment being entirely removed? I think that the information about the only viable visibility change direction and no-op effect may not be super intuitive and the documentation may help.

osidb/tests/test_query_regresion.py
```diff
@@ -420,7 +421,8 @@ def test_flaw_promote(
             major_incident_state=Flaw.FlawMajorIncident.NOVALUE,
         )
         if not embargoed:
-            flaw.set_internal()
+            flaw.acl_read = ACL.INTERNAL.uuid_read
+            flaw.acl_write = ACL.INTERNAL.uuid_write
```
osoukup: What is the advantage of replacing the shortcut with longer and less readable code?

docs/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace `ACLMixinManager` with `ACL` enum abstraction for ACL handling (OSIDB-5236)
 - Replace `embargoed` and `visibility` queryset annotations with persisted `GeneratedField` columns (OSIDB-5236)
 - Simplify serializer ACL logic: introduce `_apply_acls` and `_unembargo_if_requested` helpers (OSIDB-5236)
+- Remove `FlawEmbargoedField` DjangoQL custom field in favor of `GeneratedField` (OSIDB-5236)
```
osoukup: This is already a fourth changelog record for the same change. I do not think we need to document every step, especially when these records are pretty hard to understand for the OSIDB users who are the audience of the changelog - so they know what we released.

apps/workflows/workflow.py
```diff
@@ -244,8 +243,12 @@ def jira_status(self):
...
             self.workflow_name, self.workflow_state
         )
-        if visibility:
-            self.set_visibility(visibility)
+        if not target or self.current_acl >= target:
+            return
+        self.acl_read = target.uuid_read
+        self.acl_write = target.uuid_write
+        self.set_acls_nested()
+        self.set_acls_history()
```
osoukup, line 279: I do not think that workflow is the right place to define the visibility change procedure (nesting, history) and restrictions (widening-only). That should be the responsibility of the ACL class/module/something.

osidb/models/affect.py
```diff
@@ -1189,18 +1190,27 @@ def ps_product(self):
             return None
         return ps_module.ps_product.name
 
-    # Mirrored properties from mixins
+    @property
+    def current_acl(self):
```
osoukup, line 1171: What is the purpose of this property? To abstract that we use `read` and ignore `write`? Otherwise I do not understand how is `current_acl` different from `self.acl_read` and why to evaluate multiple `if` conditions when I want to compare two values (down there in each concrete property).

osidb/mixins.py
```diff
@@ -417,7 +395,6 @@ def set_acl_read(self, *groups):
         """
         acls = sorted(self.group2acl(group) for group in groups)
         self.acl_read = acls
-        self._visibility_dirty = True
```
osoukup: This was added in the very same PR as it is being removed. That should be squashed.

osidb/mixins.py
```diff
@@ -459,44 +435,18 @@ def set_public(self):
             ... [UUID(...), UUID(...)]
         """
 
+        if self.current_acl >= ACL.PUBLIC:
+            return
         self.set_acl_read(*settings.PUBLIC_READ_GROUPS)
         self.set_acl_write(*settings.PUBLIC_WRITE_GROUPS)
 
-    def set_embargoed(self):
-        """
```
osoukup, line 581: What is the reason this is being removed? I guess that it is unused? I am thinking that setting something embargoed is not viable anyway so probably OK to remove it.

apps/workflows/workflow.py
```diff
@@ -248,4 +248,4 @@ def adjust_acls(self):
             self.workflow_name, self.workflow_state
         )
         if visibility:
-            self.visibility = visibility
```
osoukup, line 277: The reason the setter is removed is because it got complicated with the generated field?

osoukup (changes requested):
I have a few comments which are not all necessary to get addressed in the code but at least clarify here.

osidb/mixins.py
AlejandroMinaya: I agree. WDYT of adding it to the ACL class itself since it imposes the ordering?

osidb/tests/test_query_regresion.py
AlejandroMinaya: There is no benefit in terms of readability, but the previous `set_internal` didn't enforce the "only-widen" rule. Since I updated it to do so, tests broke but not because they were testing the business logic.

It was being used for readability/shortcut to set ACLs, which is why I changed it to its more verbose counterpart.

docs/CHANGELOG.md
AlejandroMinaya: Understood! I hadn't made the link that the target was end users. How do we document this change for other developers that may use OSIDB as a dependency?

apps/workflows/workflow.py
AlejandroMinaya: I understand what you mean. This is part of a larger conversation I had with @Elkasitu. Since visbility was an abstraction of the ACLs only used in workflows, we figured it made sense to move it.

Although I can see why this doesn't seem like the workflows responsibility, it does add complexity on the ACL side just to satisfy this need.

I am okay making this compromise but I am open to suggestions too

osidb/mixins.py
AlejandroMinaya: Agreed. I added it as a separate commit due to the concerns you bring up about having it on the workflow side. It allows a clean revert in case we decide to go in a different direction

osidb/mixins.py
AlejandroMinaya: I think it was added for completeness but we don't have such scenarios after creation just like you mention. I am okay removing it too

apps/workflows/workflow.py
AlejandroMinaya: Yes. We had code that used visibility indistinctly between transient and persisted objects but with generated fields it became an issue

AlejandroMinaya:
@osoukup Thank you for the review! I believe a lot of the feedback revolves around the changes to visibility. I also think that it is worth making sure that we are all on the same page before making any compromises. Please let me know if I've offered enough context in my replies

osidb/models/affect.py
AlejandroMinaya: The intent is to abstract away the difference between our string-based groups and UUID groups. It does abstract the write away, but like you mention, it wouldn't add much value just for this purpose.

`current_acl` returns an `ACL` member instead of a list of specific values. This abstraction helps negate the need to remember if at a certain point we are dealing with regular strings or UUIDs

apps/workflows/workflow.py
Elkasitu: See https://github.com/RedHatProductSecurity/osidb/pull/1370#discussion_r3621397211 the tl;dr is that we do a bunch of generic complex logic in ACLMixin _because_ it's a mixin and doesn't know in which context these methods and properties are being used, however these are _only_ used in workflows where the code can be a lot more specific / simple.

osidb/mixins.py
```diff
@@ -486,52 +446,20 @@ def set_public(self):
...
-        e.g.:
-            >>> my_flaw.acl_read
-            ... [UUID(...), UUID(...), UUID(...), UUID(...)]
-            >>> my_flaw.set_internal()
-            >>> # note that the acl_read have been completely replaced by the
-            >>> # internal ACLs only, other ones are not kept.
-            >>> my_flaw.acl_read
-            ... [UUID(...), UUID(...)]
+        No-op if the entity is already at internal or public visibility.
         """
```
Elkasitu, line 616: Maybe the docstring wasn't explicit enough, but these methods are intended as primitives i.e. they simply change the acls, they do not care about business logic.

If we want, we can have a separate method for transitioning only if certain conditions are met or no-op.

Elkasitu (approved):
LGTM, just a minor comment.

PS: Have you checked changes in performance with this change and compared it to your previous submission? It would be interesting to document

osidb/mixins.py
AlejandroMinaya: I apologize, I didn't gather that from the docstring. If the intention is to have a shorcut that overrides business logic, maybe we could make it more generic, receiving an ACL enum member instead of having one for each list. WDYT?

AlejandroMinaya:
Thank you @Elkasitu for the review!
> PS: Have you checked changes in performance with this change and compared it to your previous submission? It would be interesting to document

I haven't done the same comparison but I agree that it would be interesting to see. Before merging I'll add it as part of the PR description like in the previous proposal

osidb/mixins.py
Elkasitu: Sure.

As a side note, I don't think it's correct to think of these methods to be a "shortcut" or an override of anything, they are simply primitives, the most basic functions of the `ACLMixin`, devoid of semantics.

osidb/migrations/0250_affect_v1_add_embargoed.py
osoukup: That would be good to confirm to prevent any possible leak.

osidb/mixins.py
osoukup: Sure, wherever it makes sense.

osidb/tests/test_query_regresion.py
osoukup: The original method could be extended to widen-only (eg. be a no-op) if applied to a public data. I would prefer readability over verbosity, AI does not care but for humans it is still beneficial.

docs/CHANGELOG.md
osoukup: Git history is the changelog for the developers.

apps/workflows/workflow.py
osoukup: So ideally there would be visibility-to-ACLs translation layer abstracting the two views. Let us leave the visibility part on the workflow side only. Still I think the ACL part should provide something like set ACLs of a target object to the ACLs of the source object. The workflow module should not care that it means updating nested and history ACLs.

osidb/models/affect.py
osoukup: I have no strong preference here so you can leave it as it is.

osoukup (commented):
I have run another review round, some of my complains still apply but they are not so serious that they should block this. Address or ignore them. I would only like a confirmation that the concern about materialized view and RLS @costaconrado raised is no issue.

osidb/tests/test_query_regresion.py
AlejandroMinaya: After discussing this with @Elkasitu we now have two separate methods:
1) A generic set_acl() that receives an ACL enum member, no business rule enforcement
2) The regular `set_internal()` and `set_public()` methods that respect the widen-only rule

osoukup: Sounds well!

PR #1371, opened 2026-07-20
OSIDB-2996: Fix DELETE api/v2/affects/bulk request body example

AlejandroMinaya:


dmonzonis (commented):
In the title, it should be `api/v2/affects/bulk`, as v1 is deprecated and doesn't support any write operations.

Also, I tried checking locally with this change and still cannot see a request body section in the schema. I am not sure if I am missing something, although I rebuilt just in case, or if the limitations explained in the comment that was removed still apply and this section just won't render for delete endpoints.

AlejandroMinaya:
You're totally right. I apologize for missing this. I checked the `drf-spectacular` [source code](https://github.com/tfranzel/drf-spectacular/blob/1a5c44a15b2d06744f49285f1f58992f938a05fa/drf_spectacular/openapi.py#L1343) and they still haven't relaxed their checks to allow request body specifications for DELETE methods. I thought this could be a quick-win but I got ahead of myself. I'm closing this PR for the time being.

PR #1377, opened 2026-07-24
feat: add django-silk profiling and sample data tooling (OSIDB-5268)

AlejandroMinaya:
## Summary
- Add opt-in django-silk profiling gated behind `OSIDB_ENABLE_SILK` env var
- Add `generate_sample_data` management command to populate dev DB with realistic flaws, affects, trackers, and related objects
- Add locust load test script (`perf/silk_profiling.py`) for generating traffic against profiled endpoints
- Add `make apply-uv-dev-sync` target to install dev deps in the osidb-service container
- Add `django-silk` and `locust` as dev dependencies

Split into separate commits per concern for easier review/revert.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/config/settings_local.py
+++ b/config/settings_local.py
@@ -1,3 +1,7 @@
+import os
+import sys
+from importlib.util import find_spec
+
 import ldap
 from django_auth_ldap.config import GroupOfNamesType, LDAPSearch
 
@@ -102,3 +106,29 @@ INSTALLED_APPS += ["osidb.tests"]
 # Email configuration
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+
+# django-silk profiling (opt-in)
+#
+# Setup:
+#   1. Add OSIDB_ENABLE_SILK=1 to your .env
+#   2. make sync-deps              (update local venv)
+#   3. make apply-uv-dev-sync      (install dev deps in osidb-service container)
+#   4. podman restart osidb-service (runs migrate + starts runserver)
+#   5. Browse http://localhost:8000/silk/
+#
+# For load testing + profiling, see perf/silk_profiling.py
+#
+# To disable: remove OSIDB_ENABLE_SILK from .env and restart.
+# Silk tables are harmless when left behind; to drop them:
+#   podman exec osidb-service python manage.py migrate silk zero
+if (
+    os.environ.get("OSIDB_ENABLE_SILK", "").strip() not in ("", "0")
+    and "pytest" not in sys.modules
+    and find_spec("silk")
+):
+    INSTALLED_APPS += ["silk"]
+    # Must be after GZipMiddleware per silk docs
+    _gzip = MIDDLEWARE.index("django.middleware.gzip.GZipMiddleware")
+    MIDDLEWARE.insert(_gzip + 1, "silk.middleware.SilkyMiddleware")
+    SILKY_PYTHON_PROFILER = True
+    SILKY_INTERCEPT_PERCENT = 100
--- a/osidb/management/commands/generate_sample_data.py
+++ b/osidb/management/commands/generate_sample_data.py
@@ -1,0 +1,834 @@
+import random
+import uuid
+from datetime import datetime, timezone
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from osidb.acls import ACL
+from osidb.core import generate_acls, set_user_acls
+from osidb.helpers import bypass_rls
+from osidb.mixins import Alert
+from osidb.models import (
+    CVSS,
+    Affect,
+    AffectCVSS,
+    Flaw,
+    FlawAcknowledgment,
+    FlawComment,
+    FlawCVSS,
+    FlawLabelV2,
+    FlawReference,
+    FlawSource,
+    Impact,
+    Package,
+    PackageVer,
+    PsModule,
+    PsProduct,
+    PsUpdateStream,
+    Tracker,
+)
+from osidb.models.affect import NotAffectedJustification
+
+# ---------------------------------------------------------------------------
+# Constants used for sample-data identification (cleanup queries)
+# ---------------------------------------------------------------------------
+SAMPLE_MODULE_NAMES = [
+    "rhel-9",
+    "rhel-8",
+    "openshift-4",
+    "satellite-6",
+    "ansible-automation-platform-2",
+]
+SAMPLE_CVE_PREFIX = "CVE-2024-5"
+SAMPLE_NO_CVE_TITLE = "[sample] flaw without CVE"
+SAMPLE_PRODUCT_NAMES = [
+    "sample_rhel_9",
+    "sample_rhel_8",
+    "sample_openshift",
+    "sample_satellite",
+    "sample_aap",
+]
+SAMPLE_TRACKER_PREFIX = "SAMPLE-"
+SAMPLE_BZ_TRACKER_START = 2200000
+
+# ---------------------------------------------------------------------------
+# Pools of values cycled / randomly picked when scaling up
+# ---------------------------------------------------------------------------
+IMPACT_POOL = [
+    Impact.CRITICAL,
+    Impact.IMPORTANT,
+    Impact.MODERATE,
+    Impact.LOW,
+    Impact.NOVALUE,
+]
+SOURCE_POOL = [
+    FlawSource.INTERNET,
+    FlawSource.CUSTOMER,
+    FlawSource.REDHAT,
+    FlawSource.GIT,
+    FlawSource.RESEARCHER,
+    FlawSource.UPSTREAM,
+    FlawSource.GOOGLE,
+]
+MI_POOL = [
+    Flaw.FlawMajorIncident.NOVALUE,
+    Flaw.FlawMajorIncident.MAJOR_INCIDENT_APPROVED,
+    Flaw.FlawMajorIncident.MAJOR_INCIDENT_REQUESTED,
+    Flaw.FlawMajorIncident.MAJOR_INCIDENT_REJECTED,
+    Flaw.FlawMajorIncident.EXPLOITS_KEV_APPROVED,
+    Flaw.FlawMajorIncident.EXPLOITS_KEV_REQUESTED,
+    Flaw.FlawMajorIncident.EXPLOITS_KEV_REJECTED,
+    Flaw.FlawMajorIncident.MINOR_INCIDENT_APPROVED,
+    Flaw.FlawMajorIncident.MINOR_INCIDENT_REQUESTED,
+    Flaw.FlawMajorIncident.MINOR_INCIDENT_REJECTED,
+]
+WF_STATE_POOL = [
+    "NEW",
+    "TRIAGE",
+    "PRE_SECONDARY_ASSESSMENT",
+    "SECONDARY_ASSESSMENT",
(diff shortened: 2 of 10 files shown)


pyproject.toml
```diff
@@ -54,12 +54,14 @@ dependencies = [
...
 dev = [
     "detect-secrets>=1.4.0",
+    "django-silk>=5.0.0",
     "django-extensions>=3.1.5",
     "django-stubs>=1.14.0",
     "djangorestframework-stubs>=1.8.0",
     "factory-boy>=3.2.1",
     "freezegun>=1.1.0",
     "k5test>=0.10.1",
+    "locust>=2.0",
```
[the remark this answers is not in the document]
AlejandroMinaya: Understood but the risk here is minimal. We're talking about dev dependencies, uv lock is used on local dev environments, and we see that other dependencies also use floor constraints. Dismissing comment

config/settings_local.py
```diff
@@ -102,3 +106,29 @@
...
+    os.environ.get("OSIDB_ENABLE_SILK", "").strip() not in ("", "0")
+    and "pytest" not in sys.modules
+    and find_spec("silk")
+):
+    INSTALLED_APPS += ["silk"]
+    # Must be after GZipMiddleware per silk docs
+    _gzip = MIDDLEWARE.index("django.middleware.gzip.GZipMiddleware")
+    MIDDLEWARE.insert(_gzip + 1, "silk.middleware.SilkyMiddleware")
+    SILKY_PYTHON_PROFILER = True
+    SILKY_INTERCEPT_PERCENT = 100
```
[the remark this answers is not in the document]
AlejandroMinaya: This is the local settings. By definition it shouldn't be used for shared instances

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,567 @@
...
+            ),
+            ("flaws", Flaw.objects.filter(cve_id__in=SAMPLE_CVE_IDS)),
+            (
+                "flaws (no CVE)",
+                Flaw.objects.filter(
+                    cve_id__isnull=True,
+                    source=FlawSource.REDHAT,
+                    impact=Impact.MODERATE,
+                ),
+            ),
```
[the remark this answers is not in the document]
AlejandroMinaya: A specific title was added to address this issue

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,595 @@
...
+    TrackerFactory,
+)
+
+SAMPLE_MODULE_NAMES = [
+    "rhel-9",
+    "rhel-8",
+    "openshift-4",
+    "satellite-6",
+    "ansible-automation-platform-2",
+]
```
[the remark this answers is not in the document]
AlejandroMinaya: No risk since it is local development and developers are notified that they may be jeopardizing their local data. Dismissing

config/settings_local.py
```diff
@@ -102,3 +106,29 @@
...
+#   podman exec osidb-service python manage.py migrate silk zero
+if (
+    os.environ.get("OSIDB_ENABLE_SILK", "").strip() not in ("", "0")
+    and "pytest" not in sys.modules
+    and find_spec("silk")
+):
+    INSTALLED_APPS += ["silk"]
+    # Must be after GZipMiddleware per silk docs
+    _gzip = MIDDLEWARE.index("django.middleware.gzip.GZipMiddleware")
+    MIDDLEWARE.insert(_gzip + 1, "silk.middleware.SilkyMiddleware")
```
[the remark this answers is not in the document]
AlejandroMinaya: This is the desired behavior. It is unlikely that we remove Gzip, but if it happens we want these lines of code to change anyway. Dismissing

perf/silk_profiling.py
```diff
@@ -0,0 +1,187 @@
...
+            "/auth/token",
+            json={
+                "username": "testuser",
+                "password": "password",  # pragma: allowlist secret
+            },
+        )
+        data = resp.json()
+        self.token = data.get("access", "")
+        if not self.token:
+            raise RuntimeError(f"Auth failed ({resp.status_code}): {data}")
```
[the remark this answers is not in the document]
AlejandroMinaya: Addressed

Elkasitu (commented):
I haven't actually looked at the code, but two remarks from the description:

1. Why not just include Django Silk and have it enabled always? It only affects the local environment and to my knowledge it shouldn't impact regular work
2. Why not use Django fixtures for the test data instead of a generation script? Declarative > Imperative IMHO

AlejandroMinaya:
> 1. Why not just include Django Silk and have it enabled always? It only affects the local environment and to my knowledge it shouldn't impact regular work

I'm not certain that it would not add overhead to regular operations but will investigate further and get back to you.


> 2. Why not use Django fixtures for the test data instead of a generation script? Declarative > Imperative IMHO

Initially I was going for a fixture but it quickly ballooned in size and I consider that makes it harder to maintain if we need to react to changes. Between updating fields, keeping track of all the UUIDs, and communicating why certain values are needed to comply with some business rule, I figured that a generative approach was worth it. 

I acknowledge that it is more complex in nature, but this is offset by the maintainability ease in comparison to the declarative file.

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,595 @@
...
+    "RHEL-12350",
+    "SAT-9001",
+    "AAP-5001",
+    "OCPBUGS-30001",
+    "2200001",
+    "2200002",
+]
+
+
+class Command(BaseCommand):
```
osoukup, line 168: The generated set seems to me to be too small and hardcoded for the meaningful profiling. It seems that it generates a few tens of models. If we create a data generation script like this, I would expect more parametrization allowing you to create 50K flaws if you want some heavy testing. Instead of the hardcoded values there should be some patterns.

osoukup (commented):
I am OK with this though I think the sample data generation script could be more scalable, not just producing the hardcoded set of data. It also seems to me that the generated data set is quite small. Does it show any useful profiling results with this amount of data?

osidb/management/commands/generate_sample_data.py
AlejandroMinaya: I see what you mean. Besides the amount of records, what would you like to see parametrized? What would be a good default to streamline the process for people that just want representative data?

I believe we need a good mix of:
- Embargoed/non-embargoed
- Trackers with different statuses
- Multi-flaw trackers

But what else?

osoukup: I am not sure how well we can parametrize the model attribute values to make it still usable. Maybe rather make them more randomized but I understand that it could be complicated and above the scope of this PR. Also maybe the randomization might make some measurements too unstable so it would need a bit of thinking. I was mostly thinking about the number of flaws and their average affects and tracker at least.

AlejandroMinaya: Understood. I think it may be straightforward enough to add parameters for amounts. I'll explore it and add it to the PR if possible

AlejandroMinaya: Done, now the command accepts parameters to control the volume of data

AlejandroMinaya:
@Elkasitu 
> Why not just include Django Silk and have it enabled always? It only affects the local environment and to my knowledge it shouldn't impact regular work

~I had misinterpreted your question. If you meant to ask why have it as optional if it is in the local settings this has to do with it being opt-in instead of opt-out. If you don't have a hard opinion I would think this is fine since it aligns with the idea of OSIDB having opt-in dependencies.~

~If you believe that this should not apply to these dev dependencies and it is just extra complexity for marginal flexibility, I can remove it. I don't have a hard opinion about either choice.~

Tests use the settings.local. When I wrote this comment I was confusing the testing-specific docker files for testing-specific settings

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,822 @@
...
+        return ps_modules, streams
+
+    # ------------------------------------------------------------------
+    # Flaws
+    # ------------------------------------------------------------------
+    def _create_flaws(self, num_flaws):
+        flaws = []
+        now = _now()
+        for i in range(num_flaws):
+            impact = IMPACT_POOL[i % len(IMPACT_POOL)]
```
osoukup, line 368: nitpick: The even distribution does not really reflect the reality where criticals are very rare and the moderates are probably the majority of the flaws.

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,822 @@
...
+            source = SOURCE_POOL[i % len(SOURCE_POOL)]
+            mi_state = MI_POOL[i % len(MI_POOL)]
```
osoukup, line 370: nitpick: MI is really rare so this is not very close to reality (does not matter for the testing).

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,822 @@
...
+            wf_state = WF_STATE_POOL[i % len(WF_STATE_POOL)]
```
osoukup, line 371: Flaws in some states are supposed to be internal. This is also maybe a nitpick but missing the internal flaws completely might actually make some cases non-testable.

osidb/management/commands/generate_sample_data.py
```diff
@@ -0,0 +1,822 @@
...
+            count = min(count, len(combos))
+            for (pm_idx, stream_idx), component in combos[:count]:
+                affectedness, resolution = random.choice(AFFECTEDNESS_RESOLUTION_POOL)  # noqa: S311
+                stream = streams[stream_idx]
+                module = ps_modules[pm_idx]
+
+                impact = Impact.NOVALUE
+                if random.random() < 0.1:  # noqa: S311
+                    impact = random.choice(  # noqa: S311
+                        [Impact.CRITICAL, Impact.IMPORTANT, Impact.MODERATE]
```
osoukup, line 450: nitpick: Why no LOWs? 🤔

osoukup (approved):
Nice work! LGTM, I have just raised a few nitpick comments on the sample data generation command which you can ignore if you want.

Elkasitu:
> @Elkasitu
> 
> > Why not just include Django Silk and have it enabled always? It only affects the local environment and to my knowledge it shouldn't impact regular work
> 
> ~I had misinterpreted your question. If you meant to ask why have it as optional if it is in the local settings this has to do with it being opt-in instead of opt-out. If you don't have a hard opinion I would think this is fine since it aligns with the idea of OSIDB having opt-in dependencies.~
> 
> ~If you believe that this should not apply to these dev dependencies and it is just extra complexity for marginal flexibility, I can remove it. I don't have a hard opinion about either choice.~
> 
> Tests use the settings.local. When I wrote this comment I was confusing the testing-specific docker files for testing-specific settings

I mean why have it gated behind an env var / feature flag when it can just always be enabled for the local environment? Worst case scenario local OSIDB goes a little slower

AlejandroMinaya:
> I mean why have it gated behind an env var / feature flag when it can just always be enabled for the local environment? Worst case scenario local OSIDB goes a little slower

When you have it enabled for tests (since it uses the local settings as well) it makes tests like `test_query_regression.py` unreliable. Silk adds queries that get counted when peforming the assert. I agree with you that it would be cleaner just to add it and forget about it, but this way we avoid unexpected surprises.

PR #1383, opened 2026-07-28
feat: add vendor-neutral OpenTelemetry tracing with embargo redaction

AlejandroMinaya:
## Summary

Adds OpenTelemetry (OTel) tracing to OSIDB with a vendor-neutral architecture. This is an **exploratory spike** to evaluate distributed tracing for production observability.

- Instruments Django, psycopg2, requests, Redis, and Celery via OTel auto-instrumentation
- Vendor-neutral: any OTLP-compatible backend (Datadog Agent, Jaeger, Grafana Tempo, ...) works — swap by changing one env var. No collector is shipped with this PR (see "No collector included" below)
- Gated behind `OSIDB_OTEL_ENABLED` (default: `false`) — completely inert when disabled
- Ships a conservative default trace sampling ratio (10%, via `OTEL_TRACES_SAMPLER`/`OTEL_TRACES_SAMPLER_ARG`) so enabling tracing doesn't spike ingestion volume before it's tuned per-environment

## Why OpenTelemetry over ddtrace

We don't want vendor lock-in. OTel is the CNCF standard for observability — Datadog, Grafana, Jaeger, and every major vendor accept OTLP natively. Switching vendors means changing `OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT`, nothing else. No Datadog-specific code anywhere in the codebase.

## No collector included

This PR ships only the producer side (SDK setup, instrumentation). It does **not** include a Jaeger/collector service — standing up and operating the OTLP backend (Datadog Agent or otherwise) is devops' call, so it's left out of `docker-compose.yml` on purpose rather than guessing at their setup.

Until a collector exists somewhere and `OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT` is pointed at it, this has **no functional impact**: `OSIDB_OTEL_ENABLED` defaults to `false`, so `configure_telemetry()` is a no-op. If someone opts in locally without a reachable collector, span export just fails/retries in `BatchSpanProcessor`'s background thread — it doesn't block requests or crash the app, it just won't produce visible traces.

## What about profiling?

OTel covers **tracing** (where time is spent across services/queries), not **profiling** (which Python code path is slow). These are complementary:

- **Local profiling**: we have `django-silk` for local per-request SQL profiling with full Python stack traces
- **Production profiling**: OTel's profiling signal is still experimental (no stable Python SDK). Options like Pyroscope (Grafana, open-source) exist but are a separate initiative. Not needed right now

## Files changed

| File | Change |
|---|---|
| `pyproject.toml` | OTel dependencies (distro, exporter, 5 instrumentors) |
| `osidb/telemetry.py` | **New** — `OtelSettings` + `instrument_django()` + `configure_telemetry()` + `_instrument_all()` |
| `config/wsgi.py` / `config/asgi.py` | Call `instrument_django()` before the WSGI/ASGI app is built, since `DjangoInstrumentor` must register before Django reads `settings.MIDDLEWARE` |
| `gunicorn_config.py` | `post_fork` hook calls `configure_telemetry()` per worker (psycopg2/requests/redis instrumentation + the OTLP exporter must be set up after fork, not in the preloaded master) |
| `config/celery.py` | `worker_process_init` hook calls `configure_telemetry()` + `CeleryInstrumentor` per worker |
| `docker-compose.yml` | `OSIDB_OTEL_*` env vars passed through to services (`OSIDB_OTEL_ENABLED` defaults to `false` when unset, so an unset env var doesn't fail `OtelSettings` validation); commented example Jaeger/Datadog Agent services for local testing |
| `osidb/tests/test_telemetry.py` | **New** — covers `instrument_django()` gating and `gunicorn_config.post_fork` surviving a `configure_telemetry()` failure |

## How to test locally

Point `OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT` at any OTLP collector you have available (e.g. run your own `jaegertracing/all-in-one` container and add it to your local compose override):

```bash
# Set in .env
OSIDB_OTEL_ENABLED=true
OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT=http://<your-collector>:4317
OSIDB_OTEL_SERVICE_NAME=osidb

# Rebuild and start
make build && make start-local
```

## Caveats

- **Only wired up for Gunicorn**: `osidb-service` in `docker-compose.yml` runs via `manage.py runserver`, which never forks, so the `post_fork` hook that sets up the exporter never fires there. Tracing only works when the app is served via Gunicorn (`scripts/run-service.sh` / `start-local-gunicorn`)
- **No custom spans**: this PR only adds auto-instrumentation. Manual spans for business logic (e.g. "flaw validation", "embargo check") would be a follow-up
- **No redaction**: traces go to the same RH-approved vendor we already send logs to, under the same trust relationship, so no span-attribute scrubbing is applied (see `osidb/telemetry.py` docstring)

## Open questions for the team

1. Is 10% the right default sample ratio, or should it start lower/higher until we've seen real volume?
2. Any concerns about the OTel dependency footprint? (7 new packages)
3. Who owns standing up the OTLP collector (devops), and on what timeline?

---

Assisted-By: Claude Code (claude-opus-4-6)

### DIFF
--- a/config/celery.py
+++ b/config/celery.py
@@ -1,7 +1,13 @@
+import logging
+
 from celery import Celery, signals
 from django.conf import settings
 from kombu import Queue
 from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from osidb.telemetry import OtelSettings, configure_telemetry, safe_instrument
+
+logger = logging.getLogger(__name__)
 
 
 class CelerySettings(BaseSettings):
@@ -62,6 +68,25 @@ app.conf.task_queues = [
 ]
 app.conf.task_default_queue = "default"
 app.conf.task_routes = ("config.celery.FIFORouter",)
+
+
+@signals.worker_process_init.connect
+def on_worker_process_init(**kwargs):
+    # CeleryInstrumentor is wired here rather than in _instrument_all()
+    # because that helper also runs from gunicorn's post_fork, i.e. in web
+    # workers. Folding it in there would make every web worker connect
+    # Celery's task signals too (harmless, but broadens instrumentation to
+    # processes that only publish tasks, not run them). Keep it scoped to
+    # actual Celery workers here instead.
+    try:
+        if not OtelSettings().enabled:
+            return
+        configure_telemetry()
+        from opentelemetry.instrumentation.celery import CeleryInstrumentor
+
+        safe_instrument(CeleryInstrumentor())
+    except Exception:
+        logger.exception("Failed to configure telemetry")
 
 
 @signals.setup_logging.connect
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -108,6 +108,9 @@ services:
         OSIDB_ROLE_ID: ${OSIDB_ROLE_ID}
         OSIDB_SECRET_ID: ${OSIDB_SECRET_ID}
         OSIDB_ENABLE_SILK: ${OSIDB_ENABLE_SILK}
+        OSIDB_OTEL_ENABLED: ${OSIDB_OTEL_ENABLED:-false}
+        OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT: ${OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT}
+        OSIDB_OTEL_SERVICE_NAME: ${OSIDB_OTEL_SERVICE_NAME}
       command: ./scripts/setup-osidb-service.sh
       volumes:
         - ${PWD}:/opt/app-root/src:z
@@ -186,6 +189,9 @@ services:
         PRODUCT_DEF_BRANCH: ${PRODUCT_DEF_BRANCH}
         PRODUCT_DEF_URL: ${PRODUCT_DEF_URL}
         PS_CONSTANTS_URL: ${PS_CONSTANTS_URL}
+        OSIDB_OTEL_ENABLED: ${OSIDB_OTEL_ENABLED:-false}
+        OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT: ${OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT}
+        OSIDB_OTEL_SERVICE_NAME: ${OSIDB_OTEL_SERVICE_NAME}
       depends_on: ["osidb-data", "osidb-service", "redis"]
       # See "NOTE about healthchecks":
       # depends_on:
@@ -265,6 +271,9 @@ services:
         OSV_COLLECTOR_ENABLED: ${OSV_COLLECTOR_ENABLED}
         JIRA_USER_MAPPING_COLLECTOR_ENABLED: ${JIRA_USER_MAPPING_COLLECTOR_ENABLED}
         JIRA_USER_MAPPING_COLLECTOR_URL: ${JIRA_USER_MAPPING_COLLECTOR_URL}
+        OSIDB_OTEL_ENABLED: ${OSIDB_OTEL_ENABLED:-false}
+        OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT: ${OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT}
+        OSIDB_OTEL_SERVICE_NAME: ${OSIDB_OTEL_SERVICE_NAME}
       depends_on: ["osidb-data", "osidb-service", "redis"]
       # See "NOTE about healthchecks":
       # depends_on:
@@ -293,6 +302,31 @@ services:
         PS_CONSTANTS_URL: ${PS_CONSTANTS_URL}
         FLAW_LABELS_URL: ${FLAW_LABELS_URL}
       depends_on: ["redis", "osidb-data"]
+
+    # OpenTelemetry tracing is inert until OSIDB_OTEL_ENABLED is set and
+    # OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT points to a collector, e.g. a local
+    # Jaeger container or the Datadog Agent. Note osidb-service here runs
+    # `manage.py runserver` (setup-osidb-service.sh), which never forks, so
+    # the post_fork hook in gunicorn_config.py that sets up the exporter
+    # never fires. To actually exercise tracing locally, run the app via
+    # Gunicorn instead (scripts/run-service.sh / start-local-gunicorn):
+    #
+    #   jaeger:
+    #     image: mirror.gcr.io/jaegertracing/all-in-one:1
+    #     ports:
+    #       - "16686:16686"  # UI at http://localhost:16686
+    #       - "4317"
+    #
+    #   datadog-agent:
+    #     image: gcr.io/datadoghq/agent:7
+    #     environment:
+    #       DD_API_KEY: ${DD_API_KEY}
+    #       DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_ENDPOINT: "0.0.0.0:4317"
+    #       DD_APM_ENABLED: "true"
+    #     ports:
+    #       - "4317"
+    #
+    # Then set OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT=http://<service>:4317 in .env.
 
     redis:
       container_name: redis
--- a/osidb/telemetry.py
+++ b/osidb/telemetry.py
@@ -1,0 +1,106 @@
+"""
+Vendor-neutral OpenTelemetry setup.
+
+No span redaction: traces go to the same RH-approved vendor we already
+send logs to, under the same trust relationship, so holding them to a
+stricter bar than logs isn't consistent, and it would leave every span
+attribute (SQL, URLs, routes) hidden.
+
+Gated on OSIDB_OTEL_ENABLED — completely inert when disabled.
+Swap tracing vendors by changing OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT alone.
+"""
+
+import logging
+import os
+
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+logger = logging.getLogger(__name__)
+
+
+class OtelSettings(BaseSettings):
+    model_config = SettingsConfigDict(env_prefix="OSIDB_OTEL_")
+
+    enabled: bool = False
+    exporter_otlp_endpoint: str = ""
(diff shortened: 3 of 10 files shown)


osidb/telemetry.py
```diff
@@ -0,0 +1,117 @@
...
+}
+
+
+def _scrub(value):
+    if not isinstance(value, str):
+        return value
+    value = CVE_RE.sub("CVE-XXXX-XXXXX", value)
+    value = UUID_RE.sub("XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX", value)
+    value = QUOTED_STR_RE.sub("'?'", value)
+    return value
```
osoukup, line 45: I am thinking about the scrubbing. Most of the time CVE is not sensitive (unless embargoed). UUID is I think never sensitive as it is random and OSIDB instance specific. I am not sure what all would the quoted strings mean. Still this may be insufficient if there is eg. a description of an embargoed flaw in the text for some weird reason. The best but I guess not really viable way would be looking at the ACLs.

osidb/telemetry.py
```diff
@@ -0,0 +1,117 @@
...
+    return span
+
+
+def configure_telemetry():
+    global _configured
+    if _configured:
+        return
+    _configured = True
+
+    if os.environ.get("OSIDB_OTEL_ENABLED", "").lower() not in ("1", "true", "yes"):
```
osoukup, line 64: The is a helper for this called `get_env` in the helpers.

docker-compose.yml
```diff
@@ -108,6 +108,9 @@ services:
         OSIDB_VAULT_ADDR: ${OSIDB_VAULT_ADDR}
         OSIDB_ROLE_ID: ${OSIDB_ROLE_ID}
         OSIDB_SECRET_ID: ${OSIDB_SECRET_ID}
+        OSIDB_OTEL_ENABLED: ${OSIDB_OTEL_ENABLED}
+        OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT}
+        OTEL_SERVICE_NAME: ${OTEL_SERVICE_NAME}
```
osoukup, line 113: Wouldn't be more consistent if all three have a prefix `OSIDB_OTEL`?

osoukup (commented):
This approach makes sense to me

osidb/telemetry.py
AlejandroMinaya: Noted

docker-compose.yml
AlejandroMinaya: Makes sense to me

AlejandroMinaya:
@coderabbitai review

AlejandroMinaya:
@coderabbitai I've addressed the comments in your review, would you mind checking it again?

AlejandroMinaya:
@coderabbitai Added the init as part of the post_fork. Please review the updates

AlejandroMinaya:
@coderabbitai check updates

config/celery.py
```diff
@@ -64,6 +71,19 @@ def route_for_task(self, task, args=None, kwargs=None):
 app.conf.task_routes = ("config.celery.FIFORouter",)
 
 
+@signals.worker_process_init.connect
+def on_worker_process_init(**kwargs):
+    configure_telemetry()
+    if not get_env("OSIDB_OTEL_ENABLED", "false", is_bool=True):
```
Elkasitu: nitpick: we should be using `pydantic_settings` these days

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+
+        def shutdown(self):
+            self._wrapped.shutdown()
+
+        def force_flush(self, timeout_millis=None):
+            return self._wrapped.force_flush(timeout_millis)
+
+    resource = Resource.create(
+        {
+            "service.name": get_env("OSIDB_OTEL_SERVICE_NAME", "osidb"),
```
Elkasitu: ditto

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+        }
+    )
+    provider = TracerProvider(resource=resource)
+    exporter = RedactingExporter(
+        OTLPSpanExporter(endpoint=get_env("OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT"))
```
Elkasitu: ditto

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+        try:
+            import importlib
+
+            mod = importlib.import_module(module_path)
+            instrumentor = getattr(mod, class_name)()
+            if not instrumentor.is_instrumented_by_opentelemetry:
+                instrumentor.instrument()
+                logger.info("Instrumented %s", class_name)
+        except Exception:
+            logger.exception("Failed to instrument %s", class_name)
```
Elkasitu: why do we load these dynamically if all are explicit dependencies?

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+        logger.exception("Failed to instrument DjangoInstrumentor")
+
+
+def configure_telemetry():
+    global _configured
+    if _configured:
+        return
+    _configured = True
+
+    if not get_env("OSIDB_OTEL_ENABLED", "false", is_bool=True):
```
Elkasitu: ditto

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+    DjangoInstrumentor works by inserting itself into settings.MIDDLEWARE,
+    which Django only reads once, when WSGIHandler is constructed. That
+    construction happens in get_wsgi_application(), which (with
+    preload_app=True) runs in the master process before post_fork. So this
+    must be called before get_wsgi_application(), not from post_fork like
+    the rest of telemetry setup (which needs to run per-worker instead).
+    Safe to call before the real TracerProvider exists: OTEL's tracer is a
+    lazy proxy that binds to whatever provider is configured later.
+    """
+    if not get_env("OSIDB_OTEL_ENABLED", "false", is_bool=True):
```
Elkasitu: ditto

osidb/telemetry.py
Elkasitu: agreed

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+QUOTED_STR_RE = re.compile(r"'[^']*'")
+
+SCRUB_ATTR_KEYS = {
+    "db.statement",
+    "http.target",
+    "http.url",
+    "url.full",
+    "url.path",
+    "url.query",
+    "http.route",
```
Elkasitu: I may not be understanding this correctly, but if we're scrubbing urls (either ingress or outgress) how can the metrics be useful? Can you explain what this does exactly? For db.statement it's fairly straightforward.

osidb/telemetry.py
```diff
@@ -0,0 +1,145 @@
...
+            instrumentor.instrument()
+    except Exception:
+        logger.exception("Failed to instrument DjangoInstrumentor")
+
+
+def configure_telemetry():
+    global _configured
+    if _configured:
+        return
+    _configured = True
```
Elkasitu: when is this necessary? we call it in pre_fork exactly once per process and exactly once per celery worker init.

it's also not thread-safe and we use gthread gunicorn workers

docker-compose.yml
```diff
@@ -108,6 +108,9 @@ services:
         OSIDB_ROLE_ID: ${OSIDB_ROLE_ID}
         OSIDB_SECRET_ID: ${OSIDB_SECRET_ID}
         OSIDB_ENABLE_SILK: ${OSIDB_ENABLE_SILK}
+        OSIDB_OTEL_ENABLED: ${OSIDB_OTEL_ENABLED}
+        OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT: ${OSIDB_OTEL_EXPORTER_OTLP_ENDPOINT}
+        OSIDB_OTEL_SERVICE_NAME: ${OSIDB_OTEL_SERVICE_NAME}
```
Elkasitu: maybe we should document that locally this only works if running `start-local-gunicorn`

config/celery.py
```diff
@@ -64,6 +71,19 @@ def route_for_task(self, task, args=None, kwargs=None):
...
+        return
+    try:
+        from opentelemetry.instrumentation.celery import CeleryInstrumentor
+
+        CeleryInstrumentor().instrument()
+    except Exception:
+        logger.exception("Failed to instrument CeleryInstrumentor")
```
Elkasitu: nitpick: purely stylistic, but this could follow the same pattern as `instrument_django()`

osidb/telemetry.py
Elkasitu: and aren't we technically instrumenting django twice?

Elkasitu:
>1. Is the redaction strategy (scrub everything) acceptable, or do we want selective redaction per-table?

Are we even sure we need scrubbing / anonymization? We don't do it for logging and we log to a third-party provider, these are RH-approved providers so I presume there's agreements and that we can get private dashboards? Is this something specifically required by the internal guidelines?

>2. Should we add sampling config now or defer to production deployment?

You'll need to set up the corresponding stuff in the ops repo along with all the environment variables introduced here, otherwise it's not usable in any deployed environment.

>3. Any concerns about the OTel dependency footprint? (7 new packages)

No.

>4. Who owns standing up the OTLP collector (devops), and on what timeline?

Whoever's available and depending on capacity 😄 It's not urgent but if we can get it in ASAP then why not, should improve debugging, incident detection, etc.

PS: maybe this warrants an ADR. There's a lot of choices being made that warrant team involvement, comparing trade-offs, etc. instead of just the few of us who happen to take a look at the PR. The PR is also just one component out of the whole system which makes it even more difficult to assess.

AlejandroMinaya:
@coderabbitai review changes since your last comments

AlejandroMinaya:
@coderabbitai It is not necessary to pin the versions in the `pyproject.toml`. As you may see, other packages follow the same "define a floor" convention. Please review again and approve it there are no issues left

AlejandroMinaya:
@coderabbitai The ADR exists in a separate repo. This is intended, not a missing reference

osoukup (approved):
LGTM. Great work!

PR #1384, opened 2026-07-28
Osidb 5239 use case srp reporting for due dates

roduran-dev:
Add missing endpoints and new needed filters.


roduran-dev:
After a refinment meet this will be done in the frontend logic.

PR #1387, opened 2026-07-30
Use stored search vector for Flaw full-text search (OSIDB-5283)

AlejandroMinaya:
## Summary

- Full-text search queries on the `osidb_flaw` table were slow due to computing `tsvector` on every request
- This PR adds a stored `SearchVectorField` populated by a `BEFORE INSERT OR UPDATE` trigger, so the vector is computed once on write instead of on every read
- A GIN index on `search_vector` supports the FTS query path; partial CVE-ID matching uses trigram similarity without a dedicated index for now
- Adds a reusable `SearchVectorMixin` (`osidb/mixins.py`): a model mixes it in, declares weighted `search_vector_fields`, and the column + trigger are wired up automatically via a `class_prepared` signal
- Removes the dead, unused `AffectManager.fts_search` / `AffectV1Manager.fts_search` helpers and their superseded test

## Tradeoffs

- **ACL index preference by the planner**: the existing GIN index on `acl_read` may be preferred over the new `search_vector` GIN index depending on the query. In practice RLS filters first through `acl_read`, and the FTS GIN index kicks in for the search predicate within that filtered set
- **Flaw-coupled default path**: the `search_helper` default path assumes the model has a `search_vector` column. A `ValueError` guard prevents silent misuse on other models; adopting another model still requires a migration adding its own `search_vector` column, even though the trigger/column setup itself is handled generically by `SearchVectorMixin`
- **Trigger maintenance**: if search fields change, both `Flaw.search_vector_fields` and migration 0251 (`osidb/migrations/0251_flaw_search_vector.py`) must be kept in sync — the comment next to `search_vector_fields` in `flaw.py` calls this out
- **Trigram similarity without index**: partial CVE matching uses `similarity()` which does not leverage a GIN trigram index — intentionally deferred pending team discussion on threshold tuning and the `%` operator
- **Trigram threshold at 0.7**: balances recall vs false positives for partial CVE matches; may need tuning based on production usage

## Test plan

- [x] `test_search_websearch_exclusion` — `-exclusion` operator with trigger-populated vector
- [x] `test_search_fts_no_match_returns_empty` — no false positives
- [x] `test_search_trigram_partial_cve` — partial CVE-ID matching via trigram
- [x] `test_search_helper_rejects_model_without_search_vector` — guard for non-Flaw models
- [x] `TestSearchVectorMixin` — a `SearchVectorMixin` subclass without `search_vector_fields` raises `ImproperlyConfigured`; a subclass with it registers the column/trigger cleanly
- [x] Existing `TestSearch` tests pass with the new query path

Assisted-By: Claude Sonnet 5 (Anthropic)


osidb/filters.py
```diff
@@ -539,38 +542,29 @@ def search_helper(
...
-    rank = SearchRank(vector, query, cover_density=True)
-    # Consider proximity of matching terms when ranking
+    rank = SearchRank(F("search_vector"), query, cover_density=True)
 
     return (
         queryset.annotate(rank=rank, similarity=similarity)
-        # The similarity threshold of 0.7 has been found by trial and error to work best with CVEs
-        .filter(Q(rank__gt=0) | Q(similarity__gt=0.7))
+        .filter(Q(search_vector=query) | Q(similarity__gt=0.7))
         .order_by("-rank")
```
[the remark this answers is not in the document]
AlejandroMinaya: Skip. The Q(search_vector=query) | Q(similarity__gt=0.7) is
  an OR, so a search like "CVE-2024-1234 -kernel" will match via trigram on cve_id even if the FTS
  branch excludes it. But this is correct behavior: the trigram matches against cve_id only (not
  title), so a CVE with a matching ID should surface regardless of text exclusions — the user is
  searching for a CVE number. Gating trigram behind a regex would break legitimate partial-CVE lookups
  mixed with text terms. The existing test test_search_trigram_partial_cve covers this.
  
Assisted-by: Claude Opus 4.6

osidb/migrations/0250_flaw_search_vector_indexes.py
```diff
@@ -0,0 +1,36 @@
...
+        ("osidb", "0249_flaw_search_vector"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="""
+            CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_osidb_flaw_search_vector
+            ON osidb_flaw USING gin (search_vector);
+            """,
+            reverse_sql="DROP INDEX IF EXISTS idx_osidb_flaw_search_vector;",
```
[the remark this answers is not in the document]
AlejandroMinaya: Skip. DROP INDEX CONCURRENTLY cannot run inside a transaction,
  and Django's migration reverse runs inside a transaction by default. The current DROP INDEX IF EXISTS
  is the correct approach for reverse migrations — it's a standard pattern. Using CONCURRENTLY in
  reverse_sql would fail.
  
  Assited-by: Claude Opus 4.6

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+
+            DROP TRIGGER IF EXISTS osidb_flaw_search_vector_trigger ON osidb_flaw;
+            CREATE TRIGGER osidb_flaw_search_vector_trigger
+            BEFORE INSERT OR UPDATE OF title, cve_id, comment_zero, cve_description, statement
+            ON osidb_flaw
+            FOR EACH ROW EXECUTE FUNCTION osidb_flaw_search_vector_update();
+            """,
+            reverse_sql="""
+            DROP TRIGGER IF EXISTS osidb_flaw_search_vector_trigger ON osidb_flaw;
+            DROP FUNCTION IF EXISTS osidb_flaw_search_vector_update();
```
osoukup: This function is something automatically created by the adding the field?

osidb/migrations/0250_flaw_search_vector_indexes.py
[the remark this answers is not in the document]
AlejandroMinaya: You're right! Fixing right away

osoukup (approved):
I am not much familiar with some parts of this PR but as far as I can understand it LGTM

osidb/migrations/0249_flaw_search_vector.py
AlejandroMinaya: No, the trigger and function were added after the fact

osidb/filters.py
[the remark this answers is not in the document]
AlejandroMinaya: How are you certain that the index is not going to be used?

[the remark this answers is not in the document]
AlejandroMinaya: I understand. I didn't know that it didn't leverage the index. Does this mean that even with the index creation, we wouldn't see any benefits compared to the behavior before this PR (talking about the trigram index specifically)

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+            if cursor.rowcount == 0:
+                break
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("osidb", "0248_ubipackage"),
+    ]
```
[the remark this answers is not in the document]
AlejandroMinaya: This is a valid concern but the risk of having partial updates in the production database offsets the risk of a one-time long-running transaction. I believe it is safe to skip it for now

osidb/filters.py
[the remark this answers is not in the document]
AlejandroMinaya: Removed the trigram index. Since it is not a regression in functionality there is no need to introduce those changes now and can be part of future contributions. The key improvement here has to do with the full-text search

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+    setweight(to_tsvector('english', COALESCE({prefix}statement, '')), 'D')
+""".strip()
+
+
+def backfill_search_vector(apps, schema_editor):
+    # set_user_acls persists for this DB connection only — safe for migration context
+    from django.conf import settings
+    from osidb.core import set_user_acls
+
+    set_user_acls(settings.ALL_GROUPS)
```
Elkasitu: these days we use the `@bypass_rls` decorator instead

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+    BATCH = 5000
+    with schema_editor.connection.cursor() as cursor:
+        while True:
+            cursor.execute(
+                f"""
+                UPDATE osidb_flaw SET search_vector =
+                    {SEARCH_VECTOR_SQL.format(prefix="")}
```
Elkasitu: it's bad practice to do string interpolation in SQL queries even when the values are not user-provided and hardcoded, is there no way to do this through the ORM?

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="flaw",
+            name="search_vector",
+            field=django.contrib.postgres.search.SearchVectorField(
+                null=True, editable=False
+            ),
+        ),
+        migrations.RunSQL(
```
Elkasitu: generally we prefer to keep django-generated migrations as-is and do any data migrations in a separate file

osidb/migrations/0249_flaw_search_vector.py
```diff
@@ -0,0 +1,78 @@
...
+            sql=f"""
+            CREATE OR REPLACE FUNCTION osidb_flaw_search_vector_update() RETURNS trigger AS $$
```
Elkasitu: one of the few benefits of `django-pghistory` is that it uses [django-pgtrigger](https://django-pgtrigger.readthedocs.io/en/4.17.0/) as a dependency, which might be better for managing these triggers than raw SQL, this also means that you can make something reusable (maybe as a mixin) and no need to check whether the caller is Flaw / has the search_vector attribute

osidb/filters.py
```diff
@@ -539,38 +542,29 @@ def search_helper(
...
+        return queryset.annotate(rank=rank).filter(rank__gt=0).order_by("-rank")
 
+    # Empty tuple or 'search' (default from django-filters when field name not specified)
+    # Search all Flaw text columns, weighted so title is most relevant
+    # To support other models, add a search_vector column with a DB trigger to populate it
+    if not hasattr(queryset.model, "search_vector"):
+        raise ValueError(
+            f"{queryset.model.__name__} has no search_vector column. "
+            "Pass explicit field_names to search_helper."
+        )
```
Elkasitu: this is not very good design OOP-wise, have you considered other options?

osidb/migrations/0252_flaw_search_vector_backfill.py
```diff
@@ -0,0 +1,37 @@
...
+
+    dependencies = [
+        ("osidb", "0250_flaw_search_vector"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=backfill_search_vector,
+            reverse_code=migrations.RunPython.noop,
+        ),
```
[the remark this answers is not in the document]
AlejandroMinaya: We spoke about this before. It is okay if it all rollsback together. This migration should always be one-off, not recurrent. We don't want inconsistent data in the DB

osidb/models/flaw/flaw.py
```diff
@@ -298,6 +300,17 @@ class FlawRequiresCVEDescription(models.TextChoices):
 
     local_updated_dt = models.DateTimeField(default=timezone.now)
 
+    # search_vector column and its DB trigger come from SearchVectorMixin, driven by
+    # search_vector_fields below. Update search_vector_fields if the weighted search
+    # fields change; the column must never be set directly in application code.
```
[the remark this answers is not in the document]
AlejandroMinaya: `pgtrigger` handles this for us. The added trigger has a matching hash that when changed it is auto-detected by Django as a migration drift. I tested this by changing the trigger and performing a `makemigrations --check`. We can drop this comment

osidb/mixins.py
```diff
@@ -129,6 +130,52 @@ def create(self, **kwargs):
...
+        f"setweight(to_tsvector('{config}', COALESCE(NEW.{column}, '')), '{weight}')"
+        for column, weight in fields
+    )
+    columns = [column for column, _ in fields]
+    return pgtrigger.Trigger(
+        name=name,
+        when=pgtrigger.Before,
+        operation=pgtrigger.Insert | pgtrigger.UpdateOf(*columns),
+        func=f"NEW.search_vector := {setweights}; RETURN NEW;",
+    )
```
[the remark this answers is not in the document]
AlejandroMinaya: Addressed by making this part of the Mixin and adding to the submethod a validation step for the weights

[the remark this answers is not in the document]
AlejandroMinaya: Config is by default `english`. It is unlikely that it will change and other things would break if the set value is not supported by Postgres. I personally believe the risk is too low since this isn't even a value that receives any kind of user input.

osidb/mixins.py
```diff
@@ -129,6 +131,74 @@ def create(self, **kwargs):
...
+    *fields: tuple[str, str],
+    name: str = "update_search_vector",
+    config: str = "english",
+) -> pgtrigger.Trigger:
+    """
+    Build a BEFORE INSERT/UPDATE trigger that keeps a `search_vector` column updated from
+    weighted document fields, e.g. _search_vector_trigger(("title", "A"), ("comment_zero", "B")).
+
+    Unlike pgtrigger.UpdateSearchVector (backed by Postgres' tsvector_update_trigger), this
+    supports per-field weights, which OSIDB search ranking relies on.
```
Elkasitu, line 151: suggestion: consider contributing something that does support per-field weights upstream

osidb/mixins.py
```diff
@@ -129,6 +131,74 @@ def create(self, **kwargs):
...
+    declares search_vector_fields. Runs on class_prepared (after _meta exists) rather
+    than from the mixin's own class body, since Meta.triggers set here would only take
+    effect for subclasses that don't define their own Meta class.
+    """
+    if issubclass(sender, SearchVectorMixin) and sender.search_vector_fields:
+        _search_vector_trigger(
+            *sender.search_vector_fields, config=sender.search_vector_config
+        ).register(sender)
+
+
```
Elkasitu, line 217: Can't you avoid this by setting up an `__init_subclass__` in `SearchVectorMixin`?

osidb/mixins.py
```diff
@@ -129,6 +131,74 @@ def create(self, **kwargs):
...
+        class Flaw(SearchVectorMixin, ...):
+            search_vector_fields = (("title", "A"), ("comment_zero", "B"))
+
+    A BEFORE INSERT/UPDATE trigger keeping the column in sync is registered automatically
+    once the model class is prepared; search_vector must never be set in application code.
+    """
+
+    search_vector = SearchVectorField(null=True, editable=False)
+    search_vector_fields: tuple = ()
```
Elkasitu, line 185: what happens if the subclass does not define `search_vector_fields`? will an "empty" trigger be created? I think we should raise an error as it's technically a programmer error.

osidb/mixins.py
Elkasitu: Here's a rough example (drafted with Gemini) on how I'd envision it:

```python3
import pgtrigger


class RequiredTriggerMixin:
    # Set to Sentinel or None on the base mixin
    trigger_condition = None

    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)

        # 1. Skip abstract models (they shouldn't have triggers registered directly)
        if getattr(cls._meta, "abstract", False):
            return

        # 2. Check if the class (or its non-mixin parents) defined the condition
        if getattr(cls, "trigger_condition", None) is None:
            raise TypeError(
                f"Model '{cls.__name__}' inherits from {RequiredTriggerMixin.__name__} "
                f"but does not define 'trigger_condition'."
            )

        # 3. Register the trigger dynamically
        pgtrigger.register(
            pgtrigger.Protect(
                name="protect_conditioned_rows",
                operation=pgtrigger.Update | pgtrigger.Delete,
                condition=cls.trigger_condition,
            )
        )(cls)
```

osidb/models/flaw/flaw.py
```diff
@@ -311,6 +324,7 @@ class Meta:
         indexes = TrackingMixin.Meta.indexes + [
             models.Index(fields=["-cve_id"]),
             GinIndex(fields=["acl_read"]),
+            GinIndex(fields=["search_vector"], name="idx_osidb_flaw_search_vector"),
```
Elkasitu: technically you can define this in the mixin and inherit here like for `TrackingMixin`

osidb/filters.py
```diff
@@ -539,38 +543,29 @@ def search_helper(
...
+        return queryset.annotate(rank=rank).filter(rank__gt=0).order_by("-rank")
 
+    # Empty tuple or 'search' (default from django-filters when field name not specified)
+    # Search all Flaw text columns, weighted so title is most relevant
+    if not issubclass(queryset.model, SearchVectorMixin):
+        raise ValueError(
+            f"{queryset.model.__name__} has no search_vector column. "
+            "Add SearchVectorMixin from osidb.mixins, or pass explicit field_names to "
+            "search_helper."
+        )
```
Elkasitu, line 557: This breaks existing behavior: Affect also uses `search_helper` but does not inherit from the mixin.

I think a better approach might be to attempt to use the stored search vector and if that fails, we log a warning and use the "old way"

osidb/filters.py
Elkasitu, line 522: nitpick: I'd love if we could move this out of here and into either its own file or into a utils file, importing this in models.py feels awkward (wrong direction).

Elkasitu: Or in fact, get rid of `fts_search()` completely, as it's only used in tests... then this method could remain here

osidb/filters.py
Elkasitu: If we remove `AffectManager.fts_search()` it technically does not break anything, but since there's no way to link `search_helper` to usage of a specific Mixin, I still think defaulting to the unoptimized query with a warning is a more robust approach

osidb/mixins.py
AlejandroMinaya: __init__subclass__ runs before Django has added `cls._meta`. 

This would raise an AttributeError or silently resolde to the parent's `_meta`, i.e. SearchVectorMixin's own `_meta`

osidb/mixins.py
AlejandroMinaya: Since by the default it is an empty tuple, it is a _falsy_ value and the guard in the trigger registration would fail.

I can see adding the exception as a way of communicating it to the programmer but it is hard for me to see a scenario where they explicitly inherit from `SearchVectorMixin` and forget to set search fields

No hard opinion. Your call

osidb/filters.py
AlejandroMinaya: Looking at the definition of `fts_search` in `AffectManager` and `AffectManagerV1`, it has a hardcoded list of `fields_to_search`. The default behavior only applied for unspecified fields (which matches the `FlawManager` method definition)

I believe that `Affect` could benefit from inheriting the `SearchVectorMixin` but I believe that is outside the scope of this PR.

AlejandroMinaya: I don't believe that this is breaking any preexisting Affect behavior. I will delete the `fts_search` methods though as per your observations in other comments

osidb/mixins.py
Elkasitu: > but it is hard for me to see a scenario where they explicitly inherit from SearchVectorMixin and forget to set search fields

you'd be surprised... if it's possible then someone will do it at some point, the cost of raising an error is cheap so let's just do it

osidb/mixins.py
Elkasitu: you're right, good catch.

in that case, could we move the signal to `signals.py`?

Elkasitu (approved):
LGTM, great work!

PR #1389, opened 2026-07-30
OSIDB-5085 Add Flaw to upstream Mapping Endpoints

svelamak:
Added flaw-to-upstream mapping endpoints list/create, update/delete.
-Add `upstream_mappings.py` for `FlawUpstreamMappingDetailView` and `FlawUpstreamMappingListCreateView`.
- Export two views in `__init__.py`.
-Register routes in `urls.py`.
-Added tests in `test_upstream_mappings.py`.
-Added entry in `CHANGELOG.md`.
-Regenerated openapi.yml.
Closes:OSIDB-5085

### DIFF
--- a/regulatory_reporting/api_views/upstream_mappings.py
+++ b/regulatory_reporting/api_views/upstream_mappings.py
@@ -1,0 +1,54 @@
+from django.shortcuts import get_object_or_404
+from rest_framework import mixins, viewsets
+from rest_framework.permissions import IsAuthenticatedOrReadOnly
+
+from osidb.api_views import RudimentaryUserPathLoggingMixin, get_valid_http_methods
+from osidb.models import Flaw
+from regulatory_reporting.models.upstream import FlawUpstreamMapping
+from regulatory_reporting.serializers.upstream import FlawUpstreamMappingSerializer
+
+
+class FlawUpstreamMappingListCreateView(
+    RudimentaryUserPathLoggingMixin,
+    mixins.ListModelMixin,
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet,
+):
+    """
+    API endpoint for listing and creating flaw-to-upstream mappings
+    """
+
+    http_method_names = get_valid_http_methods(
+        viewsets.GenericViewSet, excluded=["put", "patch", "delete"]
+    )
+    serializer_class = FlawUpstreamMappingSerializer
+    permission_classes = [IsAuthenticatedOrReadOnly]
+
+    def get_flaw(self):
+        return get_object_or_404(Flaw, uuid=self.kwargs["flaw_uuid"])
+
+    def get_queryset(self):
+        return FlawUpstreamMapping.objects.filter(flaw=self.get_flaw())
+
+    def perform_create(self, serializer):
+        serializer.save(flaw=self.get_flaw())
+
+
+class FlawUpstreamMappingDetailView(
+    RudimentaryUserPathLoggingMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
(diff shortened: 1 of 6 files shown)


regulatory_reporting/urls.py
```diff
@@ -47,5 +49,17 @@
...
+    FlawUpstreamMappingListCreateView,
+    basename="flawupstreammappings",
+)
 
+# Flaw-to-upstream mappings (update/delete)
+router.register(
+    r"flaw-upstream-mappings",
+    FlawUpstreamMappingDetailView,
+    basename="flawupstreammappingdetail",
+)
```
roduran-dev, line 73: This will not enter in the CRA_NOTIFICATIONS_ENABLED logic as it needs to be under 
`/regulatory-reporting/api/v1/notifications/ `→ needs CRA_NOTIFICATIONS_ENABLED

Any other /regulatory-reporting/api/v1/... path → needs CRA_REPORTING_ENABLED

regulatory_reporting/api_views/upstream_mappings.py
```diff
@@ -0,0 +1,54 @@
...
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet,
+):
+    """
+    API endpoint for listing and creating flaw-to-upstream mappings
+    """
+
+    http_method_names = get_valid_http_methods(
+        viewsets.GenericViewSet, excluded=["put", "patch", "delete"]
+    )
```
roduran-dev, line 23: This says to create but there is no post available for creations

roduran-dev (changes requested):
I check the notifications mapping and I have a doubt. 

This endpoint only allows to map existing contacts to a flaw but not to create a contact and map it to the flaw. Is the contact created in another step?

regulatory_reporting/api_views/upstream_mappings.py
svelamak: Post is available as `CreateModelMixin` provides `create()` and DRF's router automatically wires Post to it. `http_method_names` only excludes put/patch/delete, so post/get options are not blocked. Let me know if you  have any other thoughts.

regulatory_reporting/api_views/upstream_mappings.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    http_method_names = get_valid_http_methods(
+        viewsets.GenericViewSet, excluded=["get", "post", "patch"]
+    )
+    queryset = FlawUpstreamMapping.objects.all()
+    serializer_class = FlawUpstreamMappingSerializer
+    permission_classes = [IsAuthenticatedOrReadOnly]
+    lookup_url_kwarg = "mapping_uuid"
+    lookup_field = "uuid"
```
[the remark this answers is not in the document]
svelamak: This model does not have `acl_read/acl_write`. It only has `TrackingMixin` not `ACLMixin`. No ACL fields exist to scope by. If mapping visibility should be tied to flaw's embargo status, that would need ACLMixin added to the model itself not just to a queryset filter.

regulatory_reporting/urls.py
svelamak: As discussed over slack, no URL changes needed here for `flaw-upstream-mapping` stays under the generic prefix. 
A new ticket https://redhat.atlassian.net/browse/OSIDB-5296 to update middleware gating logic has been created.

regulatory_reporting/api_views/upstream_mappings.py
roduran-dev: Oh, ok LGTM then

roduran-dev (approved):
LGTM

PR #1400, opened 2026-08-03
OSIDB-5088 Add Upstream Maintainer Send Email Action

svelamak:
Extends `send_email` to track who sent an upstream maintainer notification and when, and sets the sender address as per OSIDB-5059 resolution.
- In `upstream.py` models added `actor` and `sent_at` to `UpstreamNotification` along with migration file.
- `actor` now records in `send_email` and also sets `from` in `regulatory_reporting/api_views/upstream_notifications.py` 
- `sent_at=timezone.now()` is set on success in `tasks.py` 
- Added tests in `test_upstream_notifications.py`.
- Added entry in `CHANGELOG.md`.
Closes: OSIDB-5088


regulatory_reporting/tasks.py
```diff
@@ -15,6 +16,7 @@ def mark_upstream_notification_sent(result, notification_uuid):
     UpstreamNotification.objects.filter(uuid=notification_uuid).update(
```
roduran-dev: to avoid collisions you should add  filter by status. (I think by QUEUED) 

If not a duplicate or later requests to the same notification uuid could override an already sent notification overriding the sent_at date or even change the status to FAILED (if it failed)

regulatory_reporting/api_views/upstream_notifications.py
roduran-dev, line 132: you are now updating the actor (line 108) but during rollback here the actor is not being updated. 

This could be intentional thought 🤔

roduran-dev (approved):
Everything LGTM,  just 2 comments but I consider those edge cases.

regulatory_reporting/tasks.py
svelamak: Good Catch, fixed. Both callbacks now filter `status=QUEUED` before updating, same pattern `send_email` already uses.

regulatory_reporting/api_views/upstream_notifications.py
svelamak: Yes, intentional. On rollback we're reverting to the pre-send state entirely, so there is no new `actor` to record since nothing was actually sent

regulatory_reporting/tests/test_upstream_notifications.py
```diff
@@ -277,6 +282,83 @@ def test_send_email_includes_confidentiality_notice_when_embargoed(
...
+        )
+        mark_upstream_notification_failed(
+            request=None,
+            exc=Exception("SMTP error"),
+            traceback=None,
+            notification_uuid=str(notification.uuid),
+        )
+        notification.refresh_from_db()
+        assert notification.status == UpstreamNotification.NotificationStatus.FAILED
+        assert notification.sent_at is None
```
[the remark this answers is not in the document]
svelamak: Fixed

PR #1407, opened 2026-08-05
Add ACE API integration to osidb

temoore-rh:
Assisted-by: Claude

Closes OSIDB-5203

### DIFF
--- a/apps/ace/client.py
+++ b/apps/ace/client.py
@@ -1,0 +1,75 @@
+from datetime import datetime, timedelta
+
+import requests
+from celery.utils.log import get_task_logger
+
+from osidb.helpers import get_env
+
+logger = get_task_logger(__name__)
+
+AFFECT_AUTOMATION_URL = get_env("AFFECT_AUTOMATION_URL", default="")
+AFFECT_AUTOMATION_TOKEN = get_env("AFFECT_AUTOMATION_TOKEN", default="")
+AFFECT_AUTOMATION_MAX_CONNECTION_AGE = get_env(
+    "AFFECT_AUTOMATION_MAX_CONNECTION_AGE", default="300"
+)
+AFFECT_AUTOMATION_TIMEOUT = get_env(
+    "AFFECT_AUTOMATION_TIMEOUT", default="60", is_int=True
+)
+
+
+class AffectAutomationConnector:
+    """
+    Connection handler for the Affect Automation microservice.
+
+    Follows the JiraConnector pattern: session reuse with max connection age.
+    """
+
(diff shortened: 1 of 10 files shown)


apps/ace/client.py
```diff
@@ -0,0 +1,75 @@
...
+    def session(self) -> requests.Session:
+        if self._session is None:
+            self._session = self._create_session()
+            self._session_timestamp = datetime.now()
+        elif AFFECT_AUTOMATION_MAX_CONNECTION_AGE is not None:
+            age = datetime.now() - self._session_timestamp
+            if age > timedelta(seconds=int(AFFECT_AUTOMATION_MAX_CONNECTION_AGE)):
+                self._session = self._create_session()
+                self._session_timestamp = datetime.now()
+        return self._session
```
AlejandroMinaya, line 55: I see the opportunity to collapse if branches or to create a submethod to initialize the session.

What I would want to avoid is unintentional divergence if we forget to update one of them

apps/ace/microservice.py
```diff
@@ -0,0 +1,94 @@
...
+            ps_update_stream = affect_data.get("ps_update_stream", "")
+            ps_component = affect_data.get("ps_component", "")
+
+            if (
+                ps_component
+                and Affect.objects.filter(
+                    flaw=flaw,
+                    ps_update_stream=ps_update_stream,
+                    ps_component=ps_component,
+                ).exists()
```
AlejandroMinaya, line 45: I think we could avoid querying the database per each iteration. We could either extract all the `ps_update_stream` and `ps_component` from the response and query the DB once filtering by the list or something in that spirit. 

I believe that it would even allow us to get the skipped count up front.

WDYT?

apps/ace/microservice.py
```diff
@@ -0,0 +1,94 @@
...
+                impact=affect_data.get("impact", flaw.impact or ""),
+                created_by="AffectCreationEngine",
+                updated_by="AffectCreationEngine",
+                assist_meta=affect_data.get("assist_meta", {}),
+            )
+
+            affectedness = affect_data.get("affectedness", "")
+            if affectedness:
+                affect.affectedness = affectedness
+
```
AlejandroMinaya, line 65: What do you think about using the walrus operation and skipping the empty string default? I'm picturing something like this:

```suggestion
            if affectedness := affect_data.get("affectedness"):
                affect.affectedness = affectedness

```

The same would go for the other setters.

temoore-rh:
Closing since ticket scope changed to move OSIDB ACE logic to affect-create repo. Followup MR opened here: https://redhat.atlassian.net/browse/OSIDB-5203

PR #1411, opened 2026-08-06
fix: trigger ACE based on workflow (OSIDB-4972, OSIDB-5307, OSIDB-5312)

Elkasitu:
Instead of being based solely on changes to the components field, it now requires the workflow_name to be "DEFAULT" (we skip rejected, embargoed and manual workflows) and at least in "TRIAGE" state so that we're sure all the required bits are available.

### DIFF
--- a/apps/ace/signals.py
+++ b/apps/ace/signals.py
@@ -22,19 +22,36 @@ def schedule_sync_flaw_affects_on_components_change(sender, instance, **kwargs) 
     if kwargs.get("raw"):
         return
 
-    update_fields = kwargs.get("update_fields")
-    if update_fields is not None and "components" not in update_fields:
-        return
-
     new_list = list(instance.components or [])
 
     if instance._state.adding:
-        components_changed = True
-    else:
-        old_list = list(Flaw.objects.get(pk=instance.pk).components or [])
-        components_changed = old_list != new_list
+        return
 
-    if not components_changed:
+    update_fields = kwargs.get("update_fields")
+    if (
+        update_fields is not None
+        and "components" not in update_fields
+        and "workflow_state" not in update_fields
+        and "workflow_name" not in update_fields
+    ):
+        return
+
+    db_flaw = Flaw.objects.get(pk=instance.pk)
+
+    if instance.workflow_name != "DEFAULT" or instance.workflow_state in (
+        "",
+        "NEW",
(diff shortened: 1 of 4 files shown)


apps/ace/signals.py
```diff
@@ -22,19 +22,36 @@ def schedule_sync_flaw_affects_on_components_change(sender, instance, **kwargs)
...
+
+    old_list = list(db_flaw.components or [])
+    components_changed = old_list != new_list
+    workflow_state_became_eligible = db_flaw.workflow_state in (
+        "",
+        "NEW",
+    ) and instance.workflow_state not in ("", "NEW", "DONE")
 
-    if not components_changed:
+    if not components_changed and not workflow_state_became_eligible:
```
osoukup: I am not sure I understand the condition well but shouldn't there be **or** instead?

osoukup (approved):
Check the one condition I commented about but otherwise LGTM

apps/ace/signals.py
Elkasitu: if components change or the workflow state transitions from "empty" or "NEW" into "TRIAGE" then we trigger auto-affects machinery, this is the negation of that.

Would `if not (components_changed or workflow_state_became_eligible):` be more readable?

osoukup: I do not see it bad to read, I just thought that both need to be true to process forward.

apps/ace/tasks.py
```diff
@@ -733,6 +733,17 @@ def sync_flaw_affects_from_newcli(flaw_id: str) -> dict[str, Any]:
...
+
+    if flaw.workflow_name != "DEFAULT" or flaw.workflow_state in ("", "NEW", "DONE"):
+        logger.info(
+            "Skipping auto-affect creation for flaw %s: "
+            "workflow_name=%r, workflow_state=%r (requires DEFAULT workflow, not NEW or DONE)",
+            flaw_id,
+            flaw.workflow_name,
+            flaw.workflow_state,
+        )
+        return {"skipped_reason": "workflow gate: not eligible"}
```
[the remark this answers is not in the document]
Elkasitu: Not an issue since rejected flaws have the REJECTED workflow and embargoed flaws have the EMBARGOED workflow, so the workflow_name != "DEFAULT" check already covers both

PR #1415, opened 2026-08-07
Fix migrations

Elkasitu:


osidb/migrations/0260_cleanup_stale_alerts.py
```diff
@@ -108,32 +107,54 @@
...
+            WHERE name NOT IN %s
+            """,
+            [tuple(VALID_ALERT_NAMES)]
+        )
+        stale_count = cursor.fetchone()[0]
+
+        if stale_count == 0:
+            logger.info("No stale alerts found")
+            return
+
```
AlejandroMinaya, line 138: What is the benefit of performing this previous check? If the DELETE doesn't match any rows, then the `RETURNING name` would be of length zero, no?

osidb/models/flaw/label.py
AlejandroMinaya, line 1: My only concern with this redefinition is having to remember to override the field in new subclasses or divergence. What do you think about a mixin or a FlawLabel abstract subclass to inherit from?

AlejandroMinaya (approved):
LGTM just left some questions

osidb/models/flaw/label.py
Elkasitu: This is only needed for existing models/tables -- new subclasses wouldn't need the manual override as they'd have a `flawlabel_ptr` field pointing to a `flawlabel_ptr_id` column. This is documented in the relevant commit message.

osidb/migrations/0260_cleanup_stale_alerts.py
Elkasitu: The only benefit is an early return

PR #1418, opened 2026-08-10
OSIDB-5189: fix workflow task creation latency (queue isolation + schedule dedup)

AlejandroMinaya:
## Summary

Celery-backed workflow task creation (Jira task on the OSIM Jira project) was taking 10-30 minutes instead of the required <1 minute SLA. Verified the actual root causes against real evidence (not just the ticket's suggested fix) and addressed the two confirmed ones:

1. **Head-of-line blocking on a shared queue.** Interactive, per-flaw tasks (Jira task sync/transition, BZ sync) shared the single `default` Celery queue with long-running periodic/bulk collector imports (Bugzilla, Jira, CVE.org, NVD, OSV, product-definitions, ...). A bulk import could occupy every worker slot, delaying interactive tasks behind it even though the interactive work itself only takes a couple of seconds.
2. **Redundant/duplicate task scheduling.** `JiraTaskSyncManager` didn't dedup like its siblings (`JiraTaskTransitionManager`), so a burst of saves on the same flaw could enqueue several redundant sync tasks, adding to the queue congestion. The dedup logic itself then had its own bug (a stale timestamp corrupting the "already scheduled" check) and a check-then-act race under real concurrency, both fixed here.

## Changes

- Route periodic/bulk collector tasks to a new `collectors` Celery queue, backed by a dedicated worker, so they can no longer starve interactive syncs on `default`.
- `JiraTaskSyncManager` now uses `SyncManagerMode.EXCLUSIVE` + the same `LockableTaskWithArgs` lock already used by `JiraTaskTransitionManager`, collapsing redundant schedules for the same flaw into a single reschedule instead of duplicate tasks.
- Fixed `SyncManager.schedule()`'s dedup check, which only worked for the first duplicate call in a burst (a stale `last_scheduled_dt` write corrupted the `is_scheduled()` check for subsequent calls).
- Serialized `schedule()`'s check-then-act sequence with `select_for_update()` so two truly concurrent calls for the same sync_id can't both pass the checks and both enqueue.

## Tests

- `osidb/tests/test_celery_routing.py`: routing-table tests (no broker needed) proving collector tasks route to `collectors` and interactive sync tasks stay on `default`.
- `osidb/tests/test_sync_manager.py`: reschedule/dedup tests for `JiraTaskSyncManager`, including a third-schedule-call regression test and a `TransactionTestCase` with two real threads proving the `select_for_update()` locking actually serializes concurrent callers.

## Not included (kept out of scope, flagged for follow-up)

- Bumping worker count / concurrency (the ticket's suggested fix) — treats the symptom, not the two root causes above; can still be layered on top if needed.
- Removing the redundant per-save `check_for_reschedules()` calls in `Flaw` (duplicates the 10-min periodic beat safety net).
- Alerting on schedule-to-start latency exceeding the 1-minute SLA.

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.db import models, transaction
 from django.db.models import F, Q
 from django.utils import timezone
+from rhubarb.exceptions import ConcurrentExecutionException
 from rhubarb.tasks import LockableTaskWithArgs
 
 from config.celery import app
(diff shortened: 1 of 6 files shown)


AlejandroMinaya:
@coderabbitai review

AlejandroMinaya:
@coderabbitai I've addressed your comments. Please review again

osidb/sync_manager.py
```diff
@@ -760,7 +815,16 @@ def __str__(self):
...
+    @app.task(
+        base=LockableTaskWithArgs,
+        name="sync_manager.jira_task_sync",
+        bind=True,
+        # Must last at least MAX_RUN_LENGTH - the same threshold
+        # is_in_progress()/check_for_reschedules() use to decide a run is
+        # stuck - or the Redis lock could expire while a still-legitimate
+        # run is in flight, letting a concurrent duplicate through.
+        lock_ttl=int(SyncManager.MAX_RUN_LENGTH.total_seconds()),
+    )
```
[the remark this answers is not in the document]
AlejandroMinaya: I've addressed this in the newly pushed commits. Please verify

osoukup (approved):
LGTM

osidb/sync_manager.py
```diff
@@ -741,6 +795,32 @@ def sync_task(task, flaw_id, **kwargs):
...
+    """
+
+    default_retry_delay = 30
+    max_retries = int(SyncManager.MAX_RUN_LENGTH.total_seconds() // 30) + 10
+
+    def before_start(self, task_id, args, kwargs):
+        try:
+            super().before_start(task_id, args, kwargs)
+        except ConcurrentExecutionException as exc:
+            raise self.retry(exc=exc, countdown=self.default_retry_delay)
```
Elkasitu, line 821: consider exponential backoff

Elkasitu (commented):
Before merging this we need a corresponding `osidb-ops` MR that either adds another Deployment with the new queue or adds the new queue to the existing Deployment, otherwise workers won't be assigned to the new queue in non-local environments.

osidb/sync_manager.py
AlejandroMinaya: I like the idea. I think we should add it in the follow-up

PR #1431, opened 2026-08-13
OSIDB-5318: Fix Jira Target Release field being ignored

JakubFrejlach:
This PR fixes Jira field Target Release being ignored and field Target Version being used instead. Hardcoded ID values of Target Relase field got stale and thus OSIDB failed to find it and since some of the Jira projects can have both Target Relase and Target Version available at the same time, at fallback wrongly to Target Version.


osoukup (approved):
LGTM

PR #1436, opened 2026-08-18
OSIDB-5400, OSIDB-5343: Fix ACE false-positive AFFECTED for container PURLs and disjoint OSV ranges

ByteHackr:
## Summary

This PR fixes two distinct ACE bugs that caused false-positive affects.

**OSIDB-5400: Container/OCI PURLs with no extractable version**

- `_osv_version_status()` only extracted version from `purls[0]`, which is empty for container/OCI PURLs. The actual dependency version in `sources[].dependencies[].version` was never checked, leading to `osv_status=no_version`.
- The `NO_VERSION` status fell through to `auto_resolve()` → `AFFECTED/DELEGATED`, creating false-positive trackers for components outside the affected range.
- Adds `_extract_version_from_result()` to extract versions from `sources.dependencies` (filtered by query component name) and `build_nvr` as fallbacks, and treats `NO_VERSION` as `NEW/NOVALUE` instead of `AFFECTED/DELEGATED`.
- Only OCI PURLs skip the PURL version; `generic` and `github` PURLs use their version directly when present.

**OSIDB-5343: Disjoint OSV version ranges silently dropped**

- `osv_entry_to_package_info()` only captured the first introduced/fixed pair per range object. Multiple disjoint intervals within the same advisory (e.g. jackson-core across 2.15.x, 2.19.x, and 3.x) were lost.
- `match_component_to_upstream()` only returned the first matching entry, discarding ranges from other `affected[]` blocks for the same component.
- Introduces `_RangeGroup` dataclass and collects all introduced/fixed pairs (including open-ended introduced-only entries with no fix yet). `affected_range()` emits `||`-separated OR groups that `parse_version_range_or()` can parse.
- Merges range groups across multiple matching upstream_purls entries into a single `OsvPackageInfo`.

### DIFF
--- a/apps/ace/tasks.py
+++ b/apps/ace/tasks.py
@@ -446,7 +446,7 @@ def _handle_go_stdlib(
         "marked_notaffected": 0,
     }
 
-    def _run_phase(phase_description, results):
+    def _run_phase(phase_description, results, query_name=""):
         stats = _sync_affects_from_results(
             flaw,
             results,
@@ -454,6 +454,7 @@ def _handle_go_stdlib(
             ps_modules,
             ecosystem="golang",
             upstream_purls=upstream_purls,
+            resolved_name=query_name or component,
         )
         for key in stats:
             totals[key] += stats[key]
@@ -472,7 +473,7 @@ def _handle_go_stdlib(
             builds_only=True,
             no_community=True,
         )
-        _run_phase("Phase 1 (golang builds)", results)
+        _run_phase("Phase 1 (golang builds)", results, query_name="golang")
     except Exception as exc:
         logger.warning("Go stdlib Phase 1 failed for flaw=%s: %s", flaw.uuid, exc)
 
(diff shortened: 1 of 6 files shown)


apps/ace/tasks.py
```diff
@@ -720,6 +766,9 @@ def _sync_affects_from_results(
                     NotAffectedJustification.VULN_CODE_NOT_PRESENT
                 )
                 marked_notaffected += 1
+            elif osv_status is OsvStatus.NO_VERSION:
+                affect.affectedness = Affect.AffectAffectedness.NEW
+                affect.resolution = Affect.AffectResolution.NOVALUE
```
Elkasitu, line 789: This means that the automation cannot progress, as one of the requirements for coming out of the `ANALYSIS` step is to have all affects resolved.

We might need to add the `manual-triage` label to the flaw so that analysts can pick up on it and resolve remaining affects so that the flaw can progress

Elkasitu (commented):
Thanks for the contribution! I believe this also fixes OSIDB-5343 so could you include that in the `CHANGELOG.md`?

Some of the tests and checks are failing, could you fix those? Once that's done we can merge.

Elkasitu:
@coderabbitai full review

ByteHackr:
This PR addresses two distinct ACE bugs. OSIDB-5400 fixes false-positive AFFECTED/DELEGATED affects for container/OCI PURLs where no package version could be extracted (osv_status=no_version) — ACE now tries sources.dependencies and build_nvr as fallback version sources, and treats genuinely versionless results as NEW/NOVALUE instead of auto-resolving them. OSIDB-5343 fixes a separate issue where ACE silently dropped disjoint OSV version ranges (e.g. multiple introduced/fixed pairs within one advisory), causing components inside a later vulnerable range to be incorrectly marked NOTAFFECTED — ACE now collects all range groups and joins them with || OR logic for correct evaluation.

Elkasitu:
@coderabbitai full review

apps/ace/tasks.py
osoukup: No problem to add such affects instead of the implicit (skipped) analysis but I agree that then we should move the flaw to the manual workflow as no affects-resolving automation exists yet.

Elkasitu:
@coderabbitai full review

AlejandroMinaya:
@coderabbitai full review

Elkasitu:
@coderabbitai full review

Elkasitu (approved):
LGTM, could you please squash your commits? Thank you! Ah, also, you need to GPG-sign your commits otherwise merging is blocked.

PR #1437, opened 2026-08-18
Cra regulatory reporting prod 5.16.3 [TEST: DO NOT MERGE]

roduran-dev:
Just to see how this Branch responds to master


.github/workflows/meta.yml
```diff
@@ -4,6 +4,7 @@ on:
     types: [opened, synchronize, reopened, labeled, unlabeled]
 jobs:
   openapi:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'breaking-openapi') }}
```
roduran-dev, line 7: This was me to avoid having an error in every difference between master, RCA and the PR with openapi

scripts/schema-check.sh
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
-make check-venv-active
-./manage.py spectacular --file openapi.yml --settings=config.settings_local &> /dev/null && git diff --quiet openapi.yml &> /dev/null
+if [[ -z "${TOX_ENV_NAME}" ]]; then
```
roduran-dev, line 2: Need to check this change.

.coderabbit.yaml
```diff
@@ -42,6 +42,7 @@ reviews:
       - "main"
       - "master"
       - "release-\\d+\\.\\d+\\.\\d+(-prep)?$"
+      - "CRA_regulatory_reporting" # temporary branch for CRA regulatory reporting [DELETE before merging to master]
```
roduran-dev, line 45: Added by me to allow coderabbit automatically, this will need to be remove before merging to master

PR #1439, opened 2026-08-19
fix: Update label filter to support FlawLabelsV2

AlejandroMinaya:
## Summary

Fixes broken DjangoQL filters for the `labels` field and its FlawLabel-subclass properties:

- `labels` is a `StrField`, so DjangoQL's UI/schema legally offers `~`, `!~`, `startswith`, `endswith` and their negated variants in addition to `=`/`!=`/`in`/`not in`. `FlawLabelsField.get_lookup()` only handled the latter four and silently returned `None` for the rest, causing an unhandled `TypeError` (500) instead of filtering.
- Fields that only exist on `FlawLabel` subclasses (`contributor`, `state`, `relevant`) were completely unreachable via DjangoQL (`Unknown field: contributor`), since the `labels` relation was severed from the schema graph in favor of the flat name-matching field.

## Changes

- `FlawLabelsField.get_lookup()` now routes all string operators through the shared `get_operator()` mapping instead of only handling `=`/`!=` explicitly.
- Added `FlawQLSchema.resolve_name()` override to allow dotted access to `labels.contributor` / `labels.state` / `labels.relevant`, without reopening the full `FlawLabel` relation graph (`labels.name` stays unreachable — `labels` alone already matches by name).
- New `FlawLabelPropertyField` mixin + concrete fields build an OR'd query across the `FlawLabel` subclasses that actually define each property, derived from `FlawLabel.TYPE_TO_MODEL` so new/changed label types stay in sync automatically. Fails loudly (`ValueError`) at import time if a field name matches no subclass.
- Tests: `test_search_flaws_by_labels_query_operators`, `test_search_flaws_by_label_properties`.

## Test plan

```
podman exec -it testrunner tox -e tests -- osidb/tests/test_search.py::TestSearch
```

### DIFF
--- a/osidb/djangoql.py
+++ b/osidb/djangoql.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.db.models import Q
 from django.utils import timezone
+from djangoql.exceptions import DjangoQLSchemaError
 from djangoql.schema import (
     BoolField,
     DateTimeField,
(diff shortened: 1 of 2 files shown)


osidb/djangoql.py
```diff
@@ -226,22 +239,91 @@ def get_lookup(self, path, operator, value):
...
         """
 
-        if operator == "=":
-            return Q(labels__name=value)
-        elif operator == "!=":
-            return ~Q(labels__name=value)
+        if operator in ("in", "not in"):
+            # Dedupe repeated operands (e.g. labels in ("a", "a")), otherwise
+            # num_labels overcounts and the AND semantics below never match.
+            unique_values = list(dict.fromkeys(value))
```
JakubFrejlach: I would use rather `set` as order preservation is not needed in this case:
```python
unique_values = list(set(value))
```
for readability and speed (negligible but still 😄 )

JakubFrejlach (approved):
Small nitpick, otherwise LGTM!

osoukup (commented):
The approach of `labels = NAME` instead of `labels.name = NAME` is just a technical decision (eg. backwards compatibility) or is there some limitation? Also searching for `labels.type` seems to be not possible (not sure if very important), is it intentional?

AlejandroMinaya:
> The approach of `labels = NAME` instead of `labels.name = NAME` is just a technical decision (eg. backwards compatibility) or is there some limitation?

The motivation is backwards compatibility but I understand that `.name` should still be valid, good catch.

>Also searching for `labels.type` seems to be not possible (not sure if very important), is it intentional?

Not left out intentionally. Will add it to this PR if it is straightforward enough

osoukup (approved):
LGTM

PR #1442, opened 2026-08-20
Cra regulatory reporting [Diff verification]

roduran-dev:


costaconrado:
@coderabbitai please dismiss all your comments in this PR, everything in this branch has been already reviewed by you, approved by at least one dev and merged into this branch.

config/settings.py
```diff
@@ -64,6 +64,7 @@
     # --- Foundation ---
     "osidb",
     # --- Business logic (no cross-app deps beyond osidb) ---
+    "regulatory_reporting",
```
Elkasitu: why does this not follow the same pattern as all other apps? why not apps.regulatory_reporting?

config/settings.py
```diff
@@ -332,6 +343,14 @@
...
 )
 
+# opt-in-flag for CRA upstream notification
+CRA_NOTIFICATIONS_ENABLED = get_env(
+    "CRA_NOTIFICATIONS_ENABLED", default="False", is_bool=True
+)
+
+# opt-in-flag for CRA upstream reporting
+CRA_REPORTING_ENABLED = get_env("CRA_REPORTING_ENABLED", default="False", is_bool=True)
+
```
Elkasitu: CRA should define a `pydantic_settings` class in its own directory and use that instead, there's no reason to have these defined here

config/settings.py
```diff
@@ -351,7 +370,7 @@
 PGHISTORY_BASE_MODEL = "osidb.models.audit_history.CustomHistoryBase"
 PGHISTORY_OBJ_FIELD = ObjForeignKey(related_name="events")
 
-
+UPSTREAM_NOTIFICATIONS_SENDER = "secalert@redhat.com"
```
Elkasitu: same as above, but additionally, is it OK for us to "impersonate" secalert?

.github/workflows/meta.yml
```diff
@@ -4,6 +4,7 @@ on:
     types: [opened, synchronize, reopened, labeled, unlabeled]
 jobs:
   openapi:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'breaking-openapi') }}
```
Elkasitu, line 7: ?

scripts/schema-check.sh
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
-make check-venv-active
-./manage.py spectacular --file openapi.yml --settings=config.settings_local &> /dev/null && git diff --quiet openapi.yml &> /dev/null
+if [[ -z "${TOX_ENV_NAME}" ]]; then
+    make check-venv-active
+fi
+./manage.py spectacular --file openapi.yml --settings=config.settings_local &> /dev/null && git diff --quiet openapi.yml &> /dev/null
```
Elkasitu, line 5: ?

.coderabbit.yaml
```diff
@@ -42,6 +42,7 @@ reviews:
       - "main"
       - "master"
       - "release-\\d+\\.\\d+\\.\\d+(-prep)?$"
+      - "CRA_regulatory_reporting" # temporary branch for CRA regulatory reporting [DELETE before merging to master]
```
Elkasitu: the comment is right!

tox.ini
```diff
@@ -70,6 +71,9 @@ allowlist_externals = /usr/bin/bash
 commands = /usr/bin/bash -c './scripts/migrations-check.sh'
 
 [testenv:schema-check]
+setenv =
+    {[testenv]setenv}
+    CRA_NOTIFICATIONS_ENABLED = true
```
Elkasitu: if you're just copying from [testenv]setenv then you don't need CRA_NOTIFICATIONS_ENABLED

regulatory_reporting/middleware.py
```diff
@@ -0,0 +1,41 @@
...
+    - /regulatory-reporting/api/v1/notifications/... is gated by CRA_NOTIFICATIONS_ENABLED
+    - /regulatory-reporting/api/v1/... CRA routes (SRP reports/milestones) are
+      gated by CRA_REPORTING_ENABLED
+
+    Flags are checked here, per request, rather than used to conditionally
+    register routes in urlpatterns at import time. Django caches the compiled
+    URLconf, so a setting read once at import time cannot be toggled at
+    runtime (e.g. by tests overriding settings) without manually clearing that
+    cache. Checking in middleware keeps urlpatterns static while still fully
+    disabling the API when the feature flag is off.
```
Elkasitu: This is an awful use of middleware, we are checking every single request just to disable/enable CRA?

Either register the route dynamically and accept the trade-off that changing the env var will require a new OSIDB rollout (perfectly acceptable) or do it at the view level with a mixin or base class that every CRA inherits from.

Also, 404 is misleading, a more accurate http error code might be 423 and by using a view mixin / baseclass you can return your own payload e.g. "CRA disabled"

Elkasitu (commented):
Maybe we should align on naming for this module/project/etc. seems like CRA and regulatory reporting are interchangeable.

docs/CHANGELOG.md
```diff
@@ -14,6 +14,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add report payloads in milestones (OSIDB-5070)
+- Add manual POST create for SRP reports with PRE_REQUIRED status and evidence field (OSIDB-5284)
+- Add owner and submitted_at fields to SRP report milestones (OSIDB-5439)
+
+### Changed
+- Remove additional information from report type (OSIDB-5437)
+- Make reference id and url optional (OSIDB-5436)
+- Remove Flaw signal to create report (OSIDB-5462)
+- Change srp report status choices (OSIDB-5442)
+- Change srp milestones status choices (OSIDB-5441)
```
JakubFrejlach: `Changed` section already exists

regulatory_reporting/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
JakubFrejlach: I suppose this file can be dropped? We don't really use django admin, right?

regulatory_reporting/filters.py
```diff
@@ -0,0 +1,153 @@
...
+            "milestone_type",
+            "status",
+            "created_dt__gte",
+            "created_dt__lte",
+            "submitted_at__gte",
+            "submitted_at__lte",
+            "owner",
+            "request_source",
+            "request_text",
+        ]
```
JakubFrejlach: If a filter is already defined as an attribute of the filterset class, listing it here won't do anything. Fields listed here are have autocreated filters based on the field type, if no filter is defined via attribute. Also this should not really include stuff like `__gte`, `__lte`, see how it is done in OSIDB filters.

This applies to all FilterSet classes in this file.

scripts/schema-check.sh
costaconrado: This PR is the result of contributions from multiple people over the past couple of months, so I don't have enough context to infer what the `?` is referring to here. Could you please add a specific question or indicate what change you'd like to see? 
That would make it much easier for me to address the concern.

.github/workflows/meta.yml
costaconrado: This PR is the result of contributions from multiple people over the past couple of months, so I don't have enough context to infer what the ? is referring to here. Could you please add a specific question or indicate what change you'd like to see?
That would make it much easier for me to address the concern.

config/settings.py
costaconrado: Good catch, I'm moving it to apps.regulatory_reporting

config/settings.py
costaconrado: I'm adding an ENV to set the appropriated email when defined, I believe this was used as placeholder with a suggested comment in a meeting and never replaced with the actual decision.

docs/CHANGELOG.md
costaconrado: I'm submitting a fix, thanks

.coderabbit.yaml
costaconrado: I'm removing it

regulatory_reporting/admin.py
costaconrado: You're correct, I'm removing it

costaconrado (approved):

costaconrado:
@coderabbitai please dismiss all your comments in this PR, everything in this branch has been already reviewed by you, approved by at least one dev and merged into this branch.

.github/workflows/meta.yml
Elkasitu: I mean why does this exist at all when the openapi check is not mandatory

scripts/schema-check.sh
Elkasitu: I mean why was this changed at all

Elkasitu (approved):

PR #1443, opened 2026-08-21
fix: Prevent duplicate/dropped Jira task sync schedules

AlejandroMinaya:
## Summary

Fixes the OSIDB-5189-revert bug class: a burst of flaw saves in quick succession could each call `SyncManager.schedule()` for the same flaw and end up enqueuing multiple duplicate Jira sync tasks, and a schedule after a failed (not just finished) run could be silently dropped instead of enqueued.

- `is_scheduled()` only excluded schedules that predated the last *success* (`last_finished_dt`); it now also excludes ones predating the last *failure* (`last_failed_dt`), so a `schedule()` call after a failed run actually enqueues instead of being read as "still scheduled".
- `schedule()`/`reschedule()` now take the `SyncManager` row lock (`select_for_update()` inside `transaction.atomic()`, bounded by a new `LOCK_WAIT_TIMEOUT` so a stuck holder fails fast instead of blocking forever) and `schedule()`'s EXCLUSIVE-mode path now explicitly checks `is_scheduled()` (skip) before `is_in_progress()` (defer) — previously a call arriving while a run was scheduled-but-not-yet-started fell through and enqueued a redundant task. The actual enqueue is extracted into `_enqueue()`, shared by a fresh schedule and by `reschedule()`. `JiraTaskSyncManager` switches to `EXCLUSIVE` mode to get this dedup behavior.
- `JiraTaskSyncManager.sync_task` runs under `LockableTaskWithArgs` (mirroring `JiraTaskTransitionManager`) so two workers can't run it concurrently for the same flaw; contention is dropped (`Reject`, no requeue) rather than retried, since Celery-level retries previously re-enqueued into the `fifo.*` queues faster than the concurrency-1 workers could drain them. `check_for_reschedules()`'s `scheduled_not_started` bucket now recovers EXCLUSIVE-mode managers on the shorter `MAX_RUN_LENGTH` window instead of `MAX_SCHEDULE_DELAY`, since a lock-rejected duplicate can bump `last_scheduled_dt` past `last_started_dt` while the original run is genuinely stuck.

## Test plan

```
podman exec -it testrunner tox -e tests -- osidb/tests/test_sync_manager.py
```

🤖 Assisted-by: Claude Sonnet 5

### DIFF
--- a/osidb/sync_manager.py
+++ b/osidb/sync_manager.py
@@ -197,6 +197,10 @@ class SyncManager(models.Model):
         Celery queue, bypassing the EXCLUSIVE-mode dedup checks in
         schedule(). Used both for genuinely new schedules and by
         reschedule(), which has already decided a new run is warranted.
+
+        Requires a SyncManager row for (cls.__name__, sync_id) to already
+        exist - the update()/get() below don't create one. Callers must
+        get_or_create() it first, as schedule() and reschedule() do.
         """
         if schedule_options is None:
             schedule_options = {}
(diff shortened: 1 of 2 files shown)


osidb/sync_manager.py
```diff
@@ -151,27 +193,73 @@ def schedule(cls, sync_id, *args, schedule_options=None, **kwargs):
...
+        # Flaw.tasksync()'s transaction.atomic(), and ATOMIC_REQUESTS wraps
+        # the whole request in an outer transaction too, so the lock stays
+        # held (and other schedule() calls for the same sync_id stay
+        # blocked) until that outermost transaction commits, well after
+        # _enqueue()/reschedule() return.
+        with transaction.atomic():
+            with cls._lock_wait_timeout():
+                _, created = SyncManager.objects.select_for_update().get_or_create(
+                    name=cls.__name__, sync_id=sync_id
                 )
```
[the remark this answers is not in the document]
AlejandroMinaya: This is intentional since we don't want errors to be obscured or masked. There are tests asserting the propagation of this OperationalError to ensure the dedup guarantee

osidb/sync_manager.py
```diff
@@ -152,27 +193,73 @@ def schedule(cls, sync_id, *args, schedule_options=None, **kwargs):
...
+        # same sync_id: without it, two callers could both pass the
+        # is_scheduled()/is_in_progress() checks below before either's
+        # write lands, and both end up enqueuing a duplicate task. The row
+        # lock is released on commit of the outermost transaction, not on
+        # return from this atomic() block: this method can run inside
+        # Flaw.tasksync()'s transaction.atomic(), and ATOMIC_REQUESTS wraps
+        # the whole request in an outer transaction too, so the lock stays
+        # held (and other schedule() calls for the same sync_id stay
+        # blocked) until that outermost transaction commits, well after
+        # _enqueue()/reschedule() return.
```
Elkasitu: I am not sure we need this optimization (the whole commit), some context:

Once upon a time, we had two-way synchronization between OSIDB Flaws and Jira tasks... this proved to be a PITA because there was no way for us to control the order of execution (so you could have a Jira->OSIDB write before an OSIDB->Jira write which would overwrite fresh data with stale data) nor guarantee execution of a given task only once at a time (so you could have two workers performing the same action at the same time).

Queue `rhubarb`, `EXCLUSIVE` mode and `FIFO` queues:
* `rhubarb` was the original solution to guarantee only a single instance of a celery task runs at any given time
* SyncManagers were introduced to solve the problem of two-way synchronization and `EXCLUSIVE` was a way to mimic the `rhubarb` behavior however it cannot guarantee that two tasks won't run at once, only that they can't be scheduled to run at the same time, but time has proven that this is not enough.
* FIFO queues were introduced to solve the out-of-order problem once and for all, all fetch/updates of a given task id (or tracker id) go to the same sequential queue, meaning if we schedule first an update and then a fetch the worker will first perform the update and then the fetch, no race condition possible.

After a while however we got rid of task two-way sync, as there was no reason for it, this leads me to believe that the correct approach would be to instead get rid of the existing SyncManagers and simply have a `LockableTaskWithArgs` that replaces both JiraTask SyncManagers.

This should get rid of duplicate tasks due to `schedule` and avoids a database lock which can affect OSIDB's day-to-day operational performance, whereas a distributed lock not so much.

WDYT?

Elkasitu (commented):
I'm OK with the first two commits, not so much with the last two for the reasons already explained in one of the review comments, I think we should in general move away from SyncManagers, they're already pretty complex and adding more complexity doesn't help... historically every time we've added complexity to it it's become less understandable and problems still persist.

osidb/sync_manager.py
AlejandroMinaya: I totally see what you mean. Thank you for the context. Locking the DB carries more service degradation risk than just having a duplicate task that gets immediately rejected. I will remove the lock logic in a fix-up commit

Elkasitu (approved):
LGTM, thank you!

PR #1445, opened 2026-08-24
fix: Skip over closed trackers instead of exiting (OSIDB-5435)

AlejandroMinaya:
`FlawSerializer.update_trackers` used `return` instead of `continue` when hitting a closed, non-unembargoing tracker, so the loop bailed out and left any trackers after it in iteration order unupdated (e.g. impact changes never reached them).

- Fix: `return` → `continue`.
- Added regression test: `test_closed_tracker_does_not_skip_later_open_trackers`.

🤖 Assisted-by: Claude Code


JakubFrejlach (approved):
LGTM! Nice catch!

AlejandroMinaya:
@coderabbitai review

PR #1448, opened 2026-08-24
implemented relation history events

costaconrado:
This PR adds relation-aware audit history for flaw audit views.

When `/audit` is queried with `include_relation_events=true&pgh_obj_model=osidb.Flaw&pgh_obj_id=<flaw_uuid>`, the response now includes related native audit rows for the flaw context:

- `AffectAudit` rows whose historical `flaw_id` matches the requested flaw, including deleted affects.
- `TrackerAudit` rows for trackers discovered through matching historical `AffectAudit.tracker_id` values.

This does not add a derived `RelationHistoryEvent` table. Related entries are returned as normal concrete audit events with their existing slugs, for example:

- `osidb.AffectAudit:<pgh_id>`
- `osidb.TrackerAudit:<pgh_id>`

The existing audit endpoint handles filtering, sorting, pagination, counts, and direct slug retrieval for these related audit rows.

ACL behavior was updated so audit history follows visibility changes. The ACL propagation is generic in `ACLMixin`: it discovers related audited ACL models through Django relation metadata and updates direct and one-hop related audit histories. Secondary targets, such as trackers, are only released when they are not still connected to embargoed ACL-managed rows, preventing mixed-visibility tracker history leaks.

The PR also adds database indexes for relation-aware audit lookups and regression tests for deleted affects, related audit slug retrieval, and mixed public/embargoed tracker visibility.

For OSIM, a follow-up task can add UI affordances around affect or tracker rows in the flaw detail page to surface this related history.

Closes OSIDB-4999.

### DIFF
--- a/osidb/api_views.py
+++ b/osidb/api_views.py
@@ -2396,8 +2396,9 @@ class AuditView(RudimentaryUserPathLoggingMixin, ReadOnlyModelViewSet):
     def list(self, request, *args, **kwargs):
         limit = self._get_limit(request)
         offset = self._get_offset(request)
-        count = self._audit_count(request.query_params)
-        rows = self._audit_rows(request.query_params, limit, offset)
+        related_context = self._related_history_context(request.query_params)
+        count = self._audit_count(request.query_params, related_context)
+        rows = self._audit_rows(request.query_params, limit, offset, related_context)
         previous_rows = self._previous_rows_for_rows(rows)
         events = [
             self._event_from_row(
@@ -2418,7 +2419,7 @@ class AuditView(RudimentaryUserPathLoggingMixin, ReadOnlyModelViewSet):
         )
 
     def retrieve(self, request, *args, **kwargs):
-        event = self._event_by_slug(kwargs[self.lookup_field])
+        event = self._event_by_slug(kwargs[self.lookup_field], request.query_params)
         if event is None:
             return Response(status=status.HTTP_404_NOT_FOUND)
         return Response(event)
@@ -2457,7 +2458,7 @@ class AuditView(RudimentaryUserPathLoggingMixin, ReadOnlyModelViewSet):
             return remove_query_param(url, "offset")
         return replace_query_param(url, "offset", previous_offset)
 
-    def _audit_count(self, params):
+    def _audit_count(self, params, related_context=None):
         pgh_slug = params.get("pgh_slug")
         if pgh_slug:
             return 1 if self._row_by_slug(pgh_slug, params) else 0
@@ -2472,10 +2473,13 @@ class AuditView(RudimentaryUserPathLoggingMixin, ReadOnlyModelViewSet):
             # Exact counts on unfiltered audit feeds are expensive on large audit tables,
             # but are preserved for API compatibility with the existing /audit contract.
             count += self._audit_queryset(audit_table, params).count()
+        if self._include_relation_events(params):
+            count += self._related_history_count(params, related_context)
         return count
 
-    def _audit_rows(self, params, limit, offset):
+    def _audit_rows(self, params, limit, offset, related_context=None):
         rows = []
+        window = limit + offset
         pgh_slug = params.get("pgh_slug")
         if pgh_slug:
             if not limit or offset:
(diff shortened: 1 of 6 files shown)


costaconrado:
@coderabbitai full review again please.

costaconrado:
@coderabbitai I believe I've addressed your actionable items, give it a second look and close the comments that you consider addressed.

costaconrado:
@coderabbitai I've checked this against the current RLS behavior and I don't think this is a relation-history regression as described. 
`Affect` rows are not practically deletable through the application with write-only ACLs because the row must first be visible/readable to load it.  I also tested a raw `DELETE FROM osidb_affect WHERE uuid = ...` under only `PUBLIC_WRITE_GROUPS`; PostgreSQL affected `0` rows because the row is not visible under the table's RLS select/update/delete visibility behavior.

Also, the existing `osidb_affectaudit` insert policy already requires both `acl_read` and `acl_write` overlap:
```
WITH CHECK (
  acl_read::uuid[] && string_to_array(current_setting('osidb.acl'), ',')::uuid[]
  AND acl_write::uuid[] && string_to_array(current_setting('osidb.acl'), ',')::uuid[]
)
```

Also, RelationHistoryEvent intentionally mirrors the existing audit-table RLS model. Relaxing only relation-history inserts to write-only would make it less strict than the source audit table and could diverge from the current audit authorization semantics.

costaconrado:
@coderabbitai I've completely re-structured the solution. Close all comments before last push and give me a full review

costaconrado:
@coderabbitai review please

osoukup (approved):
LGTM

osidb/api_views.py
```diff
@@ -2628,6 +2637,92 @@ def _previous_rows_for_rows(self, rows):
 
         return previous_rows
 
+    def _include_relation_events(self, params):
+        value = params.get("include_relation_events")
+        return isinstance(value, str) and value.lower() in {"1", "true", "yes"}
```
Elkasitu, line 2642: I think that like `include_history`, this should return `True` simply if the parameter is included (no value needed): e.g. `GET /audit?pgh_obj_id=<uuid>&include_relation_events`

osidb/migrations/0267_relation_audit_indexes.py
```diff
@@ -0,0 +1,63 @@
...
+        ("osidb", "0266_manual_triage_kernel_flaws"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="""
+                CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_affectaudit_flaw_history
+                ON osidb_affectaudit (flaw_id, pgh_created_at DESC, pgh_id DESC)
+                WHERE flaw_id IS NOT NULL
+            """,
```
Elkasitu, line 19: are any of these really necessary? audit tables inherit indexes from the source tables so e.g. flaw_id has an index

Elkasitu (commented):
I'm not blocking the PR but I think there's a better approach.

I have concerns about the performance of just dumping everything in the same API request, stitching and sorting multiple rows together, etc.

I would instead suggest that the client perform multiple queries and do the stitching client-side if necessary:

`GET /audit?pgh_obj_model=osidb.Affect&flaw_id=<uuid>`
or
`GET /audit?pgh_obj_model=osidb.Affect&related_pgh_obj_id=<target_flaw_pgh_label>`

Then the client would query:

`GET /audit?pgh_obj_model=osidb.Flaw&pgh_obj_id=<flaw_uuid>`
`GET /audit?pgh_obj_model=osidb.Affect&related_pgh_obj_id=<flaw_uuid>`
`GET /audit?pgh_obj_model=osidb.Tracker&related_pgh_obj_id=<flaw_uuid>`

In parallel, each with their own pagination, returning predictable payloads, etc.

We also should measure performance to avoid having a regression.

edit: Maybe we should reconsider using `Events.objects.references()` but with [`across()`](https://django-pghistory.readthedocs.io/en/3.9.2/aggregating_events/?h=across#filtering-event-models-using-objectsacross)

PR #1449, opened 2026-08-24
feat: provide insights into affectedness/resolution (OSIDB-5345)

Elkasitu:


osidb/signals.py
```diff
@@ -237,6 +237,57 @@ def remove_not_affected_justification(sender, instance, **kwargs):
...
+
+    # If status changed and explanation was NOT manually edited, clear it
+    if (
+        old.affectedness != instance.affectedness
+        or old.resolution != instance.resolution
+    ):
+        # Check if human manually changed the explanation
+        if old.affectedness_explanation == instance.affectedness_explanation:
+            # Explanation unchanged, so clear it since status changed
+            instance.affectedness_explanation = ""
```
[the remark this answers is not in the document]
Elkasitu: It is not currently used, and passing it directly to a `save()` should be assumed to not trigger signals correctly

Elkasitu:
@coderabbitai full review

osidb/signals.py
```diff
@@ -237,6 +237,57 @@ def remove_not_affected_justification(sender, instance, **kwargs):
...
+
+    if instance._state.adding:
+        # New instance not from auto_resolve; clear any explanation (should already be empty).
+        instance.affectedness_explanation = ""
+        return
+
+    try:
+        old = Affect.objects.get(uuid=instance.uuid)
+    except Affect.DoesNotExist:
+        return
```
AlejandroMinaya, line 265: Nit: I would rather use `.filter().first()` instead of using the try-except for control. I understand the EAFP principle but if this is not an exceptional situation I would prefer we don't handle it as such

osidb/signals.py
```diff
@@ -237,6 +237,57 @@ def remove_not_affected_justification(sender, instance, **kwargs):
...
+    if (
+        old.affectedness != instance.affectedness
+        or old.resolution != instance.resolution
+    ):
+        # Check if human manually changed the explanation
+        if old.affectedness_explanation == instance.affectedness_explanation:
+            # Explanation unchanged, so clear it since status changed
+            instance.affectedness_explanation = ""
+        # else: human edited the explanation, preserve it
+
```
AlejandroMinaya: Nit: I am big on guards, understanding the tradeoff of multiple exit points for less nesting/indentation. In this case I can see an early return such as:
```suggestion
    # If status didn't change exit
    if (
        old.affectedness == instance.affectedness
        and old.resolution == instance.resolution
    ):
        return
        
    # Check if human manually changed the explanation
    if old.affectedness_explanation == instance.affectedness_explanation:
            # Explanation unchanged, so clear it since status changed
            instance.affectedness_explanation = ""
    # else: human edited the explanation, preserve it

```
What do you generally think about this approach?

osidb/signals.py
```diff
@@ -237,6 +237,57 @@ def remove_not_affected_justification(sender, instance, **kwargs):
...
+def clear_auto_resolved_flag(sender, instance, **kwargs):
+    """
+    Clear the _auto_resolved flag after save completes.
+
+    This ensures that subsequent manual changes to affectedness/resolution
+    will properly clear stale explanations, even after an auto_resolve() + save().
+    """
+    if hasattr(instance, "_auto_resolved"):
+        delattr(instance, "_auto_resolved")
+
```
AlejandroMinaya, line 291: What is the benefit of deleting the attribute vs setting it to `False`? Is the presence of this attribute, even when false, an indicator of a certain state for the affect?

osidb/signals.py
Elkasitu: Good one, applied, thanks!

osidb/signals.py
Elkasitu: Functionally speaking there is no benefit, but it is a transient state only, we only care / check if it's truthy, it should not be used beyond this point.

osidb/signals.py
Elkasitu: I mean, we do have a guard right above which performs an early return in case the affect is in creation which is the only acceptable case for which it would be missing from the database. If the `get()` fails here it does mean something _exceptional_ happened, possibly a race condition or something, and the happy path is probably the most common case.

It's also a pretty common Django idiom, and ultimately I think it's wrong to filter by a single PK... if it were something like Affects with ps_module X then I'd agree.

AlejandroMinaya (approved):
LGTM! Thanks for going over the comments

osidb/models/affect.py
```diff
@@ -910,11 +939,37 @@ def auto_resolve(self, *, flaw_has_high_cvss_score: bool | None = None) -> None:
             )
             if impact == Impact.MODERATE and not has_high_cvss:
                 self.resolution = self.AffectResolution.DEFER
+                self.affectedness_explanation = (
+                    "Impact is MODERATE with CVSS score below 7.0. Resolution set to DEFER as we do not "
```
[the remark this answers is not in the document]
Elkasitu: no, ignore

PR #1457, opened 2026-08-27
fix: copy tracker timestamps from Jira/Bugzilla on create

jvaught-rh:
Avoid leaving the Unix-epoch placeholder that OSIM displays after tracker filing (OSIDB-5468).


jvaught-rh:
@coderabbitai review

jvaught-rh:
Closing PR as I no longer believe this is the best path forward.

PR #1458, opened 2026-08-31
fix: limit is_ubi to latest z-stream and next y-stream (OSIDB-5459)

Elkasitu:
This commit installs a special collation which allows natural order of char fields (e.g. 3.10 > 3.2) which makes finding the latest z-stream and the next y-stream trivial and efficient (done by the database), creates the necessary helpers and applies it to Affect.is_ubi(), as well as some existing helper properties / methods for retrieving z-streams and y-streams

This fix avoids false positives when marking the "ubi" label to affects, as well as bypassing the mod7 check for affects that don't meet the criteria.


osoukup (approved):
LGTM

PR #1465, opened 2026-09-01
feat: Add quantifiers (any/all) for label subfields

AlejandroMinaya:
## Summary

Add support for `labels.any.<field>` and `labels.all.<field>` DjangoQL quantifiers for label subclass properties (`contributor`, `state`, `relevant`). These allow precise filtering over polymorphic label rows that the existing bare 2-part `labels.<field>` syntax cannot express.

## Problem

The bare form `labels.contributor = X` ORs across all matching rows then negates the whole result for `!=`, making it impossible to express "find flaws with at least one unclaimed label" or "find flaws where all labels are in state Y". It always means "no label has this value" rather than "some label lacks this value".

## Solution

- `labels.any.contributor = X`: matches flaws with at least one label row satisfying the condition
- `labels.all.contributor = X`: matches flaws where every qualifying label row satisfies the condition
- Same for `.state` and `.relevant`
- Both return False (not vacuously True) for zero qualifying rows

Built on `Exists()`/`OuterRef()` subqueries across the polymorphic label subclass tables (BULabel, CollaboratorLabel, etc.).

## Test Coverage

Added 10 new test cases covering:
- Parity with bare form for positive operators
- ALL with positive/negative operators
- Zero qualifying rows behavior
- Polymorphic subclass spread (multiple label types defining the same property)
- `in`/`not in` operators
- State and relevant field types
- Validation of invalid quantifiers/fields
- Original regression case (unclaimed labels)

Changelog updated for Unreleased/Added section.

🤖 Assisted-by: Claude Haiku/Sonnet/Opus

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - make sure that the flaw impact is always at or above any affect override (OSIDB-5389)
 - Provide insights into chosen affectedness/resolution by automation (OSIDB-5345)
+- Support `labels.any.<field>` / `labels.all.<field>` DjangoQL quantifiers for
+  `contributor`, `state` and `relevant` label properties (OSIDB-5212)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


osoukup (approved):
Nice! LGTM

PR #1466, opened 2026-09-01
deal with duplicates

osoukup:
...by prioritizing the relevant trackers over duplicates.

OSIDB-5449

### DIFF
--- a/osidb/models/tracker.py
+++ b/osidb/models/tracker.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import uuid
 
 import pghistory
@@ -28,6 +29,11 @@ from .ps_module import PsModule
 from .ps_update_stream import PsUpdateStream
 
 logger = logging.getLogger(__name__)
+
+# resolutions marking a tracker as a duplicate - besides the plain duplicate also
+# the historical migrated resolution which flagged a Bugzilla tracker superseded
+# by its Jira counterpart during the one-time Bugzilla-to-Jira migration
+DUPLICATE_RESOLUTION_RE = re.compile(r"(duplicate|migrated)", re.IGNORECASE)
 
 
 class TrackerManager(TrackingMixinManager):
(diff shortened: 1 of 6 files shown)


osidb/models/tracker.py
```diff
@@ -29,6 +30,11 @@
 
 logger = logging.getLogger(__name__)
 
+# resolutions marking a tracker as a duplicate - besides the plain duplicate also
+# the historical migrated resolution which flagged a Bugzilla tracker superseded
+# by its Jira counterpart during the one-time Bugzilla-to-Jira migration
+DUPLICATE_RESOLUTION_RE = re.compile(r"(duplicate|migrated)", re.IGNORECASE)
```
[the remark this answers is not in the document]
osoukup: I am not sure there is such a resolution as `Duplicated` but that would mean the same. I do not think that this is really an issue.

osidb/models/tracker.py
```diff
@@ -514,6 +520,49 @@ def is_closed(self):
...
+
+        an affect with no tracker linked is always linked even by a duplicate
+        tracker as showing a duplicate is better than showing nothing
+        """
+        target = set(target_affects)
+        current = set(self.affects.all())
+
+        to_link = set()
+        for affect in target:
+            existing = affect.tracker
```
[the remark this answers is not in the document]
osoukup: This would be to heavy change. I think that given the low probability the risk is acceptable. Even if it happened the next sync would simply fix it.

osidb/models/tracker.py
```diff
@@ -514,6 +520,49 @@ def is_closed(self):
...
+                # affect is unlinked or already linked to this tracker
+                to_link.add(affect)
+            elif self.is_duplicate and not existing.is_duplicate:
+                # keep the relevant tracker already linked; do not steal the
+                # affect with this duplicate tracker
+                continue
+            else:
+                # this tracker is relevant while the linked one is a duplicate,
+                # or none/both are - the newly synced tracker wins
+                to_link.add(affect)
```
AlejandroMinaya:
```suggestion
            if self.is_duplicate and not existing.is_duplicate:
                # keep the relevant tracker already linked; do not steal the
                # affect with this duplicate tracker
                continue
           
            # this tracker is relevant while the linked one is a duplicate,
            # or none/both are - the newly synced tracker wins
            to_link.add(affect)
```

AlejandroMinaya (approved):

AlejandroMinaya:
Looks good! 🚀 Just a small suggestion

osidb/models/tracker.py
osoukup: Thank you for the suggestion! I think the branching conditions are probably too complicated but I am not sure we can safely remove the `existing is None` without risking `AttributeError`.

osoukup: I have simplified the conditions.

PR #1469, opened 2026-09-02
OSIDB-5332: Fix flaw component not syncing to Bugzilla

svelamak:
Flaws were getting stuck with the wrong Bugzilla component after leaving the NEW state.
-When a tracker was created with `sync_to_bz=false`, the code skipped the Bugzilla sync entirely including the component field instead of just skipping what `sync_to_bz` was meant to skip. Tested it on staging the BZ component stayed empty after creating a tracker this way.
-`serializers.py` now calls `affect.flaw.save()` for each affect same as the `sync_to_bz=true` path already does.
-In `test_trackers.py` updated `test_tracker_create_jira_bulk_enablement` to check the flaw is always synced now.
Closes: OSIDB - 5332

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fallback to Aegis' suggested ecosystem if not available in OSIDB (OSIDB-5353)
+- Fix Bugzilla flaw bug not in sync with the flaw (OSIDB-5332)
 
 ## [5.17.1] - 2026-08-31
 ### Added
--- a/osidb/serializer.py
+++ b/osidb/serializer.py
@@ -267,12 +267,13 @@ class TrackingMixinSerializer(serializers.ModelSerializer):
 
 class SyncToBzBulkEnablementMixinSerializer(serializers.ModelSerializer):
     """
-    Provides parameter "sync_to_bz" that when set to False disables sync
-    with Bugzilla for the given request. This is to be used by clients
-    during bulk actions that are carried out by repeated single-instance
-    calls (e.g. creating Trackers). The reason is that the final flaw bz
-    sync that happens after e.g. creating a tracker can take minutes for
-    huge flaws.
+    Provides parameter "sync_to_bz" that when set to False skips the
+    tracker-level Bugzilla sync for the given request. For Jira trackers,
+    the related flaws are still synced to Bugzilla automatically as part of the
+    operation, deduplicated per flaw, regardless of this flag.
+    This is to be used by clients during bulk actions that
+    are carried out by repeated single-instance calls (e.g. creating
+    Trackers). Bugzilla trackers do not go through this related-flaw sync path.
 
     The parameter does nothing if BZ sync is not enabled in the OSIDB instance.
     """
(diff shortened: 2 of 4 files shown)


svelamak:
@coderabbitai review

svelamak:
@coderabbitai review

svelamak:
@coderabbitai review

osidb/serializer.py
```diff
@@ -797,18 +811,13 @@ def create(self, validated_data):
...
-                    )
+                self._sync_related_flaws_to_bz(affects)
         else:
             # Special path for bulk tracker create. Works for Jira trackers only.
-            # Do not sync with BZ, as that can take a long time for large flaws.
-            pass
+            # Previously skipped the BZ sync entirely for performance, but that
+            # left the flaw's bz_component stale.
+            if tracker.type == Tracker.TrackerType.JIRA:
+                self._sync_related_flaws_to_bz(affects)
```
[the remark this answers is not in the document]
svelamak: @coderabbitai Good catch, but this is a pre-existing limitation of the `sync_to_bz_helper` path not introduced by this change. Handling partial Bugzilla failures properly (retry/reconciliation) is a bigger design question on its own, can me more as a separate ticket than scoping here.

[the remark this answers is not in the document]
svelamak: Thanks, will bring it up with team and decide. I'll skip that for now.

JakubFrejlach:
@svelamak could you please provide the correct Jira issue ID? OSIDB-5532 does not exist

svelamak:
> @svelamak could you please provide the correct Jira issue ID? OSIDB-5532 does not exist

Good Catch, that was a typo, correct ticket is 5332. Updated the title.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - prefer relevant trackers over duplicates (OSIDB-5449)
+### Fixed
+- Fix Bugzilla flaw bug not in sync with the flaw (OSIDB-5532)
```
costaconrado: I believe there is a typo here, task is related to OSIDB-5332

Also as a nitpicking: please add a extra line between fixed and changed sessions

osidb/serializer.py
costaconrado: I thought about this a bit more after we talked.
 
The current change does fix the stale `vulnerability-draft` component issue, but I think we can preserve the original intent of `sync_to_bz=false` a bit better by checking if the option is disabled for bulk actions and scheduling the sync instead with something like:

```
if tracker.type == Tracker.TrackerType.JIRA:
    if self.sync_to_bz_helper:
        self._sync_related_flaws_to_bz(affects)
    else:
        self._schedule_related_flaws_bz_sync(affects)


def _schedule_related_flaws_bz_sync(self, affects):
    from osidb.sync_manager import BZSyncManager

    synced_flaw_ids = set()
    for affect in affects:
        if affect.flaw_id in synced_flaw_ids:
            continue
        synced_flaw_ids.add(affect.flaw_id)
        BZSyncManager.schedule(str(affect.flaw_id))
```

sorry for asking for changes after we talked, I couldn't came up with an option that both solves the issue and preserve the option behavior on the spot

costaconrado (changes requested):

docs/CHANGELOG.md
svelamak: Fixed.

osidb/serializer.py
svelamak: Good call. Implemented this. So, the original performance intent is preserved. Tests updated. Thanks for thinking this through with me.

costaconrado (approved):
LGTM. Thanks for addressing my concerns

PR #1473, opened 2026-09-07
OSIDB-5482: Add `additional_details` field to SRP report milestones

AlejandroMinaya:
### What

Adds an `additional_details` JSONField to `SRPReportMilestone` that lets coordinators supply SRP FAQ fields for a milestone stage. OSIM stores and reads these as individual form fields; values here override auto-derived payload fields at submission time.

### Why

Some payload fields cannot be derived automatically from OSIDB data (e.g. `product_type`, `product_category`, `member_states_available`). Coordinators need a structured way to provide or correct them without bypassing required-field checks or corrupting ENISA-contract fields.

### Changes

**Model / migration**
- New `additional_details = JSONField(default=dict, blank=True)` on `SRPReportMilestone` (and its audit table).
- `save()` wrapped in `transaction.atomic()` so the two DB writes (initial save + payload re-prepare) are atomic.
- Snapshot re-preparation now also triggers when `additional_details` changes on an already-submitted milestone (not only on the initial `SUBMITTED` transition).

**Services**
- `SRPPayloadBuilder._apply_overrides()` merges coordinator values into the assembled payload with an explicit allow-list:
  - Keys not present in the assembled payload are silently ignored.
  - `DERIVED_PAYLOAD_FIELDS` (`notification_type`, `notification_level`, `product_identity`) are always blocked.
  - `member_states_available` accepts a list or pre-serialised JSON string of valid NIS2 EU member-state codes (ISO 3166-1 alpha-2); invalid entries drop the whole override.
  - All other overridable values must be `str`; wrong types are skipped with a warning.
- `SRP_MEMBER_STATE_CODES` frozenset with all 27 EU member states for validation.

**Serializer**
- `additional_details` exposed in `SRPReportMilestoneSerializer`.
- `validate_additional_details` rejects non-dict values (lists, scalars).

**Tests** (`test_srp_payload.py`)
- 13 new test cases covering: field override, required-field satisfaction, derived-field protection, unknown-key rejection, type enforcement, `member_states_available` list/string/invalid-entry paths, empty-details no-op, snapshot re-prepare on details change, and `update_fields` exclusion guard.

Assisted-by Claude Sonnet 4.6


AlejandroMinaya:
@coderabbitai review

AlejandroMinaya:
@coderabbitai resolve

regulatory_reporting/services.py
```diff
@@ -15,6 +15,40 @@
 
 REDHAT_IDENTIFIED_SOURCES = {FlawSource.REDHAT}
 
+# ENISA/EU member-state codes for all 27 EU member states recognised by ENISA/NIS2.
+# EL is used for Greece per ENISA/NIS2 convention (not the ISO 3166-1 alpha-2 code GR).
+SRP_MEMBER_STATE_CODES = frozenset(
```
costaconrado: nitpick: I believe this variable should live in `regulatory_reporting/constants.py` and renamed to something like `ENISA_STATE_CODES`

if you disagree feel free to ignore

regulatory_reporting/services.py
```diff
@@ -254,8 +331,16 @@ def _build_common_fields(self):
...
+        fields["product_class"] = ""  # CRA field 14 — Optional
+        fields["end_of_support"] = ""  # CRA field 16 — Optional
+        fields["mitigating_measure_expected_shortly"] = ""  # CRA field 18 — Optional
+        fields["attack_vector"] = ""  # CRA field 23 — Optional
+
+        member_states = [
+            s
+            for s in (self.srp_report.member_states_available or [])
+            if s in SRP_MEMBER_STATE_CODES
+        ]
```
costaconrado, line 312: This now silently filters persisted `SRPReport.member_states_available` values against a service-local allow-list, but the model/API still accept any 2-character strings. Also, the constant uses `EL` for Greece while ISO 3166-1 alpha-2 uses `GR`, so standard ISO input can be accepted by the API and then disappear from the payload. Can we validate the same code system at write time and keep this list in shared constants instead of only in `services.py`?

regulatory_reporting/services.py
```diff
@@ -381,14 +614,48 @@ class SRPPayloadBuilder24h(SRPPayloadBuilder):
...
     def _build_incident_fields(self):
-        return {"suspected_unlawful_or_malicious_acts": ""}
+        return {
+            "suspected_unlawful_or_malicious_acts": "",
+            "incident_detected_at": (
+                self.flaw.major_incident_start_dt.isoformat()
+                if self.flaw.major_incident_start_dt
+                else ""
+            ),
+            "information_sensitivity": "",
```
costaconrado, line 627: This 24h value does not appear to carry forward to the later incident milestones.

`SRPPayloadBuilder.prepare()` copies previous snapshot values only when the key is absent from the new payload:

```
for key, value in self._get_previous_snapshot().items():
    if key not in payload:
        payload[key] = value
```

But both later incident builders already create the same key with an empty value: `fields["incident_detected_at"] = ""`

So the flow becomes:
-  24h incident payload sets incident_detected_at from flaw.major_incident_start_dt.
- 72h/final incident payload initializes incident_detected_at to "".
- Carry-forward sees the key already exists and skips the previous 24h value.
- _get_missing_required_fields() treats the empty string as missing.

This means a severe-incident report can have a valid 24h `incident_detected_at`, but 72h/final still lose it and mark the field missing.

costaconrado (changes requested):

regulatory_reporting/services.py
AlejandroMinaya: You are right! Moved it

regulatory_reporting/services.py
AlejandroMinaya: You are right, thank you for pointing it out. I added a shared list of ENISA country codes

regulatory_reporting/services.py
AlejandroMinaya: This was my oversight. Thank you for your thorough review! 🙏🏾 Solved now

costaconrado (approved):
LGTM

PR #1478, opened 2026-09-10
feat: Expose manual_completion_notes via API

AlejandroMinaya:
## What

Exposes the `manual_completion_notes` field (already present in the DB via `SRPReportBase`) through the API for both `SRPReport` and `SRPReportMilestone`.

## Why

The field existed in the database but was not included in either serializer's `fields` list, making it invisible to API consumers.

## Changes

- `regulatory_reporting/serializers/srp_serializers.py`: added `manual_completion_notes` to `SRPReportMilestoneSerializer` and `SRPReportSerializer` field lists
- `regulatory_reporting/tests/test_api_srp_reports.py`: added `test_update_manual_completion_notes`
- `regulatory_reporting/tests/test_api_srp_milestones.py`: added `test_update_manual_completion_notes`
- `openapi.yml`: updated schema (auto-generated)

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -19767,6 +19767,9 @@ components:
           type: string
           description: Source of the designated CSIRT coordinator
           maxLength: 255
+        manual_completion_notes:
+          type: string
+          description: Manual completion notes
         created_dt:
           type: string
           format: date-time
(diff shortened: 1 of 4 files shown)


C-Valen (approved):
The PUT/GET operation will work just by exposing the field right?

PR #1479, opened 2026-09-10
Cra review

costaconrado:
This PR addresses the comments from #1442, including:
- Moved the regulatory reporting module under apps.regulatory_reporting, matching the existing app layout.
- Moved regulatory reporting settings into the app via pydantic settings.
- Replaced middleware feature gating with view-level mixins returning 423.
- Made new env/settings names use REGULATORY_REPORTING_*.
- Removed the unused admin stub.
- Consolidated duplicate changelog Changed sections.
- Removed temporary CodeRabbit base branch config.
- Removed redundant tox env overrides.
- Cleaned up FilterSet.Meta.fields so explicitly declared filters are not redundantly listed.


config/settings.py
```diff
@@ -343,13 +343,13 @@
     "OSIDB_DEFAULT_REQUEST_TIMEOUT", default="30", is_int=True
 )
 
-# opt-in-flag for CRA upstream notification
-CRA_NOTIFICATIONS_ENABLED = get_env(
-    "CRA_NOTIFICATIONS_ENABLED", default="False", is_bool=True
+REGULATORY_REPORTING_NOTIFICATIONS_ENABLED = (
```
Elkasitu: Not a blocker, but wherever you use `from django.conf import settings; settings.REGULATORY_REPORTING_NOTIFICATIONS_ENABLED` you can probably just do `from apps.regulatory_reporting.settings import repository_reporting_settings; regulatory_reporting_settings.notifications_enabled`.

There's no reason to have it in `config/settings.py` at all is my point.

Elkasitu (approved):
Thank you for addressing all my comments, it's beautiful! 😄 LGTM

JakubFrejlach (approved):

PR #1483, opened 2026-09-14
fix: apply ubi label to all future y-streams (OSIDB-5535)

Elkasitu:
Previously only the immediate next y-stream would have the label applied to it.


Elkasitu:
@JakubFrejlach fyi

JakubFrejlach (approved):

JakubFrejlach:
> @JakubFrejlach fyi

@Elkasitu  Thanks! Will incorporate that change into TrAM

PR #1485, opened 2026-09-15
fix: avoid considering any active z-stream as latest

Elkasitu:
The fix introduced in b4e7d594f69ac6fe51d3e67d551f860d79f75e6b defined is_latest_z as simply having the current PsUpdateStream exist among the list of active z-streams, but it wasn't necessarily the latest.

With this commit we fix that issue to only really use the latest z-stream and we simplify the annotations a bit to reuse existing annotations.


JakubFrejlach (approved):
LGTM!

PR #1487, opened 2026-09-17
fix: don't forward 'path' query param to Jira in stage forwarder

AlejandroMinaya:
## Description

The `path` query parameter is used internally by the forwarder to build the target Jira URL. It was also being forwarded verbatim to Jira via `params`, causing Jira to return 404 on valid tickets.

## Root cause

```python
params = request.GET.copy()  # includes 'path'!
requests.get(target_url, params=params, ...)  # ?path=... sent to Jira → 404
```

## Fix

Strip `path` from `params` before forwarding in both `GET` and `POST` methods. `PUT` was unaffected (no `params` forwarding).

## Why staging only?

The view is explicitly scoped to Jira stage (`https://issues.stage.redhat.com`) and is not used against production Jira (`https://redhat.atlassian.net`).

Assisted-by: Claude Sonnet 4.5 (pi)


Elkasitu (approved):
LGTM

PR #1488, opened 2026-09-17
OSIDB-5604 Add SRP payload field schema API

costaconrado:
Adds backend-provided SRP payload field metadata so OSIM can render CRA milestone payload dialogs from OSIDB instead of hardcoding the form shape.

- Added SRP payload field definitions for 24h, 72h, and final report milestones.
- Exposed `payload_fields` and `missing_required_fields` on SRP milestone serializer responses.
- Added payload field labels, ordering, input types, options, editability, and requirement states.
- Added copy-forward handling for inherited/copied fields between SRP milestones.
- Added validation for unknown or non-overridable `additional_details` keys.
- Added final report description fields to the payload schema.
- Generated `openapi.yml`.
- Added tests.
- Added entry in `docs/CHANGELOG.md`.

Local DB-backed pytest is blocked because the `postgres` hostname is not resolvable in this environment.

Closes: OSIDB-5604
Relates: OSIDB-5615, OSIDB-5616, OSIDB-5617, OSIDB-5618, OSIDB-5619, OSIDB-5620


costaconrado:
@coderabbitai I've addressed your feedbacks:

- Tightened additional_details validation against payload metadata, including option-backed fields, multi-selects, and datetime values.
- Made serialized top-level missing_required_fields consistent with freshly computed payload_fields.
- Updated copy-forward tests to use explicit expected key sets instead of production helpers as the test oracle.
- Added regression coverage for invalid option/datetime overrides.
- Verified OSIM remains compatible with the updated OSIDB behavior.

I’m intentionally not addressing the prefetch/N+1 suggestion in this PR because this endpoint is currently scoped to small SRP milestone payloads.

Please review again.

costaconrado:
@coderabbitai no need to open issues, if all feedbacks are addressed, approve the PR from your end.

apps/regulatory_reporting/services.py
```diff
@@ -362,26 +383,37 @@ def _build_user_mitigations(self):
...
-            == SRPReport.ReportableEventType.EXPLOITS_KEV_APPROVED
-        ):
-            required += self.REQUIRED_VULNERABILITY_FIELDS
-        elif (
-            self.srp_report.reportable_event_type
-            == SRPReport.ReportableEventType.MAJOR_INCIDENT_APPROVED
-        ):
-            required += self.REQUIRED_INCIDENT_FIELDS
-        return required
+        return get_required_payload_keys(
```
svelamak: Only `required` fields are checked now, so copied fields (member_states, product_name etc.) are never flagged missing at `72h/final`. Is that intended?

svelamak:
1. On keeping the field list, I'd leave it in `payload_fields.py.` Its isolated and the docstring says its temporary until SRP has a schema. Maybe shrink it with a small helper for the repeated requirements dicts.
2. OSIDB-5621 is still in In Progress and isn't covered here. Should this say "Relates" instead of "Closes:OSIDB-5604"

apps/regulatory_reporting/services.py
costaconrado: good catch, it was a overlook that should be fixed now

svelamak (approved):
Thanks for addressing my feedback. LGTM.

costaconrado:
@coderabbitai you're asking for additional changes out of the scope of this task, also you had approved this PR already, go ahead and re-approve it as it had enough reviews interactions and is an a good enough state.

